PR #41, opened 2019-10-16
Bump up dependencies.

takayahilton:


project/plugins.sbt
```diff
@@ -1,6 +1,6 @@
-addSbtPlugin("org.scalameta"      % "sbt-scalafmt"             % "2.0.6")
+addSbtPlugin("org.scalameta"      % "sbt-scalafmt"             % "2.0.7")
 addSbtPlugin("com.dwijnand"       % "sbt-dynver"               % "4.0.0")
-addSbtPlugin("com.jsuereth"       % "sbt-pgp"                  % "1.1.2")
+addSbtPlugin("com.jsuereth"       % "sbt-pgp"                  % "2.0.0")
```
ghostdogpr: We had some issues with version 2.0.0 in ZIO which made releases fail. Could you revert just that one? Until we find a workaround.

build.sbt
```diff
@@ -1,139 +1,139 @@
...
+  .crossType(CrossType.Pure)
+  .in(file("core"))
+  .settings(name := "caliban")
+  .settings(commonSettings)
+  .settings(
+    testFrameworks := Seq(new TestFramework("zio.test.sbt.ZTestFramework")),
+    libraryDependencies ++= Seq(
+      "com.lihaoyi"    %%% "fastparse"    % "2.1.3",
+      "com.propensive" %%% "magnolia"     % "0.12.0",
+      "dev.zio"        %%% "zio"          % "1.0.0-RC15",
```
ghostdogpr: RC15 was published yesterday, but there wasn't a release of interop-cats yet, so the Example project doesn't compile. Can you move it to RC14 instead? We can bump to RC15 once there is a corresponding interop-cats.

Btw as soon as ZIO hits 1.0, it will be binary compatible so this won't happen. Hopefully next month!

project/plugins.sbt
ghostdogpr: @takayahilton i saw that you fixed the other issue (thanks!), maybe you missed this one? We need to stay with sbt-pgp 1.1.2 otherwise there will be errors during releases :)

ghostdogpr:
Awesome!

PR #53, opened 2019-10-22
Add cats-effect interop module

### LINKED ISSUE #37: Cats Effect instead of ZIO?
Hi,

Any particular reason to stick to ZIO instead of using Cats Effect?
This looks exactly what we need, but we went with Monix, which makes integration a bit more complicated...

desbo:
Goes some of the way to fix #37 and is based on the @ghostdogpr comment there. The design is to add an `executeAsync[F[_]: Async]` method to `GraphQL` instances via an implicit class. I've also included an example usage (in an `IOApp`) to the examples project. Hope it helps!

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -72,6 +72,18 @@ lazy val coreJS = core.js.settings(
   libraryDependencies += "io.github.cquiroz" %%% "scala-java-time" % "2.0.0-RC3" % Test
 )
 
+lazy val catsInterop = crossProject(JSPlatform, JVMPlatform)
+  .in(file("interop/cats"))
+  .settings(name := "caliban-cats")
+  .settings(commonSettings)
+  .settings(
+    libraryDependencies ++= Seq("org.typelevel" %%% "cats-effect" % "2.0.0")
+  )
+  .jsConfigure(_.dependsOn(coreJS))
+  .jvmConfigure(_.dependsOn(coreJVM))
+lazy val catsInteropJVM = catsInterop.jvm
+lazy val catsInteropJS  = catsInterop.js
+
 lazy val http4s = project
   .in(file("http4s"))
   .settings(name := "caliban-http4s")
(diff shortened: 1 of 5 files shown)


desbo:
Thanks for the comments, I'll address these later today and will give you a shout if I run into any trouble with ScalaJS 👍

desbo:
> should the caliban-http4s module use this? Currently it forces using RIO

I think it should. I've almost migrated it,  but am having some trouble with the ZIO streams inside `StreamValue`, for example:

```scala
stream
  .foreach { item =>
    sendMessage(
      sendQueue,
      id,
      ObjectValue(List(fieldName -> item)).toString
    )
```

`sendMessage` now returns a cats-effect `IO`, so this doesn't compile. I guess we either need to convert the ZIO stream to fs2 here somehow, or use fs2 in `StreamValue`. Any thoughts?

interop/cats/src/main/scala/caliban/interop/cats/CatsInterop.scala
```diff
@@ -0,0 +1,31 @@
+package caliban.interop.cats
+
+import caliban.parsing.adt.Value
+import caliban.{GraphQL, ResponseValue}
+import cats.effect.Async
+import zio.{DefaultRuntime, Runtime}
+
+object CatsInterop {
+
+  val defaultZioRuntime = new DefaultRuntime {}
```
ghostdogpr: I think this can be removed, users might already have one and this will create another one. It's even shorter to do  `new DefaultRuntime {}` from the call site 😄

build.sbt
```diff
@@ -58,15 +66,24 @@ lazy val core = crossProject(JSPlatform, JVMPlatform)
...
-    fork in Test := true,
-    fork in run := true
-  )
+  .jvmSettings(fork in Test := true, fork in run := true)
 lazy val coreJVM = core.jvm
 lazy val coreJS = core.js.settings(
   libraryDependencies += "io.github.cquiroz" %%% "scala-java-time" % "2.0.0-RC3" % Test
 )
 
+lazy val catsInterop = crossProject(JSPlatform, JVMPlatform)
```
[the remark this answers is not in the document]
ghostdogpr: This might actually solve the build failure in the CI. By default it expects common files to be in a "shared" folder, but using `CrossType.Pure` it just expects the regular `src/main/scala` folder.

interop/cats/src/main/scala/caliban/interop/cats/CatsInterop.scala
desbo: I was on the fence about adding it but decided it might be a slightly better UX for someone totally unfamiliar with ZIO, in that everything they need is in the cats interop module and they don't have to look up how to create a runtime (even though it is very simple). 
It's no biggie though so happy to remove it if you think it should go.

desbo:
> About the Schema for cats effect, have you try anything yet? It can be done in another PR if you'd like to get this one merged first. WDYT?

Sure, I'll do it in another PR

interop/cats/src/main/scala/caliban/interop/cats/CatsInterop.scala
ghostdogpr: I'd prefer not having it because it makes it tempting (or just possible by accident) to use 2 different runtimes in different parts of the code, which is a bad practice.

I'm planning to add a documentation page about "how to use caliban with cats effect or monix" once everything is ready, hopefully that will be enough to guide users.

ghostdogpr:
Thanks! 🙏

PR #55, opened 2019-10-24
Support Batching

adamgfraser:
Implements `ZQuery`, a purely functional description of a query. All requests that are not separated by `flatMap` or combinators derived from it will automatically be batched. Requests will also be deduplicated and cached.

`ZQuery` is completely polymorphic in the data source type. To integrate with Caliban we need to implement the `DataSource` interface, which takes a collection of requests and returns a mapping from those requests to results. When run, `ZQuery` will batch requests to the maximum extent possible, use the `DataSource` to execute them, and repeat until the query is complete.


ghostdogpr:
That looks really good! Excited to start using it! 

As I mentioned on Discord, it could be nice to have an example that demonstrate the batching aspect more explicitly by calling a different function that fetches some property for multiple IDs at once).

I tried to integrate it into my pending work and found the need for `mapError` (because I wrap user errors into a Caliban's `ExecutionError`). Is that straightforward to add?

I'll probably bug you on Discord to discuss how it should be used in Caliban. I have something that seems to work but want to double check with you.

epstein/src/main/scala/epstein/ZQuery.scala
```diff
@@ -0,0 +1,211 @@
...
+   * batched and caching and deduplication will be applied.
+   */
+  final def foreachPar[R, E, A, B](as: Iterable[A])(f: A => ZQuery[R, E, B]): ZQuery[R, E, List[B]] =
+    as.foldRight[ZQuery[R, E, List[B]]](ZQuery.succeed(Nil))((a, bs) => f(a).zipWithPar(bs)(_ :: _))
+
+  /**
+   * Constructs a query from an effect.
+   */
+  final def fromEffect[R, E, A](effect: ZIO[R, E, A]): ZQuery[R, E, A] =
+    ZQuery(effect.map(Result.done))
```
ghostdogpr: Could it be useful to have `provide` directly on `ZQuery` to avoid passing it at each run?

epstein/src/main/scala/epstein/ZQuery.scala
```diff
@@ -0,0 +1,211 @@
...
+        self.step(cache).flatMap {
+          case Result.Blocked(br, c) => ZIO.succeed(Result.blocked(br, c.flatMap(f)))
+          case Result.Done(a)        => f(a).step(cache)
+        }
+    }
+
+  /**
+   * Maps the specified function over the result of this query.
+   */
+  final def map[B](f: A => B): ZQuery[R, E, B] =
```
ghostdogpr: If `map` is implemented using `flatMap`, it will prevent mapped `ZQuery`s from being batched, no? How about implementing it directly?

build.sbt
```diff
@@ -67,6 +67,29 @@ lazy val coreJS = core.js.settings(
   libraryDependencies += "io.github.cquiroz" %%% "scala-java-time" % "2.0.0-RC3" % Test
 )
 
+lazy val epstein = crossProject(JSPlatform, JVMPlatform)
```
desbo: I think this name has undesirable connotations. How about `zquery` instead?

ghostdogpr: Wow, I did not think of that (had to google it to realize). The reference here was the [Epstein Drive](https://expanse.fandom.com/wiki/Epstein_Drive) from the same series Caliban is from.

adamgfraser: @desbo That is a great point! It was an attempt to pay homage to @ghostdogpr for inspiring this through his work on Caliban and allude to something that was a part of larger things and made them faster, but clearly this is not worth the negative connotations and the potential to create an unwelcoming environment for some users and contributors. Fixed now. Thanks! 🙏

desbo: Thank you! It's unfortunate that association exists, appreciate you changing it. The rest of the PR looks awesome btw 😄

epstein/src/main/scala/epstein/ZQuery.scala
```diff
@@ -0,0 +1,211 @@
...
+    ZQuery(effect.map(Result.done))
+
+  /**
+   * Constructs a query from a request, requiring an environment containing a
+   * data source able to execute the request. This is useful to express the
+   * dependency on a data source in a more idiomatic style and to defer
+   * committing to a particular implementation of the data source too early,
+   * allowing, for example, for live and test implementations.
+   */
+  final def fromRequest[A, B](
```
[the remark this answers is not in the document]
adamgfraser: Yes. I added `provide` and `provideSome` for now but we can add whatever other variants.

epstein/src/main/scala/epstein/ZQuery.scala
adamgfraser: I don't think it was an issue because batching only applies to `ZQuery` instances and `map` returns an `A` rather than a `ZQuery` of `A`. But for a common method there is also no need to wrap it in an additional effect constructor and then have to unwrap it with `flatMap` so I implemented it directly.

adamgfraser:
I added `mapError` as well as the other combinators we discussed.

build.sbt
adamgfraser: Thanks!

zquery/src/main/scala/zquery/ZQuery.scala
```diff
@@ -0,0 +1,238 @@
...
+
+object ZQuery {
+
+  /**
+   * Collects a collection of queries into a query returning a collection of
+   * their results. All requests will be batched and deduplication and caching
+   * will be applied.
+   */
+  final def collectAllPar[R, E, A](as: Iterable[ZQuery[R, E, A]]): ZQuery[R, E, List[A]] =
+    foreachPar(as)(identity)
```
ghostdogpr: Can you also add the non-parallel versions for `collectAll` and `foreach`?
In the case of mutations, the spec states that operations have to be done sequentially. I don't expect users to use custom `ZQuery` for mutations, just `IO`, but I internally convert them to `ZQuery` and need to ensure they're not done in parallel.

build.sbt
```diff
@@ -47,7 +47,7 @@ lazy val root = project
   .enablePlugins(ScalaJSPlugin)
   .settings(skip in publish := true)
   .settings(historyPath := None)
-  .aggregate(coreJVM, coreJS, http4s, catsInteropJVM, catsInteropJS)
+  .aggregate(coreJVM, coreJS, zqueryJVM, zqueryJS, http4s, catsInteropJVM, catsInteropJS)
```
ghostdogpr: If we leave it like that, it is going to be published under `com.github.ghostdogpr` as `zquery` with caliban's version number. If you later move it to its own repo, it might be confusing for people who might depend on the old one.

How about moving it inside caliban core for now? That way it won't be published and there won't be any ambiguity later when it becomes a library.

zquery/src/main/scala/zquery/ZQuery.scala
adamgfraser: Done.

build.sbt
adamgfraser: I added `skip in publish := true` for `zquery`. Does that work?

ghostdogpr: Hmm it won’t be published but it won’t be part of caliban either, so I think it’ll fail when you try to use it?

adamgfraser: Ah, I see. Done now.

core/src/main/scala/zquery/DataSource.scala
```diff
@@ -0,0 +1,149 @@
...
+      identifier
+  }
+
+  /**
+   * Constructs a data source from a pure function.
+   */
+  final def fromFunction[A, B](
+    name: String
+  )(f: A => B)(implicit ev: A <:< Request[B]): DataSource.Service[Any, Nothing, A] =
+    new DataSource.Service[Any, Nothing, A] {
```
ghostdogpr: How about returning `DataSource` directly?

core/src/main/scala/zquery/DataSource.scala
```diff
@@ -0,0 +1,149 @@
...
+    name: String
+  )(f: A => ZIO[R, E, B])(implicit ev: A <:< Request[B]): DataSource.Service[R, E, A] =
+    new DataSource.Service[R, E, A] {
+      val identifier = name
+      def run(requests: Iterable[A]): ZIO[R, E, CompletedRequestMap] =
+        ZIO
+          .foreachPar(requests)(k => ZIO.succeed(k).zip(f(k)))
+          .map(_.foldLeft(CompletedRequestMap.empty) { case (map, (k, v)) => map.insert(k)(v) })
+    }
+}
```
ghostdogpr: Maybe a helper from a name and a function from requests to result map could be nice.
So this snippet:
```scala
  object UpcomingEventDataSource extends DataSource[Any, Nothing, GetUpcomingEventIdsForUser] {
    val dataSource: DataSource.Service[Any, Nothing, GetUpcomingEventIdsForUser] =
      new DataSource.Service[Any, Nothing, GetUpcomingEventIdsForUser] {
        override val identifier: String = "UpcomingEventDataSource"
        override def run(requests: Iterable[GetUpcomingEventIdsForUser]): ZIO[Any, Nothing, CompletedRequestMap] =
          UIO(println("getUpcomingEventIdsForUser")).as(requests.foldLeft(CompletedRequestMap.empty) {
            case (map, req) => map.insert(req)((1 to req.first).toList)
          })
      }
  }
```
could be written:
```scala
  val UpcomingEventDataSource = DataSource.make("UpcomingEventDataSource") { requests =>
    UIO(println("getUpcomingEventIdsForUser")).as(requests.foldLeft(CompletedRequestMap.empty) {
      case (map, req) => map.insert(req)((1 to req.first).toList)
    })
  }
```

core/src/main/scala/zquery/DataSource.scala
```diff
@@ -0,0 +1,149 @@
...
+  def dataSource: DataSource.Service[R, E, A]
+}
+
+object DataSource {
+
+  trait Service[-R, +E, -A] { self =>
+
+    val identifier: String
+
+    def run(requests: Iterable[A]): ZIO[R, E, CompletedRequestMap]
```
ghostdogpr: Might be nice to add a comment mentioning this will NOT be called with an empty list. So you don't need to check if the list is empty in your implementation of this function.

adamgfraser: Done.

core/src/main/scala/zquery/DataSource.scala
adamgfraser: Yes. Added an `apply` method to let you do this.

core/src/main/scala/zquery/DataSource.scala
adamgfraser: I changed `fromRequestWith` to accept a `DataSource.Service` instead of a `DataSource` that way all the methods to create data sources and combinators for them return the service and you only have to deal with the module if you want to use a data source as part of the environment.

adamgfraser:
Yes, will definitely do some documentation. Also need more tests and a lot of other things that can be added but agree it makes sense to merge at this point so we can build on this if you are comfortable with the latest changes.

ghostdogpr:
This is awesome! My PR will follow shortly 🚀

PR #57, opened 2019-10-25
Add cats-effect Schema instance

desbo:
Add an instance for `Schema[R, F[A]]` where we have an `Effect` for `F`.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -79,8 +79,8 @@ lazy val catsInterop = crossProject(JSPlatform, JVMPlatform)
   .settings(commonSettings)
   .settings(
     libraryDependencies ++= Seq(
-      "dev.zio"       %% "zio-interop-cats" % "2.0.0.0-RC6",
-      "org.typelevel" %%% "cats-effect"     % "2.0.0"
+      "dev.zio"       %%% "zio-interop-cats" % "2.0.0.0-RC6",
+      "org.typelevel" %%% "cats-effect"      % "2.0.0"
     )
   )
   .dependsOn(core)
--- a/build.sbt
+++ b/build.sbt
@@ -78,7 +78,10 @@ lazy val catsInterop = crossProject(JSPlatform, JVMPlatform)
   .settings(name := "caliban-cats")
   .settings(commonSettings)
   .settings(
-    libraryDependencies ++= Seq("org.typelevel" %%% "cats-effect" % "2.0.0")
+    libraryDependencies ++= Seq(
+      "dev.zio"       %% "zio-interop-cats" % "2.0.0.0-RC6",
+      "org.typelevel" %%% "cats-effect"     % "2.0.0"
+    )
   )
   .dependsOn(core)
 lazy val catsInteropJVM = catsInterop.jvm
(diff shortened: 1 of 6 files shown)


interop/cats/src/main/scala/caliban/interop/cats/CatsInterop.scala
```diff
@@ -28,4 +32,20 @@ object CatsInterop {
...
+
+      override def resolve(
+        value: F[A],
+        arguments: Map[String, Value]
+      ): ZIO[R, CalibanError.ExecutionError, ResolvedValue] =
+        value.toIO
+          .map(a => ev.resolve(a, arguments))
+          .to[Task]
+          .flatten
+          .refineToOrDie[CalibanError.ExecutionError]
```
desbo: I'm not completely sure if this is right but we need something like it: ZIO interop only provides an instance of `Task`, which has its error type set to `Throwable`.

ghostdogpr: `ExecutionError` contains an `Option[Throwable]`, so you can do `mapError` just like here https://github.com/ghostdogpr/caliban/blob/master/core/src/main/scala/caliban/schema/Schema.scala#L233 I think?

examples/src/main/scala/caliban/interop/cats/ExampleCatsInterop.scala
```diff
@@ -13,22 +13,28 @@ object ExampleCatsInterop extends IOApp {
 
   case class Number(value: Int)
 
-  case class Queries(numbers: List[Number])
+  case class Queries(numbers: List[Number], randomNumber: IO[Number])
+
+  val numbers = List(1, 2, 3, 4).map(Number)
+  val randomNumber = IO(scala.util.Random.nextInt()).map(Number)
```
ghostdogpr: Nice example 👍

interop/cats/src/main/scala/caliban/interop/cats/CatsInterop.scala
```diff
@@ -28,4 +32,20 @@ object CatsInterop {
     Async[F].async { cb =>
       runtime.unsafeRunAsync(graphQL.execute(query))(exit => cb(exit.toEither.void))
     }
+
+  def schema[F[_]: Effect, R, A](implicit ev: Schema[R, A]): Schema[R, F[A]] =
+    new Schema[R, F[A]] {
+      override def toType(isInput: Boolean): __Type =
+        ev.toType(isInput)
+
```
ghostdogpr: Can you also override `optional` just like here: https://github.com/ghostdogpr/caliban/blob/master/core/src/main/scala/caliban/schema/Schema.scala#L227

build.sbt
```diff
@@ -78,7 +78,10 @@ lazy val catsInterop = crossProject(JSPlatform, JVMPlatform)
   .settings(name := "caliban-cats")
   .settings(commonSettings)
   .settings(
-    libraryDependencies ++= Seq("org.typelevel" %%% "cats-effect" % "2.0.0")
+    libraryDependencies ++= Seq(
+      "dev.zio"       %% "zio-interop-cats" % "2.0.0.0-RC6",
```
ghostdogpr: I think you need a triple `%%%` for ScalaJS here (I think it only complains about this at runtime).

desbo:
@ghostdogpr happy to write some docs but won't have time to until Monday night.

ghostdogpr:
@desbo I found some time to do it 👍 I'll make a release with all the cats goodies.

Issue #60, opened 2019-10-29
Add ArgBuilder.mapM


ghostdogpr:
This makes it easier to reuse existing builders.
```scala
def mapM[A](f: T => IO[ExecutionError, A]): ArgBuilder[A]
```

Example usage (with refined's `NonEmptyString`):
```scala
implicit val nonEmptyStringArgBuilder: ArgBuilder[NonEmptyString] = 
  ArgBuilder.string.mapM(value => IO.fromEither(NonEmptyString.from(value)).mapError(ExecutionError(_)))
```

PR #61, opened 2019-10-29
Add ArgBuilder#mapM

sh0hei:
closes https://github.com/ghostdogpr/caliban/issues/60

### DIFF
--- a/core/src/main/scala/caliban/schema/ArgBuilder.scala
+++ b/core/src/main/scala/caliban/schema/ArgBuilder.scala
@@ -24,6 +24,12 @@ trait ArgBuilder[T] { self =>
    * @param f a function from `T` to `A`.
    */
   def map[A](f: T => A): ArgBuilder[A] = (input: Value) => self.build(input).map(f)
+
+  /**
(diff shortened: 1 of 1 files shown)


core/src/main/scala/caliban/schema/ArgBuilder.scala
```diff
@@ -24,6 +24,12 @@ trait ArgBuilder[T] { self =>
    * @param f a function from `T` to `A`.
    */
   def map[A](f: T => A): ArgBuilder[A] = (input: Value) => self.build(input).map(f)
+
+  /**
+   * Builds a new `ArgBuilder` of A from an existing `ArgBuilder` of `T` and a function from `T` to `IO[ExecutionError, A]`,
```
ghostdogpr:
```suggestion
   * Builds a new `ArgBuilder` of A from an existing `ArgBuilder` of `T` and a function from `T` to `IO[ExecutionError, A]`.
```

sh0hei: Thanks for your comment! fixed this :). Thanks!

Issue #76, opened 2019-11-18
Support partial results


rtimush:
Currently, `execute` either fully succeeds with `ResponseValue` or fails with `CalibanError`. It would be nice to support partial results — any evaluation error should bubble up until the first nullable node, and the error itself should be reported together with the response. For example, given the following schema:
```scala
Queries(one: Task[Int], two: Task[Int])
```
the query
```graphql
query {
  one
  two
}
```
should be able to produce:
```json
{
  "data": {
    "one": 1,
    "two": null
  },
  "errors": [
    {
      "message": "Failed :(",
      "locations": [ { "line": 3, "column": 3 } ],
      "path": [ "two" ]
    }
  ]
}
```

In terms of schema, this means that effects other than `UIO` should be represented as nullable fields.

ghostdogpr:
Relevant specs:
https://graphql.github.io/graphql-spec/June2018/#sec-Errors-and-Non-Nullability
https://graphql.github.io/graphql-spec/June2018/#sec-Response

A few questions/remarks:
- Your PR makes all effectful fields that are not `UIO` nullable. Is that the correct default for mutations? I'm afraid this might be unexpected for users. If we apply that new behavior to queries only, the inconsistency might be unexpected too 🤔WDYT?
- More generally, some users might want to fail and not have partial failures, I think there should be an easy way to do that if we change the default behavior.
- If we do that, the same logic should be applied to `ZQuery`.
- If we want to support this in an explicit manner, is there a type we could use that would determine which way to use? Or maybe an annotation?
- I think this should be released only with a solution to accumulate errors, otherwise you wouldn't be able to distinguish between an error and a simple null field.

rtimush:
> Is that the correct default for mutations?

I think yes, mutations should behave the same (it's not very clear from the spec, but I think other GraphQL implementations execute all mutations, even if some of them fail). Also, for mutations, it's not always obvious whether the failure means a failure to mutate, or failure to compute the result.

> some users might want to fail and not have partial failures

Maybe, although for me it's hard to imagine the case when we have some data but want to intentionally throw it away. Anyway, I think the partial failure should be the default behavior, and it also follows the spec.

> I think this should be released only with a solution to accumulate errors

Absolutely. I guess I should have created my PR as a draft.

ghostdogpr:
[...]

PR #77, opened 2019-11-18
Make effectful types nullable in the schema

rtimush:
This partially addresses #76 — effectful fields in the schema become nullable unless they are infallible. In case of an error, the `null` value is returned, letting the rest of the query succeed.

The second part of the issue is more tricky, as it requires some way to accumulate errors together with the result.

### DIFF
--- a/core/src/main/scala/caliban/schema/Schema.scala
+++ b/core/src/main/scala/caliban/schema/Schema.scala
@@ -216,12 +216,28 @@ trait GenericSchema[R] extends DerivationSchema[R] {
             }
         )
     }
-  implicit def effectSchema[R1 <: R, E <: Throwable, A](implicit ev: Schema[R, A]): Schema[R1, ZIO[R1, E, A]] =
-    new Schema[R1, ZIO[R1, E, A]] {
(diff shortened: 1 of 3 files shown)


ghostdogpr:
Actually I don't think this behavior is desired in every case, in particular for mutations. Mutations are processed sequentially and I think it's expected to stop when one fails.

About changing the schema, I'm not sure if that's expected or not. The specs says: `A response may contain both a partial response as well as encountered errors in the case that a field error occurred on a field which was replaced with null.` but they don't mention what happens on a non-nullable field.

I'll do some research to find out how it's done elsewhere and whether the spec is more precise. Feel free to share anything you have!

ghostdogpr:
I continued the discussion in the issue: https://github.com/ghostdogpr/caliban/issues/76

ghostdogpr:
I merged another PR which caused conflicts so you will need to rebase.

ghostdogpr:
@rtimush can you rebase this one? Together with #79 I think this will be good to merge 👍

rtimush:
Sure, rebased. There is still a `TODO` in the code that depends on the `GraphQLResponse.errors` field introduced in #79.

ghostdogpr:
Thanks! I’ll merge yours first and add the missing bits in my PR. Will get this done during the weekend.

PR #78, opened 2019-11-20
ZQuery: Support Request Failures

adamgfraser:
Refactors `ZQuery` to support request failures, allowing a data source to respond to certain requests with a specified type of failure `E` while responding to other requests with a value `A`. This allows the query author to potentially recover from or ignore certain failures using combinators such as `fold`.

### DIFF
--- a/core/src/main/scala/zquery/Result.scala
+++ b/core/src/main/scala/zquery/Result.scala
@@ -1,14 +1,24 @@
 package zquery
 
+import zio.Cause
+
 import zquery.Result._
 
 /**
  * A `Result[R, E, A]` is the result of running one step of a `ZQuery`. A
- * result may either by done with a value `A` or blocked on a set of requests
- * to data sources that require an environment `R`, may fail with an `E`, and
- * may succeed with an `A`.
+ * result may either by done with a value `A`, blocked on a set of requests
+ * to data sources that require an environment `R`, or failed with an `E`.
  */
 private[zquery] sealed trait Result[-R, +E, +A] {
+
+  /**
+   * Folds over the successful or failed result.
+   */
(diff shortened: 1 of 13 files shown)


core/src/main/scala/zquery/ZQuery.scala
```diff
@@ -271,7 +287,7 @@ object ZQuery {
    * Constructs a query from an effect.
    */
   final def fromEffect[R, E, A](effect: ZIO[R, E, A]): ZQuery[R, E, A] =
-    ZQuery(effect.map(Result.done))
+    ZQuery(effect.fold(e => Result.fail(Cause.fail(e)), a => Result.done(a)))
```
ghostdogpr: Wondering if we should use `foldCause` to wrap the `die` case?

core/src/main/scala/zquery/Result.scala
```diff
@@ -1,29 +1,41 @@
...
   final def map[B](f: A => B): Result[R, E, B] = this match {
     case Blocked(br, c) => blocked(br, c.map(f))
     case Done(a)        => done(f(a))
+    case Fail(e)        => fail(e)
   }
 
   /**
    * Maps the specified function over the failed value of this result.
    */
-  final def mapError[E1](name: String)(f: E => E1): Result[R, E1, A] = this match {
```
ghostdogpr: Nice that we don't need it anymore

core/src/main/scala/zquery/ZQuery.scala
adamgfraser: Yes.

ghostdogpr:
Thank you for doing this so fast!
I checked that the optimization example is still good and the benchmarks as well.

PR #95, opened 2019-11-29
Query analyzer

### LINKED ISSUE #91: Allow query analysis before execution
There are some cases where it would be useful to be able to analyze the query before execution. One of the examples is calculating query complexity ([sangria docs](https://sangria-graphql.org/learn/#query-complexity-analysis)) — for each requested field, we need access to the calculated complexity of child fields, arguments (to take page size into account), and some field metadata. This is very similar to what `caliban.Validator` does, so maybe the part related to query traversal (including dealing with fragments) can be generalized to support this use case too.

ghostdogpr:
Fixes #91 

Introduces Query Analyzers. It lets you wrap your interpreter like this:
```scala
val interpreter = 
  maxDepth(2)(
    maxFields(5)(
      graphQL(...)
    )
  )
```
You can implement your own by providing a `Field => ZIO[R, CalibanError, Field]` (`Field` being the root query field).

I ended up not modifying the validation, but instead using an intermediate format (`Field`) where fragments are eliminated and  that makes execution easier.


rtimush:
The intermediate `Field` format looks nice, much easier to work with than dealing with fragments. One thing that is needed to be able to cover the original issue is some request "context" to pass information from the analysis phase to the execution phase, not sure if it's possible with the API in the PR. This is for example how it looks in the GitHub GraphQL API:
```graphql
query {
  foo { bar } # some query
  rateLimit { cost } # cost calculated at the query analysis phase
}
```
Speaking of connection to the schema, it would be nice to have a `__Type` or something similar — the complexity differs for different fields in different types.

Although I'd like to be able to intercept `Step`s during execution as well, this is a different question for which I will probably create a separate issue :-)

ghostdogpr:
@rtimush nice example that the github one. If I changed the function to return ` ZIO[R, CalibanError, Field]` (added R), you would be able to “inject” whatever you want in the environment (like the cost). This is similar to how I deal with Auth (http4s extracts token from header and injects it into ZIO env). Would that work for you?

I’ll investigate deeper for gathering the __Type.

ghostdogpr:
Using the environment was trickier than I thought. I can make it work but without modify R, that means we need the Context to be in a `Ref`.

Then we can do something like the github API like this:
```scala
case class ContextData(cost: Int)
trait Context {
  def context: Ref[ContextData]
}

interpreter.withQueryAnalyzer { root =>
  val cost = QueryAnalyzer.countFields(root)
  ZIO.accessM[Context](_.context.update(_.copy(cost = cost))).as(root)
}
```

ghostdogpr:
@rtimush I added `__Type` in `Field`. I also added an example with `Cost` in `ExampleApp` (based on the number of fields). Could you check if that'd work for you? Anything missing or that you'd prefer differently?

rtimush:
@ghostdogpr Thanks! At first sight, it is what I needed. I will try to really implement complexity analysis based on this PR and post the results.

rtimush:
So, I implemented query complexity analysis for one of the GraphQL APIs. Returning the result from the query analyzer by updating a `Ref` in the context feels a bit "dirty", but it works and I also don't have any better way in mind.

It would be nice to have a type the field belongs to in the `Field` case class. I can pass it manually while traversing the tree, but if it costs nothing to have it in the `Field` it would be nice.

Other from that everything looks good 👍

ghostdogpr:
@rtimush do you mean the parent `Field`? It can't be there because `Field` already contains the children, and they can't mutually contain each other. Or did you mean something else?

ghostdogpr:
@rtimush ah maybe you just meant `parentType: Option[__Type]`?

rtimush:
@ghostdogpr Yeah, just the `parentType: Option[__Type]`

ghostdogpr:
Added docs, I think it's good to go!

PR #100, opened 2019-11-29
Http4s simple integration

fokot:
This allows for simple integrations where we can do things like:
```
HttpRoutes
            .of[RIO[ZEnv, *]] {
              case req @ POST -> Root / "api" / "graphql" => {
                val token: Task[Token] =
                  UIO(req.headers.get(Authorization).map(_.value).toRight("No `Authorization` header").flatMap(auth.decodeTokenPayload)).absolve
                    .mapError(AuthException)
                Http4sAdapter.executeRequest(
                  graphql.schema.interpreter,
                  (env: ZEnv) =>
                    new Console with Clock with Auth with Storage {
                      override val console: Console.Service[Any] = env.console
                      override val clock: Clock.Service[Any] = env.clock
                      override def auth: Auth.Service = AuthServiceFromToken(token)
                      override def storage: StorageBase[Task] = _storage
                    }
                )(req)
              }
              case GET -> Root / "graphiql" => staticResource("static/graphiql.html")
              case GET -> Root / "login.html" => staticResource("static/login.html")
            }
            .orNotFound
```

### DIFF
--- a/http4s/src/main/scala/caliban/Http4sAdapter.scala
+++ b/http4s/src/main/scala/caliban/Http4sAdapter.scala
@@ -208,6 +208,15 @@ object Http4sAdapter {
       .dimap((req: Request[F]) => req.mapK(toTask))((res: Response[Task]) => res.mapK(toF))
   }
 
+  private def wrapApp[F[_]: Effect](app: HttpApp[Task])(implicit runtime: Runtime[Any]): HttpApp[F] = {
+    val toF: Task ~> F    = λ[Task ~> F](_.toIO.to[F])
+    val toTask: F ~> Task = λ[F ~> Task](_.toIO.to[Task])
+
+    app
+      .mapK(toF)
+      .dimap((req: Request[F]) => req.mapK(toTask))((res: Response[Task]) => res.mapK(toF))
+  }
+
   def makeWebSocketServiceF[F[_], Q, M, S, E](
     interpreter: GraphQL[Any, Q, M, S, E]
   )(implicit F: Effect[F], runtime: Runtime[Any]): HttpRoutes[F] =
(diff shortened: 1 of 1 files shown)


http4s/src/main/scala/caliban/Http4sAdapter.scala
```diff
@@ -95,6 +95,19 @@ object Http4sAdapter {
     }
   }
 
+  def executeRequest[R0, R, Q, M, S, E](interpreter: GraphQL[R, Q, M, S, E], provideEnv: R0 => R): HttpApp[RIO[R0, *]] =
+    Kleisli { req =>
+      object dsl extends Http4sDsl[RIO[R0, *]]
+      import dsl._
+      for {
+        query <- req.attemptAs[GraphQLRequest].value.absolve
```
ghostdogpr: Can you extract this block into a function so that it's not repeated here and in `makeRestService`?
```scala
query <- req.attemptAs[GraphQLRequest].value.absolve
result <- execute(interpreter, query)
                   .foldCause(cause => GraphQLResponse(NullValue, cause.defects).asJson, _.asJson)
```

ghostdogpr:
Can you also add a variant `executeRequestF` for cats-effect users? Similar to `makeRestServiceF`.

fokot:
Sure, I added `executeRequestF`

http4s/src/main/scala/caliban/Http4sAdapter.scala
fokot: 👍

fokot: I looks like good idea but in executeRequest I work with Request/Response with environment R0 and executor is in environment R, so it would mean doing two `mapK` which I do not like much and I find it now more simple. Or I can to common method `provideSome` and send `indentity` from `makeRestService` which also not nice. What do you think?

ghostdogpr: Oh I see what you mean. I'd prefer to avoid an unnecessary `provideSome` for general case. How about just moving `execute(interpreter, query).foldCause(cause => GraphQLResponse(NullValue, cause.defects).asJson, _.asJson)` to a function?

ghostdogpr:
Thanks!

PR #102, opened 2019-11-29
Move circe codecs to the core module

rtimush:
Circe codecs for `GraphQLRequest` and `GraphQLResponse` are generally useful, so I moved them out of the http4s integration module. I also made `GraphQLRequest` public.

The pull request uses the trick described in https://blog.7mind.io/no-more-orphans.html — circe dependency is marked as optional, so you don't need circe to use caliban, but if you have it, the typeclass instances will be available without additional imports.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -57,13 +57,14 @@ lazy val core = crossProject(JSPlatform, JVMPlatform)
   .settings(
     testFrameworks := Seq(new TestFramework("zio.test.sbt.ZTestFramework")),
     libraryDependencies ++= Seq(
-      "com.lihaoyi"    %%% "fastparse"    % "2.1.3",
-      "com.propensive" %%% "magnolia"     % "0.12.2",
-      "com.propensive" %%% "mercator"     % "0.2.1",
-      "dev.zio"        %%% "zio"          % "1.0.0-RC17",
-      "dev.zio"        %%% "zio-streams"  % "1.0.0-RC17",
-      "dev.zio"        %%% "zio-test"     % "1.0.0-RC17" % "test",
-      "dev.zio"        %%% "zio-test-sbt" % "1.0.0-RC17" % "test",
+      "com.lihaoyi"    %%% "fastparse"        % "2.1.3",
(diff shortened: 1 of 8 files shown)


core/src/main/scala/caliban/GraphQLRequest.scala
```diff
@@ -0,0 +1,25 @@
...
+)
+
+object GraphQLRequest {
+  implicit def circeDecoder[F[_]: IsCirceDecoder]: F[GraphQLRequest] =
+    GraphQLRequestCirce.graphQLRequestDecoder.asInstanceOf[F[GraphQLRequest]]
+}
+
+private object GraphQLRequestCirce {
+  import io.circe._
+  import io.circe.derivation._
```
ghostdogpr: Since we are using circe-derivation only for this single (and simple) type, how about deriving it manually and only depend on circe-core?
Because not everyone uses circe-derivation, some people use circe-generic, others circe-magnolia. That way we would keep the dependency minimal.

build.sbt
```diff
@@ -103,7 +104,6 @@ lazy val http4s = project
       "org.http4s"    %% "http4s-circe"        % "0.21.0-M5",
       "org.http4s"    %% "http4s-blaze-server" % "0.21.0-M5",
       "io.circe"      %% "circe-parser"        % "0.12.3",
-      "io.circe"      %% "circe-derivation"    % "0.12.0-M7",
```
ghostdogpr: If we reduce the dependency to circe-core as I suggested in the other comment, this is good. But otherwise, I think we should keep that one so that people who use caliban-http4s don't need to add the dependency to circe-derivation themselves.

core/src/main/scala/caliban/GraphQLRequest.scala
rtimush: I think codecs created by circe-derivation macros depend on the circe core only, so people won't need circe-derivation as a dependency. Although, to be on the safe side I can indeed implement it manually.

ghostdogpr: Oh yeah, you must be right! All good then.

PR #106, opened 2019-12-02
graphiql

fokot:
Added graphiql to the example. It is very useful for trying out things

### DIFF
--- a/examples/src/main/scala/caliban/ExampleApp.scala
+++ b/examples/src/main/scala/caliban/ExampleApp.scala
@@ -4,8 +4,9 @@ import caliban.ExampleData._
 import caliban.GraphQL._
 import caliban.schema.Annotations.{GQLDeprecated, GQLDescription}
 import caliban.schema.GenericSchema
-import cats.data.{Kleisli, OptionT}
-import org.http4s.{EntityEncoder, HttpApp, HttpRoutes, Response, Status}
+import cats.data.Kleisli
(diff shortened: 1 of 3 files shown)


examples/src/main/scala/caliban/ExampleApp.scala
```diff
@@ -48,9 +57,10 @@ object ExampleApp extends CatsApp with GenericSchema[Console with Clock] {
       _ <- BlazeServerBuilder[ExampleTask]
             .bindHttp(8088, "localhost")
             .withHttpApp(
-              Router(
+              Router[ExampleTask](
                 "/api/graphql" -> CORS(Http4sAdapter.makeRestService(interpreter)),
-                "/ws/graphql"  -> CORS(Http4sAdapter.makeWebSocketService(interpreter))
+                "/ws/graphql"  -> CORS(Http4sAdapter.makeWebSocketService(interpreter)),
+                "/graphiql" -> Kleisli.liftF(OptionT.liftF(staticResource("graphiql.html")))
```
ghostdogpr: This code can be a bit simpler no? In https://http4s.org/v0.21/static/ it seems we can simply do something like `StaticFile.fromResource("/" + file, blocker, Some(request)).getOrElseF(NotFound())`

fokot: Fixed

Issue #72, opened 2019-11-08
Akka HTTP Adapter


ghostdogpr:
Add a new module similar to `caliban-http4s` named `caliban-akka-http` that exposes a REST and a WebSocket API for a given GraphQL interpreter.

If there is common code between the http4s and the akka one, it should be factored out.

I haven't used Akka in a while, so external contributors are welcome to pick this one up!

EDIT: HTTP is done, only WebSocket is missing

jona7o:
I am currently work on a solution with an example. 
I haven't use ZIO in a production project, so i hope i'll use the best practises in my PR later this day.
If not, please give me some tipps.

ghostdogpr:
@jona7o that would be awesome!

jona7o:
I have finished the rest part and started with the ws part. 
This week my time is running out so i created a merge request. Perhaps someone else will finish the ws part until i have time for more contribution.

jona7o:
Short hint for the faq:

if someone tries using caliban with lagom and not with akka the solution will look like this:

service-api:
`
def graphqlApi: ServiceCall[GraphQLRequest, GraphQLResponse[CalibanError]]

  override def descriptor: Descriptor = {
    import Service._
    named("graphql").
      withCalls(
        restCall(Method.POST, "/api/graphql", graphqlApi _)
      )
      .withAutoAcl(true)
      .withAcls(
        ServiceAcl.forMethodAndPathRegex(Method.OPTIONS, "/api/.*")
      )
  }
`

service-impl:
`
override def gqlApi(): ServiceCall[GraphQLRequest, GraphQLResponse[CalibanError]] = { graphQLRequest =>
    runtime
      .unsafeRunToFuture(
        interpreter
          .execute(graphQLRequest.query,
            graphQLRequest.operationName,
            graphQLRequest.variables
              .getOrElse(Map())))
      .future
  }
`
finally the DefaultRuntime can be injected while the Application is created by the LagomApplicationLoader

`
 implicit val gatewayZIORuntime = new DefaultRuntime {}

  override lazy val lagomServer: LagomServer = serverFor[graphqlGatewayService](wire[GraphQLGatewayServiceImpl])
`

phderome:
I'd like to volunteer to pursue this for approximately the period of 1/2 of February.

PR #109, opened 2019-12-04
#72 akka http without WS

jona7o:
I have added an akka http adapter like described in #72 

The good news: The rest adapter is complete and i have also added a akka http sample

The bad news: Websocket is just implemented as ping/pong message. So you can establish a websocket conenction via akka-http and the websocket stream will result with the input message data. 

Perhaps i should merge it to a seperate akka-http branch and not the master branch?


ghostdogpr:
Yeah in order to get the REST part merged first, can you remove all the WebSocket-related stuff? It can be added back in another PR when it’s ready. 

I’ll do a complete review tomorrow 👍

build.sbt
```diff
@@ -112,11 +112,32 @@ lazy val http4s = project
   )
   .dependsOn(coreJVM)
 
+lazy val akkaHttp = project
+  .in(file("akka-http"))
+  .settings(name := "caliban-akka-http")
+  .settings(commonSettings)
+  .settings(
+    libraryDependencies ++= Seq(
+      "dev.zio"           %% "zio-interop-cats" % "2.0.0.0-RC10",
```
ghostdogpr: I think the dependencies to `zio-interop-cats` and `cats-effect` are not necessary for this adapter.

How about `circe-generic`? caliban core contains instances of circe `Encoder` and `Decoder` already so we shouldn't need it?

build.sbt
```diff
@@ -112,11 +112,32 @@ lazy val http4s = project
   )
   .dependsOn(coreJVM)
 
+lazy val akkaHttp = project
```
ghostdogpr: Can you also update the config.yml file in .circleci to compile this project? (same as the http4s one)

examples/src/main/scala/caliban/akkahttp/ExampleApp.scala
```diff
@@ -0,0 +1,56 @@
...
+import scala.io.StdIn
+
+object ExampleApp extends App {
+
+  implicit val system = ActorSystem()
+  implicit val materializer = ActorMaterializer()
+  implicit val executionContext = system.dispatcher
+  implicit val defaultRuntime = new DefaultRuntime {}
+
+  case class Character(name: String)
```
ghostdogpr: How about reusing `ExampleData` and `ExampleService` so that this example is easily comparable to the http4s one?

Also you could split the current `ExampleApp` to move the http4s part into an `http4s` folder (just like you did for akka) and keep the common part in the shared folder (Queries, etc).

examples/src/main/scala/caliban/akkahttp/ExampleApp.scala
```diff
@@ -0,0 +1,56 @@
...
+   * -H 'Host: localhost:8080' \
+   * -H 'Content-Type: application/json' \
+   * -d '{
+   * "query": "query { characters { name }}"
+   * }'
+   */
+
+  val route =
+      path("api" / "graphql") {
+        AkkaHttpAdapter.makeRestService(interpreter)
```
ghostdogpr: Very easy to use! 👏

build.sbt
jona7o: resolved in 3336d63

examples/src/main/scala/caliban/akkahttp/ExampleApp.scala
jona7o: resolved in 0cbe7a7 and 430c3c9

build.sbt
jona7o: Resolved in c1137ac

jona7o:
Given suggestions resolved.

akka-http/src/main/scala/caliban/AkkaHttpAdapter.scala
```diff
@@ -0,0 +1,45 @@
...
+
+import scala.concurrent.ExecutionContext.Implicits.global
+import scala.concurrent.duration._
+import scala.concurrent.{ExecutionContext, Future}
+import scala.util.{Failure, Success}
+
+object AkkaHttpAdapter extends FailFastCirceSupport {
+
+
+  def execute[R, Q, M, S, E](
```
ghostdogpr: Can you make this one private?

examples/src/main/scala/caliban/akkahttp/ExampleApp.scala
```diff
@@ -0,0 +1,73 @@
...
+      }
```
ghostdogpr: The http4s example now includes graphiql. Would that be possible to have it here too?

ghostdogpr:
One more thing: I renamed `makeRestService` to `makeHttpService` in the http4s case, because using "REST" was causing some confusion. Can you apply the same naming here? Thanks!

jona7o:
Resolved and renamed everything.

ghostdogpr:
Awesome! Merging this one 🙏

Issue #47, opened 2019-10-19
Generate code from a GraphQL schema


ghostdogpr:
It would be nice if, given a GraphQL schema, we could generate (with an sbt command?) the corresponding case classes and sealed traits.

ghostdogpr:
Base work was done in #118 . I will create separate issues for remaining work.

PR #117, opened 2019-12-12
Generate scala code from a GraphQL schema

igorfialko:
https://github.com/ghostdogpr/caliban/issues/47


igorfialko:
Creating another one

PR #126, opened 2019-12-16
Map java.util.UUID to ID instead of String

mriceron:
The scope of the PR is to map `java.util.UUID` to `ID` instead of `String`.

I'm not sure which is the best place to add a test for this since I just took a quick look at the project, I'm open to suggestions.

### DIFF
--- a/core/src/test/scala/caliban/schema/SchemaSpec.scala
+++ b/core/src/test/scala/caliban/schema/SchemaSpec.scala
@@ -6,7 +6,7 @@ import scala.concurrent.Future
 import caliban.introspection.adt.{ __DeprecatedArgs, __Type, __TypeKind }
 import caliban.schema.SchemaSpecUtils._
 import zio.test.Assertion._
-import zio.test.{ test, _ }
+import zio.test._
 import zio.{ Task, UIO }
 
 object SchemaSpec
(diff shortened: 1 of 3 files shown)


ghostdogpr:
Awesome!

There was a test in `ExecutionSpec` and it still passes since the result is the same as before. If you want to check the returned type explicitly, you could add one here: https://github.com/ghostdogpr/caliban/blob/master/core/src/test/scala/caliban/schema/SchemaSpec.scala

core/src/test/scala/caliban/schema/SchemaSpec.scala
```diff
@@ -1,10 +1,12 @@
 package caliban.schema
 
+import java.util.UUID
+
 import scala.concurrent.Future
 import caliban.introspection.adt.{ __DeprecatedArgs, __Type, __TypeKind }
 import caliban.schema.SchemaSpecUtils._
 import zio.test.Assertion._
-import zio.test._
+import zio.test.{ test, _ }
```
ghostdogpr: This one isn't needed, is it?

mriceron: Thanks, didn't notice it 👍

ghostdogpr:
Thanks! I’ll merge it tomorrow before releasing a new version.

mriceron:
Thanks to you for the library and the support! Looking forward to the release :D

PR #128, opened 2019-12-18
Support for combining GraphQL objects

ghostdogpr:
This PR changes a few things:
- `GraphQL[R, Q, M, S, E]` is simplified into `GraphQL[R]`.
- `GraphQLInterpreter[R, E]` is introduced. It's a wrapper around `GraphQL` that allows changing the R, the E and running any wrapper around execution. It's obtained by just calling `.interpreter` on the `GraphQL` object. The `execute` method is available on this object.
- It is now possible to combine `GraphQL` objects with `|+|`. This allows splitting a large API into smaller chunks and work around Scala limitation of 22 fields.

Example:
```scala
val api1 = graphQL(...)
val api2 = graphQL(...)
val api = api1 |+| api2
```

### DIFF
--- a/core/src/main/scala/caliban/GraphQL.scala
+++ b/core/src/main/scala/caliban/GraphQL.scala
@@ -86,15 +86,20 @@ trait GraphQL[-R] { self =>
 
   /**
    * Merges this GraphQL API with another GraphQL API.
-   * In case of conflicts (same field declared on both APIs), fields from the this API will be favored.
+   * In case of conflicts (same field declared on both APIs), fields from `that` API will be used.
    * @param that another GraphQL API object
    * @return a new GraphQL API
    */
-  final def |+|[R1 <: R](that: GraphQL[R1]): GraphQL[R1] =
+  final def combine[R1 <: R](that: GraphQL[R1]): GraphQL[R1] =
     new GraphQL[R1] {
       override val schema: RootSchema[R1]                  = self.schema |+| that.schema
       override val queryAnalyzers: List[QueryAnalyzer[R1]] = self.queryAnalyzers ++ that.queryAnalyzers
     }
+
+  /**
+   * Operator alias for `combine`.
+   */
+  final def |+|[R1 <: R](that: GraphQL[R1]): GraphQL[R1] = combine(that)
 
   /**
    * Renames the root queries, mutations and subscriptions objects.
--- a/core/src/main/scala/caliban/GraphQL.scala
+++ b/core/src/main/scala/caliban/GraphQL.scala
@@ -4,7 +4,6 @@ import caliban.Rendering.renderTypes
 import caliban.execution.Executor
 import caliban.execution.QueryAnalyzer.QueryAnalyzer
 import caliban.introspection.Introspector
-import caliban.introspection.adt.__Introspection
 import caliban.parsing.Parser
 import caliban.schema.RootSchema.Operation
 import caliban.schema._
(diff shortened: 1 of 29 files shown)


core/src/main/scala/caliban/GraphQL.scala
```diff
@@ -4,111 +4,134 @@ import caliban.Rendering.renderTypes
...
-        queryAnalyzers: List[QueryAnalyzer[R1]]
-      ): URIO[R1, GraphQLResponse[E]] =
-        self.execute(query, operationName, variables, skipValidation, queryAnalyzer :: queryAnalyzers)
-      override def render: String = self.render
-    }
+  final def rename(
+    queriesName: Option[String] = None,
+    mutationsName: Option[String] = None,
+    subscriptionsName: Option[String] = None
+  ): GraphQL[R] = new GraphQL[R] {
```
paulpdaniels: This is neat, but I'm wondering if it could be potentially split off into something more general purpose traversing transformer. Of course it doesn't stop it from going in now and updating it later. But doing something like: https://www.apollographql.com/docs/apollo-server/features/schema-transforms/ might be compelling enough to hold off while fleshing this out.

ghostdogpr: I agree with you. Currently it's not doable easily because the type definition and the resolver are 2 different objects (`__Type` and `Step`) and any change in one of them should also be applied to the other one (e.g. rename or remove fields). Both are trees but they are quite different: `__Type` is entirely known, while `Step` might contain effects that return other steps, etc. I think it requires deeper changes to make this possible.

This `rename` here is more a "quick fix" for people who are not happy with the root names that are used when you combine 2 APIs. It only affect documentation since queries don't include root type names.

I definitely want to keep thinking about a way to support that, so I'm gonna create an issue for it. I'd love to hear how you use the schema transform feature btw, here or on discord. Having ideas about concrete use cases would help.

core/src/main/scala/caliban/GraphQL.scala
```diff
@@ -4,111 +4,134 @@ import caliban.Rendering.renderTypes
...
-      override def execute[R1 <: R2](
-        query: String,
-        operationName: Option[String],
-        variables: Map[String, InputValue],
-        skipValidation: Boolean,
-        queryAnalyzers: List[QueryAnalyzer[R1]] = Nil
-      ): URIO[R2, GraphQLResponse[E2]] = f(self.execute(query, operationName, variables, skipValidation))
-      override def render: String      = self.render
-    }
+  final def |+|[R1 <: R](that: GraphQL[R1]): GraphQL[R1] = combine(that)
```
jdegoes: Note this would be called `<>` in ZIO.

core/src/main/scala/caliban/GraphQL.scala
```diff
@@ -4,111 +4,134 @@ import caliban.Rendering.renderTypes
...
 /**
- * A `GraphQL[R, Q, M, S, E]` represents a GraphQL interpreter for a query type `Q`, a mutation type `M`
- * and a subscription type `S`, whose execution requires a ZIO environment of type `R` and can fail with an `E`.
+ * A `GraphQL[-R]` represents a GraphQL API whose execution requires a ZIO environment of type `R`.
  *
  * It is intended to be created only once, typically when you start your server.
  * The introspection schema will be generated when this class is instantiated.
  */
-trait GraphQL[-R, -Q, -M, -S, +E] { self =>
+trait GraphQL[-R] { self =>
```
jdegoes: Superb simplification. 👍

core/src/main/scala/caliban/introspection/adt/__Type.scala
```diff
@@ -10,4 +10,19 @@ case class __Type(
   enumValues: __DeprecatedArgs => Option[List[__EnumValue]] = _ => None,
   inputFields: Option[List[__InputValue]] = None,
   ofType: Option[__Type] = None
-)
+) {
+  def |+|(that: __Type): __Type = __Type(
```
jdegoes: This too would be called `<>` in ZIO (essentially `<+>` but without preserving the distinction between the sides). Not sure if it's a good idea here or if it's better to go with Semigroup-inspired `|+|`.

PR #141, opened 2019-12-28
Add OptionT and fs2.Stream Schema

UkonnRa:
**Target and solved questions**

Add the Schema for `cats.data.OptionT` and and `fs2.Stream` so that users can use these two data types in Query, Mutation and Subscription directly.

**Problems**

* Because I cannot find a way to convert from `fs2.Stream` to `ZStream` directly, so I just convert `fs2.Stream` to a Vector and then change it to `ZStream`. I have no idea if there will be any performance issue

### DIFF
--- a/interop/cats/src/main/scala/caliban/interop/cats/implicits/package.scala
+++ b/interop/cats/src/main/scala/caliban/interop/cats/implicits/package.scala
@@ -1,9 +1,16 @@
 package caliban.interop.cats
 
-import caliban.schema.Schema
+import caliban.introspection.adt.__Type
+import caliban.schema.Step.QueryStep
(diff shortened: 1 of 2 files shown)


interop/cats/src/main/scala/caliban/interop/cats/implicits/package.scala
```diff
@@ -1,9 +1,16 @@
 package caliban.interop.cats
```
phderome: this file does not compile on CI. You probably should investigate?

ghostdogpr:
@UkonnRa let me know if you need any help getting this one in. Closing for now.

PR #146, opened 2020-01-04
rename ObjectValue to StreamValue in text.

phderome:
`ObjectValue` wraps up a `List` and current doc says `ObjectValue` whereas here we are talking about `ZStream`.

### DIFF
--- a/vuepress/docs/docs/README.md
+++ b/vuepress/docs/docs/README.md
@@ -146,4 +146,4 @@ val subscriptions = Subscriptions(???)
 val api = graphQL(RootResolver(queries, mutations, subscriptions))
 ```
 
-All the fields of the subscription root case class MUST return `ZStream` or `? => ZStream` objects. When a subscription request is received, an output stream of `ResponseValue` will be returned wrapped in an `StreamValue`.
+All the fields of the subscription root case class MUST return `ZStream` or `? => ZStream` objects. When a subscription request is received, an output stream of `ResponseValue` (a `StreamValue`) will be returned wrapped inside an `ObjectValue `.
--- a/vuepress/docs/docs/README.md
+++ b/vuepress/docs/docs/README.md
@@ -146,4 +146,4 @@ val subscriptions = Subscriptions(???)
 val api = graphQL(RootResolver(queries, mutations, subscriptions))
 ```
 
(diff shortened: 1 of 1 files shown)


ghostdogpr:
It’s actually an `ObjectValue` wrapping a `StreamValue`. The root result can only be an `ObjectValue`.

phderome:
> It’s actually an `ObjectValue` wrapping a `StreamValue`. The root result can only be an `ObjectValue`.

so it's subtle. Not sure if the current wording needs to change then.

vuepress/docs/docs/README.md
```diff
@@ -146,4 +146,4 @@ val subscriptions = Subscriptions(???)
 val api = graphQL(RootResolver(queries, mutations, subscriptions))
 ```
 
-All the fields of the subscription root case class MUST return `ZStream` or `? => ZStream` objects. When a subscription request is received, an output stream of `ResponseValue` will be returned wrapped in an `ObjectValue`.
+All the fields of the subscription root case class MUST return `ZStream` or `? => ZStream` objects. When a subscription request is received, an output stream of `ResponseValue` will be returned wrapped in an `StreamValue`.
```
ghostdogpr:
```suggestion
All the fields of the subscription root case class MUST return `ZStream` or `? => ZStream` objects. When a subscription request is received, an output stream of `ResponseValue` (a `StreamValue`) will be returned wrapped inside an `ObjectValue `.
```

vuepress/docs/docs/README.md
```diff
@@ -146,4 +146,4 @@ val subscriptions = Subscriptions(???)
 val api = graphQL(RootResolver(queries, mutations, subscriptions))
 ```
 
-All the fields of the subscription root case class MUST return `ZStream` or `? => ZStream` objects. When a subscription request is received, an output stream of `ResponseValue` will be returned wrapped in an `ObjectValue`.
+All the fields of the subscription root case class MUST return `ZStream` or `? => ZStream` objects. When a subscription request is received, an output stream of `ResponseValue` (a `StreamValue`) will be returned wrapped inside an `ObjectValue `.
```
ghostdogpr: Oops, I inserted a spurious space before the backtick. Can you remove it?

phderome: done.

PR #147, opened 2020-01-04
Update README.md for inconsistent usage of word Character

phderome:
name at lines 37 and 47 should be the same, hence both `Character`.

I am less confident about my proposed change at lines 94-98. Given API at lines 72-75, we may request all characters using plural form `characters` without name or we may request a single character in singular form `character` and specify additionally the name as parameter. The current form seems to be a mix of the two forms instead of choosing one correctly. If we want to specify a single character I'd suggest changing `name` to `Pierre` or some common given name.

### DIFF
--- a/vuepress/docs/docs/README.md
+++ b/vuepress/docs/docs/README.md
@@ -93,7 +93,9 @@ Use `ResponseValue#toString` to get the JSON representation of the result.
 ```scala
 val query = """
   {
-    characters 
+    characters {
+       name
+    }  
   }"""
 
 for {
--- a/vuepress/docs/docs/README.md
+++ b/vuepress/docs/docs/README.md
@@ -44,7 +44,7 @@ Let's create a case class named `Queries` that will represent our API, with 2 fi
 
 ```scala
 // schema
-case class CharacterName(name: String)
+case class Character(name: String)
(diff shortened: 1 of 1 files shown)


vuepress/docs/docs/README.md
```diff
@@ -44,7 +44,7 @@ Let's create a case class named `Queries` that will represent our API, with 2 fi
 
 ```scala
 // schema
-case class CharacterName(name: String)
+case class Character(name: String)
```
ghostdogpr: `CharacterName` was not meant to be the same thing as `Character`, it was only a wrapper for the name argument. But because I simplified `Character` so much, they are identical. I suggest instead to keep `CharacterName` and add other fields in `Character`.

vuepress/docs/docs/README.md
```diff
@@ -93,9 +93,7 @@ Use `ResponseValue#toString` to get the JSON representation of the result.
 ```scala
 val query = """
   {
-    characters {
-      name
-    }
+    characters
```
ghostdogpr: Why this change? This wouldn't be valid GraphQL.

phderome: Because I didn't understand what I was doing. Sorry. I'll pay more attention next time.

PR #152, opened 2020-01-05
replace map.getOrElse with fold when considering options

phderome:
(but avoided extensions such http4s and akkahttp, but could volunteer to do similarly there)

I did a few other things along the way.

Fusing 2 consecutive maps (Field.loop), which avoids two traversals

Occasionally add temporary variables typically for String formatting to ease legibility

Broke down some long lines into multiple lines (quite often!)

Refactored a bit `resolveVariables` which was a long one liner, similarly `schema.dispatch.toType`

Took out evaluation of strings outside of one-liner formatting considerations in `Rendering.renderTypes` to ease legibility

Broke down one liner of `IntValue.apply` that has `Try` and `orElse `in it and made it similar to other code in project

### DIFF
--- a/core/src/main/scala/caliban/Rendering.scala
+++ b/core/src/main/scala/caliban/Rendering.scala
@@ -73,10 +73,12 @@ object Rendering {
   private def isBuiltinScalar(name: String): Boolean =
     name == "Int" || name == "Float" || name == "String" || name == "Boolean" || name == "ID"
 
-  private[caliban] def renderTypeName(fieldType: __Type): String =
+  private[caliban] def renderTypeName(fieldType: __Type): String = {
+    lazy val renderedTypeName = fieldType.ofType.fold("null")(renderTypeName)
     fieldType.kind match {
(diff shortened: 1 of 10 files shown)


core/src/main/scala/caliban/Rendering.scala
```diff
@@ -70,8 +75,8 @@ object Rendering {
 
   private[caliban] def renderTypeName(fieldType: __Type): String =
     fieldType.kind match {
-      case __TypeKind.NON_NULL => s"${fieldType.ofType.map(renderTypeName).getOrElse("null")}!"
-      case __TypeKind.LIST     => s"[${fieldType.ofType.map(renderTypeName).getOrElse("null")}]"
+      case __TypeKind.NON_NULL => s"${fieldType.ofType.fold("null")(renderTypeName)}!"
+      case __TypeKind.LIST     => s"[${fieldType.ofType.fold("null")(renderTypeName)}]"
```
phderome: we should merge the two cases into one here with a or of the two possible case values yielding a common outcome.

PR #154, opened 2020-01-05
final edits to README.md

phderome:
1) require is used as present 3rd pers. sing, rather than infinitive form.

2) change type discussed in `run` to be `iterable` rather than `list` to reflect code of `DataSource`

3) Put code quote for context `R` near `provide` and `provideSome` so that R stands out.

However, regarding item 2, note the Building Large DataSource code sample for `def run` has a `MatchError` latent problem in that parameter is `Iterable` and is being matched with non-empty list of size 1 using Nil that is specific to lists.


phderome:
> 👍

My last paragraph was wrong in description. I failed to notice the conversion of Iterable to List via a call to `toList`. Nothing wrong with the example.

PR #160, opened 2020-01-07
Improvements to ZQuery

adamgfraser:
Various improvements to `ZQuery` including:

- Adds new data source constructors `fromFunctionBatchedWith` and `fromFunctionBatchedWithM` that make it easier to construct batched data sources that don't return results in the same order as requests by providing a function to associate results with requests as suggested by @fokot.
- Adds `partitionM` and `partitionMPar` methods on `ZQuery` that allow for performing a collection of queries and getting back a collection of all successes along with a collection of all failures
- Adds a new `Described` data type to clean up and clarify the contract around needing to provide names for certain combinators such as `provide` and `provideSome`.
- Miscellaneous cleanup including adding `CanFail` and `NeedsEnv` to appropriate combinators and removing unnecessary `final` modifiers.

One thing I would appreciate feedback on is how much users are working with data sources in the environment. I am thinking we could simplify `DataSource` somewhat by not using the module pattern and just having all the methods on `DataSource` versus `DataSource.Service`. That would mean that users could still put the `DataSource` in the environment but it would not come "baked in" like it is today. The nice thing would be this would mean you wouldn't have to write `DataSource.Service` everywhere but if people are using it as an environmental effect then we should leave as is.

### DIFF
--- a/core/src/main/scala/zquery/ZQuery.scala
+++ b/core/src/main/scala/zquery/ZQuery.scala
@@ -1,6 +1,6 @@
 package zquery
 
-import zio.{ Cause, Ref, ZIO }
+import zio.{ CanFail, Cause, NeedsEnv, Ref, ZIO }
 
 /**
  * A `ZQuery[R, E, A]` is a purely functional description of an effectual query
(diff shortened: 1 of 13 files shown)


core/src/main/scala/zquery/ZQuery.scala
```diff
@@ -83,6 +83,14 @@ sealed trait ZQuery[-R, +E, +A] { self =>
   final def >>=[R1 <: R, E1 >: E, B](f: A => ZQuery[R1, E1, B]): ZQuery[R1, E1, B] =
     flatMap(f)
 
+  /**
+   * Returns a query whose failure and success have been lifted into an
+   * `Either`. The resulting query cannot fail, because the failure case has
+   * been exposed as part of the `Either` success case.
+   */
+  final def either(implicit ev: CanFail[E]): ZQuery[R, Nothing, Either[E, A]] =
```
ghostdogpr: Nice! It can be used in `Executor` line 145

jdegoes:
Agreed on not using environment for `DataSource`. It's better an ordinary value. 

ZIO Config used environment for its "ConfigSource" for a while, but changing that to values made everything much more natural and compositional, as well as more obvious for users.

adamgfraser:
Updated to not use the environment for `DataSource` and add additional error handling combinators `foldM` and `bimap`.

PR #165, opened 2020-01-13
Wrappers + Apollo Tracing

### LINKED ISSUE #99: Ability to wrap field execution
Just like we can wrap query execution with `wrapExecutionWith`, it would be useful to have a way to wrap each field execution.

Possible use cases: collect field-level metrics and possibly send them somewhere, field-level authorizations.

### LINKED ISSUE #133: Add module for apollo-tracing support
See specs here: https://github.com/apollographql/apollo-tracing

### LINKED ISSUE #157: Errors: add path
Add support for the `path` data for each error as defined in the [GraphQL spec](https://graphql.github.io/graphql-spec/June2018/#sec-Errors):
```json
{
  "errors": [
    {
      "message": "Name for character with ID 1002 could not be fetched.",
      "locations": [ { "line": 6, "column": 7 } ],
      "path": [ "hero", "heroFriends", 1, "name" ],
      "extensions": {
        "code": "CAN_NOT_FETCH_BY_ID",
        "timestamp": "Fri Feb 9 14:33:09 UTC 2018"
      }
    }
  ]
}
```

See initial discussion in https://github.com/ghostdogpr/caliban/issues/140

ghostdogpr:
Fixes #99 
Fixes #133 
Fixes #157 

You can use `GraphQL#withWrapper` to register a function that will wrap the execution of the query processing (or only a part of it).

There are 5 types of wrappers:
 - `OverallWrapper` to wrap the whole query processing
 - `ParsingWrapper` to wrap the query parsing only
 - `ValidationWrapper` to wrap the query validation only
 - `ExecutionWrapper` to wrap the query execution only
 - `FieldWrapper` to wrap each field execution

Each one requires a function that takes an `IO` or `ZQuery` computation together with some contextual information (e.g. the query string) and should return another computation.

Let's see how to implement a wrapper that times out the whole query if its processing takes longer that 1 minute.

```scala
val wrapper = OverallWrapper {
  case (io, query) =>
    io.timeout(1 minute)
      .map(
        _.getOrElse(
          GraphQLResponse(
            NullValue,
            List(ExecutionError(s"Query was interrupted after timeout of ${duration.render}:\n$query"))
          )
        )
      )
}

val api = graphQL(...).withWrapper(wrapper)
```

Caliban comes with a few pre-made wrappers:
- `caliban.wrappers.Wrappers.logSlowQueries` returns a wrapper that logs slow queries
- `caliban.wrappers.Wrappers.timeout` returns a wrapper that fails queries taking more than a specified time
- `caliban.wrappers.ApolloTracing.apolloTracing` returns a wrapper that adds tracing data into the `extensions` field of each response following [Apollo Tracing](https://github.com/apollographql/apollo-tracing) format.

They can be used like this:
```scala
val api =
  apolloTracing( // note: this constructor is effectful
    logSlowQueries(500 millis)(
      timeout(3 seconds)(  
        graphQL(...)
      )
    )
  )
```

### DIFF
--- a/core/src/main/scala/caliban/execution/Executor.scala
+++ b/core/src/main/scala/caliban/execution/Executor.scala
@@ -135,17 +135,11 @@ object Executor {
           reduceObject(items, fieldWrappers)
         case QueryStep(inner) =>
           ReducedStep.QueryStep(
-            inner.bimap(
-              GenericSchema.effectfulExecutionError(currentField.name, path, _),
-              reduceStep(_, currentField, arguments, path)
-            )
+            inner.bimap(GenericSchema.effectfulExecutionError(path, _), reduceStep(_, currentField, arguments, path))
           )
         case StreamStep(stream) =>
           ReducedStep.StreamStep(
-            stream.bimap(
-              GenericSchema.effectfulExecutionError(currentField.name, path, _),
-              reduceStep(_, currentField, arguments, path)
-            )
+            stream.bimap(GenericSchema.effectfulExecutionError(path, _), reduceStep(_, currentField, arguments, path))
           )
       }
 
@@ -254,8 +248,8 @@ object Executor {
     fieldWrappers: List[FieldWrapper[R]]
   ): ReducedStep[R] =
     if (!fieldWrappers.exists(_.wrapPureValues) && items.map(_._2).forall(_.isInstanceOf[PureStep]))
-      PureStep(ObjectValue(items.asInstanceOf[List[(String, PureStep)]].map {
-        case (k, v) => k -> v.value
+      PureStep(ObjectValue(items.asInstanceOf[List[(String, PureStep, FieldInfo)]].map {
+        case (k, v, _) => (k, v.value)
       }))
     else ReducedStep.ObjectStep(items)
 
--- a/core/src/main/scala/caliban/execution/Executor.scala
+++ b/core/src/main/scala/caliban/execution/Executor.scala
@@ -10,9 +10,10 @@ import caliban.parsing.adt.OperationType.{ Mutation, Query, Subscription }
 import caliban.parsing.adt._
 import caliban.schema.Step._
 import caliban.schema.{ GenericSchema, ReducedStep, RootSchema, Step }
+import caliban.wrappers.Wrapper.FieldWrapper
 import caliban.{ CalibanError, GraphQLResponse, InputValue, ResponseValue }
 import zio._
-import zquery.ZQuery
+import zquery.{ Described, ZQuery }
 
 object Executor {
 
@@ -28,7 +29,8 @@ object Executor {
     schema: RootSchema[R],
     operationName: Option[String] = None,
     variables: Map[String, InputValue] = Map(),
-    queryAnalyzers: List[QueryAnalyzer[R]] = Nil
+    queryAnalyzers: List[QueryAnalyzer[R]] = Nil,
+    fieldWrappers: List[FieldWrapper[R]] = Nil
   ): URIO[R, GraphQLResponse[CalibanError]] = {
     val fragments = document.definitions.collect {
       case fragment: FragmentDefinition => fragment.name -> fragment
@@ -66,7 +68,14 @@ object Executor {
                    .foldLeft(queryAnalyzers)(Field(op.selectionSet, fragments, variables, operationType.opType)) {
                      case (field, analyzer) => analyzer(field)
                    }
-          result <- executePlan(operationType.plan, root, op.variableDefinitions, variables, allowParallelism)
+          result <- executePlan(
+                     operationType.plan,
+                     root,
+                     op.variableDefinitions,
+                     variables,
+                     allowParallelism,
+                     fieldWrappers
+                   )
         } yield result).catchAll(fail)
     }
   }
@@ -79,13 +88,15 @@ object Executor {
     root: Field,
     variableDefinitions: List[VariableDefinition],
     variableValues: Map[String, InputValue],
-    allowParallelism: Boolean
+    allowParallelism: Boolean,
+    fieldWrappers: List[FieldWrapper[R]]
   ): URIO[R, GraphQLResponse[CalibanError]] = {
 
     def reduceStep(
       step: Step[R],
       currentField: Field,
-      arguments: Map[String, InputValue]
+      arguments: Map[String, InputValue],
+      path: List[Either[String, Int]]
     ): ReducedStep[R] =
       step match {
         case s @ PureStep(value) =>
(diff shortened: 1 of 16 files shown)


core/src/main/scala/caliban/wrappers/ApolloTracing.scala
```diff
@@ -0,0 +1,187 @@
...
+    parentType: String = "",
+    fieldName: String = "",
+    returnType: String = "",
+    startOffset: Long = 0,
+    duration: Duration = Duration.Zero
+  ) {
+    def toResponseValue: ResponseValue =
+      ObjectValue(
+        List(
+          "path" -> ListValue((Left(fieldName) :: path).reverse.map {
```
paulpdaniels: Why does it need to be reversed here but not when creating the error?

core/src/main/scala/caliban/wrappers/Wrappers.scala
```diff
@@ -0,0 +1,60 @@
...
+   * @param api a GraphQL API definition
+   */
+  def logSlowQueries[R <: Console with Clock](duration: Duration)(api: GraphQL[R]): GraphQL[R] =
+    api.withWrapper(logSlowQueriesWrapper(duration))
+
+  /**
+   * Returns a wrapper that logs slow queries
+   * @param duration threshold above which queries are considered slow
+   */
+  def logSlowQueriesWrapper(duration: Duration): OverallWrapper[Console with Clock] =
```
paulpdaniels: Nit: Naming should probably be `printSlowQueries`.

core/src/main/scala/caliban/wrappers/ApolloTracing.scala
ghostdogpr: Oops, good catch. I will add a test for the error case.

core/src/main/scala/caliban/wrappers/Wrappers.scala
ghostdogpr: Good point.

core/src/test/scala/caliban/wrappers/WrappersSpec.scala
```diff
@@ -0,0 +1,81 @@
...
+                  }
+                }
+              }""")
+          assertM(
+            interpreter >>= (_.execute(query).map(_.extensions.map(_.toString))),
+            isSome(
+              equalTo(
+                """{"tracing":{"version":1,"startTime":"1970-01-01T00:00:00.000Z","endTime":"1970-01-01T00:00:00.000Z","duration":0,"parsing":{"startOffset":0,"duration":0},"validation":{"startOffset":0,"duration":0},"execution":{"resolvers":[{"path":["hero"],"parentType":"Query","fieldName":"hero","returnType":"Hero!","startOffset":0,"duration":0},{"path":["hero","friends"],"parentType":"Hero","fieldName":"friends","returnType":"[Hero!]!","startOffset":0,"duration":0},{"path":["hero","friends",2,"name"],"parentType":"Hero","fieldName":"name","returnType":"String!","startOffset":0,"duration":0},{"path":["hero","friends",1,"name"],"parentType":"Hero","fieldName":"name","returnType":"String!","startOffset":0,"duration":0},{"path":["hero","friends",0,"name"],"parentType":"Hero","fieldName":"name","returnType":"String!","startOffset":0,"duration":0},{"path":["hero","name"],"parentType":"Hero","fieldName":"name","returnType":"String!","startOffset":0,"duration":0}]}}}"""
```
paulpdaniels: Could we add some (artificial) delays in order to see that all the duration/offset are generated correctly as well?

Maybe just change the `Hero` to take `ZIO[Clock, Nothing, String]` and then pass `ZIO.sleep(1.second) *> ZIO.succeed("Han Solo")`

examples/src/main/scala/caliban/http4s/ExampleApp.scala
```diff
@@ -38,17 +42,23 @@ object ExampleApp extends CatsApp with GenericSchema[Console with Clock] {
...
-              args => service.findCharacter(args.name)
-            ),
-            Mutations(args => service.deleteCharacter(args.name)),
-            Subscriptions(service.deletedEvents)
+  def makeApi(service: ExampleService): UIO[GraphQL[Console with Clock]] =
+    apolloTracing(                  // wrapper for https://github.com/apollographql/apollo-tracing
+      printSlowQueries(500 millis)( // wrapper that logs slow queries
+        timeout(3 seconds)(         // wrapper that fails slow queries
+          maxDepth(30)(             // query analyzer that limit query depth
+            maxFields(200)(         // query analyzer that limit query fields
```
paulpdaniels: Unrelated: But looking at this could probably allow optimization by providing some `BooleanAlgebra` to be able to do: `maxDepth(30) || maxFields(200)`

core/src/main/scala/caliban/wrappers/Wrappers.scala
```diff
@@ -0,0 +1,60 @@
...
+   * @param duration threshold above which queries are considered slow
+   */
+  def printSlowQueriesWrapper(duration: Duration): OverallWrapper[Console with Clock] =
+    OverallWrapper {
+      case (io, query) =>
+        io.timed.flatMap {
+          case (time, res) =>
+            ZIO.when(time > duration)(putStrLn(s"Slow query took ${time.render}:\n$query")).as(res)
+        }
+    }
```
paulpdaniels: Could also make one more to do generic effects on slow query `onSlowQueriesWrapper`:

```scala
  def onSlowQueriesWrapper(duration: Duration)(f: (Duration, String) => UIO[Any]): OverallWrapper[Console with Clock] =
    OverallWrapper {
      case (io, query) =>
        io.timed.flatMap {
          case (time, res) =>
            ZIO.when(time > duration)(f(time, res)).as(res)
        }
    }

```

which would allow for easy creation of logging mechanisms.

ghostdogpr: Good idea, and I can implement `printSlowQueries` using it.

core/src/test/scala/caliban/wrappers/WrappersSpec.scala
ghostdogpr: Good idea!

examples/src/main/scala/caliban/http4s/ExampleApp.scala
ghostdogpr: And/Or operations make sense for the 2 examples here but `QueryAnalyzer` is more general than that and allows any arbitrary effect to run and possibly modify the query, so I'm not sure it makes in a general context.

I like the `@@` notation though, not only is it consistent with ZIO Test, but it also reduce nesting, it's more discoverable and doesn't require me to duplicate every query analyzer / wrapper like I'm doing now. The only problematic one is `apolloTracing` because its construction is effectful so it can't be stacked with others...

ghostdogpr: Maybe I could create a new wrapper type `CombinedWrapper` that takes an optional effectful state and can contain multiple wrappers. That way I'll be able to express `apolloTracing` as a wrapper. Will try that.

core/src/main/scala/caliban/wrappers/ApolloTracing.scala
```diff
@@ -0,0 +1,187 @@
...
+object ApolloTracing {
+
+  /**
+   * Attaches to the given GraphQL API definition a wrapper that adds tracing information to every response
+   * following Apollo Tracing format: https://github.com/apollographql/apollo-tracing.
+   * @param api a GraphQL API definition
+   */
+  def apolloTracing[R](api: GraphQL[R]): UIO[GraphQL[Clock with R]] =
+    FiberRef
+      .make(Tracing())
```
paulpdaniels: Not familiar with how `FiberRef` works, so this may be a dumb question. But why can't this be lifted into an effect?

Couldn't this be:

```scala
UIO.effectSuspendTotal(FiberRef.make(Tracing()))
```

or could make an effect wrapper for `Tracing` as well.

```scala
for {
  tracing <- Tracing.make
  ref        <- FiberRef.make(tracing)
} yield api.withWrapper(apolloTracingOverallWrapper(ref))
```

ghostdogpr: About 1, `FiberRef.make` already returns an effect, no need to use `effectSuspendTotal`.
About 2, I wanted to avoid forcing users having to do that and simply do `@@ apolloTracing` or something similar.

I'm trying with 2 new wrapper types `CombinedWrapper` and `EffectfulWrapper`, so that this logic would be "hidden".

ghostdogpr: apolloTracing would be built this way:
```scala
EffectfulWrapper(
  FiberRef
    .make(Tracing())
    .map(
      ref =>
        apolloTracingOverall(ref) |+|
          apolloTracingParsing(ref) |+|
          apolloTracingValidation(ref) |+|
          apolloTracingField(ref)
    )
)
```
and then we are able to do:
```scala
graphQL(...) @@
  maxFields(200) @@
  maxDepth(30) @@
  timeout(3 seconds) @@
  printSlowQueries(500 millis) @@
  apolloTracing   
```

core/src/main/scala/caliban/execution/Executor.scala
```diff
@@ -101,46 +67,66 @@ object Executor {
...
+        wrappers match {
+          case Nil => query
+          case wrapper :: tail =>
+            ZQuery
+              .environment[R]
+              .flatMap(
+                env =>
+                  wrap(
+                    wrapper
+                      .f(query.provide(Described(env, "Wrapper")), fieldInfo)
```
paulpdaniels: Maybe provide an optional name for the `Wrapper` when creating it? Otherwise won't all wrappers end up being equal to each other?

paulpdaniels: Maybe use the same label syntax:

trait Wrapper[-R] {
  def ??(label: String): Wrapper[-R]
}

ghostdogpr: Not really, the name required by `ZQuery` has to do with caching. `ZQuery` generates a hash to ensure the same request is queried only once and otherwise cached. When we modify the environment with `provide`, the request is the same but we need to change the hash to avoid reading from the same cache. It doesn't need to be different for each wrapper.

@adamgfraser can you confirm?

ghostdogpr:
cc @rtimush FYI, I reimplemented the query analyzers using this more powerful wrapper feature.

core/src/main/scala/caliban/execution/Executor.scala
adamgfraser: Yes.

To be specific, internally we need to maintain a mapping from datasources to requests that are associated with those data sources. So if we transform the data source somehow, like by providing it with its environment, we need to modify the identifier for the data source so hash codes are consistent with equality.

We always modify the original identifier with the new descriptions. So if we had `DataSourceA` and `DataSourceB` and wrapped them each with three wrapped fields, we would get `DataSourceA.provide(Wrapper).provide(Wrapper).provide(Wrapper)` and `DataSourceB.provide(Wrapper).provide(Wrapper).provide(Wrapper)`, which would still be different.

The only situation where I could theoretically see it creating an issue is if you had the same data source and provided it with two different sets of the same number of wrappers that somehow operate on the data source, because then you could have two different data sources with the same identifier. But I don't think that is possible here because the `FieldWrapper` takes a `ZQuery[Any, _, _]` as an input, so I don't think it could even do anything to modify a data source.

adamgfraser: Could this be solved by changing the signature of `f` in `FieldWrapper` from:

```scala
(ZQuery[Any, Nothing, ResponseValue], FieldInfo) => ZQuery[R, CalibanError, ResponseValue]
```

to:

```scala
(ZQuery[R, Nothing, ResponseValue], FieldInfo) => ZQuery[R, CalibanError, ResponseValue]
```

That is, does the query being wrapper really have to have no environmental requirements at all? Or does it just not have to have any more requirements than the wrapper?

ghostdogpr: I couldn’t get the variance to work if I kept R there, it was becoming covariant. I tried the technique with `R1 <: R` but still couldn’t make it compile.

adamgfraser: Could you make `FieldWrapper` invariant and change line 123 of `Wrapper.scala` to `wrapper: FieldWrapper[R]`?

ghostdogpr: Oh right, I didn't realize I could make it invariant there while keeping it contravariant in `Wrapper`. I'll check if I can simplify the rest as well. Variance is tricky...

ghostdogpr: @adamgfraser [really nice simplification](https://github.com/ghostdogpr/caliban/pull/165/commits/b60213667f7aa7891b5222c3240658f9b900711e), thank you! I no longer need `ZQuery.environment` but since it's here I just left it.

Issue #156, opened 2020-01-06
Errors: add locations


ghostdogpr:
Add support for the `locations` data for each error as defined in the [GraphQL spec](https://graphql.github.io/graphql-spec/June2018/#sec-Errors):
```json
{
  "errors": [
    {
      "message": "Name for character with ID 1002 could not be fetched.",
      "locations": [ { "line": 6, "column": 7 } ],
      "path": [ "hero", "heroFriends", 1, "name" ],
      "extensions": {
        "code": "CAN_NOT_FETCH_BY_ID",
        "timestamp": "Fri Feb 9 14:33:09 UTC 2018"
      }
    }
  ]
}
```

This will require extracting the line and column information of each field when parsing queries.
See initial discussion in https://github.com/ghostdogpr/caliban/issues/140

phderome:
Clarification request:

I'd like to understand a few basic things about the types involved. I'd like to know whether this is a reasonable extension
 ``` 
 case class ExecutionError(
    msg: String,
    path: List[Either[String, Int]] = Nil,
    locations: List[String] = Nil,
    innerThrowable: Option[Throwable] = None
  ) extends CalibanError
```

I’d like to understand why `path` is of the proper type. it seems that the `Int` part is caliban choosing to specify a List index (from some `reduce` step on Lists that does a zip) whereas I am guessing there is no need to inject additional context for `locations`. From GraphSQL spec, this `locations` is an `array` of `objects`. How would we want to format `"locations": [ { "line": 6, "column": 7 } ]` in the CalibanError `toString` method?

phderome:
How would `GraphQLResponse!handleError` handle the simultaneous presence of `path`, `msg`, and `locations` data? This is more of a behavioural/requirement/modelling question than coding for now.

ghostdogpr:
Closed by #168

PR #168, opened 2020-01-18
Add source location mapping

paulpdaniels:
This PR adds source location mapping to fields and fragments. It is primarily used for outputting when errors occur.

### DIFF
--- a/core/src/main/scala/caliban/CalibanError.scala
+++ b/core/src/main/scala/caliban/CalibanError.scala
@@ -1,5 +1,6 @@
 package caliban
 
+import caliban.interop.circe.IsCirceEncoder
 import caliban.parsing.adt.LocationInfo
 
 /**
@@ -14,14 +15,19 @@ object CalibanError {
   /**
    * Describes an error that happened while parsing a query.
    */
-  case class ParsingError(msg: String, innerThrowable: Option[Throwable] = None) extends CalibanError {
+  case class ParsingError(
+    msg: String,
+    locationInfo: Option[LocationInfo] = None,
+    innerThrowable: Option[Throwable] = None
+  ) extends CalibanError {
     override def toString: String = s"""Parsing error: $msg ${innerThrowable.fold("")(_.toString)}"""
   }
 
   /**
    * Describes an error that happened while validating a query.
    */
-  case class ValidationError(msg: String, explanatoryText: String) extends CalibanError {
+  case class ValidationError(msg: String, explanatoryText: String, locationInfo: Option[LocationInfo] = None)
+      extends CalibanError {
     override def toString: String = s"""Validation error: $msg $explanatoryText"""
   }
 
@@ -44,4 +50,53 @@ object CalibanError {
       s"Execution error$field: $msg$inner"
     }
   }
+
+  implicit def circeEncoder[F[_]](implicit ev: IsCirceEncoder[F]): F[CalibanError] =
+    ErrorCirce.errorValueEncoder.asInstanceOf[F[CalibanError]]
+}
+
+private object ErrorCirce {
+  import io.circe._
+  import io.circe.syntax._
+
+  private def locationToJson(li: LocationInfo): Json =
+    Json.obj("line" -> li.line.asJson, "column" -> li.column.asJson)
+
+  val errorValueEncoder: Encoder[CalibanError] = Encoder.instance[CalibanError] {
+    case CalibanError.ParsingError(msg, locationInfo, _) =>
+      Json
+        .obj(
+          "message" -> msg.asJson,
+          "locations" -> Some(locationInfo).collect {
+            case Some(li) => Json.arr(locationToJson(li))
+          }.asJson
+        )
+        .dropNullValues
+    case CalibanError.ValidationError(msg, _, locationInfo) =>
+      Json
+        .obj(
+          "message" -> msg.asJson,
+          "locations" -> Some(locationInfo).collect {
+            case Some(li) => Json.arr(locationToJson(li))
+          }.asJson
+        )
+        .dropNullValues
+    case CalibanError.ExecutionError(msg, path, locationInfo, _) =>
+      Json
+        .obj(
+          "message" -> msg.asJson,
+          "locations" -> Some(locationInfo).collect {
+            case Some(li) => Json.arr(locationToJson(li))
+          }.asJson,
+          "path" -> Some(path).collect {
+            case p if p.nonEmpty =>
+              Json.fromValues(p.map {
+                case Left(value)  => value.asJson
+                case Right(value) => value.asJson
+              })
+          }.asJson
+        )
+        .dropNullValues
+  }
+
 }
--- a/core/src/main/scala/caliban/CalibanError.scala
+++ b/core/src/main/scala/caliban/CalibanError.scala
@@ -1,5 +1,7 @@
 package caliban
 
+import caliban.parsing.adt.LocationInfo
+
 /**
  * The base type for all Caliban errors.
  */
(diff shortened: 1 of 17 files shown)


core/src/main/scala/caliban/CalibanError.scala
```diff
@@ -29,6 +31,7 @@ object CalibanError {
   case class ExecutionError(
     msg: String,
     path: List[Either[String, Int]] = Nil,
+    locationInfo: Option[LocationInfo] = None,
```
ghostdogpr: How about adding it to `ParsingError` and `ValidationError`?

core/src/main/scala/caliban/schema/Schema.scala
```diff
@@ -418,9 +420,13 @@ trait DerivationSchema[R] {
 
 object GenericSchema {
 
-  def effectfulExecutionError(path: List[Either[String, Int]], e: Throwable): ExecutionError =
+  def effectfulExecutionError(
```
ghostdogpr: Not related to your code, but this one seems only used in `Executor`, how about moving it there?

core/src/main/scala/caliban/GraphQLResponse.scala
```diff
@@ -34,13 +34,19 @@ private object GraphQLResponseCirce {
 
   private def handleError(err: Any): Json =
     err match {
-      case ExecutionError(_, path, _) if path.nonEmpty =>
+      case ExecutionError(_, path, location, _) if path.nonEmpty =>
+        val locationJson =
+          location.fold(Json.obj())(
+            l => Json.obj("location" -> Json.obj("column" -> l.column.asJson, "line" -> l.line.asJson))
```
ghostdogpr: Ah, one more thing. As mentioned in #156, the spec suggests a format which is slightly different from what you did: `"locations": [ { "line": 6, "column": 7 } ]`.

core/src/main/scala/caliban/execution/Executor.scala
```diff
@@ -58,11 +58,11 @@ object Executor {
         case s @ PureStep(value) =>
           value match {
             case EnumValue(v) if mergeFields(currentField, v).collectFirst {
-                  case Field("__typename", _, _, _, _, _, _) => true
+                  case Field("__typename", _, _, _, _, _, _, _) => true
                 }.nonEmpty =>
```
phderome: I personally don't like many underscores, though I understand that was the case prior to this PR...

How about this notation?
`                  case f: Field if f.name == "__typename" => true`

core/src/main/scala/caliban/parsing/SourceMapper.scala
```diff
@@ -0,0 +1,41 @@
...
+
+  def getLocation(index: Int): LocationInfo
+
+}
+
+object SourceMapper {
+
+  /**
+   * Implementation taken from https://github.com/lihaoyi/fastparse/blob/master/fastparse/src/fastparse/ParserInput.scala
+   */
```
phderome: you may need to reproduce his MIT license somewhere if you have not taken care of that.

core/src/main/scala/caliban/validation/Validator.scala
```diff
@@ -132,7 +133,7 @@ object Validator {
   private def collectSelectionSets(selectionSet: List[Selection]): List[Selection] =
     selectionSet ++ selectionSet.flatMap {
       case _: FragmentSpread                  => Nil
-      case Field(_, _, _, _, selectionSet)    => collectSelectionSets(selectionSet)
+      case Field(_, _, _, _, selectionSet, _) => collectSelectionSets(selectionSet)
```
phderome: I'd be inclined to use a more concise notation that does not use underscores much, once again here.
`case f: Field => collectSelectionSets(f.selectionSet)`

core/src/main/scala/caliban/validation/Validator.scala
```diff
@@ -161,7 +162,7 @@ object Validator {
     IO.foreach(selectionSet) {
         case FragmentSpread(_, directives) =>
           checkDirectivesUniqueness(directives).as(directives.map((_, __DirectiveLocation.FRAGMENT_SPREAD)))
-        case Field(_, _, _, directives, selectionSet) =>
+        case Field(_, _, _, directives, selectionSet, _) =>
```
phderome: since we don't match on special values, again just `case f: Field => ...` access f members where needed

core/src/main/scala/caliban/execution/Executor.scala
```diff
@@ -58,11 +58,11 @@ object Executor {
...
               // special case of an hybrid union containing case objects, those should return an object instead of a string
               val obj = mergeFields(currentField, v).collectFirst {
-                case Field(name @ "__typename", _, _, alias, _, _, _) =>
+                case Field(name @ "__typename", _, _, alias, _, _, _, _) =>
                   ObjectValue(List(alias.getOrElse(name) -> StringValue(v)))
               }
               obj.fold(s)(PureStep(_))
```
phderome: In fact I am proposing this simplification for current lines 60-68, which pass the tests.
```
        case s @ PureStep(value) =>
          value match {
            case EnumValue(v) =>
              // special case of an hybrid union containing case objects, those should return an object instead of a string
              val obj = mergeFields(currentField, v).collectFirst {
                case f: Field if f.name == "__typename" =>
                  ObjectValue(List(f.alias.getOrElse(f.name) -> StringValue(v)))
              }
              obj.fold(s)(PureStep(_))
            case _ => s
          }
```
I don't see any need to do `mergeFields...collectFirst` twice here, once in match guard and once in evaluation optional `obj`.

phderome: Note that I am considering `s` to be the default value for `value match` in the complex case and the second match all case just like existing code.

core/src/main/scala/caliban/execution/Executor.scala
paulpdaniels: I agree it gets verbose. One concern I have about this change is that it removes exhaustive checks on the case match. I think it’s less of a concern where there are `if` constraints because I think that already disables the check as well.

core/src/main/scala/caliban/GraphQLResponse.scala
phderome: note that `location` is evaluated only when `path` is set. This may not be what you want.

core/src/main/scala/caliban/parsing/SourceMapper.scala
paulpdaniels: Good point. Is that usually at the file level or do we just need to have it once at the top level?

core/src/main/scala/caliban/GraphQLResponse.scala
paulpdaniels: Ah you’re right, though ive never actually seen multiple locations in the wild. I guess I should change it to be a `List` instead of an `Option` too

core/src/main/scala/caliban/validation/Validator.scala
paulpdaniels: Here should be fine since exhaustive checking would already be disabled from the line above

core/src/main/scala/caliban/GraphQLResponse.scala
phderome: I mean `path` and `location` are presumably independent of each other therefore the guard `if path.nonEmpty =>` looks to be dangerous as it impacts evaluation of `location`. One could consider `location` being non-empty and `path` being empty even though it's more of an intellectual exercise than a "reality".

core/src/main/scala/caliban/parsing/SourceMapper.scala
phderome: I know of firms who do it at product level. You'd have to ask ZIO principals or Pierre I guess.

ghostdogpr: I think in the file would be fine. Found an example in ZIO here https://github.com/zio/zio/blob/master/stacktracer/jvm/src/main/scala/zio/internal/stacktracer/impl/AkkaLineNumbers.scala

ghostdogpr:
Looks good to me! It needs a rebase to solve the merge conflict with master and I think we’re good.

PR #176, opened 2020-01-26
Support for circe json scalar

### LINKED ISSUE #167: Support of JSON scalar type.
In Several implementations of graphql there exists possibility to extend scalar types.
Blog: [blog](https://medium.com/open-graphql/top-5-graphql-scalars-3f8a38965b53) also mentions JSON scalar extension for an implementation of graphql.
I can’t  see any easy way how to do it with current implementation of caliban.
If you folks agree that it should be implemented, I can try to make some PR in near future.

ghostdogpr:
Fixes #167

Requires `import caliban.interop.circe.json._`.

### DIFF
--- a/core/src/main/scala/caliban/interop/circe/circe.scala
+++ b/core/src/main/scala/caliban/interop/circe/circe.scala
@@ -1,6 +1,13 @@
 package caliban.interop.circe
 
+import caliban.introspection.adt.__Type
+import caliban.schema.Step.QueryStep
+import caliban.schema.Types.makeScalar
+import caliban.schema.{ ArgBuilder, PureStep, Schema, Step }
+import caliban.{ InputValue, ResponseValue }
 import io.circe._
+import zio.ZIO
+import zquery.ZQuery
 
 /**
  * This class is an implementation of the pattern described in https://blog.7mind.io/no-more-orphans.html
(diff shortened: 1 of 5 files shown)


core/src/main/scala/caliban/interop/circe/circe.scala
```diff
@@ -19,3 +26,13 @@ private[caliban] trait IsCirceDecoder[F[_]]
 private[caliban] object IsCirceDecoder {
   implicit val isCirceDecoder: IsCirceDecoder[Decoder] = null
 }
+
+object json {
+  implicit val jsonSchema: Schema[Any, Json] = new Schema[Any, Json] {
+    override def toType(isInput: Boolean): __Type = makeScalar("Json")
+    override def resolve(value: Json): Step[Any] =
+      QueryStep(ZQuery.fromEffect(ZIO.fromEither(implicitly[Decoder[ResponseValue]].decodeJson(value))).map(PureStep))
```
fokot: Use `Decoder[ResponseValue]` instead of `implicitly[Decoder[ResponseValue]]` all cats based libs have this summoner pattern.

core/src/main/scala/caliban/interop/circe/circe.scala
```diff
@@ -19,3 +26,13 @@ private[caliban] trait IsCirceDecoder[F[_]]
...
+  }
+  implicit val jsonArgBuilder: ArgBuilder[Json] = (input: InputValue) =>
+    Right(implicitly[Encoder[InputValue]].apply(input))
```
fokot: Same here

ghostdogpr:
@fokot oops, thanks for testing, I made a copy/paste mistake. Will fix right now and add a test.

core/src/main/scala/caliban/interop/circe/circe.scala
ghostdogpr: Good point!

ghostdogpr:
@fokot should be OK now!

PR #181, opened 2020-01-28
Allow to hide unsafe call with akka-http wrapper

loicdescotte:
Very small PR to allow to hide unsafe call with akka-http wrapper


ghostdogpr:
You need to run `fmt` in sbt for linting.

About the naming, how about `makeHttpServiceM`? Like in ZIO, a lot of functions have an `M` (for Monadic) variant that takes an effect instead of a simple value.

loicdescotte:
Done + rebased the history :)

PR #192, opened 2020-01-31
Add directives to types and fields

paulpdaniels:
Adds directive declarations to various types in the adt. This is needed because both [`cache-control`](https://github.com/apollographql/apollo-cache-control) and [`federation`](https://www.apollographql.com/docs/apollo-server/federation/federation-spec/) specs use type and field level directives (though the former doesn't technically require them).


paulpdaniels:
Superseded by #225

Issue #186, opened 2020-01-31
Codegen: support input


ghostdogpr:
The sbt plugin caliban-codegen generate your case classes from a GraphQL schema. It currently doesn't support input types.

Parse inputs and generate the appropriate case classes.

You can have a look at how `type` is currently handled and apply the same logic. Let me know if you need some guidance, I'll be happy to help.

atapin:
@ghostdogpr I started working on this. It looks like we need to support of inputs in the parser as well. Is it ok to do both core and codegen changes in a single PR?

ghostdogpr:
@atapin Awesome! You're correct, each of those issues requires extending the parser to support the missing [type definitions](http://spec.graphql.org/June2018/#TypeDefinition), and then adding the corresponding code generation. Perfectly fine to do all in one PR.

PR #193, opened 2020-01-31
[WIP] Added support for input types parsing and code generation (#186)

atapin:


core/src/main/scala/caliban/parsing/Parser.scala
```diff
@@ -192,7 +192,7 @@ object Parser {
     }
 
   private def typeDefinition[_: P]: P[TypeDefinition] =
-    P("type" ~/ name ~ "{" ~ fieldDefinition.rep ~ "}").map(t => TypeDefinition(t._1, t._2.toList))
+    P(("type" | "input") ~/ name ~ "{" ~ fieldDefinition.rep ~ "}").map(t => TypeDefinition(t._1, t._2.toList))
```
ghostdogpr: Actually if you look at the spec, an input type is slightly different from a regular type:
http://spec.graphql.org/June2018/#InputValueDefinition
http://spec.graphql.org/June2018/#FieldDefinition
(no arguments possible, default value added)

ghostdogpr:
@atapin Hi! I did all of the missing types (enum, union, scalar, etc) so this is the only one left. Do you mind if I do it while I'm at it?

atapin:
@ghostdogpr sure. I was gonna ask for help with adding the types

ghostdogpr:
Thanks! I'll add it soon in #209 
Closing this one then.

PR #195, opened 2020-02-03
Fix Akka HTTP example

ghostdogpr:
The interpreter (and the service) were recreated at every API call.

### DIFF
--- a/akka-http/src/main/scala/caliban/AkkaHttpAdapter.scala
+++ b/akka-http/src/main/scala/caliban/AkkaHttpAdapter.scala
@@ -44,11 +44,6 @@ object AkkaHttpAdapter extends FailFastCirceSupport {
         .future
     )
 
-  def makeHttpServiceM[R, E](
-    interpreter: URIO[R, GraphQLInterpreter[R, E]]
-  )(implicit ec: ExecutionContext, runtime: Runtime[R]): Route =
-    makeHttpService(runtime.unsafeRun(interpreter))
-
   def makeHttpService[R, E](
     interpreter: GraphQLInterpreter[R, E]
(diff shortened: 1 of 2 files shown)


ghostdogpr:
@loicdescotte I'm wondering if we should keep `makeHttpServiceM`, since it is quite dangerous.
Because it creates the interpreter with `unsafeRun` , it should not be used inside `path { ... }` otherwise the interpreter is recreated every time. It works if you call `makeHttpServiceM` outside of `path { ... }` and store the route in a val like I did in this PR, but still I find it dangerous. What do you think?

loicdescotte:
If I understand correctly, it's more an issue with the little refactoring I've done rather than an issue with `makeHttpServiceM`, because in the first steps of my refactoring there was the same problem without `makeHttpServiceM` : 

```scala
val route =
    path("api" / "graphql") {
      runtime.unsafeRun(AkkaHttpAdapter.makeHttpService(interpreter))
    } ~ path("graphiql") {
      getFromResource("graphiql.html")
    }
```

This this code we would have the same issue, right?

ghostdogpr:
Yeah we would have the same issue, but at least it's obvious that you create the interpreter every time.

With `makeHttpServiceM` we're hiding the fact that we're doing the `unsafeRun`. We're just returning the `Route` so library users will definitely put it directly under `path` and have the issue, no?

loicdescotte:
I liked the fact that we could let the framework do the "dirty" unsafe job and let the user write an app in a pure way, but I understand your point and I don't see another way to do it so, yes maybe it needs to be removed :)

ghostdogpr:
@loicdescotte to emphasize it's not the interpreter that needs an effect, I put the `unsafeRun` around `ExampleService.make(sampleCharacters)`.

PR #196, opened 2020-02-03
rough draft wip for ws on AkkaHttp

phderome:
Minor progress on ws solution for AkkaHttpAdapter, stuck when using Altair

some potential proposal of refactoring of Http4sAdapter though there is no apparent/intentional functional change. I personally like it better, but it's not needed for sure for this PR.


examples/src/main/scala/caliban/akkahttp/ExampleApp.scala
```diff
@@ -34,21 +49,17 @@ object ExampleApp extends App with GenericSchema[Console with Clock] {
...
-            Subscriptions(service.deletedEvents)
-          )
-        ).interpreter
-      })
+      .map(service => makeApi(service).interpreter)
+
+  val apiRoute: Route = AkkaHttpAdapter.makeHttpServiceM(interpreter)
+  val wsApiRoute = {
+    // when do we create subscriptions? When do we make it effectful/unsafe?
+    //  val subscriptions: UIO[Ref[Map[String, Fiber[Throwable, Unit]]]] = Ref.make(Map.empty[String, Fiber[Throwable, Unit]])
```
ghostdogpr: `subscriptions` should be created inside the adapter, the end-user don't need to see it. A new `Ref` should be created for each WS connection.

examples/src/main/scala/caliban/akkahttp/ExampleApp.scala
```diff
@@ -59,13 +70,16 @@ object ExampleApp extends App with GenericSchema[Console with Clock] {
...
     path("api" / "graphql") {
-      AkkaHttpAdapter.makeHttpServiceM(interpreter)
+      apiRoute // do not call makeHttpServiceM here, or it will be recreated at every call
+    } ~ path("ws" / "graphql") {
+      wsApiRoute
     } ~ path("graphiql") {
       getFromResource("graphiql.html")
     }
 
+  // convention is normally 8080. Keep at 8088?
```
ghostdogpr: Because so many services use 8080, I always end up with conflicts. I'd prefer to keep 8088 for this reason.

phderome:
Comments on where we stand here.

- `onGraphQLResponse` is now almost identical to http4s version
- `processMessage` is now almost identical to http4s version except shortcoming of `stop` handling, which needs more work
 
Pending issues:
- mutations: return only partial data unlike http4s with Akka `extensions` is missing! Only `data` appears)
- lifecycle of `subscriptions` is unclear and could be quite wrong. For instance when routes are created multiple times from Caliban we see loads of duplicate answers, though that's not necessarily a subscription issue. It's not perfectly clear when it should be recreated (done in createSink)
- definition and usage `trait Protocol` with `Complete`, `Fail` is suspect (see `createSource` usage of `Source.actorRef` comment below)
- `processMessage`, connection terminate case should close cleanly and it's not immediately obvious how to do it (possibly some via hook in `createSink`/Flow)
- usage of Actor to send messages back is weak in terms of back-pressure. Akka Http and Streams users recommend instead using some kind of queue. This needs investigation.
- `createSource` usage of `Source.actorRef` requires a `CompletionMatcher` and `FailureMatcher` and it's not clear how we should handle those.

phderome:
subsumed by #219.

PR #200, opened 2020-02-05
ZQuery Improvements

adamgfraser:
Improvements to ZQuery based on recent suggestions:

- Query metrics - implements `summarized` and `timed` methods on ZQuery analogous to existing methods on ZIO. These do not interfere with query batching. 
- Optional data sources - implements `fromRequestOption` constructor on ZQuery to construct queries that return `None` instead of failing with a `QueryFailure` if a data source does not return a result for a request. Implements `fromFunctionOption`, `fromFunctionOptionM`, `fromFunctionBatchedOption`, and `fromFunctionBatchedOptionM` to create data sources that may not return a result for a request.
- More support for data sources that can fail - `fromFunctionBatchedWithM` now also allows you to use an effect that can fail. If the effect fails the failure will be returned for all requests.

### DIFF
--- a/core/src/main/scala/zquery/Cache.scala
+++ b/core/src/main/scala/zquery/Cache.scala
@@ -21,7 +21,7 @@ final class Cache private (private val state: Ref[Map[Any, Any]]) {
    * Looks up a request in the cache, failing with the unit value if the
    * request is not in the cache, succeeding with `Ref(None)` if the request is
    * in the cache but has not been executed yet, or `Ref(Some(value))` if the
-   * request has been xecuted.
+   * request has been executed.
    */
   def lookup[E, A](request: Request[E, A]): IO[Unit, Ref[Option[Either[E, A]]]] =
     state.get.map(_.get(request).asInstanceOf[Option[Ref[Option[Either[E, A]]]]]).get
--- a/core/src/main/scala/zquery/Cache.scala
+++ b/core/src/main/scala/zquery/Cache.scala
@@ -18,10 +18,10 @@ final class Cache private (private val state: Ref[Map[Any, Any]]) {
     state.update(_ + (request -> result)).unit
 
   /**
-   * Looks up a request in the cache, returning `None` if the request is not in
-   * the cache, `Some(Ref(None))` if the request is in the cache but has not
-   * been executed yet, or `Some(Ref(Some(value)))` if the request has been
-   * executed.
+   * Looks up a request in the cache, failing with the unit value if the
+   * request is not in the cache, succeeding with `Ref(None)` if the request is
+   * in the cache but has not been executed yet, or `Ref(Some(value))` if the
+   * request has been xecuted.
    */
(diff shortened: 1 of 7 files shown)


core/src/main/scala/zquery/Cache.scala
```diff
@@ -18,10 +18,10 @@ final class Cache private (private val state: Ref[Map[Any, Any]]) {
...
 
   /**
-   * Looks up a request in the cache, returning `None` if the request is not in
-   * the cache, `Some(Ref(None))` if the request is in the cache but has not
-   * been executed yet, or `Some(Ref(Some(value)))` if the request has been
-   * executed.
+   * Looks up a request in the cache, failing with the unit value if the
+   * request is not in the cache, succeeding with `Ref(None)` if the request is
+   * in the cache but has not been executed yet, or `Ref(Some(value))` if the
+   * request has been xecuted.
```
ghostdogpr:
```suggestion
   * request has been executed.
```

core/src/main/scala/zquery/ZQuery.scala
```diff
@@ -199,6 +201,24 @@ sealed trait ZQuery[-R, +E, +A] { self =>
...
+    for {
+      start <- ZQuery.fromEffect(summary)
+      value <- self
+      end   <- ZQuery.fromEffect(summary)
+    } yield (f(start, end), value)
+
+  /**
+   * Returns a new query that executes this one and times the execution.
+   */
+  final def timed: ZQuery[R with Clock, E, (Duration, A)] =
```
ghostdogpr: Nice! Do you mind calling it in `ApolloTracing.apolloTracingField`?

adamgfraser: Will do!

core/src/main/scala/zquery/Cache.scala
adamgfraser: Fixed!

adamgfraser:
All set.

core/src/main/scala/zquery/ZQuery.scala
```diff
@@ -199,6 +201,24 @@ sealed trait ZQuery[-R, +E, +A] { self =>
       a     <- runCache(cache)
     } yield (cache, a)
 
+  /**
+   * Summarizes a query by computing some value before and after execution,
+   * and then combining the values to produce a summary, together with the
+   * result of execution.
+   */
+  final def summarized[R1 <: R, E1 >: E, B, C](f: (B, B) => C)(summary: ZIO[R1, E1, B]): ZQuery[R1, E1, (C, A)] =
```
ghostdogpr: Shall we switch the parameters here too?

adamgfraser: Yes. I was slightly inclined to wait until RC18 and make conforming changes to that so we track the public ZIO API for identical methods but happy to change now as well.

ghostdogpr: Ah ok, either way's fine 👍

PR #206, opened 2020-02-08
Codegen: initial support enums

### LINKED ISSUE #185: Codegen: support enums
The sbt plugin caliban-codegen generate your case classes from a GraphQL schema. It currently doesn't support enums.

Parse enums and generate the appropriate sealed trait + case objects.

You can have a look at how `type` is currently handled and apply the same logic. Let me know if you need some guidance, I'll be happy to help.

yoohaemin:
part of #185 

Add EnumTypeDefinition

- Rename TypeDefinition to ObjectTypeDefinition


ghostdogpr:
Thanks! As we discussed, I'm merging this one because I need some of your changes to work on other types. Please submit another PR later with the missing bits 👍

Issue #66, opened 2019-11-01
Make the examples to compile


palutz:
Trying the examples (to understand the lib), but they not compile.
Is it also possible to have a build.sbt as example for running the project?

Thanks

ghostdogpr:
@palutz the examples project is part of the whole project's build.sbt but I added this README in case you want to compile only the examples: https://github.com/ghostdogpr/caliban/tree/master/examples

Does that work for you? Let me know if you still have some errors.

ghostdogpr:
Closing but feel free to reopen if the instructions aren't enough.

palutz:
Hi,
thanks for the answer.
Anyhow the code provided (with the readme) is at least solving most of the problems but still have some compile errors

> no type parameters for method apply: (http: org.http4s.Http[F,G], config: org.http4s.server.middleware.CORSConfig)(implicit F: cats.Applicative[F])org.http4s.Http[F,G] in object CORS exist so that it can be applied to arguments (org.http4s.HttpRoutes[[β$0$]zio.ZIO[zio.console.Console with zio.clock.Clock,Throwable,β$0$]], org.http4s.server.middleware.CORSConfig)
> [error]  --- because ---
> [error] argument expression's type is not compatible with formal parameter type;
> [error]  found   : org.http4s.HttpRoutes[[β$0$]zio.ZIO[zio.console.Console with zio.clock.Clock,Throwable,β$0$]]
> [error]     (which expands to)  cats.data.Kleisli[[β$0$]cats.data.OptionT[[β$0$]zio.ZIO[zio.console.Console with zio.clock.Clock,Throwable,β$0$],β$0$],org.http4s.Request[[β$0$]zio.ZIO[zio.console.Console with zio.clock.Clock,Throwable,β$0$]],org.http4s.Response[[β$0$]zio.ZIO[zio.console.Console with zio.clock.Clock,Throwable,β$0$]]]
> [error]  required: org.http4s.Http[?F,?G]
> [error]     (which expands to)  cats.data.Kleisli[?F,org.http4s.Request[?G],org.http4s.Response[?G]]
> [error]                 "/api/graphql" -> CORS(Http4sAdapter.makeRestService(interpreter)),
> 
> type mismatch;
> [error]  found   : org.http4s.HttpRoutes[[β$0$]zio.ZIO[zio.console.Console with zio.clock.Clock,Throwable,β$0$]]
[...]

PR #215, opened 2020-02-13
Fix example readme

ruurtjan:
This fixes the readme for the examples. Also see https://github.com/ghostdogpr/caliban/issues/66


ghostdogpr:
Thanks! I re-ran the CI, there was a problem unrelated to caliban.

Issue #170, opened 2020-01-23
Type names validations


fokot:
You can write output type like `case class XXXInput` and input type `case class XXX` which would produce invalid schema. `Graphiql` will fail on this.

Also another possible error is having two  output types with same name from different packages which will result in error that the first type is used in Schema, so `graphiql` does not report error but wrong type is there and it will fail when executing request if I use field which is present in first type but not in second.

ghostdogpr:
I agree some kind of check would be useful. The question is at which point do we perform that check and how do we report the problem. Most functions for building and combining apis are pure and without effects. The one that’s effectful is run for each request so it’s not ideal for this.

ghostdogpr:
Since #227 , we now validate the schema when calling `.interpreter`. So we have a place to implement these checks.

ghostdogpr:
Working on this.

PR #225, opened 2020-02-20
Add directive support

paulpdaniels:
Revamped version of #192 

This adds schema directives which are needed for such extensions as cache-control and federation. In order to test the behavior I also included an implementation of the cache-control extension using directives and the new wrappers.

### DIFF
--- a/core/src/main/scala/caliban/Rendering.scala
+++ b/core/src/main/scala/caliban/Rendering.scala
@@ -6,42 +6,61 @@ import caliban.parsing.adt.Directive
 
 object Rendering {
 
+  private implicit val kindOrdering: Ordering[__TypeKind] = Ordering
+    .by[__TypeKind, Int] {
+      case __TypeKind.SCALAR       => 1
+      case __TypeKind.NON_NULL     => 2
+      case __TypeKind.LIST         => 3
+      case __TypeKind.UNION        => 4
+      case __TypeKind.ENUM         => 5
+      case __TypeKind.INPUT_OBJECT => 6
+      case __TypeKind.INTERFACE    => 7
+      case __TypeKind.OBJECT       => 8
+    }
+
+  private implicit val renderOrdering: Ordering[(String, __Type)] = Ordering.by(o => (o._2.kind, o._2.name))
+
   /**
    * Returns a string that renders the provided types into the GraphQL format.
    */
   def renderTypes(types: Map[String, __Type]): String =
-    types.flatMap {
-      case (_, t) =>
-        t.kind match {
-          case __TypeKind.SCALAR   => t.name.flatMap(name => if (isBuiltinScalar(name)) None else Some(s"scalar $name"))
-          case __TypeKind.NON_NULL => None
-          case __TypeKind.LIST     => None
-          case __TypeKind.UNION =>
-            val renderedTypes: String =
-              t.possibleTypes
-                .fold(List.empty[String])(_.flatMap(_.name))
-                .mkString(" | ")
-            Some(s"""${renderDescription(t.description)}${renderKind(t.kind)} ${renderTypeName(t)} = $renderedTypes""")
-          case _ =>
-            val renderedDirectives: String = renderDirectives(t.directives)
-            val renderedFields: String = t
-              .fields(__DeprecatedArgs())
-              .fold(List.empty[String])(_.map(renderField))
-              .mkString("\n  ")
-            val renderedInputFields: String = t.inputFields
-              .fold(List.empty[String])(_.map(renderInputValue))
-              .mkString("\n  ")
-            val renderedEnumValues = t
-              .enumValues(__DeprecatedArgs())
-              .fold(List.empty[String])(_.map(renderEnumValue))
-              .mkString("\n  ")
-            Some(
-              s"""${renderDescription(t.description)}${renderKind(t.kind)} ${renderTypeName(t)}${renderInterfaces(t)}$renderedDirectives {
-                 |  $renderedFields$renderedInputFields$renderedEnumValues
-                 |}""".stripMargin
-            )
-        }
-    }.mkString("\n\n")
+    types.toList
+      .sorted(renderOrdering)
+      .flatMap {
+        case (_, t) =>
+          t.kind match {
+            case __TypeKind.SCALAR   => t.name.flatMap(name => if (isBuiltinScalar(name)) None else Some(s"scalar $name"))
+            case __TypeKind.NON_NULL => None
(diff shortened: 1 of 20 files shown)


core/src/main/scala/caliban/Rendering.scala
```diff
@@ -62,13 +65,40 @@ object Rendering {
     case Some(value) => if (value.contains("\n")) s"""\"\"\"\n$value\"\"\"\n""" else s""""$value"\n"""
   }
 
+  private def renderDirectiveArgument(value: InputValue): String = value match {
+    case InputValue.ListValue(values) =>
+      values.map(renderDirectiveArgument).mkString("[", ",", "]")
+    case InputValue.ObjectValue(fields) =>
+      fields.map { case (key, value) => s"$key: ${renderDirectiveArgument(value)}" }.mkString("{", ",", "}")
+    case InputValue.VariableValue(_) =>
+      throw new Exception("Variable values are not allowed in a directive declaration")
```
ghostdogpr: I'd rather not render it here than throwing an exception.

core/src/test/scala/caliban/RenderingSpec.scala
```diff
@@ -0,0 +1,62 @@
...
+import caliban.GraphQL._
+
+object RenderingSpec
+    extends DefaultRunnableSpec(
+      suite("rendering")(
+        test("it should render directives") {
+          assert(
+            graphQL(resolver).render.trim,
+            equalTo("""
+                      |union Role = Captain | Engineer | Mechanic | Pilot
```
ghostdogpr: To have a stable order, we could sort the types by kind, then by name.

core/src/main/scala/caliban/wrappers/ApolloCaching.scala
```diff
@@ -0,0 +1,145 @@
...
+import caliban.ResponseValue
+import caliban.ResponseValue.{ ListValue, ObjectValue }
+import caliban.Value.{ EnumValue, IntValue, StringValue }
+import caliban.parsing.adt.Directive
+import caliban.wrappers.Wrapper.{ EffectfulWrapper, FieldWrapper, OverallWrapper }
+import zio.Ref
+import zio.duration.Duration
+import zquery.ZQuery
+
+object CacheControl {
```
ghostdogpr: How about moving this object within `ApolloCaching`?

ghostdogpr:
This is really nice! Can you mention in the docs:
- in Schema there's the list of annotations supported, you can add the new one
- in Middleware there's the list of builtin wrapper, you can add Apollo Caching alongside Apollo Tracing

core/src/main/scala/caliban/Rendering.scala
paulpdaniels: Checking the spec after you mentioned this I realized there are a couple more things to be mindful of: https://spec.graphql.org/June2018/#sec-Type-System.Directives

> Validation

> A directive definition must not contain the use of a directive which references itself directly.
A directive definition must not contain the use of a directive which references itself indirectly by referencing a Type or Directive which transitively includes a reference to this directive.
The directive must not have a name which begins with the characters `__` (two underscores).
For each argument of the directive:
The argument must not have a name which begins with the characters `__` (two underscores).
The argument must accept a type where IsInputType(argumentType) returns true.

paulpdaniels: I don't know if it would be possible to resolve these at compile time, so it probably would have to be a runtime error, but the question is how should we handle it? Agree that throwing isn't the best option, should we have two rendering methods? one that simply suppresses errors and one that returns an `IO[ValidationError, String]`

core/src/test/scala/caliban/RenderingSpec.scala
paulpdaniels: Thats a good idea I'll give that a shot.

core/src/main/scala/caliban/wrappers/ApolloCaching.scala
paulpdaniels: I was thinking either that or moving all of it under `wrappers.caching._`. I guess for consistency with ApolloTracing I'll move it under the main object.

core/src/main/scala/caliban/Rendering.scala
ghostdogpr: This is a bit similar to issue #170 where we lack a way to report something abnormal within the schema. Ideally we should fail at the point of building the schema rather than later during query execution or rendering like here.

It bothers me to make `graphQL(...)` return an effect because attaching the wrappers with `@@` wouldn't be as smooth as it is now, but how about making `.interpreter` return an effect? That way we can build the schema easily, but when we turn it into an interpreter (which we do only once), we verify that the schema is valid and fail if it's not. What do you think?

ghostdogpr: Btw if we go for that solution, we can do it in a later PR and just remove the check for now (so that you don't need to implement the whole thing to get this PR merged).

paulpdaniels: > .interpreter return an effect

I think this makes sense, since it is the "effect" of becoming an interpretable schema.

Ok I'm fixing so that it will just remove the arguments if they aren't valid and then we can fix it with the change you mentioned above.

core/src/test/scala/caliban/TestUtils.scala
```diff
@@ -34,7 +36,13 @@ object TestUtils {
     case class Mechanic(shipName: String) extends Role
   }
 
-  case class Character(name: String, nicknames: List[String], origin: Origin, role: Option[Role])
+  @GQLDirective(Directive("key", Map("name" -> StringValue("name"))))
+  case class Character(
+    @GQLDirective(Directive("external")) name: String,
```
ghostdogpr: Do we want to keep these?

vuepress/docs/docs/middleware.md
```diff
@@ -64,9 +66,12 @@ val api =
     maxDepth(50) @@
     timeout(3 seconds) @@
     printSlowQueries(500 millis) @@
-    apolloTracing
+    apolloTracing @@
+    apolloCaching
 ```
 
+As well you can also use
```
ghostdogpr: Delete?

core/src/test/scala/caliban/TestUtils.scala
paulpdaniels: Hmm I think I was going to use them in tests but didn't end up doing so...

paulpdaniels: Yeah they are used in the RenderingSpec

PR #233, opened 2020-02-22
Add Union validation (#230)

### LINKED ISSUE #230: Schema validation: unions
In `Validator.validateSchema`, verify the [rules for union types](https://spec.graphql.org/June2018/#sec-Unions)

1. A Union type must include one or more unique member types.
2. The member types of a Union type must all be Object base types;
   Scalar, Interface and Union types must not be member types of a Union.
   Similarly, wrapping types must not be member types of a Union.

palanga:
Fixes #230

### DIFF
--- a/core/src/main/scala/caliban/CalibanError.scala
+++ b/core/src/main/scala/caliban/CalibanError.scala
@@ -28,7 +28,7 @@ object CalibanError {
    */
   case class ValidationError(msg: String, explanatoryText: String, locationInfo: Option[LocationInfo] = None)
       extends CalibanError {
-    override def toString: String = s"ValidationError Error: $msg}"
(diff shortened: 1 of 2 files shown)


core/src/main/scala/caliban/CalibanError.scala
```diff
@@ -28,7 +28,7 @@ object CalibanError {
    */
   case class ValidationError(msg: String, explanatoryText: String, locationInfo: Option[LocationInfo] = None)
       extends CalibanError {
-    override def toString: String = s"ValidationError Error: $msg}"
+    override def toString: String = s"ValidationError Error: $msg"
```
ghostdogpr: Oops! Good catch!

palanga:
Done. Thank you.

PR #234, opened 2020-02-23
Play support

### LINKED ISSUE #143: Add module for Play support
With support for Play Json serialization and routing.

yoohaemin:
close #143 

- Cross built for 2.7 and 2.8
- Json codecs for caliban datatypes
- Basic Controller

todo
- [ ] working example
- [ ] websocket (in a followup pr)
- [ ] check sbt config is fine w.r.t cross building play versions


play/src/main/scala/caliban/CalibanController.scala
```diff
@@ -0,0 +1,22 @@
...
+
+import caliban.PlayJson._
+import javax.inject.Inject
+import play.api.mvc.{ AbstractController, Action, ControllerComponents }
+import zio.Runtime
+
+class CalibanController[-R] @Inject() (cc: ControllerComponents, runtime: Runtime[R])
+    extends AbstractController(cc) {
+
+  def action[E](interpreter: GraphQLInterpreter[Any, E]): Action[GraphQLRequest] =
```
ghostdogpr: `R` instead of `Any`

build.sbt
```diff
@@ -142,6 +142,17 @@ lazy val http4s = project
   )
   .dependsOn(coreJVM)
 
+lazy val play_base = project
+  .in(file("play"))
+  .settings(name := "caliban-play", commonSettings)
```
ghostdogpr: Isn’t it going to try to publish the same binary twice?

core/src/main/scala/caliban/CalibanError.scala
```diff
@@ -58,35 +58,26 @@ private object ErrorCirce {
...
-                case Left(value)  => value.asJson
-                case Right(value) => value.asJson
-              })
-          }.asJson
+          "message"   -> msg.asJson,
+          "locations" -> locationInfo.fold(Json.Null)(li => Json.arr(locationToJson(li))),
+          "path" -> (path match {
+            case path if path.isEmpty => Json.Null
+            case path                 => Json.fromValues(path.map(_.fold(_.asJson, _.asJson)))
+          })
```
ghostdogpr: Nice simplifications!

ghostdogpr:
Also, I have never used Play so maybe @paulpdaniels and @fagossa can give more feedback.

build.sbt
paulpdaniels: No this should just do it once because only the specific builds are being added to aggregate (is my understanding at least)

play/src/main/scala/caliban/PlayJson.scala
```diff
@@ -0,0 +1,123 @@
...
+    case boolean: JsBoolean => BooleanValue(boolean.value)
+    case JsNumber(value) =>
+      Try(value.toIntExact).fold(_ => FloatValue(value), IntValue(_))
+    case JsString(value) => StringValue(value)
+    case JsArray(array) =>
+      InputValue.ListValue(array.toList.map(jsonToInputValue))
+    case JsObject(obj) =>
+      val builder = Predef.Map.newBuilder[String, InputValue]
+      builder.sizeHint(obj)
+      obj.foreach { case (k, v) => builder += (k -> jsonToInputValue(v)) }
```
paulpdaniels: I generally prefer `foldLeft` for this. i.e.

```scala
InputValue.ObjectValue(obj.foldLeft(builder) { case (k, v) => builder += (k -> jsonToInputValue(v)) }.result())

```

play/src/main/scala/caliban/PlayJson.scala
```diff
@@ -0,0 +1,123 @@
...
+
+  implicit val readsInputValue: Reads[InputValue] =
+    json => JsSuccess(jsonToInputValue(json))
+
+  implicit val readsGraphQLRequest: Reads[GraphQLRequest] =
+    for {
+      query         <- (__ \ "query").read[String]
+      operationName <- (__ \ "operationName").readNullable[String]
+      variables     <- (__ \ "variables").readNullable[Map[String, InputValue]]
+    } yield GraphQLRequest(query, operationName, variables)
```
paulpdaniels: This shouldn't be necessary should it? Can't this just be:

```scala
Json.reads[GraphQLRequest]
```

Since all the `Reads` are already present.

play/src/main/scala/caliban/PlayJson.scala
```diff
@@ -0,0 +1,123 @@
...
+  implicit val writesResponseValue: Writes[ResponseValue] = responseValueToJson
+
+  private def handleError(err: Any): JsValue = err match {
+    case ce: CalibanError => errorValueEncoder.writes(ce)
+    case _                => Json.obj("message" -> JsString(err.toString))
+  }
+
+  // CalibanError
+  private def locationToJson(li: LocationInfo): JsValue =
+    JsObject(List("line" -> li.line.asJson, "column" -> li.column.asJson))
```
paulpdaniels: Hint: use `Json.obj()` which already has magnets, then you don't need the `asJson` suffixes

play/src/main/scala/caliban/PlayJson.scala
```diff
@@ -0,0 +1,123 @@
...
+        "extensions" -> responseValueToJson(extensions.asInstanceOf[ResponseValue])
+      )
+    case GraphQLResponse(data, errors, None) =>
+      Json.obj("data" -> responseValueToJson(data), "errors" -> JsArray(errors.map(handleError)))
+    case GraphQLResponse(data, errors, Some(extensions)) =>
+      Json.obj(
+        "data"       -> responseValueToJson(data),
+        "errors"     -> JsArray(errors.map(handleError)),
+        "extensions" -> responseValueToJson(extensions.asInstanceOf[ResponseValue])
+      )
```
paulpdaniels: The `Writes` should automatically remove `None` so you only need:

```
case GraphQLResponse(data, Nil, maybeExtensions) =>
  Json.obj("data" -> responseValueToJson(data), "extensions" -> maybeExtensions.map(responseValueToJson)
case GraphQLResponse(data, errors, extensions) =>
  Json.obj("data" -> ...)
```

play/src/main/scala/caliban/PlayJson.scala
```diff
@@ -0,0 +1,123 @@
...
+  // CalibanError
+  private def locationToJson(li: LocationInfo): JsValue =
+    JsObject(List("line" -> li.line.asJson, "column" -> li.column.asJson))
+
+  val errorValueEncoder: Writes[CalibanError] = {
+    case CalibanError.ParsingError(msg, locationInfo, _) =>
+      JsObject(
+        List("message" -> s"Parsing Error: $msg".asJson) ++
+          locationInfo.map(li => "locations" -> Json.arr(locationToJson(li)))
+      )
```
paulpdaniels: Can be simplified to:

```scala
Json.obj(
  "message" -> "Parsing Error: $msg",
  "locations" -> locationInfo.map(li => Json.arr(locationToJson(li))
)
```

play/src/main/scala/caliban/PlayJson.scala
paulpdaniels: Actually why not just turn this into a private `Writes` then below you can just write `locationInfo.map(Json.arr(_))` or even better have the `Writes` return a `JsArray`

play/src/main/scala/caliban/PlayJson.scala
```diff
@@ -0,0 +1,123 @@
...
+    case CalibanError.ExecutionError(msg, path, locationInfo, _) =>
+      JsObject(
+        List("message" -> msg.asJson) ++
+          locationInfo.map(li => "locations" -> Json.arr(locationToJson(li))) ++
+          (path match {
+            case p if p.nonEmpty => Some("path" -> JsArray(p.map(_.fold(_.asJson, _.asJson))))
+            case _               => None
+          })
+      )
```
paulpdaniels: Hint:

```scala
Json.obj(
  "message" -> msg,
   // If using the suggested method above to create a `Writes`
  "locations" -> locationInfo,
   (path match {
     case p if p.nonEmpty => Some("path" -> JsArray(p.map(_.fold(_.asJson, _.asJson))))
     case _               => None
   })
)
```

play/src/main/scala/caliban/PlayJson.scala
```diff
@@ -0,0 +1,123 @@
...
+    case v: Value                          => valueToJson(v)
+    case ResponseValue.ListValue(values)   => JsArray(values.map(responseValueToJson))
+    case ResponseValue.ObjectValue(fields) => JsObject(fields.map { case (k, v) => k -> responseValueToJson(v) })
+    case s: ResponseValue.StreamValue      => JsString(s.toString)
+  }
+
+  private def valueToJson(value: Value): JsValue = value match {
+    case Value.NullValue     => JsNull
+    case value: IntValue     => JsNumber(BigDecimal(value.toBigInt.bigInteger))
+    case value: FloatValue   => JsNumber(value.toBigDecimal)
```
paulpdaniels: Don't these types autoconvert?

play/src/main/scala/caliban/PlayJson.scala
```diff
@@ -0,0 +1,123 @@
...
+import scala.util.Try
+
+object PlayJson {
+
+  //// JSON Reads
+  private def jsonToInputValue(json: JsValue): InputValue = json match {
+    case JsNull             => NullValue
+    case boolean: JsBoolean => BooleanValue(boolean.value)
+    case JsNumber(value) =>
+      Try(value.toIntExact).fold(_ => FloatValue(value), IntValue(_))
```
paulpdaniels: I'm actually not sure if this conversion is always true, but `Try` seems kind of expensive if it is a `Float` since it will have to capture a thrown `Exception` seems like it could just split this into two case blocks:

```scala
case JsNumber(value) if value.isWhole => IntValue(value.intValue)
case JsNumber(value) => FloatValue(value)
```

play/src/main/scala/caliban/CalibanController.scala
```diff
@@ -0,0 +1,22 @@
...
+    Action.async(parse.json[GraphQLRequest]) { req =>
+      runtime
+        .unsafeRunToFuture(
+          interpreter
+            .execute(req.body.query, req.body.operationName, req.body.variables.getOrElse(Map.empty))
+            .map(Ok(_))
```
paulpdaniels: Need to handle the `E` case as well.

play/src/main/scala/caliban/PlayJson.scala
yoohaemin: Ah thanks for the tip

play/src/main/scala/caliban/PlayJson.scala
yoohaemin: My intention was to make it very loud and clear that there's a side effect happening, however local and contained it is, but it's not a big deal to change too.

play/src/main/scala/caliban/PlayJson.scala
yoohaemin: Could you elaborate? I don't see how `play-json` would know how to deal with `caliban` values

play/src/main/scala/caliban/PlayJson.scala
yoohaemin: Great suggestion! But I think checking only with `isWhole` is not safe, I could use `isValidXXX` methods alongside.

play/src/main/scala/caliban/CalibanController.scala
```diff
@@ -0,0 +1,25 @@
...
+class PlayCaliban[R](runtime: Runtime[R]) {
+
+  class Controller @Inject() (cc: ControllerComponents) extends AbstractController(cc) {
+
+    def action[E](interpreter: GraphQLInterpreter[R, E]): Action[GraphQLRequest] =
+      Action.async(parse.json[GraphQLRequest]) { req =>
+        runtime
+          .unsafeRunToFuture(
+            interpreter
+              .execute(req.body.query, req.body.operationName, req.body.variables.getOrElse(Map.empty))
```
paulpdaniels: Potential follow up areas: supporting the persisted-query and the file upload specs.

play/src/main/scala/caliban/PlayJson.scala
```diff
@@ -0,0 +1,105 @@
...
+    case ce: CalibanError => errorValueEncoder.writes(ce)
+    case _                => Json.obj("message" -> JsString(err.toString))
+  }
+
+  // CalibanError
+  private implicit val locationToJson: Writes[LocationInfo] =
+    li => Json.obj("locations" -> Json.arr(Json.obj("line" -> li.line, "column" -> li.column)))
+
+  private def encodeLocationInfoAndMessage(li: Option[LocationInfo], message: String) =
+    li.fold(Json.obj())(li => Json.obj("locations" -> li)) ++ Json.obj("message" -> message)
```
paulpdaniels: Can simplify to `Json.obj("message" -> message, "locations" -> li)`

yoohaemin: Won't that change the behavior?

yoohaemin: On second thought, it does change the behavior however trivial it is. I feel better leaving it behave the same with circe, wdyt?

paulpdaniels: How does the behavior change?

paulpdaniels: Actually why does the locationinfo writes also pass it’s key name?

yoohaemin: Thanks for the catch! 
I meant your simplification would add field `"locations": null`, when `LocationInfo` is empty, where currently it drops the key altogether. 
Current circe impl drops the null values.

ghostdogpr:
This [change](https://github.com/ghostdogpr/caliban/pull/260) needs to be applied in this adapter too.

vpavkin:
@yoohaemin Hi!
FYI, play-json support was merged. I didn't touch controllers though, so I guess this PR is still viable, just with a more narrow scope.

ghostdogpr:
Closing this one as most code related to play-json has been merged already. The issue remains open for creating the adapter part.

Issue #229, opened 2020-02-22
Schema validation: interfaces


ghostdogpr:
In `Validator.validateSchema`, verify the [rules for interface types](https://spec.graphql.org/June2018/#sec-Interfaces)

1. An Interface type must define one or more fields.
2. For each field of an Interface type:
   1. The field must have a unique name within that Interface type;
      no two fields may share the same name.
   2. The field must not have a name which begins with the
      characters {"__"} (two underscores).
   3. The field must return a type where {IsOutputType(fieldType)}
      returns {true}.
   4. For each argument of the field:
      1. The argument must not have a name which begins with the
         characters {"__"} (two underscores).
      2. The argument must accept a type where {IsInputType(argumentType)}
         returns {true}.

TobiasPfeifer:
I'd like to work on this

TobiasPfeifer:
From the given rules its not 100% clear to me if empty names for fields and arguments are valid. Empty names don't make sense imo though. Other thoughts on this?

ghostdogpr:
@TobiasPfeifer I don't think we need to check for empty names because they are derived from Scala case class parameters and those can't be empty anyway.

ghostdogpr:
Closed by #243

Issue #228, opened 2020-02-22
Schema validation: objects


ghostdogpr:
In `Validator.validateSchema`, verify the [rules for object types](https://spec.graphql.org/June2018/#sec-Objects)

1. An Object type must define one or more fields.
2. For each field of an Object type:
   1. The field must have a unique name within that Object type;
      no two fields may share the same name.
   2. The field must not have a name which begins with the
      characters {"__"} (two underscores).
   3. The field must return a type where {IsOutputType(fieldType)} returns {true}.
   4. For each argument of the field:
      1. The argument must not have a name which begins with the
         characters {"__"} (two underscores).
      2. The argument must accept a type where {IsInputType(argumentType)}
         returns {true}.
4. An object type may declare that it implements one or more unique interfaces.
5. An object type must be a super-set of all interfaces it implements:
   1. The object type must include a field of the same name for every field
      defined in an interface.
      1. The object field must be of a type which is equal to or a sub-type of
         the interface field (covariant).
         1. An object field type is a valid sub-type if it is equal to (the same
            type as) the interface field type.
         2. An object field type is a valid sub-type if it is an Object type and
            the interface field type is either an Interface type or a Union type
            and the object field type is a possible type of the interface field
            type.
         3. An object field type is a valid sub-type if it is a List type and
            the interface field type is also a List type and the list-item type
            of the object field type is a valid sub-type of the list-item type
            of the interface field type.
         4. An object field type is a valid sub-type if it is a Non-Null variant
            of a valid sub-type of the interface field type.
      2. The object field must include an argument of the same name for every
         argument defined in the interface field.
         1. The object field argument must accept the same type (invariant) as
            the interface field argument.
      3. The object field may include additional arguments not defined in the
         interface field, but any additional argument must not be required, e.g.
         must not be of a non-nullable type.

zach-albia:
Working on this if someone isn't already.

PR #243, opened 2020-02-25
Schema validation: interfaces #229

TobiasPfeifer:
initial schema validation method layout for interfaces for #229 

@ghostdogpr can you do an early stage review of `private def validateInterface(t: __Type)`


core/src/main/scala/caliban/validation/Validator.scala
```diff
@@ -563,6 +564,49 @@ object Validator {
...
+        duplicateFieldName(fields) match {
+          case Some(value) =>
+            IO.fail(
+              ValidationError(
+                s"message with $value",
+                "explanatory"
+              )
+            )
+          case None =>
+            fields.foldRight[IO[ValidationError, Unit]](IO.unit) { (field, result) =>
```
ghostdogpr: I think you can simply do `IO.foreach(fields){ ...}.unit` here

core/src/main/scala/caliban/validation/Validator.scala
```diff
@@ -563,6 +564,49 @@ object Validator {
...
+    def validateFieldArgumentsName(args: List[__InputValue]): IO[ValidationError, Unit]      = ???
+    def validateFieldArgumentsInputType(args: List[__InputValue]): IO[ValidationError, Unit] = ???
+
+    def validateInterfaceField(field: __Field): IO[ValidationError, Unit] =
+      for {
+        _ <- validateFieldName(field)
+        _ <- validateFieldReturnOutputType(field.`type`)
+        _ <- validateFieldArgumentsName(field.args)
+        r <- validateFieldArgumentsInputType(field.args)
+      } yield r
```
ghostdogpr: You can just `yield ()` and no need to have `r`

core/src/main/scala/caliban/validation/Validator.scala
```diff
@@ -563,6 +564,49 @@ object Validator {
...
+
+    def duplicateFieldName(fields: List[__Field]): Option[__Field] =
+      fields.groupBy(_.name).collectFirst { case (_, List(f, _, _*)) => f }
```
ghostdogpr: slightly shorter: `collectFirst { case (_, f :: _ :: _) => f `

TobiasPfeifer:
@ghostdogpr  I've implemented fieldname and argumentname validation. I'm now thinking about return type and input type validation. Should I try to lift this validation to the type level?

A `__Field` could define a `` `type` `` of type `__ReturnType` that has a `kind: __ReturnTypeKind` that only has instances of the valid return type kinds.
Same for the `__InputValue` thats `` `type` `` can be `__InputType` that has `kind: __InputTypeKind`

However it's possible for me to leave the types as they are and validate on them at runtime

ghostdogpr:
@TobiasPfeifer doing it at the type level would be better in the general case, however `__Type` and other classes prefixed with `__` are modeled the [introspection schema](https://spec.graphql.org/June2018/#sec-Schema-Introspection) and they are directly used for introspection in Caliban. If you use a sealed trait, `__Type` would become a union or an interface and wouldn't match with the schema from the spec. For that reason I think we have to stick to that and perform the check at runtime. Makes sense?

core/src/main/scala/caliban/validation/Validator.scala
```diff
@@ -563,6 +564,60 @@ object Validator {
         )
     }
 
+  private def validateInterface(t: __Type): IO[ValidationError, Unit] = {
+    def validateName(name: String) = {
+      if (name.isEmpty) {
```
ghostdogpr: You could use `IO.whenCase` to make this even shorter and avoid the `else IO.unit`

zach-albia:
@ghostdogpr Something came up so I haven't gotten around to pushing a WIP PR #228 yet. Looking to get back to it this Saturday.

ghostdogpr:
@zach-albia no problem. There are a few rules in common in your 2 issues (and the one @mriceron is doing too) so keep an eye on the other PRs. Whoever does it first, the others can reuse 😄

TobiasPfeifer:
@ghostdogpr I'll add tests to ensure correct ValidationError when calling `.interpreter`

> @TobiasPfeifer doing it at the type level would be better in the general case, however `__Type` and other classes prefixed with `__` are modeled after the [introspection schema](https://spec.graphql.org/June2018/#sec-Schema-Introspection) and they are directly used for introspection in Caliban. If you use a sealed trait, `__Type` would become a union or an interface and wouldn't match with the schema from the spec. For that reason I think we have to stick to that and perform the check at runtime. Makes sense?

Sure. I'll implement runtime validation with this PR. If I feel challenged to find a way to do this on the type level without altering the `__XYZ` types for the introspection schema, I'll tackle this in a separate PR

ghostdogpr:
#244 was merged so you can re-use the same file for tests (and maybe some of the code)

core/src/main/scala/caliban/validation/Validator.scala
TobiasPfeifer: this got replaced by reusing existing code of #244

TobiasPfeifer:
@ghostdogpr I restructured my code to match the structure of #244 and reused common validation methods

Regarding the tests: I think it's not possible to construct an `Interface` that has a `field` returning an `INPUT_OBJECT`, so there is no test for that rule. Same for the rule "no two fields may share the same name"

core/src/main/scala/caliban/validation/Validator.scala
```diff
@@ -545,61 +546,117 @@ object Validator {
...
-        .collectFirst { case (_, f :: _ :: _) => f }
-        .fold[IO[ValidationError, Unit]](IO.unit)(duplicateField =>
-          failValidation(
-            s"InputObject has repeated fields: ${duplicateField.name}",
-            "The input field must have a unique name within that Input Object type; no two input fields may share the same name"
-          )
+    t.fields(__DeprecatedArgs(Some(true))) match {
+      case None | Some(Nil) =>
+        failValidation(
+          s"message",
```
ghostdogpr: Add proper messages here

core/src/main/scala/caliban/validation/Validator.scala
```diff
@@ -545,61 +546,117 @@ object Validator {
...
-          "An Input Object type must define one or more input fields"
+          s"${errorType.name.getOrElse("")} is of kind ${errorType.kind}, must be an OutputType",
+          """The input field must accept a type where IsOutputType(type) returns true, https://spec.graphql.org/June2018/#IsInputType()"""
         )
-      case Some(fields) => validateFields(fields)
     }
   }
 
+  private def noDuplicateFieldName(fields: List[__Field]) =
+    noDuplicateName[__Field](fields, _.name)
```
ghostdogpr: Is it necessary to set the type explicitly?

core/src/main/scala/caliban/validation/Validator.scala
```diff
@@ -545,61 +546,117 @@ object Validator {
...
+  private def noDuplicateInputValueName(inputValues: List[__InputValue]) =
+    noDuplicateName[__InputValue](inputValues, _.name)
+
+  private def noDuplicateName[T](listOfNamed: List[T], nameExtractor: T => String): IO[ValidationError, Unit] =
+    listOfNamed
+      .groupBy(nameExtractor(_))
+      .collectFirst { case (_, f :: _ :: _) => f }
+      .fold[IO[ValidationError, Unit]](IO.unit)(duplicate =>
+        failValidation(
+          s"InputObject has repeated fields: ${nameExtractor(duplicate)}",
```
ghostdogpr: The message here seems okay for input objects but not for interfaces.

core/src/main/scala/caliban/validation/Validator.scala
```diff
@@ -545,61 +546,117 @@ object Validator {
...
+          "The input field must have a unique name within that Input Object type; no two input fields may share the same name"
+        )
+      )
+
+  private def doesNotStartWithUnderscore(name: String, typeName: String, inType: String): IO[ValidationError, Unit] =
+    IO.when(name.startsWith("__"))(
+      failValidation(
+        s"A $typeName in $inType can't start with '__': $name",
```
ghostdogpr: Maybe remove "A " to avoid the having the grammatical mistake "A Argument"?

core/src/main/scala/caliban/validation/Validator.scala
TobiasPfeifer: `noDuplicateName(fields, _.name)' and `noDuplicateName(fields, arg => arg.name)'  results in "missing parameter type"

`noDuplicateName(fields, (arg: __Field) => arg.name)` is compiling fine but longer than explicitly setting type parameter T on the method

I actually don't know why the compiler can't infer the type for T. Is there a better way to do it?

ghostdogpr: Ah OK, that's not a big deal. I thought the compiler would infer it but if it doesn't, no problem with leaving it.

core/src/main/scala/caliban/validation/Validator.scala
TobiasPfeifer: added arguments `messageBuilder: T=> String` and `explanatoryText: String` for the caller to build the messages

core/src/test/scala/caliban/validation/ValidationSchemaSpec.scala
```diff
@@ -13,12 +13,12 @@ object ValidationSchemaSpec
...
         assertM(gql.interpreter.run, fails[ValidationError](hasField("msg", _.msg, equalTo(expectedMessage))))
 
-      suite("ValidationSchemaSpec")({
+      suite("ValidationSchemaSpec")(
         suite("InputObjects")(
           testM("name can't start with '__'") {
             check(
               graphQL(resolverWrongMutationUnderscore),
-              "InputObject can't start with '__': __name"
+              "input value in InputObject can't start with '__': __name"
```
ghostdogpr: I’m nitpicking but can you make the first letter uppercase, to be consistent with the rest?

TobiasPfeifer: what is your opinion on the `nameExtractor` and `messageBuilder` stuff? I could do the same for this validation with overloaded methods one for `__Field` and one for `__InputValue`

core/src/main/scala/caliban/validation/Validator.scala
```diff
@@ -545,61 +546,130 @@ object Validator {
...
-          s"InputObject ${t.name.getOrElse("")} does not have fields",
-          "An Input Object type must define one or more input fields"
+          s"${errorType.name.getOrElse("")} is of kind ${errorType.kind}, must be an OutputType",
+          """The input field must accept a type where IsOutputType(type) returns true, https://spec.graphql.org/June2018/#IsInputType()"""
         )
-      case Some(fields) => validateFields(fields)
     }
   }
 
+  private def noDuplicateFieldName(fields: List[__Field]) = {
```
ghostdogpr: Maybe should be called `noDuplicateFieldInterface` since the same will be needed for object types. 
Or move it inside the function the validates interfaces to avoid confusion.

core/src/test/scala/caliban/validation/ValidationSchemaSpec.scala
ghostdogpr: I don't have a strong opinion on it, you can do as you feel. As long as the result message looks good 😄

core/src/main/scala/caliban/validation/Validator.scala
TobiasPfeifer: moved inside the interface validation method

core/src/test/scala/caliban/validation/ValidationSchemaSpec.scala
TobiasPfeifer: I improved the error messages to also provide the context where the validation error occurred
The interface and input object validation error messages follow the same schema now and lead to cleaner code

Issue #231, opened 2020-02-22
Schema validation: input objects


ghostdogpr:
In `Validator.validateSchema`, verify the [rules for input object types](https://spec.graphql.org/June2018/#sec-Input-Objects)

1. An Input Object type must define one or more input fields.
2. For each input field of an Input Object type:
   1. The input field must have a unique name within that Input Object type;
      no two input fields may share the same name.
   2. The input field must not have a name which begins with the
      characters {"__"} (two underscores).
   3. The input field must accept a type where {IsInputType(inputFieldType)}
      returns {true}.

mriceron:
I can work on this if no ones is

ghostdogpr:
@mriceron sure! 👍

PR #244, opened 2020-02-27
Add validation for input objects

mriceron:
Closes https://github.com/ghostdogpr/caliban/issues/231
Not sure how to test the duplicate names rule without building an invalid Input Object by hand.

### DIFF
--- a/core/src/main/scala/caliban/validation/Validator.scala
+++ b/core/src/main/scala/caliban/validation/Validator.scala
@@ -39,6 +39,9 @@ object Validator {
       }
       .unit
 
+  def failValidation[T](msg: String, explanatoryText: String): IO[ValidationError, T] =
+    IO.fail(ValidationError(msg, explanatoryText))
+
   /**
    * Prepare the request for execution.
    * Fails with a [[caliban.CalibanError.ValidationError]] otherwise.
@@ -70,19 +73,19 @@ object Validator {
       }
 
       operation match {
-        case Left(error) => IO.fail(ValidationError(error, ""))
+        case Left(error) => failValidation(error, "")
         case Right(op) =>
           (op.operationType match {
             case Query => IO.succeed(rootSchema.query)
             case Mutation =>
               rootSchema.mutation match {
                 case Some(m) => IO.succeed(m)
-                case None    => IO.fail(ValidationError("Mutations are not supported on this schema", ""))
+                case None    => failValidation("Mutations are not supported on this schema", "")
               }
             case Subscription =>
               rootSchema.subscription match {
                 case Some(m) => IO.succeed(m)
-                case None    => IO.fail(ValidationError("Subscriptions are not supported on this schema", ""))
+                case None    => failValidation("Subscriptions are not supported on this schema", "")
               }
           }).map(operation =>
             ExecutionRequest(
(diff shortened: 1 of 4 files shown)


core/src/main/scala/caliban/validation/Validator.scala
```diff
@@ -591,6 +592,67 @@ object Validator {
...
 
+  private def validateInputObject(t: __Type): IO[ValidationError, Unit] = {
+    // https://spec.graphql.org/June2018/#IsInputType()
+    def isInputType(t: __Type): Either[__Type, Unit] = t.kind match {
+      case __TypeKind.LIST | __TypeKind.NON_NULL                         => t.ofType.fold[Either[__Type, Unit]](Left(t))(isInputType)
+      case __TypeKind.SCALAR | __TypeKind.ENUM | __TypeKind.INPUT_OBJECT => Right(())
+      case _                                                             => Left(t)
+    }
+
+    def failValidation(msg: String, explanatoryText: String): IO[ValidationError, Unit] =
```
ghostdogpr: If that's not too much a hassle, could you extract it and use it elsewhere in this file?

core/src/main/scala/caliban/validation/Validator.scala
```diff
@@ -591,6 +592,67 @@ object Validator {
...
+      IO.fail(ValidationError(msg, explanatoryText))
+
+    def validateFields(fields: List[__InputValue]): IO[ValidationError, Unit] =
+      duplicateFieldName(fields).flatMap(_ =>
```
ghostdogpr: `.flatMap(_ =>` could be just `<*`, which would let you remove `.unit` too

core/src/main/scala/caliban/validation/Validator.scala
```diff
@@ -591,6 +592,67 @@ object Validator {
 
   }
 
+  private def validateInputObject(t: __Type): IO[ValidationError, Unit] = {
+    // https://spec.graphql.org/June2018/#IsInputType()
+    def isInputType(t: __Type): Either[__Type, Unit] = t.kind match {
+      case __TypeKind.LIST | __TypeKind.NON_NULL                         => t.ofType.fold[Either[__Type, Unit]](Left(t))(isInputType)
```
ghostdogpr: `Option[__Type]` would be simpler no?

ghostdogpr:
Ah also I think it's okay not to test the duplicate names rule since Caliban will only generate something that doesn't break the rule.

core/src/main/scala/caliban/validation/Validator.scala
mriceron: I didn't like it since it's not super clear, by just looking at it, when it's an error or not (having `Some` would be the error state)

mriceron:
> Ah also I think it's okay not to test the duplicate names rule since Caliban will only generate something that doesn't break the rule.

Yeah I thought about that after writing it , I'll remove it

core/src/main/scala/caliban/validation/Validator.scala
```diff
@@ -572,25 +529,78 @@ object Validator {
...
+      case __TypeKind.SCALAR | __TypeKind.ENUM | __TypeKind.INPUT_OBJECT => Right(())
+      case _                                                             => Left(t)
+    }
+
+    def validateFields(fields: List[__InputValue]): IO[ValidationError, Unit] =
+      duplicateFieldName(fields).<*(
```
ghostdogpr: why not just `<*` without `.` and `()`?

mriceron: No reason :)

PR #246, opened 2020-02-27
Allow errors to send an error code in the response extension

### LINKED ISSUE #158: Errors: support extensions
Add support for the `extensions` data for each error as defined in the [GraphQL spec](https://graphql.github.io/graphql-spec/June2018/#sec-Errors):
```json
{
  "errors": [
    {
      "message": "Name for character with ID 1002 could not be fetched.",
      "locations": [ { "line": 6, "column": 7 } ],
      "path": [ "hero", "heroFriends", 1, "name" ],
      "extensions": {
        "code": "CAN_NOT_FETCH_BY_ID",
        "timestamp": "Fri Feb 9 14:33:09 UTC 2018"
      }
    }
  ]
}
```

See initial discussion in https://github.com/ghostdogpr/caliban/issues/140

DJLemkes:
A first attempt to allow errors to define an error code which is renderd in the `extensions` section of error elements. Closely related to #158 

I have no idea whether this is solution is on the right track. If not, no hard feelings on rejections :). If it is on the right track, I can put in a little more effort to include more error information as mentioned in #158

Closes #158

### DIFF
--- a/core/src/main/scala/caliban/CalibanError.scala
+++ b/core/src/main/scala/caliban/CalibanError.scala
@@ -1,5 +1,6 @@
 package caliban
 
+import caliban.ResponseValue.ObjectValue
 import caliban.interop.circe.IsCirceEncoder
 import caliban.parsing.adt.LocationInfo
 
@@ -18,7 +19,8 @@ object CalibanError {
   case class ParsingError(
     msg: String,
     locationInfo: Option[LocationInfo] = None,
-    innerThrowable: Option[Throwable] = None
+    innerThrowable: Option[Throwable] = None,
+    extensions: Option[ObjectValue] = None
   ) extends CalibanError {
     override def toString: String = s"Parsing Error: $msg ${innerThrowable.fold("")(_.toString)}"
   }
@@ -26,8 +28,12 @@ object CalibanError {
   /**
    * Describes an error that happened while validating a query.
    */
-  case class ValidationError(msg: String, explanatoryText: String, locationInfo: Option[LocationInfo] = None)
-      extends CalibanError {
+  case class ValidationError(
+    msg: String,
+    explanatoryText: String,
+    locationInfo: Option[LocationInfo] = None,
+    extensions: Option[ObjectValue] = None
+  ) extends CalibanError {
     override def toString: String = s"ValidationError Error: $msg"
   }
 
@@ -38,7 +44,8 @@ object CalibanError {
     msg: String,
     path: List[Either[String, Int]] = Nil,
     locationInfo: Option[LocationInfo] = None,
-    innerThrowable: Option[Throwable] = None
+    innerThrowable: Option[Throwable] = None,
+    extensions: Option[ObjectValue] = None
   ) extends CalibanError {
     override def toString: String = s"Execution Error: $msg ${innerThrowable.fold("")(_.toString)}"
   }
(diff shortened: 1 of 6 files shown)


ghostdogpr:
@DJLemkes thanks for volunteering on this! I think the idea of using a trait extending `Throwable` works. But I would make the function return a `Option[ObjectValue]` (the same type we used for extensions in `GraphQLResponse`). That way, users are able to return a `Map` containing anything they want, instead of us forcing them to return something specific (extensions have no specified fields, anything could be used).

I had another approach in mind, let me explain it and you can tell me what you think. If we added `Option[ObjectValue]` (or maybe directly `ObjectValue`) directly in `CalibanError` (and each of the 3 case classes), users would be able to use `mapError` on the interpreter and put anything they want into it without having to use a custom `Throwable` and modify their business code. They could easily remove the items Caliban adds (e.g. explanation text for validation errors) or add their own based on the type of innerThrowable which could be their own business error type. What do you think?

Tip: run `fmt` in sbt to format sources and allow the CI to run the tests

DJLemkes:
@ghostdogpr Like your idea better! Will give it a shot in the next couple of days.

DJLemkes:
@ghostdogpr sorry, had the flue hence the delay. Could you have a look and tell me if this is what you had in mind?

ghostdogpr:
@DJLemkes no problem, I'll have a closer look tomorrow!

core/src/main/scala/caliban/CalibanError.scala
```diff
@@ -55,25 +62,27 @@ private object ErrorCirce {
...
-    case CalibanError.ParsingError(msg, locationInfo, _) =>
+    case CalibanError.ParsingError(msg, locationInfo, _, extensions) =>
       Json
         .obj(
           "message" -> s"Parsing Error: $msg".asJson,
           "locations" -> Some(locationInfo).collect {
             case Some(li) => Json.arr(locationToJson(li))
-          }.asJson
+          }.asJson,
+          "extensions" -> extensions.asInstanceOf[Option[ResponseValue]].asJson.dropNullValues
```
ghostdogpr: `(extensions: Option[ResponseValue])` would be better than `asInstanceOf` (here and below)

examples/src/main/scala/caliban/http4s/ExampleApp.scala
```diff
@@ -56,15 +70,29 @@ object ExampleApp extends CatsApp with GenericSchema[Console with Clock] {
       maxDepth(30) @@                 // query analyzer that limit query depth
       timeout(3 seconds) @@           // wrapper that fails slow queries
       printSlowQueries(500 millis) @@ // wrapper that logs slow queries
-      apolloTracing                   // wrapper for https://github.com/apollographql/apollo-tracing
+      apolloTracing // wrapper for https://github.com/apollographql/apollo-tracing
+
+  def withErrorCodeExtensions[R](
```
ghostdogpr: Great example! However I think it makes the "sample" use case a bit too complex. I also would like to keep the http4s, akka and finch examples identical. Instead of adding this everywhere, how about moving it to a unit test, or making a dedicated example here https://github.com/ghostdogpr/caliban/tree/master/examples/src/main/scala/caliban

core/src/main/scala/caliban/CalibanError.scala
DJLemkes: Makes a lot of sense but I can't seem to get the encoder for an `Option[ResponseValue]` in implicit scope such that `.asJson` works. Thought that the implicit `circeEncoder` of `ResponseValue` would be able to pick this one up? I got the `.asInstanceOf[ResponseValue]` trick from `GraphQLResponse` actually.

examples/src/main/scala/caliban/http4s/ExampleApp.scala
DJLemkes: Removed the example code and created a little alinea in the docs for it. If that doesn't work I can move it to an example. Started with docs because creating an example for it felt quite a big thing to do straight away.

core/src/main/scala/caliban/CalibanError.scala
ghostdogpr: `(extensions: Option[ResponseValue]).asJson.dropNullValues` doesn't work? We were about to remove the `asInstanceOf` in this PR by using that same trick: https://github.com/ghostdogpr/caliban/pull/234/files#diff-16ab822f0162e843af921d1e9d0c34dfR23

vuepress/docs/docs/middleware.md
```diff
@@ -92,4 +92,32 @@ val i4: GraphQLInterpreter[MyEnv with Clock, CalibanError] =
       _.getOrElse(GraphQLResponse(NullValue, List(ExecutionError("Timeout!"))))
     )
   )
+```
+
+## Customizing error responses
+
+During various phases of executing a query, an error may occur. Caliban renders the different instances of `CalibanError` to a GraphQL spec compliant response. As a user, you will most likely encounter `ExecutionError` at some point because this will encapsulate the errors in the error channel of your effects. For Caliban to be able to render some basic message about the error that occured during query execution, it is important that your errror extends `Throwable`.
```
ghostdogpr: typo: errror

vuepress/docs/docs/middleware.md
```diff
@@ -92,4 +92,32 @@ val i4: GraphQLInterpreter[MyEnv with Clock, CalibanError] =
...
+
+For more meaningful error handling, GraphQL spec allows for an [`extension`](http://spec.graphql.org/June2018/#example-fce18) object in the error response. This object may include, for instance, `code` information to model enum-like error codes that can be handled by a front-end. In order to generate this information, one can use the `mapError` function on a `GraphQLInterpreter`. An example is provided below in which we map a custom domain error within an `ExecutionError` to a meaningful error code.
```
ghostdogpr: Nice doc!

core/src/main/scala/caliban/CalibanError.scala
DJLemkes: Aha, I was trying to put it in the match clause. Apparently didn't fully understand what you meant. Changed it now and it works!

vuepress/docs/docs/middleware.md
DJLemkes: fixed.

ghostdogpr:
@DJLemkes thanks for your contribution!

DJLemkes:
@ghostdogpr Thanks for you reviews and patience! And for creating the library of course :)

PR #253, opened 2020-02-29
Fix input values encoding

### LINKED ISSUE #248: Error when receiving a response with no data field
Example of valid response giving a decoding failure: 
```json
{"errors":[{"message":"Parse error on \"direction\" (STRING) at [1, 107]","locations":[{"line":1,"column":107}]}]}
```

This is because we always expect `data`.

### LINKED ISSUE #249: Input objects are not properly encoded
Input object fields are surrounded with double quotes, but they should not.

`"field": value` should be `field: value`

The problem is that we use `Value#toString` to encode input values, but that gives a JSON-like representation which is not the one we want here.

### LINKED ISSUE #250: Input enums should not be surrounded with quotes
An input enum is encoded into `direction:"DESC"` instead of `direction:DESC`.

ghostdogpr:
Fixes #249 
Fixes #250

### DIFF
--- a/client/src/main/scala/caliban/client/Value.scala
+++ b/client/src/main/scala/caliban/client/Value.scala
@@ -14,6 +14,9 @@ object Value {
   case class NumberValue(value: BigDecimal) extends Value {
     override def toString: String = s"$value"
   }
+  case class EnumValue(value: String) extends Value {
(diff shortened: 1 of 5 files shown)


kitlangton:
👍 Fixes the problem for me!

PR #257, opened 2020-03-01
Resolve and validate variables deeply inside lists and objects

### LINKED ISSUE #255: Validation error when a variable is used inside an array or an object
This mutation triggers a validation error: "Variable 'day' is not used" even though it is used.

```graphql
mutation CreateDay($day: NewDayInput!) {
    modifyDays{
        create(items: [$day]){
            __typename
        }
    }
}
```

Krever:
Fixes #255 
In the process of fixing validation, I found issue with variables resolution in execution itself.

### DIFF
--- a/core/src/main/scala/caliban/execution/Executor.scala
+++ b/core/src/main/scala/caliban/execution/Executor.scala
@@ -153,8 +153,9 @@ object Executor {
   ): Map[String, InputValue] = {
     def resolveVariable(value: InputValue): InputValue =
       value match {
-        case InputValue.ListValue(values)   => InputValue.ListValue(values.map(resolveVariable))
-        case InputValue.ObjectValue(fields) => InputValue.ObjectValue(fields.mapValues(resolveVariable).toMap)
+        case InputValue.ListValue(values) => InputValue.ListValue(values.map(resolveVariable))
+        case InputValue.ObjectValue(fields) =>
+          InputValue.ObjectValue(fields.map({ case (k, v) => k -> resolveVariable(v) }))
         case InputValue.VariableValue(name) =>
           lazy val defaultInputValue = (for {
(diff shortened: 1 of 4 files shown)


core/src/main/scala/caliban/execution/Executor.scala
```diff
@@ -151,19 +150,22 @@ object Executor {
...
+          InputValue.ObjectValue(fields.map({ case (k, v) => k -> resolveVariable(v) }))
+        case InputValue.VariableValue(name) =>
+          lazy val defaultInputValue = (for {
+            definition <- variableDefinitions.find(_.name == name)
+            inputValue <- definition.defaultValue
+          } yield inputValue) getOrElse value
+          variableValues.getOrElse(name, defaultInputValue)
+        case value: Value => value
+      }
+    arguments.map({ case (k, v) => k -> resolveVariable(v) })
```
ghostdogpr: This is nitpicking but why the parenthesis around the braces? You find this more readable? I usually keep only the braces.

Krever: Recently I grew a strong reluctance against whitespace-based function application. It feels to irregular and prone to issues (e.g. the order of execution) and so I try to avoid it. 
It's even worse for such an application of functions expecting `PartialFunction` as an argument. I haven't yet seen a junior developer who properly understands that `map { case xx => ..|}`  is, in fact, a partial function literal passed to a `map`. So I prefer to make it explicit and have the regular syntax across the codebase. 

But please, feel free to adjust to your style preference, I wouldn't mind at all!

ghostdogpr: Interesting point. I'm so used to `map { }` that I haven't considered that. I don't mind keeping it that way.

PR #259, opened 2020-03-03
Schema validation for GraphQL objects

### LINKED ISSUE #228: Schema validation: objects

zach-albia:
This is for #228, I'm reusing code written by @TobiasPfeifer for the other schema validation as there's a significant amount of overlap in the rules. I'm planning on adding the rest of the rules in the next commits.

Also, @ghostdogpr I'm looking to add some tests for the whole of schema validation but I'm a little stuck and am going to need your input to get it going.

Closes #228


zach-albia:
@ghostdogpr can you come take a look at the test errors for jdk8 and jdk11? Looks like it has to do with the UzHttpAdapter. I don't think it has anything to do with the pull request.

ghostdogpr:
@zach-albia hmm the UzHttpAdapter is not merged to master. Did you merge the wrong branch?

zach-albia:
@ghostdogpr ah it makes sense now. I merged it wrong somewhere, but somehow still ended up with the UzHttpAdapter. My bad.

core/src/main/scala/caliban/validation/Validator.scala
```diff
@@ -561,6 +561,7 @@ object Validator {
             types.filterNot(isObject).map(_.name.getOrElse("")).filterNot(_.isEmpty).mkString("", ", ", "."),
           s"The member types of a Union type must all be Object base types."
         )
+      // shouldn't we be validating each object here?
       case _ => IO.unit
```
zach-albia, line 565: @ghostdogpr just noticed `validateUnion` doesn't check if each object member of the union is valid. We should run `validateObject` on the object members, right?

ghostdogpr: I think they are already validated, because we run the validation on the list of all types that we build recursively.

ghostdogpr:
Hey @zach-albia, how is it going? 😄 I see you added a lot of tests, that's great. Do you think it will be done soon? Let me know if you need any help.

zach-albia:
Hey @ghostdogpr, sorry for this taking so long. I'm looking to finish it on or before the 28th.  I will let you know if I need help.

ghostdogpr:
No problem. Thanks for working on this 🙏

zach-albia:
@ghostdogpr please check 🙏

core/src/main/scala/caliban/introspection/adt/__TypeKind.scala
```diff
@@ -11,4 +11,19 @@ object __TypeKind {
...
+      case __TypeKind.LIST         => 3
+      case __TypeKind.UNION        => 4
+      case __TypeKind.ENUM         => 5
+      case __TypeKind.INPUT_OBJECT => 6
+      case __TypeKind.INTERFACE    => 7
+      case __TypeKind.OBJECT       => 8
+    }
+
+  implicit val renderOrdering: Ordering[__Type] =
+    Ordering.by(o => (o.kind, o.name.getOrElse("")))
```
ghostdogpr, line 28: Minor: call it `typeOrdering`

core/src/main/scala/caliban/validation/Validator.scala
```diff
@@ -30,17 +31,22 @@ object Validator {
...
-    IO.foreach(types) { t =>
-      t.kind match {
-        case __TypeKind.ENUM         => validateEnum(t)
-        case __TypeKind.UNION        => validateUnion(t)
-        case __TypeKind.INTERFACE    => validateInterface(t)
-        case __TypeKind.INPUT_OBJECT => validateInputObject(t)
-        case _                       => IO.unit
-      }
-    } *> validateClashingTypes(types) *> validateDirectives(types) *> validateRootQuery(schema)
+    IO.foreach(types.sorted(renderOrdering))(validateType) *>
```
ghostdogpr, line 34: Is it necessary to pass `renderOrdering`? The implicit should be found no?

core/src/main/scala/caliban/validation/Validator.scala
```diff
@@ -588,45 +594,122 @@ object Validator {
...
+
+        def isValidSubtype(supertypeFieldType: __Type, objectFieldType: __Type) = {
+          val supertypePossibleTypes = supertypeFieldType.possibleTypes.toList.flatten
+          (supertypeFieldType == objectFieldType) || supertypePossibleTypes.contains(objectFieldType)
+        }
+
+        IO.foreach_(objectFields) { objField =>
+          val fieldContext = s"Field '${objField.name}'"
+
+          supertypeFields.find(_.name == objField.name).fold[IO[ValidationError, Unit]](IO.unit) { superField =>
```
ghostdogpr, line 656: You could use `IO.whenCase` to simplify this a bit:
```scala
IO.whenCase(supertypeFields.find(_.name == objField.name)) {
  case Some(superField) =>
    ...
```

core/src/main/scala/caliban/validation/Validator.scala
```diff
@@ -588,45 +594,122 @@ object Validator {
...
+                  "An object list item field type must be equal to or a possible" +
+                    " type of the interface list item field type."
+                )
+              case _ if !extraArgsAreValid =>
+                val argNames = extraArgs.filter(_.`type`().kind == __TypeKind.NON_NULL).map(_.name).mkString(", ")
+                failValidation(
+                  s"$fieldContext with extra non-nullable arg(s) '$argNames' in $objectContext is invalid",
+                  "Any additional field arguments must not be of a non-nullable type."
+                )
+              case _ => IO.unit
```
ghostdogpr, line 687: Can be removed

zach-albia:
@ghostdogpr pushed the changes you asked for

PR #266, opened 2020-03-06
Add Parser support for Type System Extensions

### LINKED ISSUE #238: Support type system extensions in Parser
Schema extensions will make the codegen commands fail. We need to add support for it in the parser.

See http://spec.graphql.org/June2018/#TypeSystemExtension

jorge-vasquez-2301:
closes #238

### DIFF
--- a/core/src/main/scala/caliban/parsing/Parser.scala
+++ b/core/src/main/scala/caliban/parsing/Parser.scala
@@ -9,6 +9,8 @@ import caliban.parsing.adt.Definition.ExecutableDefinition._
 import caliban.parsing.adt.Definition.TypeSystemDefinition.DirectiveLocation._
 import caliban.parsing.adt.Definition.TypeSystemDefinition._
 import caliban.parsing.adt.Definition.TypeSystemDefinition.TypeDefinition._
+import caliban.parsing.adt.Definition.TypeSystemExtension._
+import caliban.parsing.adt.Definition.TypeSystemExtension.TypeExtension._
 import caliban.parsing.adt.Selection._
 import caliban.parsing.adt.Type._
 import caliban.parsing.adt._
@@ -116,7 +118,7 @@ object Parser {
   private def argument[_: P]: P[(String, InputValue)]     = P(name ~ ":" ~ value)
   private def arguments[_: P]: P[Map[String, InputValue]] = P("(" ~/ argument.rep ~ ")").map(_.toMap)
 
-  private def directive[_: P]: P[Directive] = P(Index ~ "@" ~/ name ~ arguments.?).map {
+  private def directive[_: P]: P[Directive] = P(Index ~ "@" ~ name ~ arguments.?).map {
     case (index, name, arguments) => Directive(name, arguments.getOrElse(Map()), index)
   }
   private def directives[_: P]: P[List[Directive]] = P(directive.rep).map(_.toList)
@@ -231,11 +233,10 @@ object Parser {
         EnumValueDefinition(description.map(_.value), enumValue, directives.getOrElse(Nil))
     }
 
+  private def enumName[_: P]: P[String] = name.filter(s => s != "true" && s != "false" && s != "null")
+
   private def enumTypeDefinition[_: P]: P[EnumTypeDefinition] =
-    P(
-      stringValue.? ~ "enum" ~/ name
-        .filter(s => s != "true" && s != "false" && s != "null") ~ directives.? ~ "{" ~ enumValueDefinition.rep ~ "}"
-    ).map {
+    P(stringValue.? ~ "enum" ~/ enumName ~ directives.? ~ "{" ~ enumValueDefinition.rep ~ "}").map {
       case (description, name, directives, enumValuesDefinition) =>
         EnumTypeDefinition(description.map(_.value), name, directives.getOrElse(Nil), enumValuesDefinition.toList)
     }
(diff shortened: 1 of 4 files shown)


core/src/main/scala/caliban/parsing/Parser.scala
```diff
@@ -266,6 +268,62 @@ object Parser {
         )
     }
 
+  private def schemaExtension[_: P]: P[SchemaExtension] =
+    P("extend schema" ~/ directives.? ~ "{" ~ rootOperationTypeDefinition.rep ~ "}").map {
```
ghostdogpr: It should be possible to extend a schema and only provide directives without any operation: 
http://spec.graphql.org/June2018/#sec-Schema-Extension (2nd case)

core/src/main/scala/caliban/parsing/Parser.scala
```diff
@@ -266,6 +268,62 @@ object Parser {
...
+    }
+
+  private def scalarTypeExtension[_: P]: P[ScalarTypeExtension] =
+    P("extend scalar" ~/ name ~ directives).map {
+      case (name, directives) =>
+        ScalarTypeExtension(name, directives)
+    }
+
+  private def objectTypeExtension[_: P]: P[ObjectTypeExtension] =
+    P("extend type" ~/ name ~ implements.? ~ directives.? ~ "{" ~ fieldDefinition.rep ~ "}").map {
```
ghostdogpr: Here too, there could be a case where we provide no fields and only implements/directives: http://spec.graphql.org/June2018/#sec-Object-Extensions

Same for the others below

jorge-vasquez-2301:
@ghostdogpr Thanks for looking into this! And yeah, I totally missed the alternative cases, so I'll make sure to add them 😀

core/src/main/scala/caliban/parsing/Parser.scala
```diff
@@ -266,6 +267,161 @@ object Parser {
...
+          Nil,
+          Nil,
+          fields.toList
+        )
+    }
+
+  private def objectTypeExtension[_: P]: P[ObjectTypeExtension] =
+    P(
+      "extend type" ~/ (
+        objectTypeExtensionWithInterfacesAndOptionalFields |
```
ghostdogpr: I think you can reduce to 3 cases:
- fields + optional interfaces + optional directives
- no fields + directives + optional interfaces
- no fields, no directives, interfaces

core/src/main/scala/caliban/parsing/Parser.scala
```diff
@@ -266,6 +267,161 @@ object Parser {
...
+      SchemaExtension(
+        Nil,
+        opsMap.get(OperationType.Query).map(_.name),
+        opsMap.get(OperationType.Mutation).map(_.name),
+        opsMap.get(OperationType.Subscription).map(_.name)
+      )
+    }
+
+  private def schemaExtension[_: P]: P[SchemaExtension] =
+    P("extend schema" ~/ (schemaExtensionWithDirectivesAndOptionalOperations | schemaExtensionWithOperations))
```
ghostdogpr: How about sticking to the spec definitions and have:
- operations + optional directives
- no operations, directives

core/src/main/scala/caliban/parsing/Parser.scala
jorge-vasquez-2301: @ghostdogpr Yeah, actually my first attempt was modelling the Parser more strictly after the specs, like this:
```scala
private def objectTypeExtensionWithOptionalInterfacesOptionalDirectivesAndFields[_: P]: P[ObjectTypeExtension] =
    P(name ~ implements.? ~ directives.? ~ "{" ~ fieldDefinition.rep ~ "}").map {
      case (name, implements, directives, fields) =>
        ObjectTypeExtension(
          name,
          implements.getOrElse(Nil),
          directives.getOrElse(Nil),
          fields.toList
        )
    }

  private def objectTypeExtensionWithOptionalInterfacesAndDirectives[_: P]: P[ObjectTypeExtension] =
    P(name ~ implements.? ~ directives).map {
      case (name, implements, directives) =>
        ObjectTypeExtension(
          name,
          implements.getOrElse(Nil),
          directives,
          Nil
        )
    }

  private def objectTypeExtensionWithInterfaces[_: P]: P[ObjectTypeExtension] =
    P(name ~ implements).map {
      case (name, implements) =>
        ObjectTypeExtension(
          name,
          implements,
          Nil,
          Nil
        )
    }


  private def objectTypeExtension[_: P]: P[ObjectTypeExtension] =
    P(
      "extend type" ~/ (
        objectTypeExtensionWithOptionalInterfacesOptionalDirectivesAndFields |
          objectTypeExtensionWithOptionalInterfacesAndDirectives |
          objectTypeExtensionWithInterfaces
      )
    )
```

jorge-vasquez-2301: However, that didn't work as expected, when running unit tests I've got failures for some cases. Here are the error messages I got (I've enabled verboseFailures temporally):

jorge-vasquez-2301:
```
 extend type with interfaces and directives
[info]     Fiber failed.
[info]     A checked error was not handled.
[info]     Parsing Error: Expected document:1:1 / objectTypeExtension:1:1 / objectTypeExtensionWithOptionalInterfacesOptionalDirectivesAndFields:1:13 / "{":1:58, found "" 
```

jorge-vasquez-2301:
```
extend type with directives
[info]     Fiber failed.
[info]     A checked error was not handled.
[info]     Parsing Error: Expected document:1:1 / objectTypeExtension:1:1 / objectTypeExtensionWithOptionalInterfacesOptionalDirectivesAndFields:1:13 / "{":1:33, found "" 
```

jorge-vasquez-2301: I've tried other combinations as well, but none of them worked for all cases, that's why I had to implement the parser rules in a different way.

jorge-vasquez-2301: What do you think @ghostdogpr? Maybe I'm missing something?

core/src/main/scala/caliban/parsing/Parser.scala
jorge-vasquez-2301: Similarly to the above case, my first version was more like the spec definitions:

```scala
private def schemaExtensionWithOptionalDirectivesAndOperations[_: P]: P[SchemaExtension] =
    P(directives.? ~ "{" ~ rootOperationTypeDefinition.rep ~ "}").map {
      case (directives, ops) =>
        val opsMap = ops.toMap
        SchemaExtension(
          directives.getOrElse(Nil),
          opsMap.get(OperationType.Query).map(_.name),
          opsMap.get(OperationType.Mutation).map(_.name),
          opsMap.get(OperationType.Subscription).map(_.name)
        )
    }

  private def schemaExtensionWithDirectives[_: P]: P[SchemaExtension] =
    P(directives).map(SchemaExtension(_, None, None, None))

  private def schemaExtension[_: P]: P[SchemaExtension] =
    P("extend schema" ~/ (schemaExtensionWithOptionalDirectivesAndOperations | schemaExtensionWithDirectives))
```

core/src/main/scala/caliban/parsing/Parser.scala
ghostdogpr: Hmm that’s strange indeed. I can’t see anything wrong with the code at first sight but tomorrow I’ll get your branch and try to figure it out in debug.

core/src/main/scala/caliban/parsing/Parser.scala
jorge-vasquez-2301: However, this didn't work for all cases, I got an error for one of the unit tests:

```
extend schema with directives
[info]     Fiber failed.
[info]     A checked error was not handled.
[info]     Parsing Error: Expected document:1:1 / schemaExtension:1:1 / schemaExtensionWithOptionalDirectivesAndOperations:1:15 / "{":1:30, found "" 
```

jorge-vasquez-2301: I tried other combinations that were more similar to the specs too, but none of them worked

core/src/main/scala/caliban/parsing/Parser.scala
jorge-vasquez-2301: And, I got similar issues when I tried to define parsers for the other type extension cases

jorge-vasquez-2301: Awesome, thanks for your help @ghostdogpr!

ghostdogpr:
@jorge-vasquez-2301 OK, I found the reason. You just need to change the `~/` into `~` in `def directive`. Otherwise it takes the first path and "cuts" which means it won't go back and try the other possibilities.

jorge-vasquez-2301:
Ohhhh, didn't realize that, thanks @ghostdogpr! I'll make the adjustments

PR #271, opened 2020-03-07
Add provideLayer variants to GraphQLInterpreter

ghostdogpr:
Implements `provideLayer`, `provideCustomLayer`, and `provideSomeLayer` methods.


ghostdogpr:
@adamgfraser good point! I'll update.

PR #278, opened 2020-03-11
Upgrade Finch and add support for 2.13

ghostdogpr:
@hderms FYI 😄


hderms:
@ghostdogpr oh man, exciting!

PR #285, opened 2020-03-16
Protect some field names and properties

vascorsd:
When trying to generate from github idl some issues
came up:
  * reserved scala keywords are used in multiple places
    which need escaping with ``.
  * multiline strings on the docs, so added the annotations
    to always use the triple quotes """xx"""

### DIFF
--- a/codegen/src/test/scala/caliban/codegen/SchemaWriterSpec.scala
+++ b/codegen/src/test/scala/caliban/codegen/SchemaWriterSpec.scala
@@ -217,11 +217,11 @@ userList: () => List[Option[User]]
           """
              "role"
              union Role = Captain | Pilot
-             
+
              type Captain {
                "ship" shipName: String!
              }
-             
+
              type Pilot {
                shipName: String!
              }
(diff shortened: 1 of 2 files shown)


vascorsd:
I tried to add a small test but it seems there are problems with the triple quotes. Not sure how to fix, help welcomed.

codegen/src/test/scala/caliban/codegen/SchemaWriterSpec.scala
```diff
@@ -293,6 +293,52 @@ object Types {
...
+}
+"""
+          )
+        )
+      },
+      testM("schema with multiline comment") {
+        val schema =
+          """
+             input CharacterArgs {
+               \"\"\"
```
ghostdogpr: They don't get escaped that way. I am not sure it's doable to escape triples quotes inside triple quotes. Maybe you have to use regular double quotes outside. Can you try?

ghostdogpr:
@vascorsd Any chance you could finish this one? Someone on Discord just hit that issue today 😄 If not, let me know and I will do it.

vascorsd:
@ghostdogpr Sorry, I haven't had the time to go back to this lately. I put on hold the experiment I was doing with graphql for a while. So be free to pick it up if you want it ;)

ghostdogpr:
OK, thanks!

PR #288, opened 2020-03-20
Add spray-json support

vpavkin:
Adds support for spray-json, the same way it's done for circe. My ultimate goal here is to make specific http modules (namely akka-http) to have pluggable json backends.

The only problem ATM is that spray-json is not published for ScalaJS yet. So there're some hoops to jump to make it work with fully cross-built caliban.
Support for ScalaJS in spray is actually already implemented and merged, we just lack a release (https://github.com/spray/spray-json/issues/327). So I think we can wait until it's released to make this PR prettier.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -89,7 +89,9 @@ lazy val core = crossProject(JSPlatform, JVMPlatform)
     fork in Test := true,
     fork in run := true
   )
-lazy val coreJVM = core.jvm
+lazy val coreJVM = core.jvm.settings(
+  libraryDependencies += "io.spray" %% "spray-json" % "1.3.5" % Optional
+)
 lazy val coreJS = core.js.settings(
   libraryDependencies += "io.github.cquiroz" %%% "scala-java-time" % "2.0.0-RC5" % Test
 )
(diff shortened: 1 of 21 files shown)


ghostdogpr:
Actually I'm a bit concerned about ScalaJS support. Currently caliban is still on ScalaJS 0.6.x because some dependencies are not yet published for 1.x (namely: ZIO and sttp). But both of them should arrive soon-ish so I'm hoping to upgrade ScalaJS to 1.x as soon as they are available. Spray-json is currently on 0.6x on master. Considering it's not published very often (the PR to add ScalaJS support was merged 9 months ago), I'm afraid this becomes a blocker.

What do you think? Or maybe we could merge this once spray-json is published for ScalaJS 1.x and Caliban upgrades?

An alternative is to make it a separate module, JVM only.

vpavkin:
Yep, I'm on the same page with you here - don't want this to become a blocker in future. I pinged spray maintainers wherever I could, so let's see how fast and involved they actually are these days.

No rush here.

vpavkin:
Plan changed to proceed with play-json instead of spray. closing this one

PR #289, opened 2020-03-21
Move adapters to a single folder and reduce duplication in examples

ghostdogpr:
- Make the project a little less messy by moving the adapters to a single folder.
- Simplify the examples and reduce duplicated code.


ghostdogpr:
cc @yoohaemin FYI, this will impact the play adapter

PR #292, opened 2020-03-23
Decouple akka-http adapter from hard circe dependency

vpavkin:
My initial attempt at making akka-http adapter work with different json backends. 

These "glue" interfaces are a bit loosely typed with all those strings flying around, but I don't expect the end user working with them much - additional adapters can just be added here later.


build.sbt
```diff
@@ -218,6 +218,8 @@ lazy val examples = project
   .settings(skip in publish := true)
   .settings(
     libraryDependencies ++= Seq(
+      "de.heikoseeberger"            %% "akka-http-circe"               % "1.31.0",
```
ghostdogpr: Hmm... For someone not using Circe on their own and just wants caliban exposed with akka, this is easy to miss (that they need to add these 2 dependencies). It is quite different from the circe interop in core where you can only trigger the circe-related code if you use Circe.

I had another idea to support this problem of having too many combinations of backends and JSON libraries: integration with [tapir](https://github.com/softwaremill/tapir) or [endpoints](https://github.com/julienrf/endpoints). That would be another adapter that gives a tapir/endpoints route, then you can use it with the backend and json library of your choice. Not sure they support WebSocket though. The idea is similar to what I did with caliban-client where I just give an sttp request and users can use what they want. What do you think?

vpavkin:
@ghostdogpr pinging you here to not unncessarily distract:)

I managed to make a compile-safe all-included mixin, as shown here:
https://github.com/ghostdogpr/caliban/pull/292/commits/f277407a2cc81cbd636737f28f0846ce4e332ead

When you comment out akka-http-circe dependency, then it will fail to compile with 
```
Symbol 'type de.heikoseeberger.akkahttpcirce.FailFastCirceSupport' is missing from the classpath.
```

Does it look ok to you? If it is, then I'll polish the docs :)

ghostdogpr:
@vpavkin yep, looks good! You can remove the dependency to `circe-parser` in akka-http as well.

vpavkin:
@ghostdogpr Added docs and removed the unnecessary dependency. I can also help you with release notes (how to migrate existing akka-http integrations).

ghostdogpr:
@vpavkin looks good to me! Can you add the needed import here too? https://github.com/ghostdogpr/caliban/tree/master/examples

Also before I merge, since there are no tests for the adapter, did you test with a tool like Altair that it was still working well (including the subscriptions)? :D

ghostdogpr:
Btw now that I merged your other PR, wanna add the play backend here?

vpavkin:
I tested only http part, will check subscriptions today :) As for the play backend - that's my goal, yep. will add as soon as we merge everything that is required :)

vpavkin:
@ghostdogpr I checked the subscription in example app. It works, but for some reason each deletion triggers 4 identical subscription events in Altair.

Like this

image

Was it the same behavior previously? I really doubt that extracting out json specifics could change behaviour in this way :thinking:

ghostdogpr:
@vpavkin I think that's a bug in Altair that creates multiple subscriptions -_- If you close Altair and just subscribe once, it works no?

vpavkin:
You are correct:) Thanks for help - after restart it's just one message. So it works fine

vpavkin:
Also added that dependency to the readme you mentioned :)

PR #294, opened 2020-03-24
Add play-json integration

vpavkin:
This is a part of #143 - only play-json integration. 

It's done in a different way to #234 - there seems to be no objective reason to not be using optional dependency approach (as with circe).

As far as I understand, cross-building to different play versions (as done in #234) also has nothing to do with json - it's related to controllers api changes between play 2.7 and 2.8. Json sould work fine with either.

@yoohaemin, please let me know if you have any concerns with this implementation :)

### DIFF
--- a/core/src/main/scala/caliban/CalibanError.scala
+++ b/core/src/main/scala/caliban/CalibanError.scala
@@ -117,7 +117,8 @@ private object ErrorPlayJson {
     path: Option[JsArray]
   )
 
-  implicit val locationInfoWrites: Writes[LocationInfo] = Json.writes[LocationInfo].transform((v: JsValue) => Json.arr(v))
+  implicit val locationInfoWrites: Writes[LocationInfo] =
+    Json.writes[LocationInfo].transform((v: JsValue) => Json.arr(v))
 
   private implicit val errorDTOWrites = Json.writes[ErrorDTO]
 
--- a/core/src/main/scala/caliban/CalibanError.scala
+++ b/core/src/main/scala/caliban/CalibanError.scala
@@ -109,50 +109,38 @@ private object ErrorCirce {
 
 private object ErrorPlayJson {
   import play.api.libs.json._
-  import play.api.libs.json.Json.toJson
 
-  private def locationToJson(locInfo: Option[LocationInfo]): Map[String, JsValue] =
-    Some(locInfo).collect {
-      case Some(li) => Json.arr(Json.obj("line" -> toJson(li.line), "column" -> toJson(li.column)))
-    }.fold(Map.empty[String, JsValue])(v => Map("locations" -> v))
+  private final case class ErrorDTO(
+    message: String,
+    extensions: Option[ResponseValue],
+    locations: Option[LocationInfo],
+    path: Option[JsArray]
+  )
 
-  val errorValueWrites: Writes[CalibanError] = Writes {
+  implicit val locationInfoWrites: Writes[LocationInfo] = Json.writes[LocationInfo].transform(Json.arr(_))
+
+  private implicit val errorDTOWrites = Json.writes[ErrorDTO]
+
+  val errorValueWrites: Writes[CalibanError] = errorDTOWrites.contramap[CalibanError] {
     case CalibanError.ParsingError(msg, locationInfo, _, extensions) =>
-      val strictFields = Map(
-        "message"    -> toJson(s"Parsing Error: $msg"),
-        "extensions" -> toJson[Option[ResponseValue]](extensions)
-      )
-      // excluding to avoid nulls
-      val optionalFields = locationToJson(locationInfo)
-
-      JsObject(strictFields ++ optionalFields)
+      ErrorDTO(s"Parsing Error: $msg", extensions, locationInfo, None)
 
     case CalibanError.ValidationError(msg, _, locationInfo, extensions) =>
-      val strictFields = Map(
-        "message"    -> toJson(msg),
-        "extensions" -> toJson[Option[ResponseValue]](extensions)
-      )
-      // excluding to avoid nulls
-      val optionalFields = locationToJson(locationInfo)
-
-      JsObject(strictFields ++ optionalFields)
+      ErrorDTO(msg, extensions, locationInfo, None)
 
     case CalibanError.ExecutionError(msg, path, locationInfo, _, extensions) =>
-      val strictFields = Map(
-        "message"    -> toJson(msg),
-        "extensions" -> toJson[Option[ResponseValue]](extensions)
-      )
-      // excluding to avoid nulls
-      val optionalFields = locationToJson(locationInfo) ++
+      ErrorDTO(
+        msg,
+        extensions,
+        locationInfo,
         Some(path).collect {
           case p if p.nonEmpty =>
             JsArray(p.map {
               case Left(value)  => JsString(value)
               case Right(value) => JsNumber(value)
             })
-        }.fold(Map.empty[String, JsValue])(v => Map("path" -> v))
-
-      JsObject(strictFields ++ optionalFields)
+        }
+      )
   }
 
 }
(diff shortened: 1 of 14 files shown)


core/src/main/scala/caliban/GraphQLRequest.scala
```diff
@@ -26,3 +29,16 @@ private object GraphQLRequestCirce {
...
+  import play.api.libs.json._
+  import play.api.libs.json.Reads._
+  import play.api.libs.functional.syntax._
+
+  val graphQLRequestReads: Reads[GraphQLRequest] =
+    (
+      (JsPath \ "query").read[String] and
+        (JsPath \ "operationName").readNullable[String] and
+        (JsPath \ "variables").readNullable[Map[String, InputValue]]
+    )(GraphQLRequest.apply _)
```
paulpdaniels: Why can't the macro be used here? It should be able to derive these already.

core/src/main/scala/caliban/CalibanError.scala
```diff
@@ -102,3 +106,53 @@ private object ErrorCirce {
...
 }
+
+private object ErrorPlayJson {
+  import play.api.libs.json._
+  import play.api.libs.json.Json.toJson
+
+  private def locationToJson(locInfo: Option[LocationInfo]): Map[String, JsValue] =
+    Some(locInfo).collect {
+      case Some(li) => Json.arr(Json.obj("line" -> toJson(li.line), "column" -> toJson(li.column)))
+    }.fold(Map.empty[String, JsValue])(v => Map("locations" -> v))
```
paulpdaniels: I would just make this a `Writes`

```scala
private implicit val locationInfoWrites: Writes[LocationInfo] =
  Json.writes[LocationInfo].transform(Json.arr(_))

```

core/src/main/scala/caliban/CalibanError.scala
```diff
@@ -102,3 +106,53 @@ private object ErrorCirce {
...
+
+  val errorValueWrites: Writes[CalibanError] = Writes {
+    case CalibanError.ParsingError(msg, locationInfo, _, extensions) =>
+      val strictFields = Map(
+        "message"    -> toJson(s"Parsing Error: $msg"),
+        "extensions" -> toJson[Option[ResponseValue]](extensions)
+      )
+      // excluding to avoid nulls
```
paulpdaniels: If you use the `Writes` approach above it should automatically remove `None` meaning you can just do:

```scala
JsObject(
  // Strict fields
  "message" -> s"Parsing Error: $msg",
  "extensions" -> toJson[Option[ResponseValue]](extensions)
  // Optional
  "locations" -> locationInfo
)

```

core/src/main/scala/caliban/GraphQLResponse.scala
```diff
@@ -38,3 +41,29 @@ private object GraphQLResponseCirce {
...
 }
+
+private object GraphQLResponsePlayJson {
+  import play.api.libs.json._
+  import play.api.libs.json.Json.toJson
+
+  val graphQLResponseWrites: Writes[GraphQLResponse[Any]] = Writes {
+    case GraphQLResponse(data, Nil, None) => Json.obj("data" -> toJson(data))
+    case GraphQLResponse(data, Nil, Some(extensions)) =>
+      Json.obj("data" -> toJson(data), "extensions" -> toJson(extensions.asInstanceOf[ResponseValue]))
```
paulpdaniels: Don't need `toJson` as long as the `implicit` writes is available in scope

core/src/test/scala/caliban/execution/ExecutionSpec.scala
```diff
@@ -348,6 +349,19 @@ object ExecutionSpec extends DefaultRunnableSpec {
...
+        import caliban.interop.play.json._
+        case class Queries(test: JsValue)
+
+        val interpreter = graphQL(RootResolver(Queries(Json.obj(("a", JsNumber(333)))))).interpreter
+        val query       = gqldoc("""
+             {
+               test
+             }""")
+
+        assertM(interpreter.flatMap(_.execute(query)).map(_.data.toString))(equalTo("""{"test":{"a":333}}"""))
```
paulpdaniels: Could you also add a field that returns `None` so that there is a regression for `nullable` fields.

core/src/main/scala/caliban/CalibanError.scala
```diff
@@ -102,3 +106,53 @@ private object ErrorCirce {
...
+      val optionalFields = locationToJson(locationInfo) ++
+        Some(path).collect {
+          case p if p.nonEmpty =>
+            JsArray(p.map {
+              case Left(value)  => JsString(value)
+              case Right(value) => JsNumber(value)
+            })
+        }.fold(Map.empty[String, JsValue])(v => Map("path" -> v))
```
paulpdaniels: Again could simplify because `play` removes `None` already.

core/src/main/scala/caliban/CalibanError.scala
vpavkin: Oh, I just didn't realize that it was written manually for circe to not depend in circe-generic/circe-derivation, and for play macros already included so I can use them.

Thanks for good suggestions, will clean it up :+1:

core/src/test/scala/caliban/execution/ExecutionSpec.scala
vpavkin: There's already one in `GraphQLResponseSpec` and it actually fails when I just put option values instead of separating optional fields.
I checked about None-fields behavior in play and it seems that it doesn't remove null fields by default. Moreover, it seems there's not simple way to turn that off - you have to manually process json AST before printing to do it.

vpavkin: To expand - the behavior you describe works for macro-derived codecs. If you create object manually and specify a field to be None you still get a null.
See an example:

https://scastie.scala-lang.org/oCuWojs4TjajMl0laRYKSg

vpavkin: I just used an intermediate dto class with derived codec so that these rules apply. Now it looks even better.

core/src/main/scala/caliban/GraphQLRequest.scala
vpavkin: fixed

vpavkin:
@ghostdogpr thanks! Added play-json entry to the table.
@paulpdaniels I addressed all your comments so far :)

core/src/main/scala/caliban/interop/play/play.scala
```diff
@@ -0,0 +1,45 @@
...
+}
+
+object json {
+  implicit val jsonSchema: Schema[Any, JsValue] = new Schema[Any, JsValue] {
+    private def parse(value: JsValue) =
+      implicitly[Reads[ResponseValue]]
+        .reads(value)
+        .asEither
+        .left
+        .map(errs => new Throwable(s"Couldn't decode json: $errs"))
```
paulpdaniels: Maybe not for this PR, but general question to @ghostdogpr should this be a typed error? Seems like we should surface this in a structured way to the user.

core/src/main/scala/caliban/interop/play/play.scala
```diff
@@ -0,0 +1,45 @@
...
+
+    override def toType(isInput: Boolean): __Type = makeScalar("Json")
+    override def resolve(value: JsValue): Step[Any] =
+      QueryStep(ZQuery.fromEffect(ZIO.fromEither(parse(value))).map(PureStep))
+  }
+  implicit val jsonArgBuilder: ArgBuilder[JsValue] = (input: InputValue) =>
+    Right(implicitly[Writes[InputValue]].writes(input))
```
paulpdaniels: Could also simplify to `Right(Json.toJson(input))`

vpavkin: Thanks:) Fixed

core/src/main/scala/caliban/interop/play/play.scala
ghostdogpr: This is fine, `QueryStep` expects a `Throwable` and will wrap it into an `ExecutionError`.

PR #302, opened 2020-03-29
Apollo Persisted Queries

### LINKED ISSUE #300: Support for Apollo Persisted Query
See specs: https://github.com/apollographql/apollo-link-persisted-queries

This requires a change of `WrappingFunction` to make it more flexible.

ghostdogpr:
Closes #300

### DIFF
--- a/core/src/main/scala/caliban/wrappers/ApolloPersistedQueries.scala
+++ b/core/src/main/scala/caliban/wrappers/ApolloPersistedQueries.scala
@@ -1,0 +1,66 @@
+package caliban.wrappers
+
+import caliban.CalibanError.ValidationError
+import caliban.Value.{ NullValue, StringValue }
+import caliban.wrappers.Wrapper.OverallWrapper
+import caliban.{ GraphQLRequest, GraphQLResponse, InputValue }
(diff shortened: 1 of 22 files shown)


core/src/main/scala/caliban/wrappers/ApolloPersistedQueries.scala
```diff
@@ -0,0 +1,66 @@
...
+      .getOrElse(Map())
+      .get("persistedQuery")
+      .flatMap {
+        case InputValue.ObjectValue(fields) =>
+          fields.get("sha256Hash").flatMap {
+            case StringValue(hash) => Some(hash)
+            case _                 => None
+          }
+        case _ => None
+      }
```
paulpdaniels: Can be slightly reduced to:

```scala
request.extensions
  .flatMap(_.get("persistedQuery"))
  .flatMap {
      fields.get("sha256Hash").collectFirst {
       case StringValue(hash) => hash
      }
    case _ => None
  }
```

ghostdogpr: 👍

PR #317, opened 2020-04-11
Subscription keep alive

rleibman:
Added keep alive capabilities to akka-http handler


adapters/akka-http/src/main/scala/caliban/AkkaHttpAdapter.scala
```diff
@@ -121,16 +126,39 @@ trait AkkaHttpAdapter {
...
+                      val ack: ZIO[Clock with Random, Throwable, Any] =
+                        IO.fromFuture(_ => queue.offer(TextMessage("""{"type":"connection_ack"}""")))
+
+                      val withKeepAlive = keepAliveTime.fold(ack) { time =>
+                        ack
+                          .flatMap(_ =>
+                            IO.fromFuture(_ =>
+                                queue
+                                  .offer(
+                                    TextMessage("""{"type":"ka"}""")
```
ghostdogpr, line 148: The type should be `connection_keep_alive` I think?

adapters/akka-http/src/main/scala/caliban/AkkaHttpAdapter.scala
```diff
@@ -121,16 +126,39 @@ trait AkkaHttpAdapter {
...
+                                  )
+                              )
+                              .repeat(Schedule.spaced(time).jittered)
+                              .forkDaemon
```
ghostdogpr, line 152: This fiber will go on forever, you need to capture it and interrupt it when the connection is stopped.

adapters/akka-http/src/main/scala/caliban/AkkaHttpAdapter.scala
```diff
@@ -155,6 +183,7 @@ trait AkkaHttpAdapter {
         }
 
         val flow = Flow.fromSinkAndSource(sink, source).watchTermination() { (_, f) =>
+          //TODO: Shouldn't it send GQL_COMPLETE here before terminating?
```
ghostdogpr, line 186: Hmm not here, but indeed the spec says that for subscriptions, GQL_COMPLETE should be sent when unsubscribing and this seems missing so far.

ghostdogpr: I thought there might be better: https://github.com/ghostdogpr/caliban/blob/master/adapters/akka-http/src/main/scala/caliban/AkkaHttpAdapter.scala#L150 but maybe this is called too so it might be fine if it's called on unsubscribing.

adapters/akka-http/src/main/scala/caliban/AkkaHttpAdapter.scala
rleibman: That would make more sense to me too, but all the protocol documents I could find said 'ka', e.g. https://medium.com/@rob.blackbourn/writing-a-graphql-websocket-subscriber-in-javascript-4451abb9cd60, https://github.com/apollographql/subscriptions-transport-ws/blob/faa219cff7b6f9873cae59b490da46684d7bea19/src/message-types.ts

adapters/akka-http/src/main/scala/caliban/AkkaHttpAdapter.scala
rleibman: I'll leave it up to you, and at any rate, it should be a different PR

adapters/akka-http/src/main/scala/caliban/AkkaHttpAdapter.scala
rleibman: Please take a look at my change, something like that?

adapters/akka-http/src/main/scala/caliban/AkkaHttpAdapter.scala
rleibman: Maybe because you can potentially get tons of these and `ka` conserves bandwidth?

adapters/akka-http/src/main/scala/caliban/AkkaHttpAdapter.scala
ghostdogpr: Fair enough, I'll do it in another PR. You can remove the TODOs then.

adapters/akka-http/src/main/scala/caliban/AkkaHttpAdapter.scala
ghostdogpr: You're right, they say `The keep alive message type does not follow the standard due to connection optimizations`

adapters/akka-http/src/main/scala/caliban/AkkaHttpAdapter.scala
```diff
@@ -113,26 +118,52 @@ trait AkkaHttpAdapter {
...
+          Source.queue[Message](0, OverflowStrategy.fail).preMaterialize()
+        val subscriptions = runtime.unsafeRun(Ref.make(Map.empty[String, Fiber[Throwable, Unit]]))
         val sink = Sink.foreach[Message] {
           case TextMessage.Strict(text) =>
             val io = for {
-              msg     <- Task.fromEither(json.parseWSMessage(text))
-              msgType = msg.messageType
+              msg               <- Task.fromEither(json.parseWSMessage(text))
+              msgType           = msg.messageType
+              keepAliveFiberRef <- Ref.make(None: Option[Fiber.Runtime[Throwable, Int]])
```
ghostdogpr, line 136: Use `Fiber` directly.

adapters/akka-http/src/main/scala/caliban/AkkaHttpAdapter.scala
```diff
@@ -113,26 +118,52 @@ trait AkkaHttpAdapter {
...
+                      val withClientDuration = clientDuration.fold(withKeepAlive) { time =>
+                        //TODO: Shouldn't it send GQL_COMPLETE here before terminating?
+                        withKeepAlive.flatMap(_ => IO.effect(queue.complete())).delay(time).forkDaemon
+                      }
+
+                      withClientDuration
                     case "connection_terminate" =>
-                      IO.effect(queue.complete())
+                      IO.effect(queue.complete()).map { _ =>
+                        keepAliveFiberRef.map(_.map(_.interrupt))
```
ghostdogpr, line 164: You're not running these effects.

adapters/akka-http/src/main/scala/caliban/AkkaHttpAdapter.scala
```diff
@@ -82,8 +85,10 @@ trait AkkaHttpAdapter {
 
   def makeWebSocketService[R, E](
     interpreter: GraphQLInterpreter[R, E],
-    skipValidation: Boolean = false
-  )(implicit ec: ExecutionContext, runtime: Runtime[R], materializer: Materializer): Route = {
+    skipValidation: Boolean = false,
+    keepAliveTime: Option[Duration] = None,
+    clientDuration: Option[Duration] = None
```
ghostdogpr, line 90: I don't think we need that one. Apollo Server only has the keep-alive parameter.

adapters/akka-http/src/main/scala/caliban/AkkaHttpAdapter.scala
```diff
@@ -148,13 +179,18 @@ trait AkkaHttpAdapter {
                       subscriptions
                         .modify(map => (map.get(msg.id), map - msg.id))
                         .flatMap(fiber => IO.whenCase(fiber) { case Some(fiber) => fiber.interrupt })
+                        .map { _ =>
+                          keepAliveFiberRef.map(_.map(_.interrupt))
+                          keepAliveFiberRef.set(None)
```
ghostdogpr, line 184: This is stopping a single subscription, we don't want to interrupt the keep alive here.

adapters/akka-http/src/main/scala/caliban/AkkaHttpAdapter.scala
```diff
@@ -148,13 +179,18 @@ trait AkkaHttpAdapter {
...
+                        }
                   }
             } yield ()
             runtime.unsafeRun(io)
           case _ => ()
         }
 
         val flow = Flow.fromSinkAndSource(sink, source).watchTermination() { (_, f) =>
+          //TODO: Shouldn't it send GQL_COMPLETE here before terminating?
           f.onComplete(_ => runtime.unsafeRun(subscriptions.get.flatMap(m => IO.foreach(m.values)(_.interrupt).unit)))
```
ghostdogpr, line 194: I think the interruption should be done here actually.

adapters/akka-http/src/main/scala/caliban/AkkaHttpAdapter.scala
```diff
@@ -113,24 +117,45 @@ trait AkkaHttpAdapter {
...
+                sendMessage(sendTo, message.id, other, result.errors) *> IO
+                  .fromFuture(_ => sendTo.offer(TextMessage(s"""{"type":"complete","id":"${message.id}"}""")))
             }
       } yield ()
 
     get {
       extractUpgradeToWebSocket { upgrade =>
-        val (queue, source) = Source.queue[Message](0, OverflowStrategy.fail).preMaterialize()
-        val subscriptions   = runtime.unsafeRun(Ref.make(Map.empty[String, Fiber[Throwable, Unit]]))
+        val (queue: SourceQueueWithComplete[Message], source) =
```
ghostdogpr, line 127: Do we need the explicit type annotation?

adapters/akka-http/src/main/scala/caliban/AkkaHttpAdapter.scala
```diff
@@ -82,8 +85,9 @@ trait AkkaHttpAdapter {
 
   def makeWebSocketService[R, E](
     interpreter: GraphQLInterpreter[R, E],
-    skipValidation: Boolean = false
-  )(implicit ec: ExecutionContext, runtime: Runtime[R], materializer: Materializer): Route = {
+    skipValidation: Boolean = false,
+    keepAliveTime: Option[Duration] = None
+  )(implicit ec: ExecutionContext, runtime: Runtime[R with Clock with Random], materializer: Materializer): Route = {
```
ghostdogpr, line 90: No need for `Random` I guess?

adapters/akka-http/src/main/scala/caliban/AkkaHttpAdapter.scala
```diff
@@ -113,24 +117,45 @@ trait AkkaHttpAdapter {
...
+                        ack.flatMap(_ =>
+                          IO.fromFuture(_ =>
+                              queue
+                                .offer(
+                                  TextMessage("""{"type":"ka"}""")
+                                )
+                            )
+                            .repeat(Schedule.spaced(time).jittered)
+                            .forkDaemon
+                            .map(keepAliveFiber =>
```
ghostdogpr, line 153: Shouldn't it be `flatMap` here?

adapters/akka-http/src/main/scala/caliban/AkkaHttpAdapter.scala
```diff
@@ -113,24 +117,45 @@ trait AkkaHttpAdapter {
...
+                              subscriptions.update(_.updated("_keepAliveFiber", keepAliveFiber.unit))
```
ghostdogpr, line 154: Maybe a map of `Option[String]` would be more elegant? And we use `None` for this case.

adapters/akka-http/src/main/scala/caliban/AkkaHttpAdapter.scala
rleibman: No, I was trying to figure out what type of queue it was, forgot to take it out.

adapters/akka-http/src/main/scala/caliban/AkkaHttpAdapter.scala
rleibman: You do, otherwise you can't add jittered to the timing

adapters/akka-http/src/main/scala/caliban/AkkaHttpAdapter.scala
rleibman: Yeah, that's much better.

adapters/akka-http/src/main/scala/caliban/AkkaHttpAdapter.scala
rleibman: That works.

adapters/akka-http/src/main/scala/caliban/AkkaHttpAdapter.scala
ghostdogpr: Oh, I see. But why using jittered at all? You expect the keep-alive to be regular no?

rleibman: Not really, adding jitter help prevent congestion, it's not part of the spec per-se, but it's common in other keep-alive mechanisms I've seen elsewhere.

ghostdogpr:
@rleibman Thanks for your work! I'll merge it and do a couple adjustments + uzhttp support.

PR #318, opened 2020-04-12
Allow codegen tool to get a schema by introspecting a URL

### LINKED ISSUE #290: Codegen: generate from a GraphQL server URL using introspection
Currently the code generation tool works from a file in the GraphQL IDL format. Instead, we could simply request the URL of a GraphQL server and use introspection to get the schema.

Getting the introspection data can be done using the Caliban client module.

ghostdogpr:
Closes #290

### DIFF
--- a/codegen/src/main/scala/caliban/codegen/CodegenPlugin.scala
+++ b/codegen/src/main/scala/caliban/codegen/CodegenPlugin.scala
@@ -15,9 +15,8 @@ object CodegenPlugin extends AutoPlugin {
 
   def genCommand(name: String, helpMsg: String, writer: (Document, String, Option[String]) => String): Command =
     Command.args(name, helpMsg) { (state: State, args: Seq[String]) =>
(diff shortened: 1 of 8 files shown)


codegen/src/main/scala/caliban/codegen/CodegenPlugin.scala
```diff
@@ -57,23 +60,31 @@ object CodegenPlugin extends AutoPlugin {
       case _                                         => putStrLn(helpMsg)
     }
 
+  def getSchema(path: String): Task[Document] =
```
paulpdaniels, line 63: Depending on their configuration they may need to pass additional headers when introspecting, should we include that?

ghostdogpr: I won't do it right now but I agree. The current code for parsing options isn't very convenient, maybe I should use something like https://github.com/scopt/scopt. Will make an issue.

PR #319, opened 2020-04-13
Add federation to release

paulpdaniels:
Missed this in my last PR but found it while testing locally. Forgot to add the federation module to the aggregate for publishing

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -65,7 +65,8 @@ lazy val root = project
     monixInterop,
     clientJVM,
     clientJS,
-    codegen
+    codegen,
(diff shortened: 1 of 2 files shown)


ghostdogpr:
Wow, I also missed it, good catch!

ghostdogpr:
Scalafmt seems unhappy.

ghostdogpr:
It's `FederationSpec`, not sure why it didn't complain before 🤔

paulpdaniels:
Not sure why that happened, but I fixed the spec to make it happy.

PR #324, opened 2020-04-15
Remove scala.js for caliban core

ghostdogpr:
It doesn't seem really useful to have caliban server work with Scala.js and prevents me from upgrading Magnolia because of Scala.js versioning. None of the adapters support Scala.js anyway.

I initially did this because I though caliban-client would depend on caliban but it ended up not being the case (only the codegen tool does).


paulpdaniels:
LGTM

ghostdogpr:
Tests fail because of an issue with Magnolia `ReadOnlyParam` which doesn't return annotations as expected.

PR #327, opened 2020-04-16
split codegen plugin into library and sbt modules

### LINKED ISSUE #326: Codegen: make it usable without sbt
Move the sbt-specific code to a new module `caliban-codegen-sbt` and make `caliban-codegen` a regular module that can be used without sbt.

joprice:
Closes #326


build.sbt
```diff
@@ -102,15 +103,28 @@ lazy val codegen = project
...
     crossScalaVersions := Seq("2.12.10"),
     testFrameworks := Seq(new TestFramework("zio.test.sbt.ZTestFramework")),
     libraryDependencies ++= Seq(
       "org.scalameta"                %% "scalafmt-dynamic"              % "2.4.2",
       "org.scalameta"                %% "scalafmt-core"                 % "2.4.2",
       "com.softwaremill.sttp.client" %% "async-http-client-backend-zio" % sttpVersion,
-      "dev.zio"                      %% "zio-test"                      % zioVersion % "test",
-      "dev.zio"                      %% "zio-test-sbt"                  % zioVersion % "test"
+      "dev.zio"                      %% "zio-test"                      % zioVersion % "test"
+    )
```
ghostdogpr, line 113: You need to keep zio-test-sbt here to run the tests.

build.sbt
```diff
@@ -102,15 +103,28 @@ lazy val codegen = project
   .settings(name := "caliban-codegen")
   .settings(commonSettings)
   .settings(
-    sbtPlugin := true,
     crossScalaVersions := Seq("2.12.10"),
```
ghostdogpr, line 106: This line can be remove (only needed for the plugin).

build.sbt
```diff
@@ -102,15 +103,28 @@ lazy val codegen = project
...
       "com.softwaremill.sttp.client" %% "async-http-client-backend-zio" % sttpVersion,
-      "dev.zio"                      %% "zio-test"                      % zioVersion % "test",
-      "dev.zio"                      %% "zio-test-sbt"                  % zioVersion % "test"
+      "dev.zio"                      %% "zio-test"                      % zioVersion % "test"
+    )
+  )
+  .dependsOn(coreJVM, clientJVM)
+
+lazy val codegenSbt = project
+  .dependsOn(codegen)
```
ghostdogpr, line 118: Move that to below with the others.

build.sbt
```diff
@@ -66,6 +66,7 @@ lazy val root = project
     clientJVM,
     clientJS,
     codegen,
+    codegenSbt,
```
ghostdogpr, line 69: Can you also add it to .circleci/config.yml in the test212_jdk8 task?

build.sbt
joprice: Ah I hadn't seen that module and thought it was something to do with sbt plugins.

build.sbt
joprice: I moved it to the end and removed the others since they're transitively depended on.

build.sbt
joprice: I added it as `codegenSbt/test`, although there are currently no tests. Should it be compile for now instead?

joprice:
I moved the two methods into a new object in the codegen module.

build.sbt
ghostdogpr: That's fine, it will just compile in that case.

build.sbt
ghostdogpr: You missed this one (I meant `crossScalaVersions` should be removed, this was to force the sbt plugin to be built on 2.12 only but a regular module can be cross-published.

joprice: I missed that one. Updated

Issue #232, opened 2020-02-22
Schema validation: directives


ghostdogpr:
In `Validator.validateSchema`, verify the [rules for directives](https://spec.graphql.org/June2018/#sec-Type-System.Directives)

1. A directive definition must not contain the use of a directive which
   references itself directly.
2. A directive definition must not contain the use of a directive which
   references itself indirectly by referencing a Type or Directive which
   transitively includes a reference to this directive.
3. The directive must not have a name which begins with the characters
   {"__"} (two underscores).
4. For each argument of the directive:
   1. The argument must not have a name which begins with the
      characters {"__"} (two underscores).
   2. The argument must accept a type where {IsInputType(argumentType)}
      returns {true}.

spf3000:
I would like to attempt this

javimartinez:
Working on this

ghostdogpr:
Closed by #333

PR #333, opened 2020-04-18
Add schema directives validation

javimartinez:
Following the schema validation rules for directives posted in #232, I've implemented the 3 and 4.1 rules in this PR. 

I think that due to the type model, it isn't necessary to check the rest of the rules, but I'm no pretty sure, am I missing something?

### DIFF
--- a/core/src/main/scala/caliban/validation/Validator.scala
+++ b/core/src/main/scala/caliban/validation/Validator.scala
@@ -747,21 +747,18 @@ object Validator {
       } yield ()
     }
 
-    def validateDirectives[T](
-      t: T,
-      directivesExtractor: T => Option[List[Directive]],
+    def validateDirectives(
+      directives: Option[List[Directive]],
       errorContext: String
     ): IO[ValidationError, Unit] =
-      IO.foreach_(directivesExtractor(t).getOrElse(List.empty))(validateDirective(_, errorContext))
+      IO.foreach_(directives.getOrElse(List.empty))(validateDirective(_, errorContext))
 
     def validateInputValueDirectives(
       inputValues: List[__InputValue],
       errorContext: String
     ): IO[ValidationError, Unit] = {
       val inputValueErrorContextBuilder = (name: String) => s"InputValue '$name' of $errorContext"
-      IO.foreach_(inputValues)(iv =>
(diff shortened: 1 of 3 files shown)


core/src/main/scala/caliban/validation/Validator.scala
```diff
@@ -722,6 +728,45 @@ object Validator {
...
+      errorContext: String
+    ): IO[ValidationError, Unit] =
+      IO.foreach_(directivesExtractor(t).getOrElse(List.empty))(validateDirective(_, errorContext))
+
+    IO.foreach_(types) { t =>
+      val typeErrorContext                    = s"Type '${t.name.getOrElse("")}'"
+      val fieldErrorContext: String => String = fieldName => s"Field '$fieldName' of $typeErrorContext"
+      for {
+        _ <- validateDirectives[__Type](t, _.directives, typeErrorContext)
+        _ <- IO.foreach_(t.fields(__DeprecatedArgs(Some(true))).getOrElse(List.empty[__Field]))(f =>
```
ghostdogpr, line 762: There can also be directives in `__Type.inputFields` and `__Field.args`: both contain `__InputValue` which has `directives`.

javimartinez: I haven't seen those, added!

core/src/main/scala/caliban/validation/Validator.scala
```diff
@@ -722,6 +728,65 @@ object Validator {
...
+    def validateDirective(directive: Directive, errorContext: String) = {
+      val directiveErrorContext = s"Directive '${directive.name}' of $errorContext"
+      for {
+        _ <- doesNotStartWithUnderscore(directive, directiveErrorContext)
+        _ <- validateArguments(directive.arguments, directiveErrorContext)
+      } yield ()
+    }
+
+    def validateDirectives[T](
+      t: T,
```
ghostdogpr, line 751: You pass `T` and `T => Option[List[Directive]]`, why not just pass `Option[List[Directive]]`? Then you won't need the type parameter anymore.

javimartinez: done :)

PR #339, opened 2020-04-19
ZQuery: Execute Arbitrary Effects in Parallel in Parallelism Combinators

### LINKED ISSUE #338: ZQuery parallel combinators run effects sequentially
`ZQuery` parallel combinators run queries using Datasource in parallel, but arbitrary effects run sequentially, which is not really expected.

adamgfraser:
Resolves #338.

The "Apollo Tracing" test is failing, I think because the ordering of requests performed in parallel is not guaranteed.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,6 +8,7 @@ tags
 
 .bloop
 .metals
+project/metals.sbt
 *.class
 *.log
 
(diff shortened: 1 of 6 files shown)


ghostdogpr:
Thanks! I'll fix the test and merge tonight.

PR #351, opened 2020-04-25
fix #348 Support variables in subscription

### LINKED ISSUE #348: Variables in Subscription
Hi. Thank you for the great library! 

I started playing around with caliban and  tried to use a variable in my subscription query like this one.

```graphql
subscription OnEvent($topic: String!) {
    onEvent(topic: $topic) {
        name
    }
}
```

But caliban fails to execute query because it can not resolve the variable "$topic". I checked the WebSocket message and it contains variable definitions as expected.

```json
{
  "id": "1",
  "type": "start",
  "payload": {
    "variables": {
      "topic": "mytopic"
    },
    "extensions": {},
    "operationName": "OnEvent",
    "query": "subscription OnEvent($topic: String!) { onEvent(topic: $topic) {name}}"
  }
}
```

The code for Http4sAdapter ignores the variables in payload is this intentional?

https://github.com/ghostdogpr/caliban/blob/dac5817f687479493e2a200dee2e5fa214411247/adapters/http4s/src/main/scala/caliban/Http4sAdapter.scala#L178

When I change the line to something like this it can execute a query

```scala
val payload = msg.hcursor.downField("payload")
val id = msg.hcursor.downField("id").success.flatMap(_.value.asString).getOrElse("")
Task.whenCase(payload.as[GraphQLRequest]) {
   case Right(req) =>
       (for {
          result <- interpreter.execute(req.query.get, req.operationName, req.variables.getOrElse(Map.empty), req.extensions.getOrElse(Map.empty), skipValidation = skipValidation)
```

mitsutaka-takeda:
This commit fixes the issue #348 for 3 adapaters; Akka, Http4s, and ushttp.

### DIFF
--- a/adapters/akka-http/src/main/scala/caliban/interop/circe/CirceWSMessage.scala
+++ b/adapters/akka-http/src/main/scala/caliban/interop/circe/CirceWSMessage.scala
@@ -4,6 +4,5 @@ import caliban._
 import io.circe.ACursor
 
 private[caliban] final case class CirceWSMessage(id: String, messageType: String, payload: ACursor) extends WSMessage {
-  lazy val operationName: Option[String] = payload.downField("operationName").success.flatMap(_.value.asString)
-  lazy val query: Option[String]         = payload.downField("query").success.flatMap(_.value.asString)
+  lazy val request: Option[GraphQLRequest] = payload.as[GraphQLRequest].fold(_ => None, Some(_))
 }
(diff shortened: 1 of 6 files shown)


adapters/akka-http/src/main/scala/caliban/interop/circe/CirceWSMessage.scala
```diff
@@ -4,6 +4,5 @@ import caliban._
 import io.circe.ACursor
 
 private[caliban] final case class CirceWSMessage(id: String, messageType: String, payload: ACursor) extends WSMessage {
-  lazy val operationName: Option[String] = payload.downField("operationName").success.flatMap(_.value.asString)
-  lazy val query: Option[String]         = payload.downField("query").success.flatMap(_.value.asString)
+  lazy val request: Option[GraphQLRequest] = payload.as[GraphQLRequest].fold(_ => None, Some(_))
```
ghostdogpr, line 7: Nice, this actually simplifies the implementation!

PR #357, opened 2020-04-28
Add codegen client options for scalafmtPath and headers

### LINKED ISSUE #322: Codegen: add options to sbt command
Add an option to specify headers when using a URL. Also make the scalafmt path an option.

The current code isn't very convenient for dealing with options, maybe use something like https://github.com/scopt/scopt?

kiranbayram:
Closes #322

- Some refactoring (mostly, making private things private, moving public members to top)
- Define `--scalafmtPath` option for scalafmt conf path
- Define `--headers` option to provide headers when schemaPath is a URL
- Defined `Options` case class and moved all arguments to that
- Use `zio-config` to parse options


kiranbayram:
Hey @ghostdogpr , I'm not completely done yet but still wanted to create the draft PR. It would be nice if you can have a look. Thanks 😊

codegen/src/main/scala/caliban/codegen/Codegen.scala
```diff
@@ -7,30 +7,30 @@ import caliban.parsing.adt.Document
...
-      Task(scala.io.Source.fromFile(path))
-        .bracket(f => UIO(f.close()), f => Task(f.mkString))
-        .flatMap(Parser.parseQuery)
-    }
-
   def generate(
-    schemaPath: String,
-    toPath: String,
-    fmtPath: Option[String],
+    arguments: CommandLineArguments,
```
ghostdogpr, line 11: How about calling this something less specific like `Config` or `Options`? Since `caliban-codegen` could be used outside of sbt.

kiranbayram: Fair point, I'll change that 👍

kiranbayram:
I was actually thinking if it'd be better to do the parsing from command line arguments (`CommandLineArguments.fromargs`) in `codegen-sbt`, we can rename the case class `CommandLineArguments` and leave it in `codegen` but parsing and zio-config dependency can go to `codegen-sbt`. WDYT? @ghostdogpr

ghostdogpr:
> I was actually thinking if it'd be better to do the parsing from command line arguments (`CommandLineArguments.fromargs`) in `codegen-sbt`, we can rename the case class `CommandLineArguments` and leave it in `codegen` but parsing and zio-config dependency can go to `codegen-sbt`. WDYT? @ghostdogpr

I was initially going to suggest that, then I hesitated because I thought someone using `codegen` (the library) might be happy to use the parsing as well. I don't have a strong preference, either way is fine for me.

kiranbayram:
Alright, let's keep it in library then

kiranbayram:
I'm done for now, could you please have another look at it? @ghostdogpr

kiranbayram:
Updated ✅

PR #364, opened 2020-04-30
Pass request context to execution of subscription for Http4sAdapter

mitsutaka-takeda:
By processing a WebSocket message in a child fiber of request handler, we can pass information
extracted by http4s middlewares.

### DIFF
--- a/adapters/http4s/src/main/scala/caliban/Http4sAdapter.scala
+++ b/adapters/http4s/src/main/scala/caliban/Http4sAdapter.scala
@@ -149,10 +149,11 @@ object Http4sAdapter {
       )
 
     def processMessage(
+      receivingQueue: fs2.concurrent.Queue[Task, WebSocketFrame],
       sendQueue: fs2.concurrent.Queue[Task, WebSocketFrame],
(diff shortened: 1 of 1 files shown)


adapters/http4s/src/main/scala/caliban/Http4sAdapter.scala
```diff
@@ -208,30 +207,30 @@ object Http4sAdapter {
                     }
                   case "stop" =>
                     val id = msg.hcursor.downField("id").success.flatMap(_.value.asString).getOrElse("")
-                    subscriptions
```
ghostdogpr, line 211: I think that part is needed. When we receive an `unsubscribe` request (`stop`), we want to interrupt the fiber that was consuming the stream for that ID (and keep the other fibers running). 
We no longer need to use `subscriptions` for the keep-alive fiber but I think it's still needed for each subscription so that we can stop them before the WebSocket is closed.

mitsutaka-takeda: You're right! I'll fix this.

PR #370, opened 2020-05-05
allow configuring package name in codegen

joprice:


joprice:
@ghostdogpr I updated with what I think are the right places to put this. Let me know if you prefer different wording.

Issue #374, opened 2020-05-08
tapir integration operation name conflict


fkowal:
Let me just start by thanking you for this amazing work that you've done.
Caliban is absolutely amazing.

Here's an issue I ran into, and after some exploration, I didn't find a way to resolve it.

Here are two endpoint i described using `tapir`
```
val getOrderEndpoint: Endpoint[(UsernamePassword, String), String, OrderResponse, Nothing] =
    orders.get
      .in("orders" / path[String]("orderId")) // <- orders
      .out(jsonBody[OrderResponse])

  val getOrderListEndpoint: Endpoint[UsernamePassword, String, OrderListResponse, Nothing] =
    orders.get
      .in("orders") // <- orders
      .out(jsonBody[OrderListResponse])
```

as far as I can tell [extractPath](https://github.com/ghostdogpr/caliban/blob/master/interop/tapir/src/main/scala/caliban/interop/tapir/package.scala#L165) is responsible for defining the name of the graphql "field"

As a result the generated schema contains

```
{
              "name" : "orders",
              "description" : null,
              "args" : [
                {
                  "name" : "Authorization",
                  "description" : null,
                  "type" : {
                    "kind" : "NON_NULL",
                    "name" : null,
                    "ofType" : {
                      "kind" : "INPUT_OBJECT",
                      "name" : "UsernamePasswordInput",
                      "ofType" : null
                    }
                  },
                  "defaultValue" : null
                },
                {
                  "name" : "orderId",
                  "description" : null,
                  "type" : {
                    "kind" : "NON_NULL",
                    "name" : null,
                    "ofType" : {
                      "kind" : "SCALAR",
                      "name" : "String",
                      "ofType" : null
                    }
                  },
                  "defaultValue" : null
                }
              ],
              "type" : {
                "kind" : "OBJECT",
                "name" : "OrderResponse",
                "ofType" : null
              },
              "isDeprecated" : false,
              "deprecationReason" : null
            },
            {
              "name" : "orders",
              "description" : null,
              "args" : [
                {
                  "name" : "username",
                  "description" : null,
                  "type" : {
                    "kind" : "NON_NULL",
                    "name" : null,
                    "ofType" : {
                      "kind" : "SCALAR",
                      "name" : "String",
                      "ofType" : null
                    }
                  },
                  "defaultValue" : null
                },
                {
                  "name" : "password",
                  "description" : null,
                  "type" : {
                    "kind" : "SCALAR",
                    "name" : "String",
[...]

PR #375, opened 2020-05-08
Allow overriding the name of graphql query/operation name

fkowal:
by using endpoint.info.name if it exists

Problem: Having more than one operation with the same name
doesn't play well with graphiql and conflicts with caliban execution

#374 
Issue: https://github.com/ghostdogpr/caliban/issues/374

### DIFF
--- a/interop/tapir/src/main/scala/caliban/interop/tapir/package.scala
+++ b/interop/tapir/src/main/scala/caliban/interop/tapir/package.scala
@@ -164,7 +164,7 @@ package object tapir {
 
   private def extractPath[I](endpointName: Option[String], input: EndpointInput[I]): String =
     endpointName
-      .flatMap(toCamel)
+      .map(replaceIllegalChars)
       .getOrElse(
         input
           .asVectorOfBasicInputs(includeAuth = false)
@@ -178,14 +178,8 @@ package object tapir {
         }
       )
 
-  private def toCamel(s: String): Option[String] =
-    s.replaceAll("\\W", "_")
-      .split("_")
-      .filterNot(_.isEmpty) match {
-      case Array(head, tail @ _*) =>
-        Some(head ++ tail.map(s => s"${s.head.toUpper}${s.tail.toLowerCase()}").mkString(""))
-      case _ => None
-    }
+  private def replaceIllegalChars(s: String): String =
+    s.replaceAll("\\W+", "_")
 
   private def extractArgNames[I](input: EndpointInput[I]): Map[String, Option[(String, Option[String])]] =
     input.traverseInputs {
--- a/interop/tapir/src/main/scala/caliban/interop/tapir/package.scala
+++ b/interop/tapir/src/main/scala/caliban/interop/tapir/package.scala
@@ -181,10 +181,10 @@ package object tapir {
   private def toCamel(s: String): Option[String] =
     s.replaceAll("\\W", "_")
       .split("_")
-      .filterNot(_.isEmpty)
-      .map(s => s"${s.head.toUpper}${s.tail.toLowerCase()}") match {
-      case Array(head, tail @ _*) => Some((head.toLowerCase ++ tail).mkString(""))
-      case _                      => None
+      .filterNot(_.isEmpty) match {
+      case Array(head, tail @ _*) =>
+        Some(head ++ tail.map(s => s"${s.head.toUpper}${s.tail.toLowerCase()}").mkString(""))
+      case _ => None
     }
 
   private def extractArgNames[I](input: EndpointInput[I]): Map[String, Option[(String, Option[String])]] =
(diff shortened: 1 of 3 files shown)


interop/tapir/src/main/scala/caliban/interop/tapir/package.scala
```diff
@@ -162,17 +162,19 @@ package object tapir {
     override protected val additionalDirectives: List[__Directive] = Nil
   }
 
-  private def extractPath[I](input: EndpointInput[I]): String =
-    input
-      .asVectorOfBasicInputs(includeAuth = false)
-      .collect {
-        case EndpointInput.FixedPath(s, _, _) => s
+  private def extractPath[I](endpointName: Option[String], input: EndpointInput[I]): String =
+    endpointName.getOrElse(
```
ghostdogpr, line 166: Maybe add some sanitization in case the name contains some spaces? Like turning `Add a book` into `AddABook`? In case people use actual sentences there.

interop/tapir/src/main/scala/caliban/interop/tapir/package.scala
```diff
@@ -162,17 +162,19 @@ package object tapir {
...
+      input
+        .asVectorOfBasicInputs(includeAuth = false)
+        .collect {
+          case EndpointInput.FixedPath(s, _, _) => s
```
ghostdogpr, line 170: I'm wondering if we should catch `PathCapture` here and use the name for the field. With your example I imagine there would be a `PathCapture` with name `orderId`, so maybe we could generate a field `ordersByOrderId` or `ordersWithOrderId`?

interop/tapir/src/main/scala/caliban/interop/tapir/package.scala
fkowal: handled this part, pls have a look

as for `PathCapture` i'll try to do it

interop/tapir/src/main/scala/caliban/interop/tapir/package.scala
```diff
@@ -162,16 +162,29 @@ package object tapir {
...
+          case head :: tail => head ++ tail.map(_.capitalize).mkString
+        }
+      )
+
+  private def toCamel(s: String): Option[String] =
+    s.replaceAll("\\W", "_")
+      .split("_")
+      .filterNot(_.isEmpty)
+      .map(s => s"${s.head.toUpper}${s.tail.toLowerCase()}") match {
+      case Array(head, tail @ _*) => Some((head.toLowerCase ++ tail).mkString(""))
```
ghostdogpr, line 186: If someone named it well like `addBook`, it would turn it into `addbook`, no? Maybe we should preserve casing for the tail?

interop/tapir/src/main/scala/caliban/interop/tapir/package.scala
ghostdogpr: By the way I’m fine if you prefer splitting this into 2 PRs and handle PathCapture later. 

Also, can you modify the docs in vuepress/docs/docs/interop.md to mention the new rule for naming fields?

interop/tapir/src/main/scala/caliban/interop/tapir/package.scala
fkowal: i started with
```
if (name.matches(".*\\W")) doNameFixing(name) else name
```
because overall the less a library forces naming conventions the better

it'd rather the library enable customization, and not enforce custom rules

so if a user wants `snake_case_is_superior` that should ok

What do you think?

I tent to of look at such issues from testing perspective.
Should I test the `addBook -> addbook` ? I am not convinced

Doing only `name.replaceAll("\\W","_")` could be enough.
This, of course, is only valid if the library doesn't already follow a certain style.

I am good with either case

interop/tapir/src/main/scala/caliban/interop/tapir/package.scala
fkowal: Yes I'd much rather do the second part as a separate PR

This somewhat depends on what we agree on in my other comment https://github.com/ghostdogpr/caliban/pull/375#discussion_r422998153

interop/tapir/src/main/scala/caliban/interop/tapir/package.scala
ghostdogpr: I agree, we probably shouldn't do too much and only prevent invalid graphql schema to be generated. So replacing the spaces with `_` is fine with me.

interop/tapir/src/main/scala/caliban/interop/tapir/package.scala
ghostdogpr: Sounds good. Thanks for adding the docs.

interop/tapir/src/main/scala/caliban/interop/tapir/package.scala
fkowal: so I think this is all for this PR

the second part i will have to dig myself a bit more into the lovely codebase

PR #382, opened 2020-05-11
Add Play Support

### LINKED ISSUE #143: Add module for Play support

javimartinez:
Closes #143

* Added new module for PlayAdapter 
* Added example app using an embedded Akka Http Server 
* Docs updated to add the link for Play example


build.sbt
```diff
@@ -263,10 +278,13 @@ lazy val examples = project
       "de.heikoseeberger"            %% "akka-http-circe"               % "1.32.0",
       "com.softwaremill.sttp.client" %% "async-http-client-backend-zio" % sttpVersion,
       "com.softwaremill.sttp.tapir"  %% "tapir-json-circe"              % tapirVersion,
-      "io.circe"                     %% "circe-generic"                 % circeVersion
+      "io.circe"                     %% "circe-generic"                 % circeVersion,
+      "com.typesafe.play"            %% "play-akka-http-server"         % playVersion,
+      "com.typesafe.akka"            %% "akka-actor-typed"              % akkaVersion,
```
javimartinez, line 283: I had to add both `akka-actor-typed` and `akka-slf4j` dependencies to avoid the error below:
`
java.lang.IllegalStateException: You are using version 2.6.5 of Akka, but it appears you (perhaps indirectly) also depend on older versions of related artifacts. You can solve this by adding an explicit dependency on version 2.6.5 of the [akka-actor-typed, akka-slf4j] artifacts to your project. See also: https://doc.akka.io/docs/akka/current/common/binary-compatibility-rules.html#mixed-versioning-is-not-allowed
`

adapters/play/src/main/scala/caliban/PlayAdapter.scala
```diff
@@ -0,0 +1,197 @@
...
+import play.api.mvc.{ Action, ActionBuilder, AnyContent, PlayBodyParsers, Request, RequestHeader, Result, WebSocket }
+import play.api.mvc.Results.Ok
+import zio.{ CancelableFuture, Fiber, IO, RIO, Ref, Runtime, Schedule, Task, ZIO }
+import zio.clock.Clock
+import zio.duration.Duration
+import scala.concurrent.{ ExecutionContext, Future }
+
+trait PlayAdapter {
+
+  def json: JsonBackend
```
ghostdogpr, line 18: I think you can simplify the API and not expose this, instead just use the play json backend directly.

build.sbt
```diff
@@ -61,6 +63,7 @@ lazy val root = project
     http4s,
     akkaHttp,
     uzhttp,
+    play,
```
ghostdogpr, line 66: Can you also add this to the config file in `.circleci`?

adapters/play/src/main/scala/caliban/PlayAdapter.scala
```diff
@@ -0,0 +1,197 @@
...
+  def makePostAction[R, E](
+    interpreter: GraphQLInterpreter[R, E],
+    skipValidation: Boolean = false,
+    enableIntrospection: Boolean = true
+  )(implicit runtime: Runtime[R]): Action[GraphQLRequest] =
+    actionBuilder.async(parse.json[GraphQLRequest])(req =>
+      executeRequest(
+        interpreter,
+        req.body,
+        skipValidation: Boolean,
```
ghostdogpr, line 47: Why specifying the types?

build.sbt
```diff
@@ -234,6 +237,18 @@ lazy val uzhttp = project
...
   .dependsOn(core)
 
+lazy val play = project
+  .in(file("adapters/play"))
+  .settings(name := "caliban-play")
+  .settings(commonSettings)
+  .settings(
+    libraryDependencies ++= Seq(
+      "com.typesafe.play" %% "play"                % playVersion,
+      "de.heikoseeberger" %% "akka-http-play-json" % "1.32.0"
```
ghostdogpr, line 247: Why do we need the dependency to `akka-http-play-json`?

build.sbt
ghostdogpr: It's caused by `akka-http-play-json` that uses an older version of Akka but I wonder why we need that one, see my other comment.

examples/src/main/scala/caliban/play/CalibanController.scala
```diff
@@ -0,0 +1,51 @@
+package caliban.play
+
+import akka.actor.ActorSystem
+import akka.stream.Materializer
+import caliban.{
```
ghostdogpr, line 5: You can import `._` at this point.

adapters/play/src/main/scala/caliban/PlayAdapter.scala
javimartinez: done!

build.sbt
javimartinez: done!

adapters/play/src/main/scala/caliban/PlayAdapter.scala
javimartinez: I overlooked them :( Removed.

build.sbt
javimartinez: It's because I'm using [PlayJsonBackend](https://github.com/ghostdogpr/caliban/blob/master/adapters/akka-http/src/main/scala/caliban/interop/play/PlayJsonBackend.scala) from AkkaHttp module, Maybe we just copy/paste this code to avoid that dependency?

examples/src/main/scala/caliban/play/CalibanController.scala
javimartinez: done!

build.sbt
ghostdogpr: Can it be written without the Akka stuff as a pure play thing? If that’s the case I think it’s better to avoid the dependency.

ghostdogpr:
Ah there’s a conflict with master so you’ll need a rebase :)

build.sbt
javimartinez: Yes! I've written it already (you can check it out in the last commit). The only thing I'm not pretty sure if it should be public for the users o not. What do you think?

build.sbt
ghostdogpr: I guess those 2 aren't needed anymore?

examples/src/main/scala/caliban/play/ExampleApp.scala
```diff
@@ -0,0 +1,45 @@
...
+      DefaultControllerComponents(
+        components.defaultActionBuilder,
+        components.playBodyParsers,
+        components.messagesApi,
+        components.langs,
+        components.fileMimeTypes,
+        components.executionContext
+      )
+    )(components.actorSystem, components.materializer)
+    val router: PartialFunction[RequestHeader, Handler] = {
```
ghostdogpr, line 30: Is it possible to serve graphiql from the resources like the other examples?

adapters/play/src/main/scala/caliban/PlayWSMessage.scala
```diff
@@ -0,0 +1,7 @@
+package caliban
+
+import play.api.libs.json.JsObject
+
+final case class PlayWSMessage(id: String, messageType: String, payload: Option[JsObject]) {
+  lazy val request: Option[GraphQLRequest] = payload.flatMap(_.validate[GraphQLRequest].asOpt)
```
paulpdaniels, line 6: Hint: `_.validate.asOpt` can just be `_.asOpt[GraphQLRequest]`

adapters/play/src/main/scala/caliban/PlayJson.scala
```diff
@@ -0,0 +1,67 @@
...
+
+trait PlayJson {
+
+  implicit def writableGraphQLResponse[E](implicit wr: Writes[GraphQLResponse[E]]): Writeable[GraphQLResponse[E]] =
+    Writeable.writeableOf_JsValue.map(wr.writes)
+
+  private def parseJson(s: String): Try[JsValue] =
+    Try(Json.parse(s))
+
+  def parseHttpRequest(
```
paulpdaniels, line 18: There should already be a reads for this.

adapters/play/src/main/scala/caliban/PlayJson.scala
```diff
@@ -0,0 +1,67 @@
...
+    )
+
+  def encodeWSError(id: String, error: Throwable): String =
+    Json.stringify(
+      Json
+        .obj(
+          "id"      -> id,
+          "type"    -> "complete",
+          "payload" -> error.toString
+        )
```
paulpdaniels, line 65: Make this a `Writes`, also I don't think this is correct, the `type` should be `error` if this is an error message, and the payload should be an error object not a string.

adapters/play/src/main/scala/caliban/PlayJson.scala
```diff
@@ -0,0 +1,67 @@
...
+    Json.stringify(
+      Json
+        .obj(
+          "id"      -> id,
+          "type"    -> "data",
+          "payload" -> GraphQLResponse(data, errors)
+        )
+    )
+
+  def encodeWSError(id: String, error: Throwable): String =
```
paulpdaniels, line 58: I'd suggest using a `CalibanError` here instead

examples/src/main/scala/caliban/play/ExampleApp.scala
```diff
@@ -0,0 +1,45 @@
...
+      case POST(p"/api/graphql") => controller.graphqlPost
+      case GET(
+          p"/api/graphql" ? q"query=$query" & q_o"variables=$variables" & q_o"operation=$operation" & q_o"extensions=$extensions"
+          ) =>
+        controller.graphqlGet(query, variables, operation, extensions)
+      case GET(p"/ws/graphql") => controller.webSocket()
```
paulpdaniels, line 36: Couldn't we just roll this up into a `Router` ourselves?

examples/src/main/scala/caliban/play/CalibanController.scala
```diff
@@ -0,0 +1,42 @@
...
+  implicit val ec      = actorSystem.dispatcher
+
+  val service: ZLayer[Any, Nothing, ExampleService] = ExampleService.make(ExampleData.sampleCharacters)
+
+  val interpreter: GraphQLInterpreter[ZEnv, CalibanError] = runtime.unsafeRun(
+    ExampleService
+      .make(ExampleData.sampleCharacters)
+      .memoize
+      .use(layer => ExampleApi.api.interpreter.map(_.provideCustomLayer(layer)))
+  )
```
paulpdaniels, line 28: Usually I move the `layer` into the runtime, then bind the shutdown hook to the runtime shutdown, which cleans up the resources, but that is a matter of preference I suppose.

adapters/play/src/main/scala/caliban/PlayJson.scala
```diff
@@ -0,0 +1,67 @@
...
+      variablesJs.map(js => "variables"   -> js) ++
+      extensionsJs.map(js => "extensions" -> js)
+    JsObject(fields)
+      .validate[GraphQLRequest]
+      .asEither
+      .left
+      .map(parsingException)
+  }
+
+  def encodeGraphQLResponse(r: GraphQLResponse[Any]): String = Json.toJson(r).toString()
```
paulpdaniels, line 37: Doesn't appear to be used?

adapters/play/src/main/scala/caliban/PlayJson.scala
```diff
@@ -0,0 +1,67 @@
...
+
+  def parseWSMessage(text: String): Either[Throwable, PlayWSMessage] =
+    parseJson(text).toEither.map { json =>
+      PlayWSMessage(
```
paulpdaniels, line 41: Make a `Reads` for this.

examples/src/main/scala/caliban/play/ExampleApp.scala
javimartinez: Sure! done :)

adapters/play/src/main/scala/caliban/PlayWSMessage.scala
javimartinez: Applied :)

adapters/play/src/main/scala/caliban/PlayJson.scala
javimartinez: I was using this function to get the GraphqlRequest from query params, probably the name wasn't intuitive. I've replaced it with https://github.com/ghostdogpr/caliban/pull/382/files#diff-e5658f60c6f05951486935dea5234209R29

adapters/play/src/main/scala/caliban/PlayJson.scala
javimartinez: removed

adapters/play/src/main/scala/caliban/PlayJson.scala
javimartinez: I've made Writes and Reads for PlayWSMessage,  the code has been simplified, and it seems more idiomatic. Thank you for the tip!

javimartinez: I'm not sure if the `type` should be `error` here. I've checked the other adapters, and they use `complete` in this case. Maybe we need to check the protocol, but I haven't found a lot of docs about it (only https://github.com/apollographql/subscriptions-transport-ws/blob/master/PROTOCOL.md) Do you have more info or any doc @ghostdogpr?

adapters/play/src/main/scala/caliban/PlayJson.scala
javimartinez: done

examples/src/main/scala/caliban/play/ExampleApp.scala
javimartinez: Do you mean using [Sird Router](https://www.playframework.com/documentation/2.8.x/ScalaSirdRouter#Binding-sird-Router)?

adapters/play/src/main/scala/caliban/PlayJson.scala
ghostdogpr: Yeah the docs are not very detailed in this case. Looking at the code, I think this can never happen actually. We have a `Throwable` because we wrap Futures but the only thing they do is offering to a queue. I think you can keep it consistent with the other adapters and we might clean that up later.

examples/src/main/scala/caliban/play/ExampleApp.scala
paulpdaniels: Yeah, I'm thinking the easiest out-of-the-box setup would just be like:

```
val router = CalibanRouter(
  subscriptions = true,
  playground = true,
  allowGETRequests = true
)
```

paulpdaniels: Keep the others accessible, because people might want to manipulate configuration, but that would allow pretty much drop in usage.

examples/src/main/resources/application.conf
```diff
@@ -0,0 +1,2 @@
+# Used by the Play example
+play.http.secret.key = "changeme"
```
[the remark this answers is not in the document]
javimartinez: Not really, line removed. However, we need to maintain the `application.conf` because it seems that play loads it by default. Without it, I'm getting the next error: 
```[error] Configuration error: Configuration error[application: application.conf: java.io.IOException: resource not found on classpath: application.conf, application.json: java.io.IOException: resource not found on classpath: application.json, application.properties: java.io.IOException: resource not found on classpath: application.properties]```

File: build.sbt [hunk 17]
```diff
@@ -265,10 +279,13 @@ lazy val examples = project
       "de.heikoseeberger"            %% "akka-http-circe"               % "1.32.0",
       "com.softwaremill.sttp.client" %% "async-http-client-backend-zio" % sttpVersion,
       "com.softwaremill.sttp.tapir"  %% "tapir-json-circe"              % tapirVersion,
-      "io.circe"                     %% "circe-generic"                 % circeVersion
+      "io.circe"                     %% "circe-generic"                 % circeVersion,
+      "com.typesafe.play"            %% "play-akka-http-server"         % playVersion,
+      "com.typesafe.akka"            %% "akka-actor-typed"              % akkaVersion,
```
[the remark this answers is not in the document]
javimartinez: I've removed `akka-slf4j`, but sadly without `Akka-actor-typed` it doesn't work. We still have `akka-http-circe` here, with a different Akka version  (2.6.4). [akka-http-json](https://github.com/hseeberger/akka-http-json/blob/master/build.sbt#L176) has updated the version of akka in master. Hopefully, we'll have the next version soon, and we will be able to remove the `Akka-actor-typed` dependency.

examples/src/main/scala/caliban/play/ExampleApp.scala
javimartinez: ahh got it, it's a good point. I'll try to do it.

javimartinez: I've added a PlayRouter. Could you check my last commit out? and let me know what do you think.

adapters/play/src/main/scala/caliban/PlayJson.scala
javimartinez: This is no longer needed since now we are using Writes for WSMessages

adapters/play/src/main/scala/caliban/PlayWSMessage.scala
```diff
@@ -0,0 +1,27 @@
...
+
+  implicit val playWSMessageReads: Reads[PlayWSMessage] =
+    ((__ \ "type").read[String] and
+      (__ \ "id").readNullable[String]
+      and (__ \ "payload").readNullable[JsValue])(PlayWSMessage.apply _)
+
+  implicit val playWSMessageWrites: Writes[PlayWSMessage] =
+    ((JsPath \ "type").write[String] ~
+      (JsPath \ "id").writeNullable[String] ~
+      (JsPath \ "payload").writeNullable[JsValue])(unlift(PlayWSMessage.unapply))
```
paulpdaniels, line 23: Can collapse these `Reads` and `Writes` into a `Format`

either: 
`implicit val playWSMessageFormat: Format[PlayWSMessage] = Json.format[PlayWSMessage]`

or if the `macro` won't work because there are two apply methods then can do:

```scala
  implicit val playWSMessageReads: Reads[PlayWSMessage] =
    ((__ \ "type").format[String] and
      (__ \ "id").formatNullable[String]
      and (__ \ "payload").formatNullable[JsValue])(PlayWSMessage.apply _, unlift(PlayWSMessage.unapply))
```

adapters/play/src/main/scala/caliban/PlayRouter.scala
```diff
@@ -0,0 +1,44 @@
...
+import akka.stream.Materializer
+import play.api.mvc.{ ActionBuilder, AnyContent, ControllerComponents, PlayBodyParsers, Request, Results }
+import play.api.routing.Router.Routes
+import play.api.routing.SimpleRouter
+import play.api.routing.sird._
+import zio.Runtime
+import zio.duration.Duration
+import scala.concurrent.ExecutionContext
+
+case class PlayRouter[R, E](
```
paulpdaniels, line 12: Awesome! Love it 💯

adapters/play/src/main/scala/caliban/PlayAdapter.scala
```diff
@@ -0,0 +1,218 @@
...
+    op: Option[String],
+    vars: Option[String],
+    exts: Option[String]
+  ): Either[Throwable, GraphQLRequest] = {
+    val variablesJs  = vars.flatMap(parseJson(_).toOption)
+    val extensionsJs = exts.flatMap(parseJson(_).toOption)
+    val fields = List("query" -> JsString(query)) ++
+      op.map(o => "operationName"         -> JsString(o)) ++
+      variablesJs.map(js => "variables"   -> js) ++
+      extensionsJs.map(js => "extensions" -> js)
```
paulpdaniels, line 40: Could this just be:

```scala
Json.obj(
  "query" -> query,
  "operationName" -> op,
  "variables" -> variablesJs,
  "extensions" -> extensionsJs
).validate[GraphQLRequest]
```

Also technically, query can be empty if this is a `persisted-query`, so I think you can make `query` optional and let the interpreter handle it.

paulpdaniels:
Few nits but LGTM

adapters/play/src/main/scala/caliban/PlayWSMessage.scala
javimartinez: done!

adapters/play/src/main/scala/caliban/PlayAdapter.scala
javimartinez: I agree with you, `query` should be optional. I've done it already. I think we need to check this in the other adapters since there is a mismatch between them. If @ghostdogpr agrees too, I can adjust it in another PR.

ghostdogpr: It was already supposed to be optional, and I think it's the case in http4s/uzhttp adapters. It's a bug that it's not in the Akka one 😄 If you can fix it at the same time that'd be great!

ghostdogpr: Since this PR is ready and a release is coming soon, I'll merge it and this can be fixed in another PR as you suggested.

PR #385, opened 2020-05-13
move codegen console output into sbt module

joprice:
This moves the output that the codegen process writes to stdout into the sbt plugin, in order to let other build tools consuming the core codegen module to decide control what and whether output is printed.

### DIFF
--- a/codegen/src/main/scala/caliban/codegen/Codegen.scala
+++ b/codegen/src/main/scala/caliban/codegen/Codegen.scala
@@ -1,7 +1,7 @@
 package caliban.codegen
 
 import java.io.{ File, PrintWriter }
-import zio.{ RIO, Task, UIO }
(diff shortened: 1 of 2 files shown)


codegen/src/main/scala/caliban/codegen/Codegen.scala
```diff
@@ -2,27 +2,25 @@ package caliban.codegen
...
-import zio.console.{ putStrLn, Console }
 import caliban.parsing.adt.Document
 import caliban.parsing.Parser
 
 object Codegen {
   def generate(
     arguments: Options,
     writer: (Document, String, Option[String]) => String
-  ): RIO[Console, Unit] =
+  ): RIO[Any, Unit] = {
```
ghostdogpr, line 12: Minor: change this to a `Task`

PR #427, opened 2020-05-29
Fix nested type name computation

### LINKED ISSUE #425: Schema derivation doesn't handle nested types
Hello there, as already discussed in Discord, opening the issue here:
```
@ final case class ApiTranslated[T](value: T )
defined class ApiTranslated

@ case class Data2(a: ApiTranslated[Option[Double]], b: ApiTranslated[Option[Int]])
defined class Data2

@ val data2 = Data2(ApiTranslated(Some(1.0)), ApiTranslated(Some(5)))
data2: Data2 = Data2(ApiTranslated(Some(1.0)), ApiTranslated(Some(5)))

@ GraphQL.graphQL(RootResolver(data2))
res44: GraphQL[Any] = caliban.GraphQL$$anon$5@1ef8f1ca

@ res44.render
res45: String = """schema {
  query: Data2
}

type ApiTranslatedOption {
  value: Float
}

type Data2 {
  a: ApiTranslatedOption!
  b: ApiTranslatedOption!
}
```

Second field's(`Data2.b`) type is ApiTranslated[Option[Int]] but it's mapped to Float in schema

kiranbayram:
Closes #425

### DIFF
--- a/core/src/main/scala/caliban/schema/Schema.scala
+++ b/core/src/main/scala/caliban/schema/Schema.scala
@@ -452,8 +452,12 @@ trait DerivationSchema[R] {
     getDirectives(ctx.annotations)
 
   private def getName(annotations: Seq[Any], typeName: TypeName): String =
-    annotations.collectFirst { case GQLName(name) => name }
-      .getOrElse(typeName.short + typeName.typeArguments.map(_.short).mkString)
+    annotations.collectFirst { case GQLName(name) => name }.getOrElse {
+      typeName.typeArguments match {
+        case Nil  => typeName.short
+        case args => typeName.short + args.map(getName(Nil, _)).mkString
+      }
+    }
 
   private def getName[Typeclass[_], Type](ctx: ReadOnlyCaseClass[Typeclass, Type]): String =
     getName(ctx.annotations, ctx.typeName)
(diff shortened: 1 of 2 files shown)


kiranbayram:
Gave this a go @ghostdogpr It seems to be working according to the test i added but couldn't make the function tail recursive at this moment

core/src/main/scala/caliban/schema/Schema.scala
```diff
@@ -452,8 +452,12 @@ trait DerivationSchema[R] {
     getDirectives(ctx.annotations)
 
   private def getName(annotations: Seq[Any], typeName: TypeName): String =
-    annotations.collectFirst { case GQLName(name) => name }
-      .getOrElse(typeName.short + typeName.typeArguments.map(_.short).mkString)
+    annotations.collectFirst { case GQLName(name) => name }.getOrElse {
+      typeName.typeArguments match {
+        case Nil  => typeName.short
+        case args => typeName.short + args.map(getName(annotations, _)).mkString
```
ghostdogpr, line 458: I think you can pass an empty list for annotations, to avoid searching something that won’t be found.

kiranbayram: good point 👍 updated.

PR #430, opened 2020-05-30
Unwrap value classes by default

### LINKED ISSUE #426: Improve automatic derivation of value classes
The following value class
```scala
case class OrganizationId(value: Long) extends AnyVal
```
gives the following schema
```graphql
events { organizationId { value } }
```

Magnolia exposes a boolean `isValueClass`, so it should be possible to change the default derivation of case classes and use the underlying `Schema` when `isValueClass` is true.

palanga:
Closes #426

### DIFF
--- a/core/src/main/scala/caliban/schema/Schema.scala
+++ b/core/src/main/scala/caliban/schema/Schema.scala
@@ -1,22 +1,24 @@
 package caliban.schema
 
 import java.util.UUID
-import scala.annotation.implicitNotFound
-import scala.concurrent.Future
-import scala.language.experimental.macros
+
 import caliban.CalibanError.ExecutionError
 import caliban.ResponseValue._
 import caliban.Value._
 import caliban.introspection.adt._
 import caliban.parsing.adt.Directive
-import caliban.schema.Annotations.{ GQLDeprecated, GQLDescription, GQLDirective, GQLInputName, GQLInterface, GQLName }
+import caliban.schema.Annotations._
 import caliban.schema.Step._
 import caliban.schema.Types._
 import caliban.{ InputValue, ResponseValue }
 import magnolia._
-import zio.{ URIO, ZIO }
-import zio.stream.ZStream
 import zio.query.ZQuery
+import zio.stream.ZStream
+import zio.{ URIO, ZIO }
+
+import scala.annotation.implicitNotFound
+import scala.concurrent.Future
+import scala.language.experimental.macros
 
 /**
  * Typeclass that defines how to map the type `T` to the according GraphQL concepts: how to introspect it and how to resolve it.
@@ -313,7 +315,8 @@ trait DerivationSchema[R] {
 
   def combine[T](ctx: ReadOnlyCaseClass[Typeclass, T]): Typeclass[T] = new Typeclass[T] {
     override def toType(isInput: Boolean = false): __Type =
-      if (isInput)
+      if (ctx.isValueClass && ctx.parameters.nonEmpty) ctx.parameters.head.typeclass.toType(isInput)
+      else if (isInput)
         makeInputObject(
           Some(ctx.annotations.collectFirst { case GQLInputName(suffix) => suffix }
             .getOrElse(customizeInputTypeName(getName(ctx)))),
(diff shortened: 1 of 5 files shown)


palanga:
* Some import reorgs were made by IntelliJ, I can revert those if you want.
* Should we put something in the docs ?

core/src/main/scala/caliban/schema/Schema.scala
```diff
@@ -313,7 +315,8 @@ trait DerivationSchema[R] {
 
   def combine[T](ctx: ReadOnlyCaseClass[Typeclass, T]): Typeclass[T] = new Typeclass[T] {
     override def toType(isInput: Boolean = false): __Type =
-      if (isInput)
+      if (ctx.isValueClass) ctx.parameters.headOption.map(_.typeclass.toType(isInput)).get
```
palanga, line 318: We can make `ctx.parameters.head.typeclass.toType(isInput)` instead. I guess that would be more efficient but unsafe. The las `.get` is unsafe also, but a little bit less unsafe.

core/src/main/scala/caliban/schema/Schema.scala
```diff
@@ -356,6 +359,7 @@ trait DerivationSchema[R] {
 
     override def resolve(value: T): Step[R] =
       if (ctx.isObject) PureStep(EnumValue(getName(ctx)))
+      else if (ctx.isValueClass) ctx.parameters.headOption.map(p => p.typeclass.resolve(p.dereference(value))).get
```
palanga, line 362: ditto

core/src/main/scala/caliban/schema/Schema.scala
ghostdogpr: It shouldn't be possible to have anything else than 1 parameter but just in case, maybe you can add a check in the `if`? Then you can just use `head` safely.

core/src/main/scala/caliban/schema/Schema.scala
```diff
@@ -1,22 +1,24 @@
...
 import caliban.schema.Types._
 import caliban.{ InputValue, ResponseValue }
 import magnolia._
-import zio.{ URIO, ZIO }
-import zio.stream.ZStream
 import zio.query.ZQuery
+import zio.stream.ZStream
+import zio.{ URIO, ZIO }
+
+import scala.annotation.implicitNotFound
```
ghostdogpr, line 19: I used to have java imports, then scala imports, then the rest but maybe better to stick with IntelliJ defaults which is what you did.

ghostdogpr:
You can add "Closes #426" in the PR description so that it closes the issue automatically when the PR is merged.

About the docs, it would be nice to mention it in this page: https://github.com/ghostdogpr/caliban/blob/master/vuepress/docs/docs/schema.md

palanga:
> You can add "Closes #426" in the PR description so that it closes the issue automatically when the PR is merged.

I forgot that again 🤦 sorry.

core/src/main/scala/caliban/schema/Schema.scala
```diff
@@ -356,7 +359,11 @@ trait DerivationSchema[R] {
 
     override def resolve(value: T): Step[R] =
       if (ctx.isObject) PureStep(EnumValue(getName(ctx)))
-      else ObjectStep(getName(ctx), ctx.parameters.map(p => p.label -> p.typeclass.resolve(p.dereference(value))).toMap)
+      else if (ctx.isValueClass && ctx.parameters.nonEmpty) {
+        val head = ctx.parameters.head
+        head.typeclass.resolve(head.dereference(value))
```
palanga, line 364: For some reason, the above compiles but this doesn't: `ctx.parameters.head.typeclass.resolve(ctx.parameters.head.dereference(value))`

We get:
```
[error] /Users/palan/code/caliban/core/src/main/scala/caliban/schema/Schema.scala:363:78: type mismatch;
[error]  found   : magnolia.ReadOnlyParam[DerivationSchema.this.Typeclass,T]#PType
[error]  required: _1.PType where val _1: magnolia.ReadOnlyParam[DerivationSchema.this.Typeclass,T]
[error]         ctx.parameters.head.typeclass.resolve(ctx.parameters.head.dereference(value))
[error]                                                                              ^
```

vuepress/docs/docs/schema.md
```diff
@@ -191,6 +191,8 @@ implicit val localDateArgBuilder: ArgBuilder[LocalDate] = {
 }
 ```
 
+Value classes (`case class SomeWrapper(self: SomeType) extends AnyVal`) will be unwrapped by default.
+
```
palanga, line 195: Thumbs up or thumbs down ?

PR #436, opened 2020-06-07
Schema diff

### LINKED ISSUE #372: Is there a way to diff two GraphQL schemas?
Such feature is quite important for versioning the GraphQL schemas, when one would like to compare the differences between two schema versions.

ghostdogpr:
Closes #372 

Example usage:
```scala
val schema: String =
  """
  type Hero {
    name(pad: Int!): String!
    nick: String!
    bday: Int
  }
  
  type Query {
    hero: Hero!
  }
  """

case class NameArgs(pad: Int)
case class Hero(name: NameArgs => String, nick: String, bday: Option[Int])
case class Query(hero: Hero)

val api = graphQL(RootResolver(Query(Hero(_ => "name", "nick", None))))

val from = SchemaLoader.fromString(schema)
val to   = SchemaLoader.fromCaliban(api)

assertM(for {
  diff <- SchemaComparison.compare(from, to)
} yield diff)(equalTo(Nil))
```


tools/src/test/scala/caliban/tools/SchemaComparisonSpec.scala
```diff
@@ -0,0 +1,143 @@
...
+            |""".stripMargin
+
+        case class NameArgs(pad: Int)
+        case class Hero(name: NameArgs => String, nick: String, bday: Option[Int])
+        case class Query(hero: Hero)
+
+        val api = graphQL(RootResolver(Query(Hero(_ => "name", "nick", None))))
+
+        assertM(for {
+          diff <- SchemaComparison.compare(SchemaLoader.fromString(schema), SchemaLoader.fromCaliban(api))
```
ghostdogpr, line 139: Example Usage

PR #440, opened 2020-06-10
Fix object type validation

ghostdogpr:
Fix some false positive validation error that happened with objects implementing interfaces.


ghostdogpr:
@zach-albia if you have a chance, can you review this?

zach-albia:
Looks good. Thanks for the fix!

PR #443, opened 2020-06-12
Add --effect option to codegen

AlixBa:
In order not to rely only on synchronous code execution, it can be useful to inject a specific effect type in order to integrate well with existing codebase


codegen-sbt/src/main/scala/caliban/codegen/CodegenPlugin.scala
```diff
@@ -59,7 +59,7 @@ object CodegenPlugin extends AutoPlugin {
   def execGenCommand(
     helpMsg: String,
     args: List[String],
-    writer: (Document, String, Option[String]) => String
+    writer: (Document, String, Option[String], String) => String
```
ghostdogpr, line 62: Can you also add the new option in the Help message a few lines above this? (only for `calibanGenSchema`)

ghostdogpr:
Actually I was planning to do a release today and it would be pity not to include this, so I'm gonna merge it and add the help message on master 😄

PR #449, opened 2020-06-16
Make the schema type of a ZStream a List

avdv:
Hi.

Given this Queries object:
```scala
case class Queries(
      @GQLDescription("return all images")
      images: ImagesArgs => ZStream[ExampleService, Throwable, Image]
  )
```
The introspection would return this piece of information:
```json
{
            "name": "images",
            "description": "return all images",
            "args": [
               ...
            ],
            "type": {
              "kind": "NON_NULL",
              "name": null,
              "ofType": { "kind": "OBJECT", "name": "Image", "ofType": null }
            },
            "isDeprecated": false,
            "deprecationReason": null
          }
```
which, AFAIU, means that it returns a single `Image`, but the stream emits multiple images of course.

### DIFF
--- a/core/src/main/scala/caliban/schema/Schema.scala
+++ b/core/src/main/scala/caliban/schema/Schema.scala
@@ -297,7 +297,10 @@ trait GenericSchema[R] extends DerivationSchema[R] {
   ): Schema[R, ZStream[R1, E, A]] =
     new Schema[R, ZStream[R1, E, A]] {
       override def optional: Boolean                          = ev.optional
-      override def toType(isInput: Boolean = false): __Type   = ev.toType(isInput)
+      override def toType(isInput: Boolean = false): __Type   = {
+        val t = ev.toType(isInput)
+        makeList(if (ev.optional) t else makeNonNull(t))
+      }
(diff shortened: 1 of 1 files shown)


ghostdogpr:
That’s because ZStreams were initially supposed to be used with Subscriptions only and in that case they shouldn’t return a List.

At some point I made them return a list if used in a Query but yeah the type doesn’t match. The problem is that at this place we have no idea if the type is being used in a Query or a Subscription.

avdv:
> That’s because ZStreams were initially supposed to be used with Subscriptions only and in that case they shouldn’t return a List.
> 
> At some point I made them return a list if used in a Query but yeah the type doesn’t match. The problem is that at this place we have no idea if the type is being used in a Query or a Subscription.

OK, that's too bad... I wanted to use this from another tool which generates code that validates the return values and of course blows up. Not a big deal for me, but it would be nice to resolve this some day.

ghostdogpr:
@avdv I created https://github.com/ghostdogpr/caliban/issues/450 to keep track of this problem. A workaround for you if you don't use Subscriptions is to just copy that schema instead of using the one from Caliban.

ghostdogpr:
@avdv just opened a PR for that https://github.com/ghostdogpr/caliban/pull/457

PR #466, opened 2020-06-23
Serialize into case object when there is no field

AlixBa:


ghostdogpr:
Hmm, this shouldn't be possible as per the spec: `An Object type must define one or more fields.` (http://spec.graphql.org/June2018/#sec-Objects)

AlixBa:
Oh right, I knew that was weird.
I think I'll change the online linter I usually use.

PR #490, opened 2020-07-04
Performance improvements

ghostdogpr:
Includes the following improvements:
- Avoid re-generating `__Type` objects all the time
- Make calls to `sourceMapper.getLocation` lazy since it's only needed in case of error
- Use mutable data structures in internal code that's on the hot path
- Optimize the whitespace parsing by fastparse by @yoohaemin 

Benchmarks show that the introspection query is executed 4x faster.

Before:
```
[info] Benchmark                             Mode  Cnt      Score     Error  Units
[info] GraphQLBenchmarks.introspectCaliban  thrpt    5    565.365 ±  27.284  ops/s
[info] GraphQLBenchmarks.introspectSangria  thrpt    5    485.646 ±   7.157  ops/s
[info] GraphQLBenchmarks.simpleCaliban      thrpt    5  15850.452 ± 554.077  ops/s
[info] GraphQLBenchmarks.simpleSangria      thrpt    5   7710.546 ± 140.220  ops/s
```
After:
```
[info] Benchmark                             Mode  Cnt      Score     Error  Units
[info] GraphQLBenchmarks.introspectCaliban  thrpt    5   2249.019 ±  30.014  ops/s
[info] GraphQLBenchmarks.introspectSangria  thrpt    5    495.380 ±  14.566  ops/s
[info] GraphQLBenchmarks.simpleCaliban      thrpt    5  20622.959 ± 774.391  ops/s
[info] GraphQLBenchmarks.simpleSangria      thrpt    5   7939.774 ±  43.622  ops/s
```

### DIFF
--- a/core/src/main/scala/caliban/execution/Executor.scala
+++ b/core/src/main/scala/caliban/execution/Executor.scala
@@ -182,7 +182,8 @@ object Executor {
     // ugly mutable code but it's worth it for the speed ;)
     val array = ArrayBuffer.empty[Field]
     val map   = collection.mutable.Map.empty[String, Int]
-    var index = 0
+
+    val reusedFieldBuilder: ArrayBuffer[Field] = ArrayBuffer.empty[Field]
 
     field.fields.foreach { field =>
       if (field.condition.forall(_ == typeName)) {
@@ -191,22 +192,18 @@ object Executor {
           case None =>
             // first time we see this field, add it to the array
             array += field
-            index += 1
           case Some(index) =>
             // field already existed, merge it
             val f = array(index)
-            array(index) = f.copy(fields = f.fields ++ field.fields)
+            reusedFieldBuilder ++= f.fields
+            reusedFieldBuilder ++= field.fields
+            array(index) = f.copy(fields = reusedFieldBuilder.toList)
+            reusedFieldBuilder.clear()
         }
       }
     }
 
-    val result = List.newBuilder[Field]
-    var i      = 0
-    while (i < index) {
-      result += array(i)
-      i += 1
-    }
-    result.result()
+    array.toList
   }
 
   private def fieldInfo(field: Field, path: List[Either[String, Int]], fieldDirectives: List[Directive]): FieldInfo =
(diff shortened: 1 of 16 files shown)


core/src/main/scala/caliban/execution/Executor.scala
```diff
@@ -178,20 +179,34 @@ object Executor {
...
-      .toList
+    }
+
+    val result = List.newBuilder[Field]
+    var i      = 0
+    while (i < index) {
+      result += array(i)
+      i += 1
+    }
+    result.result()
```
yoohaemin, line 209: I propose some improvements over the current implementation:
```scala
  private[caliban] def mergeFields(field: Field, typeName: String): List[Field] = {
    // ugly mutable code but it's worth it for the speed ;)
    val array = ArrayBuffer.empty[Field]
    val map   = collection.mutable.Map.empty[String, Int]

    val reusedFieldBuilder: ArrayBuffer[Field] = ArrayBuffer.empty[Field]

    field.fields.foreach { field =>
      if (field.condition.forall(_ == typeName)) {
        val name = field.alias.getOrElse(field.name)
        map.get(name) match {
          case None =>
            // first time we see this field, add it to the array
            array += field
          case Some(index) =>
            // field already existed, merge it
            val f = array(index)
            reusedFieldBuilder ++= f.fields
            reusedFieldBuilder ++= field.fields
            array(index) = f.copy(fields = reusedFieldBuilder.toList)
            reusedFieldBuilder.clear()
        }
      }
    }

    array.toList
  }
```

Comments:
- `ArrayBuffer` does not require tracking its size manually (unlike preallocated arrays)
- `array(index) = f.copy(fields = f.fields ++ field.fields)` could be improved.
- Having a `ListBuilder` at the end is unnecessary, conversion to a `List` is already optimized (inside `IndexedSeqOptimized`)

ghostdogpr: Thanks! Applied your suggestion.

yoohaemin: Hmm... come to think of it, maybe just ::: could be better than using a reusedFieldBuilder and make the code simpler. WDYT?

ghostdogpr: Yeah maybe. It didn’t make a difference in benchmarks anyway.

ghostdogpr: Simplified. This branch isn't even in the hot path.

PR #494, opened 2020-07-09
akka-http: access http request inside effect world

darl:
My main use-case is parsing opentracing's span from request and bypass is up to datasource (via FiberRef).

Another use-case is user authentication.

### DIFF
--- a/adapters/akka-http/src/main/scala/caliban/AkkaHttpAdapter.scala
+++ b/adapters/akka-http/src/main/scala/caliban/AkkaHttpAdapter.scala
@@ -9,6 +9,7 @@ import akka.http.scaladsl.server.{ RequestContext, Route }
 import akka.http.scaladsl.unmarshalling.FromEntityUnmarshaller
 import akka.stream.scaladsl.{ Flow, Sink, Source, SourceQueueWithComplete }
 import akka.stream.{ Materializer, OverflowStrategy, QueueOfferResult }
+import caliban.AkkaHttpAdapter.Around
 import caliban.ResponseValue.{ ObjectValue, StreamValue }
 import caliban.Value.NullValue
 import zio._
@@ -62,8 +63,7 @@ trait AkkaHttpAdapter {
     interpreter: GraphQLInterpreter[R, E],
     skipValidation: Boolean = false,
     enableIntrospection: Boolean = true,
-    around: RequestContext => URIO[R, HttpResponse] => URIO[R, HttpResponse] = (_: RequestContext) =>
-      (e: URIO[R, HttpResponse]) => e
+    around: Around = Around.empty
   )(request: GraphQLRequest)(implicit ec: ExecutionContext, runtime: Runtime[R]): Route =
     extractRequestContext { ctx =>
       complete(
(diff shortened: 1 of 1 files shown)


darl:
> How do you think about contextWrapper: ContextWrapper?

I think it's a better name. And it also aligned with Wrappers from core.

darl:
Thanks!
Will there be a release soon?

ghostdogpr:
Yeah I’ll do one this weekend, there are quite a few things ready including performance improvements.

ghostdogpr:
@darl if you see this, we are having a discussion on Discord about this. It seems you can't use `R` inside a wrapper since the `apply` function must work for any `R`. Looks like `ContextWrapper` should be parameterized in `R`. Or did we miss something? How were you able to inject the context in ZIO env?

darl:
Actually i'm using FiberRef for context propagation. I find this way more useful to bypass context trough layers that don't aware about tracing. 
About `R` in  `ContextWrapper`: my main issue was inference in `makeHttpService` method. It should work if `R` will be marked as contravariant parameter.

ghostdogpr:
Yeah the idea is to put the `FiberRef` inside the `R`. @moonkev submitted a PR that lets you do that: https://github.com/ghostdogpr/caliban/pull/519 can you confirm it's okay for you as well?

darl:
If you put `FiberRef` inside `R` you don't actually need `FiberRef`.
`FiberRef` is actually like `ThreadLocal` for zio. So you only need one instance of it.
So my code looks like it:
```scala
def opentracing(tracing: Tracing.Service): ContextWrapper =
    new ContextWrapper {
      override def apply[R](ctx: RequestContext)(e: URIO[R, HttpResponse]): URIO[R, HttpResponse] = {
        ctx match {
          case context: my.company.ApiRequestContext => tracing.withSpan(context.span)(e)
          case _ => tracing.withSpan(operationName = "graphql")(e)
        }
      }
    }
```
Where `Tracing.Service` is similar to [Opentracing](https://github.com/zio/zio-telemetry/blob/master/opentracing/src/main/scala/zio/telemetry/opentracing/OpenTracing.scala#L15) one.

I'm totally ok with changes in #519

ghostdogpr:
In the use case we were discussing with @moonkev, we extract authentication information from the context and want to access it from the business logic behind Caliban resolvers, so the environment is the only way to pass such information. But a `Ref` would be share between all users which is why we need `FiberRef`.

darl:
Didn't `.provide` do the trick?
```scala
val wrapper = new ContextWrapper[Has[Auth]] {
    override def apply[R1 <: Has[Auth]](
      ctx: RequestContext
    )(effect: URIO[R1, HttpResponse]): URIO[R1, HttpResponse] =
      for {
        response <- effect.provide(Has(Auth("foo")))
      } yield response
  }
```

darl:
Or you can also pass FiberRef in method argument:
```scala
  def wrapper(authRef: FiberRef[Auth]) = new ContextWrapper[Any] {
    override def apply[R1](
      ctx: RequestContext
    )(effect: URIO[R1, HttpResponse]): URIO[R1, HttpResponse] =
      for {
        _        <- authRef.set(Auth("foo"))
        response <- effect
      } yield response
  }
```

And do something like this on construction phase:
```scala
for {
  authRef <- FiberRef.make(Auth("no auth"))
  interpreter <- ???
} yield {
  adapter.makeHttpService(interpreter, contextWrapper = wrapper(authRef))
}
```

ghostdogpr:
Yeah that works too.

Issue #451, opened 2020-06-17
Support multipart request


ghostdogpr:
See spec: https://github.com/jaydenseric/graphql-multipart-request-spec

paulpdaniels:
We have one PR to support this in play #500

We also need adapters for
- [ ] Http4s
- [ ] Akka

ghostdogpr:
Note that [batching](https://github.com/jaydenseric/graphql-multipart-request-spec#operations-2) is currently not supported in the Play Adapter, this is another TODO.

PR #500, opened 2020-07-13
add multipart request for play

igorfialko:
https://github.com/ghostdogpr/caliban/issues/451


adapters/play/src/main/scala/caliban/Uploads.scala
```diff
@@ -0,0 +1,137 @@
...
+        }
+
+//        files.foldLeft(vars) {
+//          case (acc, (key, StringValue(value))) => acc.get(key) match {
+//            case Some(ListValue(mv)) => acc.updated(key, mv.patch(value, List(mv), 1))
+//          }
+//          case (acc, _) => acc
+//
+//        }
+//        vars ++ files
```
paulpdaniels, line 107: Still need?

adapters/play/src/test/scala/caliban/AdapterSpec.scala
```diff
@@ -0,0 +1,221 @@
...
+
+  override def spec: ZSpec[TestEnvironment, Any] =
+    suite("Requests")(
+      testM("multipart request with one file") {
+        val fileHash         = "64498927ff9cd735daefebe7175ed1567650399e58648a6b8340f636243962c0"
+        val fileName: String = s"$fileHash.png"
+        val fileURL: URL     = getClass.getResource(s"/$fileName")
+
+        val query: String =
+          """{ "query": "mutation ($file: Upload!) { uploadFile(file: $file) { hash, path, filename, mimetype } }",   "variables": {  "file": "File" }}"""
```
paulpdaniels, line 142: Nit: The spec indicates that the usages of the value should be replaced with `null`, thus it should be `"file": null` I believe

adapters/play/src/test/scala/caliban/AdapterSpec.scala
```diff
@@ -0,0 +1,221 @@
...
+      DefaultControllerComponents(
+        components.defaultActionBuilder,
+        components.playBodyParsers,
+        components.messagesApi,
+        components.langs,
+        components.fileMimeTypes,
+        components.executionContext
+      )
+    )(runtime, components.materializer).routes
+  }
```
paulpdaniels, line 132: You can package this up into a `Layer` then provide it to the `Spec` via a `provideSomeLayerShared`

build.sbt
```diff
@@ -244,11 +244,20 @@ lazy val play = project
...
-      "com.typesafe.play" %% "play" % playVersion
+      "dev.zio"           %% "zio"                   % zioVersion,
+      "com.typesafe.play" %% "play"                  % playVersion,
+      "dev.zio"           %%% "zio-test"             % zioVersion % "test",
+      "com.typesafe.play" %% "play-akka-http-server" % playVersion % "test",
+      "org.http4s"        %% "http4s-client"         % "0.21.6" % "test",
+      "io.circe"          %% "circe-core"            % circeVersion % "test",
+      "io.circe"          %% "circe-generic"         % circeVersion % "test",
+      "org.http4s"        %% "http4s-blaze-client"   % "0.21.6" % "test",
+      "org.http4s"        %% "http4s-circe"          % "0.21.6" % "test"
```
paulpdaniels, line 257: `Sttp` is used elsewhere, and it has a backend that supports `ZIO` natively, so you wouldn't need the extra interop.

adapters/play/src/test/scala/caliban/AdapterSpec.scala
paulpdaniels: Expanding on this you can do something like this instead:

```scala
val apiLayer = ZLayer.fromEffect(for {
  interpreter <- TestAPI.api.interpreter
} yield AkkaHttpServer.fromRouterWithComponents(
    ServerConfig(
      mode = Mode.Dev,
      port = Some(8088),
      address = "127.0.0.1"
    )
  ) { components =>
    PlayRouter(
      interpreter,
      DefaultControllerComponents(
        components.defaultActionBuilder,
        components.playBodyParsers,
        components.messagesApi,
        components.langs,
        components.fileMimeTypes,
        components.executionContext
      )
    )(runtime, components.materializer).routes
  }


val specLayer = AsyncHttpZioBackend.layer() ++ ((Uploads.Service.empty ++  ZEnv.any) >>> apiLayer)


suite("Requests")(...).provideCustomLayerShared[ZEnv](specLayer)
```

can use `fromManaged` as well, if you want to provide a shutdown hook to terminate the `AkkaServer`

paulpdaniels:
@igorfialko are you planning to continue work on this?

igorfialko:
@paulpdaniels Looks like the only thing left for play is the usage of other testing library. Is that correct?

paulpdaniels:
That was all I had yes @igorfialko

igorfialko:
@paulpdaniels Ok, I'll try to squeeze it this week. Maybe further splitting https://github.com/ghostdogpr/caliban/issues/451 by framework would make sense?

adapters/play/src/main/scala/caliban/Uploads.scala
```diff
@@ -0,0 +1,128 @@
...
+  val allBytes: RIO[Uploads with Blocking, Chunk[Byte]] =
+    Uploads.stream(name).run(ZSink.foldLeftChunks(Chunk[Byte]())(_ ++ (_: Chunk[Byte])))
+
+  val meta: URIO[Uploads, Option[FileMeta]] =
+    Uploads.fileMeta(name)
+
+}
+
+object Uploads {
+  type Uploads = Has[Service.Service]
```
paulpdaniels, line 25: I think the typical pattern is for this to be a package object with `Uploads` under it, so then it would be used like 

```scala
import caliban.uploads._

Uploads.empty
```

or 

```scala
uploads.stream("filename")
```

This seems slightly odd that it is `Service.Service`

adapters/play/src/test/scala/caliban/AdapterSpec.scala
```diff
@@ -0,0 +1,214 @@
...
+  val runtime: Runtime[Console with Clock with Blocking with Random with Uploads] =
+    Runtime.unsafeFromLayer(
+      Console.live ++ Clock.live ++ Blocking.live ++ Random.live ++ Uploads.Service.empty,
+      Platform.default
+    )
+
+  val apiLayer = ZLayer.fromEffect(
+    for {
+      interpreter <- TestAPI.api.interpreter
+    } yield AkkaHttpServer.fromRouterWithComponents(
```
paulpdaniels, line 104: Does the `AkkaHttpServer` need to be shutdown as well?

adapters/play/src/test/scala/caliban/AdapterSpec.scala
```diff
@@ -0,0 +1,214 @@
...
+          components.messagesApi,
+          components.langs,
+          components.fileMimeTypes,
+          components.executionContext
+        )
+      )(runtime, components.materializer).routes
+    }
+  )
+
+  val specLayer: ZLayer[Any with zio.ZEnv, CalibanError.ValidationError, Has[Server]] =
```
paulpdaniels, line 125: You can safely remove the `Any` from the type signature

adapters/play/src/test/scala/caliban/AdapterSpec.scala
```diff
@@ -0,0 +1,214 @@
...
+          hasField("hash", (fl: List[TestAPI.File]) => fl(0).hash, equalTo(file1Hash)) &&
+            hasField("hash", (fl: List[TestAPI.File]) => fl(1).hash, equalTo(file2Hash)) &&
+            hasField("filename", (fl: List[TestAPI.File]) => fl(0).filename, equalTo(file1Name)) &&
+            hasField("filename", (fl: List[TestAPI.File]) => fl(1).filename, equalTo(file2Name)) &&
+            hasField("mimetype", (fl: List[TestAPI.File]) => fl(0).mimetype, equalTo("image/png")) &&
+            hasField("mimetype", (fl: List[TestAPI.File]) => fl(1).mimetype, equalTo("text/plain"))
+        )
+      }
+    ).provideCustomLayer(AsyncHttpClientZioBackend.layer())
+      .provideCustomLayerShared(specLayer)
```
paulpdaniels, line 211: How does this work, wouldn't the first `provideCustomLayer` erase the requirements for everything _except_ `ZEnv` but then you provide the additional `specLayer` is it because the tests don't actually require it? We are just running for the effect?

build.sbt
```diff
@@ -243,11 +243,20 @@ lazy val play = project
...
   .settings(commonSettings)
   .settings(
+    testFrameworks := Seq(new TestFramework("zio.test.sbt.ZTestFramework")),
     libraryDependencies ++= Seq(
-      "com.typesafe.play" %% "play" % playVersion
+      "dev.zio"                      %% "zio"                           % zioVersion,
+      "com.typesafe.play"            %% "play"                          % playVersion,
+      "dev.zio"                      %%% "zio-test"                     % zioVersion % "test",
+      "com.typesafe.play"            %% "play-akka-http-server"         % playVersion % "test",
+      "org.http4s"                   %% "http4s-client"                 % "0.21.6" % "test",
```
paulpdaniels, line 252: Can remove now right?

build.sbt
```diff
@@ -243,11 +243,20 @@ lazy val play = project
...
+      "io.circe"                     %% "circe-core"                    % circeVersion % "test",
+      "io.circe"                     %% "circe-generic"                 % circeVersion % "test",
+      "com.softwaremill.sttp.client" %% "async-http-client-backend-zio" % sttpVersion % "test",
+      "com.softwaremill.sttp.client" %%% "circe"                        % sttpVersion % "test"
```
paulpdaniels, line 256: Is the `%%%` necessary? I don't think we build this for scala.js

build.sbt
```diff
@@ -243,11 +243,20 @@ lazy val play = project
...
     )
   )
-  .dependsOn(core)
+  .dependsOn(core, catsInterop)
```
paulpdaniels, line 259: What do we need `catsInterop` for?

adapters/play/src/main/scala/caliban/Uploads.scala
```diff
@@ -0,0 +1,128 @@
...
+import caliban.InputValue.ListValue
+import caliban.Uploads.{ FileMeta, Uploads }
+import caliban.Value.{ NullValue, StringValue }
+import caliban.schema.Annotations.GQLName
+import zio.blocking.Blocking
+import zio.stream.{ Stream, ZSink, ZStream }
+import zio.{ Chunk, Has, Layer, RIO, UIO, URIO, ZIO, ZLayer }
+
+@GQLName("Upload")
+sealed case class Upload(name: String) {
```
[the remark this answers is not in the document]
igorfialko: sealed forbids inheritance outside of this file, final forbids overriding in case of inheritance. do we want users at all to extend this case class?

adapters/play/src/test/scala/caliban/AdapterSpec.scala
igorfialko: is this not how actually the layer is "executed"?

igorfialko: I refactored slightly, but I still don't understand what you mean by erase

paulpdaniels: I meant because the signature for `provideCustomLayer` is:

```scala
  final def provideCustomLayer[E1 >: E, R1 <: Has[_]](
    layer: ZLayer[ZEnv, E1, R1]
  )(implicit ev: ZEnv with R1 <:< R, tagged: Tag[R1]): ZIO[ZEnv, E1, A]
```

So the resulting effect doesn't _require_ the stuff from `specLayer` (at least according to the signature), so I was just curious if it was because we are just running the `Layer` for an effect rather than to get the resource it creates.

adapters/play/src/main/scala/caliban/Uploads.scala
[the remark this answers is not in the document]
paulpdaniels: https://stackoverflow.com/a/31133245/2521865

IIRC `sealed` is mainly for `trait` hierarchies because they can be used by the compiler to generate exhaustivity checks and `final` can have some useful properties during compilation (though I don't know if that would apply to such a simple case class).

adapters/play/src/main/scala/caliban/PlayAdapter.scala
```diff
@@ -27,8 +33,140 @@ trait PlayAdapter[R] {
...
+      }
+    }
+
+  private def makeParser(
+    runtime: Runtime[Blocking with Random]
+  ): BodyParser[Either[GraphQLUploadRequest, GraphQLRequest]] =
+    parse.using { req =>
+      implicit val ec: ExecutionContext = runtime.platform.executor.asEC
+      req.contentType.map(_.toLowerCase(Locale.ENGLISH)) match {
+        case Some("text/json") | Some("application/json") =>
```
paulpdaniels, line 68: Nit: There should be some constants for these from the play framework

adapters/play/src/main/scala/caliban/Uploads.scala
```diff
@@ -0,0 +1,124 @@
...
+//            values.lift(idx).fold[InputValue](NullValue)(loop(_, path.drop(1), name))
+//          case _ => NullValue
+//        }
+//      case None =>
+//        // If we are out of values then we are at the end of the path, so we need to replace this current node
+//        // with a string node containing the file name
+//        StringValue(name)
+//    }
+//
+//}
```
paulpdaniels, line 124: Delete?

paulpdaniels (approved):
A few small nits from me otherwise LGTM

adapters/play/src/test/scala/caliban/AdapterSpec.scala
igorfialko: I guess you are right then. In this case we run for the effect.

adapters/play/src/main/scala/caliban/Uploads.scala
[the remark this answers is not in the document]
igorfialko: Let it be final then, I don't have all the information either to make a perfect decision.

build.sbt
```diff
@@ -243,8 +243,16 @@ lazy val play = project
   .settings(name := "caliban-play")
   .settings(commonSettings)
   .settings(
+    testFrameworks := Seq(new TestFramework("zio.test.sbt.ZTestFramework")),
     libraryDependencies ++= Seq(
-      "com.typesafe.play" %% "play" % playVersion
+      "dev.zio"                      %% "zio"                           % zioVersion,
```
ghostdogpr, line 248: Some of these dependencies are obtained transitively from core, we don't need to specify them again, do we?

ghostdogpr (commented):
LGTM, just a tiny remark on dependencies that can be simplified. Also I wonder if there's an easy way to make the dependency to Blocking and Random optional without duplicating code? If not this is fine.

paulpdaniels:
I do recall there is a `+++` operator that lets you select between a provided left or right environment, but I don't know if that would apply here because we actually require the `Runtime` upfront so we wouldn't be able to switch it out.

ghostdogpr (approved):
Thanks for your work on this!

PR #506, opened 2020-07-16
Add location and path info to error message

paulpdaniels:
This adds the location and path info to an `ExecutionError` that is thrown during field resolution. I'm not sure if we want to have some sort of "opt-out" mechanism in case users don't want this behavior, but in general it is the minimum we should have when showing errors, since it allows the client to piece it back together with the original query.


ghostdogpr:
There's one test that needs to be updated, otherwise very nice 👍

Issue #421, opened 2020-05-27
Support "query" query string parameter in POST endpoint 


ghostdogpr:
From https://graphql.org/learn/serving-over-http/:

> If the "query" query string parameter is present (as in the GET example above), it should be parsed and handled in the same way as the HTTP GET case.

To be done in each adapter:
- [x] http4s
- [x] akka http
- [x] finch
- [ ] play

ghostdogpr:
If anyone wants to work on this, you can totally do one adapter only, no need to do all at once😄

jessenr:
I'll take the play adaptor.

Issue #420, opened 2020-05-27
Add support for application/graphql content type


ghostdogpr:
From https://graphql.org/learn/serving-over-http/:

> If the "application/graphql" Content-Type header is present, treat the HTTP POST body contents as the GraphQL query string.

To be done in each adapter:
- [x] http4s
- [x] akka http
- [x] finch
- [ ] play

ghostdogpr:
If anyone wants to work on this, you can totally do one adapter only, no need to do all at once😄

dgeyer-mdsol:
Hi 👋  I would like to pick up this ticket for the hackathon, I see play is the only one pending, am I right?

ghostdogpr:
@dgeyer-mdsol This is right, we only need to add it in the Play Adapter. You can have a look at the other PRs linked here to see how it was added to the other adapters. Let me know if you have any question 😄

PR #514, opened 2020-07-20
add additional cases for POST handling in Akka HTTP

moonkev:
Hello

This pull request is to handle additional POST request cases when using the Akka HTTP adapter.  The additional functionalities relate to tickets #420 and #421 

Please note that this does have 1 small implication on the current API.  The `AkkaHttpAdapter.makeHttpService` call now requires an additional implicit `akka.stream.Materializer` so that it may materialize the stream which reads the POST body contents.  Currently, `AkkaHttpAdapter.makeWebSocketService` also requires an implicit `Materializer`, so this shouldn't be an issue for those using subscriptions.  However those only performing queries and mutations, may have to make a change.  However, even this is unlikely as probably most will have the `ActorSystem` in implicit scope when calling `makeHttpService` which will satisfy the `Materializer` parameter.

We could always just extract the materializer from the request context just as `MarshallingDirectives.entity` does - i.e.

```scala
            case _ if request.entity.contentType.mediaType == `application/graphql` =>
              extractRequestContext { ctx =>
                onComplete(
                  request.entity.dataBytes
                    .runFold(ByteString.empty)(_ ++ _)(ctx.materializer)
                    .map(_.utf8String)
                ) {
```

I would say however, this seems slightly more obtrusive, but thats more a matter of opinion.

### DIFF
--- a/adapters/akka-http/src/main/scala/caliban/AkkaHttpAdapter.scala
+++ b/adapters/akka-http/src/main/scala/caliban/AkkaHttpAdapter.scala
@@ -1,20 +1,21 @@
 package caliban
 
-import scala.concurrent.ExecutionContext
 import akka.http.scaladsl.model.MediaTypes.`application/json`
 import akka.http.scaladsl.model.ws.{ Message, TextMessage }
-import akka.http.scaladsl.model.{ HttpEntity, HttpResponse, StatusCodes }
+import akka.http.scaladsl.model._
 import akka.http.scaladsl.server.Directives.{ complete, extractRequestContext }
 import akka.http.scaladsl.server.{ RequestContext, Route }
 import akka.http.scaladsl.unmarshalling.FromEntityUnmarshaller
 import akka.stream.scaladsl.{ Flow, Sink, Source, SourceQueueWithComplete }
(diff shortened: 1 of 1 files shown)


adapters/akka-http/src/main/scala/caliban/AkkaHttpAdapter.scala
```diff
@@ -107,14 +110,47 @@ trait AkkaHttpAdapter {
...
+                  failWith,
+                  completeRequest(
+                    interpreter,
+                    skipValidation = skipValidation,
+                    enableIntrospection = enableIntrospection,
+                    contextWrapper = contextWrapper
+                  )
+                )
+            case _ if request.entity.contentType.mediaType == `application/graphql` =>
+              onComplete(
```
ghostdogpr, line 128: Could we maybe use `entity(as[String])` here?

moonkev: Haha, I had initially thought that using that would fail on an invalid json string since the post body would not be double quoted and have newlines escaped, and then in true programmer form, was too lazy to even give it a shot.  But of course, it does in fact work 😄 
I've updated the PR accordingly.

PR #519, opened 2020-07-25
Add support for context wrapper in subscriptions

moonkev:
This pull request is to lift the environment parameter to the trait level, and also parameterize the success type, so that it can be used for more than HttpResponse.  Most notably, this allows us to now wrap executeRequest call in subscriptions.

### DIFF
--- a/adapters/akka-http/src/main/scala/caliban/AkkaHttpAdapter.scala
+++ b/adapters/akka-http/src/main/scala/caliban/AkkaHttpAdapter.scala
@@ -64,7 +64,7 @@ trait AkkaHttpAdapter {
     interpreter: GraphQLInterpreter[R, E],
     skipValidation: Boolean = false,
     enableIntrospection: Boolean = true,
-    contextWrapper: ContextWrapper = ContextWrapper.empty
+    contextWrapper: ContextWrapper[R, HttpResponse] = ContextWrapper.empty
   )(request: GraphQLRequest)(implicit ec: ExecutionContext, runtime: Runtime[R]): Route =
     extractRequestContext { ctx =>
       complete(
@@ -87,7 +87,7 @@ trait AkkaHttpAdapter {
     interpreter: GraphQLInterpreter[R, E],
     skipValidation: Boolean = false,
     enableIntrospection: Boolean = true,
-    contextWrapper: ContextWrapper = ContextWrapper.empty
+    contextWrapper: ContextWrapper[R, HttpResponse] = ContextWrapper.empty
   )(implicit ec: ExecutionContext, runtime: Runtime[R]): Route = {
     import akka.http.scaladsl.server.Directives._
 
@@ -149,8 +149,10 @@ trait AkkaHttpAdapter {
     interpreter: GraphQLInterpreter[R, E],
     skipValidation: Boolean = false,
     enableIntrospection: Boolean = true,
-    keepAliveTime: Option[Duration] = None
+    keepAliveTime: Option[Duration] = None,
+    contextWrapper: ContextWrapper[R, GraphQLResponse[E]] = ContextWrapper.empty
   )(implicit ec: ExecutionContext, runtime: Runtime[R], materializer: Materializer): Route = {
+
     def sendMessage(
       sendQueue: SourceQueueWithComplete[Message],
       id: String,
@@ -165,13 +167,16 @@ trait AkkaHttpAdapter {
       messageId: String,
       request: GraphQLRequest,
       sendTo: SourceQueueWithComplete[Message],
-      subscriptions: Ref[Map[Option[String], Fiber[Throwable, Unit]]]
+      subscriptions: Ref[Map[Option[String], Fiber[Throwable, Unit]]],
+      ctx: RequestContext
     ): RIO[R, Unit] =
       for {
-        result <- interpreter.executeRequest(
-                   request,
-                   skipValidation = skipValidation,
-                   enableIntrospection = enableIntrospection
+        result <- contextWrapper(ctx)(
+                   interpreter.executeRequest(
+                     request,
+                     skipValidation = skipValidation,
+                     enableIntrospection = enableIntrospection
+                   )
                  )
         _ <- result.data match {
               case ObjectValue((fieldName, StreamValue(stream)) :: Nil) =>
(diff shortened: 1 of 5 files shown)


adapters/akka-http/src/main/scala/caliban/AkkaHttpAdapter.scala
```diff
@@ -149,96 +149,104 @@ trait AkkaHttpAdapter {
...
+        extractUpgradeToWebSocket { upgrade =>
+          def sendMessage(
+            sendQueue: SourceQueueWithComplete[Message],
+            id: String,
+            data: ResponseValue,
+            errors: List[E]
+          ): Task[QueueOfferResult] =
+            IO.fromFuture(_ => sendQueue.offer(TextMessage(json.encodeWSResponse(id, data, errors))))
+
+          def startSubscription(
```
ghostdogpr, line 169: How about just adding `ctx` as a parameter to this function and keep it outside as it was? Just to reduce allocations done at each request and keep it slightly more readable.

adapters/akka-http/src/main/scala/caliban/AkkaHttpAdapter.scala
```diff
@@ -251,18 +259,18 @@ object AkkaHttpAdapter {
   /**
    * ContextWrapper provides a way to pass context from http request into Caliban's query handling.
    */
-  trait ContextWrapper { self =>
-    def apply[R](ctx: RequestContext)(e: URIO[R, HttpResponse]): URIO[R, HttpResponse]
+  trait ContextWrapper[-R] { self =>
```
darl, line 262: :+1:

moonkev:
Good point about the subscription functions, I have moved them back out of the routing, and pass in the request context as a parameter as suggested.  There was one thing that came up though when working on the example.  To mimic the behavior of the http4s adapter, we need to be able to short circuit the request with a 403.  To do that we need type awareness on the success type (HttpResponse).  As such, I had to lift the success type parameter back to the trait level.  Let me know your thoughts.  Thank you also @darl for reviewing.

When we are happy with the Akka HTTP implementation, I most definitely can port this over to the play adapter.

examples/src/main/scala/caliban/akkahttp/AuthExampleApp.scala
```diff
@@ -0,0 +1,71 @@
...
+  type Auth = Has[FiberRef[Option[AuthToken]]]
+
+  object AuthWrapper extends ContextWrapper[Auth, HttpResponse] {
+    override def apply[R <: Auth](ctx: RequestContext)(effect: URIO[R, HttpResponse]): URIO[R, HttpResponse] =
+      ctx.request.headers.collectFirst {
+        case header if header.name.toLowerCase == "token" => header.value
+      } match {
+        case Some(token) =>
+          for {
+            _      <- ZIO.accessM[Auth](_.get.set(Some(AuthToken(token))))
```
ghostdogpr, line 32: Could be simplified to `ZIO.accessM[Auth](_.get.set(Some(AuthToken(token)))) *> effect`

examples/src/main/scala/caliban/akkahttp/AuthExampleApp.scala
```diff
@@ -0,0 +1,71 @@
...
+            result <- effect
+          } yield result
+        case _ =>
+          URIO.fromFunction(_ => HttpResponse(StatusCodes.Forbidden))
```
ghostdogpr, line 36: Why not just `URIO.succeed`?

adapters/akka-http/src/main/scala/caliban/AkkaHttpAdapter.scala
```diff
@@ -251,18 +258,18 @@ object AkkaHttpAdapter {
...
+    def |+|[R1 <: R](that: ContextWrapper[R1, A]): ContextWrapper[R1, A] = new ContextWrapper[R1, A] {
+      override def apply[R2 <: R1](ctx: RequestContext)(e: URIO[R2, A]): URIO[R2, A] =
         that.apply(ctx)(self.apply(ctx)(e))
     }
   }
 
   object ContextWrapper {
-    def empty: ContextWrapper = new ContextWrapper {
-      override def apply[R](ctx: RequestContext)(effect: URIO[R, HttpResponse]): URIO[R, HttpResponse] =
+    def empty[A]: ContextWrapper[Any, A] = new ContextWrapper[Any, A] {
```
ghostdogpr, line 271: If you make `ContextWrapper` covariant in `A`, you can remove the `[A]` type parameter and fix it to `Nothing`.

examples/src/main/scala/caliban/akkahttp/AuthExampleApp.scala
moonkev: We must retain the environment in the return type, and thus succeed will not satisfy.

adapters/akka-http/src/main/scala/caliban/AkkaHttpAdapter.scala
moonkev: Don't think we can do this since the appender function's `that` parameter is in a contravariant position

examples/src/main/scala/caliban/akkahttp/AuthExampleApp.scala
moonkev: Oh cool, still quite new to ZIO, so still learning the operators.  PR Updated.

moonkev:
Have added changes to play adapter as well and added example if you could please review.  If you could please review, and if there are still changes to make in akka http version, I'll make same changes there.

adapters/akka-http/src/main/scala/caliban/AkkaHttpAdapter.scala
ghostdogpr: You can, you just have to use the same trick we use for `R`:
```scala
  trait ContextWrapper[-R, +A] { self =>
    def apply[R1 <: R, A1 >: A](ctx: RequestContext)(e: URIO[R1, A1]): URIO[R1, A1]
  }
```

examples/src/main/scala/caliban/akkahttp/AuthExampleApp.scala
ghostdogpr: If you use `succeed`, it returns a `ZIO` with `R` = `Any`, which is valid because R is contravariant.

adapters/akka-http/src/main/scala/caliban/AkkaHttpAdapter.scala
moonkev: Many thanks for the advice.  If I change this I am struggling to find the right combination to be able to take in `e` effect parameter in the `apply` definition.  For example, something like below does not work

```scala
    def |+|[R1 <: R, A1 >: A](that: ContextWrapper[R1, A1]): ContextWrapper[R1, A1] = new ContextWrapper[R1, A1] {
      override def apply[R2 <: R1, A2 >: A1](ctx: RequestContext)(e: URIO[R2, A2]): URIO[R2, A2] =
        that.apply(ctx)(self.apply(ctx)(e))
    }
```

ghostdogpr: Does it help to specify the types explicitly? `apply[R2, A2]`

moonkev: Yes, thanks, that did the trick!  I have pushed the suggested changes.

moonkev:
Thanks a lot for the patience and help.  On that note, made one more change to the play side.  Since we always just return `Result`, no need for the success type parameter.  I'll squash down the commit if/after @paulpdaniels is able to review, just left a second one for now so that we're confident I didn't change anything else since you last reviewed.

ghostdogpr:
@moonkev thanks!

PR #529, opened 2020-07-30
send complete message upon completion of a stream

moonkev:
As per subscriptions-transport-ws spec, send GQL_COMPLETE message to client, so that client can initiate GQL_CONNECTION_TERMINATE.  In case of failure, close the send queue, so that server terminates websocket connection.

### DIFF
--- a/adapters/akka-http/src/main/scala/caliban/AkkaHttpAdapter.scala
+++ b/adapters/akka-http/src/main/scala/caliban/AkkaHttpAdapter.scala
@@ -11,6 +11,7 @@ import akka.stream.{ Materializer, OverflowStrategy, QueueOfferResult }
 import caliban.AkkaHttpAdapter.{ `application/graphql`, ContextWrapper }
 import caliban.ResponseValue.{ ObjectValue, StreamValue }
 import caliban.Value.NullValue
+import zio.Exit.Failure
 import zio._
 import zio.clock.Clock
 import zio.duration._
@@ -182,6 +183,15 @@ trait AkkaHttpAdapter {
               case ObjectValue((fieldName, StreamValue(stream)) :: Nil) =>
                 stream
                   .foreach(item => sendMessage(sendTo, messageId, ObjectValue(List(fieldName -> item)), result.errors))
+                  .onExit {
+                    case Failure(cause) if !cause.interrupted =>
+                      IO.fromFuture(_ =>
+                          sendTo.offer(TextMessage(json.encodeWSError(messageId, cause.squash.toString)))
+                        )
+                        .orDie
+                    case _ =>
+                      IO.fromFuture(_ => sendTo.offer(TextMessage(s"""{"type":"complete","id":"$messageId"}"""))).orDie
+                  }
                   .forkDaemon
                   .flatMap(fiber => subscriptions.update(_.updated(Option(messageId), fiber)))
               case other =>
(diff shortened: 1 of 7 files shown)


paulpdaniels:
Does play need this as well?

ghostdogpr:
Ah sorry, I told @moonkev that only akka/http4s/uzhttp had WebSocket support (forgot the Play part was done)! It should be easy since it's based on Akka Http.

adapters/http4s/src/main/scala/caliban/Http4sAdapter.scala
```diff
@@ -198,7 +198,11 @@ object Http4sAdapter {
                                 case ObjectValue((fieldName, StreamValue(stream)) :: Nil) =>
                                   stream.foreach { item =>
                                     sendMessage(sendQueue, id, ObjectValue(List(fieldName -> item)), result.errors)
-                                  }.fork.flatMap(fiber => subscriptions.update(_.updated(id, fiber)))
+                                  }.onExit { _ =>
+                                    println("HERE")
```
ghostdogpr, line 202: cleanup

adapters/akka-http/src/main/scala/caliban/AkkaHttpAdapter.scala
```diff
@@ -182,6 +183,13 @@ trait AkkaHttpAdapter {
               case ObjectValue((fieldName, StreamValue(stream)) :: Nil) =>
                 stream
                   .foreach(item => sendMessage(sendTo, messageId, ObjectValue(List(fieldName -> item)), result.errors))
+                  .onExit {
+                    case Success(_) =>
+                      IO.fromFuture(_ => sendTo.offer(TextMessage(s"""{"type":"complete","id":"$messageId"}"""))).orDie
+                    case Failure(_) =>
+                      IO.fromFuture(_ => sendTo.offer(TextMessage(json.encodeWSError(messageId, "Upstream failed"))))
```
ghostdogpr, line 190: Should maybe the error message contain what was inside `Failure(_)`?

adapters/http4s/src/main/scala/caliban/Http4sAdapter.scala
```diff
@@ -141,19 +142,19 @@ object Http4sAdapter {
...
     import dsl._
 
-    def sendMessage(
+    def sendMessage[T <: Json](
       sendQueue: fs2.concurrent.Queue[Task, WebSocketFrame],
+      messageType: String,
       id: String,
-      data: ResponseValue,
-      errors: List[E]
+      payload: T
```
ghostdogpr, line 149: Could `T` just be replaced by `Json`?

adapters/akka-http/src/main/scala/caliban/AkkaHttpAdapter.scala
ghostdogpr, line 190: In the case of a clean `stop` where we cancel the stream execution, I think there will be a spurious error message, no?

adapters/http4s/src/main/scala/caliban/Http4sAdapter.scala
moonkev: Yes, you are right, will change.

adapters/uzhttp/src/main/scala/caliban/UzHttpAdapter.scala
```diff
@@ -157,11 +165,11 @@ object UzHttpAdapter {
...
   }
 
-  private def sendMessage[E](
+  private def sendMessage[T <: Json](
     sendQueue: Queue[Take[Nothing, Frame]],
+    messageType: String,
     id: String,
-    data: ResponseValue,
-    errors: List[E]
+    payload: T
```
ghostdogpr, line 172: You can use `Json` here too.

moonkev: sorry, missed fixing this one, updated.

PR #530, opened 2020-07-31
Add client persisted queries

paulpdaniels:
This PR adds [persisted query] support to the client. I added it as a separate module because it requires some additional dependencies that we don't want to pollute the core module and because it is still a bit experimental. I'll admit that I'm not terribly familiar with scala.js so hopefully someone with more experience can tell me where i've messed up. 

Also this is pretty much the bare minimum to get it working, so I'm marking this as WIP in case anyone has any ideas around performance or ergonomics.


ghostdogpr:
Hmm... honestly I don't know if this should be a part of this repository. Wondering if anyone would actually use that. On Apollo side, the [client plugin](https://github.com/apollographql/apollo-link-persisted-queries) has 272 stars (vs 14.4K for apollo-client) and wasn't updated in 2 years so I'm not even sure this is super popular in the JS world? Did you plan on using it yourself?
My concern is that it makes the project organization more complex, adds dependencies and code we are not so familiar with. Maybe it should just be a repo on its own?

paulpdaniels:
Will move to a separate repo

PR #542, opened 2020-08-08
Add SBT welcome message

yoohaemin:
Credit to @reibitto , copied over mostly from https://github.com/reibitto/command-center/blob/master/project/ConsoleHelper.scala


reibitto:
👍 

Also just to note, I'm creating an sbt-plugin for this with a few extra features. But it might take me a week or two to actually release it.

PR #544, opened 2020-08-08
Use sbt-ci-release to publish snapshots

yoohaemin:
This will start publishing snapshots for all commits on master


ghostdogpr:
Let's pray that it works 🙏

PR #561, opened 2020-08-18
Make basic schema definitions do not depend on GenericSchema's type param

darl:
With this change `Schema[R, List[T]]` can be inferred for any `R` without need to extend from `GenericSchema`.


ghostdogpr:
@darl it looks good but I wonder, since you need to extend `GenericSchema` for case class derivation anyway, is there some case where this changes anything? Do you have an example?

darl:
Sometimes i get strange compilation issues:
Once i got conflict between `GenericSchema.gen` and implicit schema in companion object.
Other time `Option[MyClass]` resolved via magnolia `gen` instead of `GenericSchema.optionSchema` and `union` was generated instead of optional value.
It's very unstable and I can not make repro.

So I started experimenting with avoiding wildcard imports from `object schema extends GenericSchema[Env]`.
(I use `R` in `Schema` a lot)

ghostdogpr:
I see. Well, if it reduces weird compilation errors, it sounds good!

ghostdogpr (approved):

PR #569, opened 2020-08-27
Allow passing errors to wrappers

paulpdaniels:
This is my attempt at allowing errors to be passed through to the `FieldWrapper` this should allow wrapper authors to deal with errors arising from execution.

### DIFF
--- a/core/src/main/scala/caliban/execution/Executor.scala
+++ b/core/src/main/scala/caliban/execution/Executor.scala
@@ -101,21 +101,20 @@ object Executor {
 
     def makeQuery(step: ReducedStep[R], errors: Ref[List[CalibanError]]): ZQuery[R, Nothing, ResponseValue] = {
 
-      def eraseError(error: CalibanError): UQuery[ResponseValue] =
+      def handleError(error: CalibanError): UQuery[ResponseValue] =
         ZQuery.fromEffect(errors.update(error :: _)).as(NullValue)
 
       @tailrec
-      def wrap(query: ZQuery[R, ExecutionError, ResponseValue])(
+      def wrap(query: ZQuery[R, CalibanError, ResponseValue])(
         wrappers: List[FieldWrapper[R]],
         fieldInfo: FieldInfo
-      ): ZQuery[R, ExecutionError, ResponseValue] =
+      ): ZQuery[R, CalibanError, ResponseValue] =
         wrappers match {
           case Nil => query
           case wrapper :: tail =>
             wrap(
               wrapper
                 .f(query, fieldInfo)
-                .foldM(eraseError, ZQuery.succeed(_))
             )(tail, fieldInfo)
 
         }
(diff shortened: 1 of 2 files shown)


core/src/main/scala/caliban/wrappers/Wrapper.scala
```diff
@@ -67,7 +67,7 @@ object Wrapper {
    * If false, simple pure values will be ignored.
    */
   case class FieldWrapper[R](
-    f: (ZQuery[R, Nothing, ResponseValue], FieldInfo) => ZQuery[R, CalibanError, ResponseValue],
+    f: (ZQuery[R, CalibanError, ResponseValue], FieldInfo) => ZQuery[R, CalibanError, ResponseValue],
```
ghostdogpr, line 70: We could restrict to `ExecutionError` I think?

core/src/main/scala/caliban/execution/Executor.scala
```diff
@@ -101,42 +101,63 @@ object Executor {
 
     def makeQuery(step: ReducedStep[R], errors: Ref[List[CalibanError]]): ZQuery[R, Nothing, ResponseValue] = {
 
+      def eraseError(error: CalibanError): UQuery[ResponseValue] =
```
ghostdogpr, line 104: Maybe name is `handleError` or something like that since "erase" sounds like we just throw it away.

core/src/main/scala/caliban/execution/Executor.scala
```diff
@@ -101,42 +101,63 @@ object Executor {
...
-      ): ZQuery[R, Nothing, ResponseValue] =
+      ): ZQuery[R, ExecutionError, ResponseValue] =
         wrappers match {
           case Nil => query
           case wrapper :: tail =>
             wrap(
               wrapper
                 .f(query, fieldInfo)
-                .foldM(error => ZQuery.fromEffect(errors.update(error :: _)).as(NullValue), ZQuery.succeed(_))
+                .foldM(eraseError, ZQuery.succeed(_))
```
ghostdogpr, line 118: Why do we need to remove the error here? It should be passed to all the wrappers I think?

ghostdogpr (commented):
Glad it was not too complicated! Can you check if you see any impact on performance?

core/src/main/scala/caliban/execution/Executor.scala
paulpdaniels: I tried but wrapper returns a `CalibanError`, should I make that return `ExecutionError` too?

paulpdaniels: That's a fair point we can all the downstream wrappers to clean up errors from the upstream.

core/src/main/scala/caliban/execution/Executor.scala
paulpdaniels: That's a fair point we can all the downstream wrappers to clean up errors from the upstream.

core/src/main/scala/caliban/wrappers/Wrapper.scala
paulpdaniels: I tried but wrapper returns a `CalibanError`, should I make that return `ExecutionError` too?

paulpdaniels:
Performance seems stable as well after this change

core/src/main/scala/caliban/wrappers/Wrapper.scala
ghostdogpr: Ah no you're right.

ghostdogpr (approved):
Looks good!

PR #578, opened 2020-09-04
Codegen args root

### LINKED ISSUE #560: [CodeGen] Arguments name do not match for root level operation
This is related to issue #552 . The previous PR solved most of the mismatches, but in the current snapshot (`0.9.1+20-40e9f3a2-SNAPSHOT`) arguments in root level operations still do not match.

Example
```
object Operations {
  @GQLDescription("query root")
  case class query_root(
      posts: PostsArgs => zio.UIO[List[Post]]
  )
}

case class Query_rootPostsArgs(
    ...
)
```

I can work on this if you want

galales:
This PR closes #560

Fix misalignment in arguments type names also at schema root levels (`Query`, `Mutation` and `Subscription`).
All arguments types names are now generated using the same function, so there should not be other similar issues.

The function prepends the "parent" name to the arguments type, in order to avoid possible duplications.

**Note**: names generated since this version will be different from the previous ones.

### Example

The schema
```
type Query {
   user(id: Int): User!
}

type Mutation {
   setMessage(message: String): String!
}
```

is now generated as

<pre>
case class Query(
   user: <b>Query</b>UserArgs => zio.UIO[User]
)

case class Mutation(
   setMessage: <b>Mutation</b>SetMessageArgs => zio.UIO[String]
)
</pre>

while in the previous versions was

```
case class Query(
   user: UserArgs => zio.UIO[User]
)

case class Mutation(
   setMessage: SetMessageArgs => zio.UIO[String]
)
```

### DIFF
--- a/tools/src/main/scala/caliban/tools/SchemaWriter.scala
+++ b/tools/src/main/scala/caliban/tools/SchemaWriter.scala
@@ -91,28 +91,28 @@ object SchemaWriter {
   def reservedType(typeDefinition: ObjectTypeDefinition): Boolean =
     typeDefinition.name == "Query" || typeDefinition.name == "Mutation" || typeDefinition.name == "Subscription"
 
-  def writeRootField(field: FieldDefinition, effect: String): String = {
-    val argsName = if (field.args.nonEmpty) s" ${field.name.capitalize}Args =>" else ""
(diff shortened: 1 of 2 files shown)


ghostdogpr (approved):
Looks good! Thanks for getting to the bottom of this!

PR #584, opened 2020-09-08
Fix federation orphan types

paulpdaniels:
In federation it is possible that you declare a type which has no entry point in the graph but which is nonetheless required to be part of the outgoing schema. Consider:

```graphql
// Schema A
@key(fields:"id")
type User {
  id: String!
}

type Query {
  me: User!
}

```

```graphql
// Schema B
@key(fields:"id")
extends type User {
  id: String! @external
  team: Team!
}

type Query {
  team(id: String): Team
}
```

In the above graphs they should become federated with `User` gaining the `team` field through declaration, however currently the implementation ignores types which don't have entry points into the graph from `Query` so the users extension would be skipped.

This PR allows the schema to include these orphan types as long as there is an entity resolver defined for the typel.

### DIFF
--- a/federation/src/main/scala/caliban/federation/Federation.scala
+++ b/federation/src/main/scala/caliban/federation/Federation.scala
@@ -96,11 +96,13 @@ trait Federation {
       _fieldSet: FieldSet = FieldSet("")
     )
 
+    val withSDL = original.withAdditionalTypes(resolvers.map(_.toType))
+
     GraphQL.graphQL(
       RootResolver(
         Query(
           _entities = args => args.representations.map(rep => _Entity(rep.__typename, rep.fields)),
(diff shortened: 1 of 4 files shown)


federation/src/main/scala/caliban/federation/Federation.scala
```diff
@@ -96,11 +96,13 @@ trait Federation {
       _fieldSet: FieldSet = FieldSet("")
     )
 
+    val withSDL = original.withAdditionalTypes(resolvers.map(_.toType))
```
paulpdaniels, line 99: We add these only here to avoid polluting the original graph since these new types don't have any utility since they couldn't be accessed through a normal query anyway.

ghostdogpr (approved):
Looks good!

PR #593, opened 2020-09-30
http4s adapter update

fokot:
This might me helpful for others too. We needed to parse token with the help of configured `AuthService` from common top layer in our project. It is just widening app so backwards compatible.

### DIFF
--- a/adapters/http4s/src/main/scala/caliban/Http4sAdapter.scala
+++ b/adapters/http4s/src/main/scala/caliban/Http4sAdapter.scala
@@ -292,7 +292,7 @@ object Http4sAdapter {
    */
   def provideSomeLayerFromRequest[R <: Has[_], R1 <: Has[_]](
     route: HttpRoutes[RIO[R with R1, *]],
-    f: Request[RIO[R, *]] => TaskLayer[R1]
+    f: Request[RIO[R, *]] => RLayer[R, R1]
   )(implicit tagged: Tag[R1]): HttpRoutes[RIO[R, *]] =
     Kleisli { req: Request[RIO[R, *]] =>
       val to: RIO[R, *] ~> RIO[R with R1, *] = FunctionK.lift[RIO[R, *], RIO[R with R1, *]](identity)


adapters/http4s/src/main/scala/caliban/Http4sAdapter.scala
```diff
@@ -292,7 +292,7 @@ object Http4sAdapter {
    */
   def provideSomeLayerFromRequest[R <: Has[_], R1 <: Has[_]](
     route: HttpRoutes[RIO[R with R1, *]],
-    f: Request[RIO[R, *]] => TaskLayer[R1]
+    f: Request[RIO[R, *]] => RLayer[R, R1]
```
ghostdogpr, line 295: How about `provideLayerFromRequest` above? We can change it the same way, right?

fokot: I tried it and now I looked at method type definition and I'm not sure if it is possible. Look
```
  def provideLayerFromRequest[R <: Has[_]](route: HttpRoutes[RIO[R, *]], f: Request[Task] => TaskLayer[R])(
    implicit tagged: Tag[R]
  ): HttpRoutes[Task] =
```
```
type HttpRoutes[F[_]] = Http[OptionT[F, *], F]
```
```
type Http[F[_], G[_]] = Kleisli[F, Request[G], Response[G]]
```
This means that when I use HttpRoutes, the inner (request body) and outer (request returning) effect needs to be same. So function `f: Request[F]` needs to have same effect as result value which for sure must be `HttpRoutes[Task]`. So I can't change effect for Request body. I would work if I change result type to something like Http[OptionT[RIO[R, *], *],  Task] so outer and inner effect would not be the same. But it might be more complicated for users. But I might be wrong, so pls correct me if so.

ghostdogpr: Oh I see. Let's merge this one then.

ghostdogpr (approved):

PR #594, opened 2020-09-30
Wrap derived schema in Exported wrapper

darl:
This PR fights with issue described on this issues:
  - https://github.com/propensive/magnolia/issues/89 
  - https://github.com/propensive/magnolia/issues/213
  - https://github.com/propensive/magnolia/issues/107

Short version: macro derivation sometimes conflicts with builtin schemas defined in Schema companion object.

On this PR I followed solution described by @neko-kai on [this comment](https://github.com/propensive/magnolia/issues/107#issuecomment-589289260).

In first commit I added failing test so it can be tested it in isolation.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -74,6 +74,18 @@ lazy val root = project
     federation
   )
 
+lazy val macros = project
+  .in(file("macros"))
+  .settings(name := "caliban")
+  .settings(commonSettings)
+  .settings(
+    testFrameworks := Seq(new TestFramework("zio.test.sbt.ZTestFramework")),
+    libraryDependencies ++= Seq(
+      "com.propensive" %% "magnolia" % "0.17.0",
+      "com.propensive" %% "mercator" % "0.2.1"
+    )
+  )
+
 lazy val core = project
   .in(file("core"))
   .settings(name := "caliban")
@@ -94,6 +106,7 @@ lazy val core = project
       compilerPlugin("com.olegpy" %% "better-monadic-for" % "0.3.1")
     )
   )
+  .dependsOn(macros)
   .settings(
     fork in Test := true,
     fork in run := true
--- a/core/src/main/scala/caliban/schema/Schema.scala
+++ b/core/src/main/scala/caliban/schema/Schema.scala
@@ -89,7 +89,11 @@ trait Schema[-R, T] { self =>
   }
 }
 
-object Schema extends GenericSchema[Any]
+object Schema extends GenericSchema[Any] with LowPriorityExportedSchema
+
+private[schema] trait LowPriorityExportedSchema {
+  implicit def exportedSchema[R, T](implicit exported: Exported[Schema[R, T]]): Schema[R, T] = exported.instance
+}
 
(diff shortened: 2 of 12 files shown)


core/src/main/scala/caliban/schema/Schema.scala
```diff
@@ -542,8 +546,7 @@ trait DerivationSchema[R] {
   private def getDescription[Typeclass[_], Type](ctx: ReadOnlyParam[Typeclass, Type]): Option[String] =
     getDescription(ctx.annotations)
 
-  implicit def gen[T]: Typeclass[T] = macro Magnolia.gen[T]
-
+  implicit def gen[T]: Exported[Typeclass[T]] = macro ExportedMagnolia.exportedMagnolia[Typeclass, T]
```
ghostdogpr, line 549: How about calling this differently and make `gen` return the same thing as before? So that we don't need to change every existing call.

build.sbt
```diff
@@ -74,6 +74,18 @@ lazy val root = project
     federation
   )
 
+lazy val macros = project
+  .in(file("macros"))
```
ghostdogpr, line 78: Why is it needed?

ghostdogpr (commented):
Interesting! I have some suggestion to reduce breaking changes.

build.sbt
darl: Macros cannot be invoken in same compilation unit where they defined.
`Schema.gen` used in `caliban.introspection.Introspector`

ghostdogpr: Makes sense 👍

core/src/main/scala/caliban/schema/Schema.scala
darl: Have some experiments with this.

1. Old method
```scala
implicit def gen[T]: Typeclass[T] = Magnolia.gen[T]
```
Will pollute implicit scope when imported.

2. Without implicit
```scala
def gen[T]: Typeclass[T] = Magnolia.gen[T]
```
Different semantic: magnolia disables recursive generation without `implicit` keyword
https://github.com/propensive/magnolia/blob/master/core/shared/src/main/scala/magnolia.scala#L249-L251

3. Using implicit
```scala
def gen[T](implicit e: Exported[Typeclass[T]]) = e.instance
```
Works with default schema (`Schema.gen[MyClass]`) but not work with custom schemas like this:
```scala
object mySchema extends GenericSchema[Blocking]
mySchema.gen[MyClass] // this will fail 
```

I will experiment with this a bit more on weekend.

ghostdogpr: What if you renamed the implicit `gen` into e.g. `autogen` and added a new `def gen` that is not implicit and just a shortcut to avoid calling `.instance`?

ghostdogpr: I think the 3rd one is acceptable, actually the caliban docs say you should `import mySchema._` when you use this pattern, which makes the implicit in scope. I opened https://github.com/ghostdogpr/caliban/pull/658 to get this in.

ghostdogpr:
Closing since #658 was merged

PR #595, opened 2020-10-02
Add a pure selection builder helper

paulpdaniels:
Adds a pure selection helper to the `SelectionBuilder`, this enables building more expressive queries.

For instance say you wanted to emulate the `@include`/`@skip` directive. While you could do it with `withDirective` it is somewhat clumsy.

Now you can do:

```scala
case class Customer(id: String, subscription: Option[Subscription])

// Generated client...

//Query
(Customer.id ~ (if(skip) Customer.subscription else SelectionBuilder.pure(None)))
```

This also helps to align case classes in instances where you have an additional field that you want to include but it isn't part of the schema (and you don't want to perform intermediate transforms).


ghostdogpr (approved):
Looks good! Can you check the test failure?

ghostdogpr (approved):

PR #597, opened 2020-10-02
Fix: custom directives can be added and used in wrappers

hwielenberg:
I encountered some issues while trying to add a custom directive to our server.
This fixes those issues.

Directives can be added on many levels - I've only tested on the level of `__DirectiveLocation .QUERY` and `__DirectiveLocation .MUTATION`. Maybe it's worth adding tests for other DirectiveLocations as well.


ghostdogpr (approved):
Looks good! Thanks for the fix!

(FYI a snapshot will be generated as I merge this)

PR #607, opened 2020-10-09
Handle nested orphans

paulpdaniels:
Missed this when I did #584 , only the top level orphans were getting added to the `sdl` schema even though they were still appearing in the introspection call. This makes it so that all children of the orphaned types will also be added.


ghostdogpr (approved):
👍

PR #610, opened 2020-10-14
Clarifies CodeGen sbt dependency and plugin instructions in client and schema docs

favorable-mutation:
The `caliban-codegen-sbt` plugin needs to be added via the `project/plugins.sbt` file and enabled through the `build.sbt` file in order to be used in an sbt project. This change clarifies that the call to `addSbtPlugin` should be in `project/plugins.sbt`, and the call to `enablePlugins` should be in `build.sbt`.


favorable-mutation:
Added another change to the client docs:

In order to use caliban-client with ScalaJS, it needs to be included in `build.sbt` using three percent signs (`%%%`) instead of two (`%%`). This change makes that difference explicit for a user reading the docs.

ghostdogpr (approved):
Looks good, thank you!

PR #628, opened 2020-11-08
Keep information about interface

### LINKED ISSUE #627: Mutation schema overrides types from Query schema
If the mutation resolver has an operation with a subtype of an interface, the information about the subtype is overwritten. 

Interface definition:
```scala
@GQLInterface
sealed trait MyInterface
object MyInterface {
  case class A(common: Int, different: String)  extends MyInterface 
  case class B(common: Int, different: Boolean) extends MyInterface
}
```

### Everything works as expected when the interface is a part of the mutation:

```scala
case class Queries(findInterface: MyInterface)
case class Mutations(create: String => MyInterface, createA: String => MyInterface.A)

val resolver = RootResolver(Queries(...), Mutation(...))
GraphQL.graphQL(resolver)
```

### Information about the interface is missing for the subtype:

```scala
case class Queries(findInterface: MyInterface)
case class Mutations(create: String => MyInterface.A)

val resolver = RootResolver(Queries(...), Mutation(...))
GraphQL.graphQL(resolver)
// definition says MyInterface.A does not implement MyInterface
```

### Expected behavior

Information about the interface is always present for the subtype

iRevive:
Closes #627

### DIFF
--- a/core/src/main/scala/caliban/GraphQL.scala
+++ b/core/src/main/scala/caliban/GraphQL.scala
@@ -26,6 +26,9 @@ trait GraphQL[-R] { self =>
   protected val wrappers: List[Wrapper[R]]
   protected val additionalDirectives: List[__Directive]
 
+  private[caliban] def validateRootSchema: IO[ValidationError, RootSchema[R]] =
+    Validator.validateSchema(schemaBuilder)
+
   /**
    * Returns a string that renders the API types into the GraphQL format.
    */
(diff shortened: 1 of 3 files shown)


iRevive:
The build fails due to https://github.com/sbt/sbt/issues/6075

ghostdogpr:
Thanks for the PR! I tried restarting the build but same result. Hopefully the domain will be restored soon.

ghostdogpr (approved):

ghostdogpr:
Thanks for the fix! This will be available in a snapshot.

PR #640, opened 2020-11-21
Fix for Issue 603

### LINKED ISSUE #603: Exception classes should override getMessage instead of toString
Exceptions sub classes overrides `toString` instead of `getMessage` for instance in https://github.com/ghostdogpr/caliban/blob/master/client/src/main/scala/caliban/client/CalibanClientError.scala#L13

As a result when these exceptions are logged through SLF4J, the stack trace does not contain the actual error message. For instance I get:

```
caliban.client.CalibanClientError$CommunicationError: null
	at caliban.client.SelectionBuilder.$anonfun$toRequestWithExtensions$2(SelectionBuilder.scala:95)
	at scala.util.Either$LeftProjection.map(Either.scala:573)
	at caliban.client.SelectionBuilder.$anonfun$toRequestWithExtensions$1(SelectionBuilder.scala:95)
        ...
```

I'll try to contribute a PR.

tdrozdowski:
Closes #603 - Added getMessage to CalibanClientError trait and implemented on Error classes; added spec for each implementation

### DIFF
--- a/client/src/main/scala/caliban/client/CalibanClientError.scala
+++ b/client/src/main/scala/caliban/client/CalibanClientError.scala
@@ -33,7 +33,7 @@ object CalibanClientError {
         .map(e =>
           s"${e.message} ${e.locations
             .getOrElse(Nil)
-            .map(loc => s" at line ${loc.line} and column ${loc.column}")
+            .map(loc => s"at line ${loc.line} and column ${loc.column}")
             .mkString(" ")}${e.path.fold("")(p =>
             s" at path ${p.map {
               case Left(value)  => s"/$value"
--- a/client/src/main/scala/caliban/client/CalibanClientError.scala
+++ b/client/src/main/scala/caliban/client/CalibanClientError.scala
@@ -4,8 +4,7 @@ package caliban.client
  * The base type for all Caliban Client errors.
  */
 sealed trait CalibanClientError extends Throwable with Product with Serializable {
-  def msg: String
-  override def getMessage: String = msg
+  override def getMessage: String = toString
 }
 
 object CalibanClientError {
@@ -28,8 +27,6 @@ object CalibanClientError {
    * A GraphQL error returned by the backend (e.g. failure during query parsing, validation or execution)
    */
   case class ServerError(errors: List[GraphQLResponseError]) extends CalibanClientError {
-
-    override def msg: String = toString
 
     override def toString: String =
       s"Server Error: ${errors
--- a/client/src/main/scala/caliban/client/CalibanClientError.scala
+++ b/client/src/main/scala/caliban/client/CalibanClientError.scala
@@ -3,7 +3,10 @@ package caliban.client
 /**
  * The base type for all Caliban Client errors.
  */
-sealed trait CalibanClientError extends Throwable with Product with Serializable
+sealed trait CalibanClientError extends Throwable with Product with Serializable {
+  def msg: String
+  override def getMessage: String = msg
+}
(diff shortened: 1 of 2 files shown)


client/src/main/scala/caliban/client/CalibanClientError.scala
```diff
@@ -3,7 +3,10 @@ package caliban.client
 /**
  * The base type for all Caliban Client errors.
  */
-sealed trait CalibanClientError extends Throwable with Product with Serializable
+sealed trait CalibanClientError extends Throwable with Product with Serializable {
+  def msg: String
+  override def getMessage: String = msg
```
ghostdogpr, line 8: How about just `override def getMessage: String = toString`?

So we get the most details including for communication errors and decoding errors?

client/src/test/scala/caliban/client/CalibanClientErrorSpec.scala
```diff
@@ -0,0 +1,41 @@
...
+          val graphQLResponseErrors = List(
+            GraphQLResponseError(
+              message = "Error1",
+              locations = Option(List(Location(line = 1, column = 1))),
+              path = Option(List(Left("somewhere"))),
+              extensions = Option.empty
+            )
+          )
+          val error = ServerError(graphQLResponseErrors)
+          assert(error.getMessage)(equalTo(error.toString))
```
ghostdogpr, line 36: How checking the actual string here? To make sure the `toString` is producing what we expect? Otherwise you are just testing that `getMessage` points to `toString`.

client/src/main/scala/caliban/client/CalibanClientError.scala
tdrozdowski: Done

client/src/test/scala/caliban/client/CalibanClientErrorSpec.scala
tdrozdowski: Done

client/src/test/scala/caliban/client/CalibanClientErrorSpec.scala
```diff
@@ -0,0 +1,41 @@
...
+          val graphQLResponseErrors = List(
+            GraphQLResponseError(
+              message = "Error1",
+              locations = Option(List(Location(line = 1, column = 1))),
+              path = Option(List(Left("somewhere"))),
+              extensions = Option.empty
+            )
+          )
+          val error = ServerError(graphQLResponseErrors)
+          assert(error.getMessage)(equalTo("Server Error: Error1  at line 1 and column 1 at path /somewhere"))
```
ghostdogpr, line 36: There's an extra space that can be fixed 😄  That test proved useful after all~

tdrozdowski: The extra space due to processing the optional cause?

ghostdogpr: I think it's the one between `${e.message} ${e.locations` that is not needed.

ghostdogpr: Unless it's the one before "at". To be sure you could test one with 2 locations.

tdrozdowski: Yeah, found it all - I think I have it now.

ghostdogpr (approved):
LGTM! Thank you!

PR #641, opened 2020-11-21
Query param in post endpoint for the play adaptor

### LINKED ISSUE #421: Support "query" query string parameter in POST endpoint

jessenr:
Closes #421


ghostdogpr (approved):
LGTM! Thanks for the contribution!

@paulpdaniels can you double check it since you're the play expert 😄

paulpdaniels (approved):

Issue #543, opened 2020-08-08
[CodeGen] name clashe with imported fields (EnumValue)


MasseGuillaume:
Another day, another weird GraphQL schema!

I'm working with an e-commerce platform called commercetools. They have a graphql schema that looks like this:
https://github.com/commercetools/sunrise-spa/blob/master/graphql.schema.json (They are not even using this type ...).

They decided to define `EnumValue` and this clashes with the import: `import caliban.client.Value._`.

I think this is correct because the specs refer to `__EnumValue`. https://spec.graphql.org/June2018/#sec-The-__EnumValue-Type

Generated code

```scala
// evil part
type EnumValue
object EnumValue {
  def key: SelectionBuilder[EnumValue, String] = Field("key", Scalar())
  def label: SelectionBuilder[EnumValue, String] = Field("label", Scalar())
}

// usage that breaks ...
sealed trait SubscriptionHealthStatus
      extends scala.Product
      with scala.Serializable
  object SubscriptionHealthStatus {
    case object TemporaryError extends SubscriptionHealthStatus
    case object ConfigurationErrorDeliveryStopped
        extends SubscriptionHealthStatus
    case object ConfigurationError extends SubscriptionHealthStatus
    case object Healthy extends SubscriptionHealthStatus

    //...
    implicit val encoder: ArgEncoder[SubscriptionHealthStatus] =
      new ArgEncoder[SubscriptionHealthStatus] {
        override def encode(value: SubscriptionHealthStatus): Value =
          value match {
            case SubscriptionHealthStatus.TemporaryError =>
              EnumValue("TemporaryError")
            case SubscriptionHealthStatus.ConfigurationErrorDeliveryStopped =>
              EnumValue("ConfigurationErrorDeliveryStopped")
            case SubscriptionHealthStatus.ConfigurationError =>
              EnumValue("ConfigurationError")
            case SubscriptionHealthStatus.Healthy => __EnumValue("Healthy")
          }
        override def typeName: String = "SubscriptionHealthStatus"
      }
  }

```

Compilation issue

```
[error] /home/gui/automat/commercetools-graphql-scala/src/main/scala/client.scala:646:24: com.automat.product.commercetools.graphql.Client.EnumValue.type does not take parameters
[error]               EnumValue("ConfigurationErrorDeliveryStopped")
[error]                        ^
```

Proposal:

```diff
diff --git a/src/main/scala/client.scala b/src/main/scala/client.scala
index feac5e6..be290fe 100644
[...]

PR #642, opened 2020-11-21
Rename classes to avoid name clashes

xirc:
This PR solves the issue https://github.com/ghostdogpr/caliban/issues/543.

https://github.com/ghostdogpr/caliban/issues/543#issuecomment-670833351
> How about renaming those internal classes to __???Value directly?

As suggested by @ghostdogpr, I've changed the internal class names.

- NullValue to __NullValue
- NumberValue to __NumberValue
- EnumValue to __EnumValue
- StringValue to __StringValue
- BooleanValue to __BooleanValue
- ListValue to __ListValue
- ObjectValue to __ObjectValue

### DIFF
--- a/client/src/main/scala/caliban/client/Value.scala
+++ b/client/src/main/scala/caliban/client/Value.scala
@@ -23,7 +23,7 @@ object Value {
   case class __BooleanValue(value: Boolean) extends Value {
     override def toString: String = value.toString
   }
-  case class ListValue(values: List[Value]) extends Value {
+  case class __ListValue(values: List[Value]) extends Value {
     override def toString: String = values.map(_.toString).mkString("[", ",", "]")
   }
   case class ObjectValue(fields: List[(String, Value)]) extends Value {
@@ -37,7 +37,7 @@ object Value {
       __BooleanValue,
       number => __NumberValue(number.toBigDecimal getOrElse BigDecimal(number.toDouble)),
       __StringValue,
-      array => Value.ListValue(array.toList.map(jsonToValue)),
+      array => Value.__ListValue(array.toList.map(jsonToValue)),
       obj => Value.ObjectValue(obj.toList.map { case (k, v) => k -> jsonToValue(v) })
     )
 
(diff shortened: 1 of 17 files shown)


ghostdogpr:
Nice! There's one more thing to do: modifying the code generation to use those new names. The relevant code is there: https://github.com/ghostdogpr/caliban/blob/master/tools/src/main/scala/caliban/tools/ClientWriter.scala 

Let me know if you need more explanations.

xirc:
@ghostdogpr Thanks for the review! I will try it tomorrow morning (JST+9:00).
Because the CI is success, I guess that the code generation is not tested yet, right?
I think it's better to test the code generation.
So, I will also try to test it if there is no technical problem.

ghostdogpr:
It is tested (here: https://github.com/ghostdogpr/caliban/blob/master/tools/src/test/scala/caliban/tools/ClientWriterSpec.scala) but the output is a string (generated code), and we just verify that this is the expected code, we don't actually verify that it compiles. That's why the tests passed. If you have a better idea for testing this, feel free to add it 👍

xirc:
Thanks! I got it.
In this PR, I will focus on changing the code generation and its existing test because I think writing a test verifying that generated code compiles is out of topic in this PR and makes the PR huge.

client/src/main/scala/caliban/client/Value.scala
```diff
@@ -8,47 +8,47 @@ import io.circe.{ Decoder, Encoder, Json }
 sealed trait Value
```
xirc, line 8: Is it better to rename `Value` to `__Value`?
I think the type may have the potential to collide with a user type name.

ghostdogpr: Yeah, I agree.

ghostdogpr (approved):
Perfect! Thanks for the contribution!

jeffrey-aguilera:
Ugg. This broke my build. Should call out breaking changes like this in the release notes.

PR #643, opened 2020-11-21
Add support for application/graphql content type in PlayAdapter

### LINKED ISSUE #420: Add support for application/graphql content type

dgeyer-mdsol:
Hi,

This is for ticket #420

Closes #420

### DIFF
--- a/adapters/play/src/main/scala/caliban/PlayAdapter.scala
+++ b/adapters/play/src/main/scala/caliban/PlayAdapter.scala
@@ -28,8 +28,6 @@ trait PlayAdapter[R] {
   implicit def writableGraphQLResponse[E](implicit wr: Writes[GraphQLResponse[E]]): Writeable[GraphQLResponse[E]] =
     Writeable.writeableOf_JsValue.map(wr.writes)
 
-  implicit val graphQLReqReader = Json.reads[GraphQLRequest]
-
   private def parseJson(s: String): Try[JsValue] =
     Try(Json.parse(s))
 
--- a/adapters/play/src/main/scala/caliban/PlayAdapter.scala
+++ b/adapters/play/src/main/scala/caliban/PlayAdapter.scala
@@ -1,19 +1,19 @@
 package caliban
 
-import akka.stream.{ Materializer, OverflowStrategy, QueueOfferResult }
 import akka.stream.scaladsl.{ Flow, Sink, Source, SourceQueueWithComplete }
+import akka.stream.{ Materializer, OverflowStrategy, QueueOfferResult }
 import caliban.PlayAdapter.RequestWrapper
 import caliban.ResponseValue.{ ObjectValue, StreamValue }
 import caliban.Value.NullValue
 import caliban.interop.play.json.parsingException
 import play.api.http.Writeable
-import play.api.libs.json.{ JsError, JsString, JsSuccess, JsValue, Json, Writes }
-import play.api.mvc._
+import play.api.libs.json.{ JsValue, Json, Writes }
 import play.api.mvc.Results.Ok
+import play.api.mvc._
 import zio.Exit.Failure
-import zio.{ CancelableFuture, Fiber, IO, RIO, Ref, Runtime, Schedule, Task, URIO, ZIO }
 import zio.clock.Clock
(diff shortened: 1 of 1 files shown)


adapters/play/src/main/scala/caliban/PlayAdapter.scala
```diff
@@ -7,8 +7,18 @@ import caliban.ResponseValue.{ ObjectValue, StreamValue }
 import caliban.Value.NullValue
 import caliban.interop.play.json.parsingException
 import play.api.http.Writeable
-import play.api.libs.json.{ JsValue, Json, Writes }
-import play.api.mvc.{ Action, ActionBuilder, AnyContent, PlayBodyParsers, Request, RequestHeader, Result, WebSocket }
+import play.api.libs.json.{ JsError, JsString, JsSuccess, JsValue, Json, Writes }
+import play.api.mvc.{
```
ghostdogpr, line 11: Shall we just `import  play.api.mvc._` since there are so many?

adapters/play/src/main/scala/caliban/PlayAdapter.scala
```diff
@@ -70,15 +83,28 @@ trait PlayAdapter[R] {
...
+    req.contentType match {
+      case Some(`application/graphql`) =>
+        executeRequest(
+          interpreter,
+          req.withBody(GraphQLRequest(query = req.body.asText)),
+          skipValidation,
+          enableIntrospection
+        )
+      case _ =>
+        Json.fromJson[GraphQLRequest](req.body.asJson.getOrElse(JsString("Empty request body"))) match {
```
ghostdogpr, line 96: Any reason not to use `parse.json[GraphQLRequest])` anymore? It was more concise.

ghostdogpr (commented):
Thanks for the contribution! I've never used Play myself so I will let @paulpdaniels give it a look.

adapters/play/src/main/scala/caliban/PlayAdapter.scala
dgeyer-mdsol: Thanks for reviewing, I could not manage to use it because it returns a BodyParser[GraphQLRequest] and I couldn't build the request body from it

paulpdaniels: You can create a `BodyParser` that looks like this:

```scala
private def graphqlBodyParser(implicit ec: ExecutionContext): BodyParser[GraphQLRequest] = parse.using { rh =>
  rh.contentType match {
    case Some(`application/graphql`) => parse.text.map(text => GraphQLRequest(query = Some(text))
    case _ => parse.json[GraphQLRequest]
  }
}
```

ghostdogpr:
@dgeyer-mdsol can you check the CI errors? It's just some unused imports.

adapters/play/src/main/scala/caliban/PlayAdapter.scala
```diff
@@ -1,32 +1,35 @@
...
 
+  val `application/graphql` = "application/graphql"
   def actionBuilder: ActionBuilder[Request, AnyContent]
   def parse: PlayBodyParsers
   def requestWrapper: RequestWrapper[R]
 
   implicit def writableGraphQLResponse[E](implicit wr: Writes[GraphQLResponse[E]]): Writeable[GraphQLResponse[E]] =
     Writeable.writeableOf_JsValue.map(wr.writes)
 
+  implicit val graphQLReqReader = Json.reads[GraphQLRequest]
```
paulpdaniels, line 31: I don't think this is needed, there should already be a `Reads` available for `GraphQLRequest`

dgeyer-mdsol: You're right, is a leftover for the previous changes

ghostdogpr (approved):
All green now. Thanks for the contribution!

PR #644, opened 2020-11-22
Client: Guarantee case-insensitive name uniqueness

### LINKED ISSUE #633: Client: Guarantee case-insensitive name uniqueness?
Hi!

I tried to regenerate the sources for the GitLab API again (for https://github.com/kubukoz/caliban-gitlab/), but noticed I'm getting compilation warnings:

```[E] [E6] core/src/main/scala/io/pg/gitlab/graphql.scala:961:17
[E]      Generated class io.pg.gitlab.graphql$IssueSort$UPDATED_ASC$ differs only in case from io.pg.gitlab.graphql$IssueSort$updated_asc$.
[E]        Such classes will overwrite one another on case-insensitive filesystems.
[E]      L961:     case object UPDATED_ASC extends IssueSort
[E]                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

and some more.

Context:

```scala
object IssueSort {
    case object updated_desc extends IssueSort
    case object updated_asc extends IssueSort //here
    case object created_desc extends IssueSort
    case object created_asc extends IssueSort
    case object UPDATED_DESC extends IssueSort
    case object UPDATED_ASC extends IssueSort //here
    case object CREATED_DESC extends IssueSort
    case object CREATED_ASC extends IssueSort
    case object PRIORITY_ASC extends IssueSort
    case object PRIORITY_DESC extends IssueSort
    case object LABEL_PRIORITY_ASC extends IssueSort
    case object LABEL_PRIORITY_DESC extends IssueSort
    case object MILESTONE_DUE_ASC extends IssueSort
    case object MILESTONE_DUE_DESC extends IssueSort
    case object DUE_DATE_ASC extends IssueSort
    case object DUE_DATE_DESC extends IssueSort
    case object RELATIVE_POSITION_ASC extends IssueSort
    case object SEVERITY_ASC extends IssueSort
    case object SEVERITY_DESC extends IssueSort
    case object WEIGHT_ASC extends IssueSort
    case object WEIGHT_DESC extends IssueSort
    case object PUBLISHED_ASC extends IssueSort
    case object PUBLISHED_DESC extends IssueSort
    case object SLA_DUE_AT_ASC extends IssueSort
    case object SLA_DUE_AT_DESC extends IssueSort
// ...
}
```

The API in question: https://gitlab.com/api/graphql

As you can see in [https://docs.gitlab.com/ee/api/graphql/reference/index.html#issuesort](the documentation for this type), the lowercase values have been deprecated, so in theory shouldn't appear in the responses.

Still, it would be great to avoid generating classes / objects with the same name. Do you think this is something the Caliban generator should take care of?

Reproduction at https://github.com/kubukoz/caliban-gitlab/tree/da346a5440cb62b93f09d12c72e0718f3e34c626

javimartinez:
Closes #633


tools/src/main/scala/caliban/tools/ClientWriter.scala
```diff
@@ -94,104 +107,184 @@ object ClientWriter {
        |}""".stripMargin
   }
 
+  private def addSufixesClashesNames(list: List[String]): Map[String, String] = {
```
ghostdogpr, line 110: typo: `sufix` => `suffix`

ghostdogpr (commented):
Wow, this is nice! The code is tricky but the impressive test suite increases the confidence 😄 

To be 200% sure it works correctly would you be able to run it on the original API from the issue (https://gitlab.com/api/graphql)?

ghostdogpr:
If you get the latest master there's now a scripted test to verify the generated code compiles 😄 https://github.com/ghostdogpr/caliban/pull/647

javimartinez:
I've tried https://github.com/kubukoz/caliban-gitlab/ with this version, and the problems @kubukoz reported has gone! but sadly  I've found another one: 

```scala
error: Nil.scala: org.scalafmt.dynamic.exceptions.PositionExceptionImpl: Nil.scala:1919: error: identifier expected but private found
          case object private extends VisibilityLevelsEnum
                      ^
Caused by: Nil.scala:1919: error: identifier expected but private found
          case object private extends VisibilityLevelsEnum
```

Due to the [VisibilityLevelsEnum](https://docs.gitlab.com/ee/api/graphql/reference/index.html#visibilitylevelsenum) has a member called `private` 

Would you mind if I open an issue for addressing that? I don't have much time today,  but If no one else tackles it, I can do the next week. 

#647 is amazing, I had no idea that this could be done either, in fact, I was doing by hand yesterday 😅  

I'm gonna rebase with master and move the last test  of ClientWriterSpec to this approach

tools/src/main/scala/caliban/tools/ClientWriter.scala
javimartinez: done

ghostdogpr:
Ah it's weird, `private` is already included in `reservedKeywords`. Isn't it a regression from this PR?

javimartinez:
you are right, it's a regression from this PR, I've just caught it. Should we add the Gitlab use case in the sbt-test?

ghostdogpr:
That would be awesome! Already proven useful 😄

codegen-sbt/src/sbt-test/codegen/test-compile/test
```diff
@@ -8,5 +8,7 @@ $ mkdir src/main/scala/client
 $ exists src/main/scala/Client.scala
 > calibanGenClient project/schema-to-check-name-uniqueness.graphql src/main/scala/client/ClientNameUniqueness.scala --packageName client
 $ exists src/main/scala/client/ClientNameUniqueness.scala
+> calibanGenClient https://gitlab.com/api/graphql src/main/scala/client/ClientGitLab.scala --packageName client
```
ghostdogpr, line 11: It might be safer to copy the schema to prevent the test from failing if their URL changes or if they are down.

javimartinez: it's true! done

ghostdogpr (approved):
Thanks!

PR #649, opened 2020-11-22
Add zio-json interop

paulpdaniels:
Since we are a ZIO project figured we should probably support zio-json 😸 . This PR adds support for zio-json, through the standard optional NMO channel. There are a couple areas that I'm not certain if I chose the most efficient path so I'll be asking the zio-json team to review this one as well.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -18,6 +18,7 @@ val zioVersion            = "1.0.3"
 val zioInteropCatsVersion = "2.2.0.1"
 val zioConfigVersion      = "1.0.0-RC31-1"
 val zqueryVersion         = "0.2.6"
+val zioJsonVersion        = "0.0.1"
 
 inThisBuild(
   List(
@@ -105,6 +106,7 @@ lazy val core = project
       "dev.zio"           %% "zio-test-sbt" % zioVersion % "test",
       "io.circe"          %% "circe-core"   % circeVersion % Optional,
       "com.typesafe.play" %% "play-json"    % playJsonVersion % Optional,
+      "dev.zio"           %% "zio-json"     % zioJsonVersion % Optional,
       compilerPlugin("com.olegpy" %% "better-monadic-for" % "0.3.1")
     )
   )
--- a/core/src/main/scala/caliban/CalibanError.scala
+++ b/core/src/main/scala/caliban/CalibanError.scala
@@ -1,8 +1,10 @@
 package caliban
 
 import caliban.ResponseValue.ObjectValue
+import caliban.Value.{ IntValue, StringValue }
 import caliban.interop.circe.IsCirceEncoder
 import caliban.interop.play.IsPlayJsonWrites
+import caliban.interop.zio.IsZIOJsonEncoder
 import caliban.parsing.adt.LocationInfo
 
 /**
@@ -59,6 +61,9 @@ object CalibanError {
 
   implicit def playJsonWrites[F[_]](implicit ev: IsPlayJsonWrites[F]): F[CalibanError] =
     ErrorPlayJson.errorValueWrites.asInstanceOf[F[CalibanError]]
+
+  implicit def zioJsonEncoder[F[_]](implicit ev: IsZIOJsonEncoder[F]): F[CalibanError] =
+    ErrorZioJson.errorValueEncoder.asInstanceOf[F[CalibanError]]
 }
 
 private object ErrorCirce {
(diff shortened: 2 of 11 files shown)


.sbtopts
```diff
@@ -1 +1 @@
--J-XX:MaxMetaspaceSize=3G
\ No newline at end of file
+-XX:MaxMetaspaceSize=3G
```
paulpdaniels, line 1: This one was weird it didn't work locally for me, and it looks like a strange option, was the `-J` intentional?

build.sbt
```diff
@@ -180,6 +182,7 @@ lazy val tapirInterop = project
       "com.softwaremill.sttp.tapir" %% "tapir-core"   % tapirVersion,
       "dev.zio"                     %% "zio-test"     % zioVersion % "test",
       "dev.zio"                     %% "zio-test-sbt" % zioVersion % "test",
+      "dev.zio"                     %% "zio-query"    % zqueryVersion,
```
paulpdaniels, line 185: Another weirdness, I think I can remove, but my local sbt wasn't pulling in the transitive zio-query dependencies for some reason.

core/src/main/scala/caliban/CalibanError.scala
```diff
@@ -1,9 +1,12 @@
 package caliban
 
 import caliban.ResponseValue.ObjectValue
+import caliban.Value.{ IntValue, StringValue }
 import caliban.interop.circe.IsCirceEncoder
 import caliban.interop.play.IsPlayJsonWrites
+import caliban.interop.zio.IsZIOJsonEncoder
 import caliban.parsing.adt.LocationInfo
+import zio.json.internal.Write
```
paulpdaniels, line 9: Whoops, I'll move these into the private objects

ghostdogpr:
Good idea! Now that it's released, I should give it a closer look...

paulpdaniels:
@ghostdogpr Yeah! It actually a very pleasant experience, though there are some aspects I'm still a little shaky. I think if they manage to separate the zio dependency we could also consider using it for caliban-client.

paulpdaniels:
@ghostdogpr you seen those errors before, tapir build doesn't seem to resolve `ZQuery`

ghostdogpr:
> @ghostdogpr you seen those errors before, tapir build doesn't seem to resolve `ZQuery`

I really don't get it, it has no relation with your change at all 🤔 You reproduce this locally?

ghostdogpr:
It doesn't find ZIO either: `not found: type ZIO`.

paulpdaniels:
Yeah I see it locally as well. Let me try to minimize the issue, Seemed to only occur after I added the `zio-json` dependency

paulpdaniels:
Figured it out, I added `caliban.interop.zio` which broke the import resolution for the tapir interop package. fixed by doing `_root_.`

core/src/main/scala/caliban/Value.scala
```diff
@@ -272,3 +287,144 @@ private object ValuePlayJson {
...
+      case InputValue.ObjectValue(fields) =>
+        JsonEncoder.map(JsonFieldEncoder.string, inputValueEncoder).unsafeEncode(fields, indent, out)
+      case InputValue.VariableValue(name) => JsonEncoder.string.unsafeEncode(name, indent, out)
+    }
+
+  val responseValueDecoder: JsonDecoder[ResponseValue] = (trace: List[JsonDecoder.JsonError], in: RetractReader) => {
+    val c = in.nextNonWhitespace()
+    in.retract()
+    (c: @switch) match {
+      case 'n'       => Null.decoder.unsafeDecode(trace, in)
```
ghostdogpr, line 408: We need something that low level?

paulpdaniels: I don't know if it is a bug or not, but I couldn't get the `orElse` operator to work when trying to do a regular composition. I found this code in the `ast` subtree in `zio-json` and ported it. On the plus side I think it should be really fast

core/src/main/scala/caliban/GraphQLResponse.scala
```diff
@@ -67,3 +71,45 @@ private object GraphQLResponsePlayJson {
     }
 
 }
+
+private object GraphQLResponseZioJson {
```
ghostdogpr, line 75: How about moving part of this code to the interop folder? (same for circe and play, as well as code in `Value` and `GraphQLRequest`) This is becoming quite bloated here.

paulpdaniels: Sure, I can do that. That will be a breaking change though right? Since it will no longer be available as part of the companion object?

ghostdogpr: Why wouldn't it be available? We can keep the `implicit def` in the companion object and move the implementation, no?

paulpdaniels: Doesn't `private object` need to be in the same file?

ghostdogpr: Does it need to be in a private object there? If yes we can still have the code inside that object call a function in interop I think?

core/src/main/scala/caliban/Value.scala
paulpdaniels: Discussed in the `zio-json` channel on discord. There are two approaches, one is to use the adt that is provided by zio-json which we would then transform into the caliban ADT. According to the experts we already lose many of the security benefits because of the dynamic nature of the queries. If we go with the ADT then we can remove some of this low level code but we sacrifice the remaining performance benefits (gained by directly converting into our ADT). I can go either way on this, I think the performance gain is a definite plus, but the real reason for this PR is convenience for people wanting to use `zio` all the way through their stack.

paulpdaniels:
@ghostdogpr all comments addressed, the one outstanding is whether to use the low-level zio-json API or to use the built in ast and convert after deserializing.

ghostdogpr:
I don't mind keeping the low level API if there are benefits. I'll review the code more deeply after the holidays, hopefully by Sunday. Merry Christmas! 🎅

core/src/main/scala/caliban/GraphQLRequest.scala
```diff
@@ -37,3 +40,9 @@ private object GraphQLRequestPlayJson {
 
   val graphQLRequestReads: Reads[GraphQLRequest] = Json.reads[GraphQLRequest]
 }
+
```
ghostdogpr, line 43: Just for consistency, how about moving this part too?

ghostdogpr (approved):
LGTM 👍  Just a small suggestion.

Issue #631, opened 2020-11-14
Support for Scala 3


ghostdogpr:
There are currently 2 blockers caused by usage of macros in dependencies:
- the parser is written using `fastparse` which heavily relies on macros and needs a rewrite. There are some efforts at porting Li Haoyi stack https://contributors.scala-lang.org/t/porting-ammonite-to-dotty/4469/11 but nothing on fastparse yet. An alternative would be using recently released [cats-parse](https://github.com/typelevel/cats-parse) that already publishes for dotty
- schemas are derived using `magnolia` which is also using macros. There are also some ongoing works https://github.com/propensive/magnolia/pull/249 but nothing usable yet. An alternative would be using dotty native typeclass derivation features to derive the schema without magnolia.

Possible action items:
- [ ] migrate the [Parser](https://github.com/ghostdogpr/caliban/blob/master/core/src/main/scala/caliban/parsing/Parser.scala) from fastparse to cats-parse (which supports Dotty) and compare the performance (run the benchmarks project and see if there's an impact). If there is no performance impact, we can simply replace it. Otherwise we can make a Scala 3 branch and use it only there.
- [ ] make a POC (proof of concept) for deriving caliban `Schema` using [Scala 3 typeclass derivation](http://dotty.epfl.ch/docs/reference/contextual/derivation.html). Relevant code in Caliban is https://github.com/ghostdogpr/caliban/blob/master/core/src/main/scala/caliban/schema/Schema.scala#L366. This can be done step by step, first focusing on `toType()` and converting the Scala type to a GraphQL type.

ghostdogpr:
About Dotty, I did some tests about it but I hit some dotty bugs (https://github.com/lampepfl/dotty/issues/8183) and the lack of documentation and examples threw me off for now. But I definitely want to revisit when it is more mature (or if someone is motivated to collaborate on this).

ghostdogpr:
I added 2 action items. The first one can be done in Scala 2 while the second one requires exploring new territories.

lvitaly:
@ghostdogpr, I would like to start with the first one. I will migrate the Parser from fastparse to cats-parse.

ghostdogpr:
@lvitaly awesome! Let me know if you have any question!

ghostdogpr:
Super helpful article on typeclass derivation with Scala 3: https://blog.philipp-martini.de/blog/magic-mirror-scala3/

javimartinez:
[...]

PR #650, opened 2020-11-22
the Parser migration from fastparse to cats-parse

lvitaly:
Relates to #631


YulawOne:
I think it worth to make some performance comparison between fastparse and cats-parse solution

timzaak:
If it's going on?

ghostdogpr:
@timzaak I don't think there was much progress, in case you're interested to contribute ;)

timzaak:
I tried, but cats-parse api is now very different from fastparse. Almostly need rewrite all the code. and need wait this issue: https://github.com/typelevel/cats-parse/issues/186
I will try to pass the `Parser` test cases. https://github.com/timzaak/caliban/tree/cat-parse

Honestly, It's may be a good idea to wait fastparse support Scala 3.

lvitaly:
@timzaak @ghostdogpr I rewrote Parser with cats-parse but stuck with whitespace implementation. I will commit the last changes soon.

timzaak:
cats-parse 0.3.1 has changed a lot comparing with 0.1 
you may update it firstly

timzaak:
@lvitaly I changed a lot depends on your work.

timzaak:
https://github.com/timzaak/caliban/blob/cat-parse/core/src/main/scala/caliban/parsing/Parser2.scala#L32 the whitespaces can work

timzaak:
@lvitaly @ghostdogpr 
[Parser2.scala](https://github.com/timzaak/caliban/blob/cat-parse/core/src/main/scala/caliban/parsing/Parser2.scala) now can pass all the test cases( except invalid syntax, becase cats-parse error is different from fastparse)


The performance would improve by handle `whitespaces` and `trackback` more carefully.  

PS:
1. string value implementation is from cats-parse JsonString.
2. I firstly use cats-parse, so it needs code review.

timzaak:
ref: https://github.com/typelevel/cats-parse/issues/188#issuecomment-808873555

timzaak:
I rewrite it, and it looks good to me, I don't know if it is good to replace fastparse, or just provide an alternative choice for GraphQL parse.

ghostdogpr:
@timzaak one thing that would be interesting is to run the caliban benchmark project with fastparse vs cats-parse. If there is not much difference maybe we could migrate entirely. If cats-parse is much slower maybe we only adopt it temporarily inside a "scala3" branch.

ghostdogpr:
Closing in favor of https://github.com/ghostdogpr/caliban/pull/850

PR #651, opened 2020-11-23
Add mapEither to the client

ghostdogpr:
It gives more flexibility for mapping to your own types, for example getting rid of `Option` by failing when the value is `None`.


kubukoz:
I love how simple this is :)

PR #658, opened 2020-11-26
Improve laziness of typeclass derivation

ghostdogpr:
Continuation of #594 and @darl's work to make sure Magnolia uses the custom typeclass instances before it tries to derive it.


paulpdaniels (approved):

ghostdogpr:
There is one drawback to have one level of indirection when calling magnolia, which is that you don't get Magnolia detailed error messages when calling `gen[A]` explicitly. Instead, you get an error saying it can't find the implicit `Derived`.

You can get the detailed error by calling `autogen[A].instance` but that's less straightforward.

ghostdogpr:
Maybe it's fine if I document it. `gen[A]` will still work so no breaking change, and `autogen[A].schema` can be recommended for debugging.

@paulpdaniels any opinion on this?

paulpdaniels:
I'm fine with this, it's all still magic to me. I wonder though could we have some way of surfacing the `autogen` distinction in the error message? Or perhaps we could add it to the scaladocs on `gen`?

ghostdogpr:
I don't think we can surface the error message because `gen` is not calling the macro itself, it is only checking if there is an implicit in scope.

Good point on the documentation. I think I will rename `autogen` to `genMacro` too, the difference will be clearer.

PR #671, opened 2020-12-05
fix: error in Monix interop example

patseev:
Hey guys. Stumbled upon an error in Monix interop example. There was an error in creation of default ZIO runtime. Fixed so nobody else has to struggle after mindlessly copying the code.


ghostdogpr:
Good catch, thanks!

I am not going to deploy the docs right now because they include stuff about unreleased features, but they will be updated at the next release which should happen very soon.

PR #680, opened 2020-12-10
Revert schema derivation changes

ghostdogpr:
It seems to bring new problems and it doesn't totally prevent previous issues (see https://scastie.scala-lang.org/g8tUeZOpRkKf3XWB0CjcRw)


paulpdaniels (approved):

ghostdogpr:
@paulpdaniels actually I've heard about one case where this change really helped so I am not sure what to do with it... Will see if I can make the recursive one work...

ghostdogpr:
Okay so it has been useful for @iRevive and his team, and we found a workaround for the recursion issues, so I'm closing this.

PR #685, opened 2020-12-11
fixes GraphQL#render to add deprecated fields and enum vals

### LINKED ISSUE #669: Fields annotated with @GQLDeprecated disappear from GraphQL#render
Fields annotated with `@GQLDeprecated` disappear from GraphQL#render, while expected behavior is fields should remain and annotation `@deperecated` should be set on the field.

heyrutvik:
Closes #669


ghostdogpr (approved):
Thanks!

PR #687, opened 2020-12-13
Add handrolled schema helpers

paulpdaniels:
Makes it easier to craft handrolled schema's directly when the `gen` isn't sufficient (usually in the case of mutually- or self- recursive types).

### DIFF
--- a/core/src/main/scala/caliban/schema/Schema.scala
+++ b/core/src/main/scala/caliban/schema/Schema.scala
@@ -4,7 +4,6 @@ import java.time.{ Instant, LocalDate, LocalDateTime, LocalTime, OffsetDateTime,
 import java.time.format.DateTimeFormatter
 import java.time.temporal.Temporal
 import java.util.UUID
-
 import caliban.CalibanError.ExecutionError
 import caliban.ResponseValue._
 import caliban.Value._
@@ -18,12 +17,11 @@ import caliban.{ InputValue, ResponseValue }
 import magnolia._
 import zio.query.ZQuery
 import zio.stream.ZStream
-import zio.{ URIO, ZIO }
+import zio.{ Chunk, NonEmptyChunk, URIO, ZIO }
 
 import scala.annotation.implicitNotFound
 import scala.concurrent.Future
 import scala.language.experimental.macros
-import zio.Chunk
 
 /**
  * Typeclass that defines how to map the type `T` to the according GraphQL concepts: how to introspect it and how to resolve it.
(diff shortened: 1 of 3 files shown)


core/src/main/scala/caliban/schema/Schema.scala
```diff
@@ -231,30 +293,14 @@ trait GenericSchema[R] extends DerivationSchema[R] with TemporalSchema {
...
-              else makeNonNull(evB.toType_(isInput, isSubscription))
-          ) ->
-            ((tuple: (A, B)) => evB.resolve(tuple._2))
-        )
+      Some(s"A tuple of $typeAName and $typeBName")
+    )(implicit ft =>
+      NonEmptyChunk(
+        field[(A, B)]("_1", Some("First element of the tuple"))(_._1),
+        field[(A, B)]("_2", Some("Second element of the tuple"))(_._2)
+      )
```
ghostdogpr, line 303: Could the helper be used for `lazy val kvSchema: Schema[RA with RB, (A, B)] = objectSchema(` a few lines below?

core/src/main/scala/caliban/schema/Schema.scala
```diff
@@ -683,3 +729,56 @@ trait TemporalSchema {
     )(formatter)
 
 }
+
+case class FieldType(isInput: Boolean, isSubscription: Boolean)
```
ghostdogpr, line 733: How about `FieldAttributes`? Just because `Type` is used a lot with a different meaning

core/src/main/scala/caliban/schema/Schema.scala
```diff
@@ -683,3 +729,56 @@ trait TemporalSchema {
...
+  def apply[R, V1](fn: V => V1)(implicit ev: Schema[R, V1], ft: FieldType): (__Field, V => Step[R]) =
+    (
+      __Field(
+        name,
+        description,
+        Nil,
+        () =>
+          if (ev.optional) ev.toType_(ft.isInput, ft.isSubscription)
+          else Types.makeNonNull(ev.toType_(ft.isInput, ft.isSubscription)),
+        directives = Some(directives).filter(_.nonEmpty)
```
ghostdogpr, line 749: Should it be `None` if empty? not sure if it makes a difference

core/src/main/scala/caliban/schema/Schema.scala
```diff
@@ -683,3 +729,56 @@ trait TemporalSchema {
...
+}
+
+case class PartiallyAppliedFieldWithArgs[V, A](
+  name: String,
+  description: Option[String],
+  directives: List[Directive]
+) {
+  def apply[R, V1](
+    fn: V => V1
+  )(implicit ev1: Schema[R, V1], ev2: Schema[R, A], ft: FieldType): (__Field, V => Step[R]) = {
```
ghostdogpr, line 762: If you take a `Schema[R, V1 => A]` maybe you can use it avoid duplicating that part from `functionSchema`?

vuepress/docs/docs/schema.md
```diff
@@ -241,3 +241,48 @@ The generated code will be formatted with Scalafmt using the configuration defin
 The package of the generated code is derived from the folder of `outputPath`. This can be overridden by providing an alternative package with the `--packageName` option.
 
 By default, each Query and Mutation will be wrapped into a `zio.UIO` effect. This can be overridden by providing an alternative effect with the `--effect` option.
+
+## Building Schemas by hand
+
+Sometimes the for what ever reason schema generation fails. This can sometimes happen if your schema has co-recursive types and Magnolia is unable
```
ghostdogpr, line 247:
```suggestion
Sometimes for whatever reason schema generation fails. This can happen if your schema has co-recursive types and Magnolia is unable
```

core/src/main/scala/caliban/schema/Schema.scala
```diff
@@ -150,6 +148,70 @@ trait GenericSchema[R] extends DerivationSchema[R] with TemporalSchema {
...
+   *      field("id")(_.id),
+   *      field("group")(_.group)
+   *    )
+   *  )
+   *
+   * }}}
+   *
+   */
+  def obj[R1, V](name: String, description: Option[String] = None, directives: List[Directive] = Nil)(
+    fields: FieldType => NonEmptyChunk[(__Field, V => Step[R1])]
```
ghostdogpr, line 205: How about using `List` instead of `NonEmptyChunk`? The schema validation will catch the case where an empty list is passed, and that type is friendlier to people who don't know ZIO. I don't have a strong opinion on this though, it's just that I know a bunch of people using Caliban without ZIO types.

core/src/main/scala/caliban/schema/Schema.scala
```diff
@@ -231,30 +293,14 @@ trait GenericSchema[R] extends DerivationSchema[R] with TemporalSchema {
     lazy val typeAName: String = Types.name(evA.toType_())
     lazy val typeBName: String = Types.name(evB.toType_())
 
-    objectSchema(
+    obj[RA with RB, (A, B)](
```
ghostdogpr, line 296: This is so nice.

core/src/main/scala/caliban/schema/Schema.scala
```diff
@@ -683,3 +729,56 @@ trait TemporalSchema {
...
 }
+
+case class FieldType(isInput: Boolean, isSubscription: Boolean)
+
+case class PartiallyAppliedField[V](
+  name: String,
+  description: Option[String],
+  directives: List[Directive]
+) {
+  def apply[R, V1](fn: V => V1)(implicit ev: Schema[R, V1], ft: FieldType): (__Field, V => Step[R]) =
```
ghostdogpr, line 740: How about a variant that could be used for `eitherSchema`? It could take a `V => Either[Step[R], V1`. Is that too specific?

core/src/main/scala/caliban/schema/Schema.scala
paulpdaniels: I don't remember why we used `Option` in some cases and `empty` in others tbh. So I just copied what I had elsewhere.

core/src/main/scala/caliban/schema/Schema.scala
paulpdaniels: Fair point, I liked NEC because it enforces the behavior at compile time, but yes didn't account for people looking to use without `ZIO` types.

core/src/main/scala/caliban/schema/Schema.scala
paulpdaniels: I'll see if that would work, my only concern would be that it would confuse the inference in some cases.

ghostdogpr: If the inference is degraded, maybe we name the other something else than `apply`?

core/src/main/scala/caliban/schema/Schema.scala
paulpdaniels: Not sure I understand how this would work, could you show an example of what you mean?

ghostdogpr: Sorry I meant `A => V1`:
```scala
def apply[R, A, V1](
    fn: V => (A, V1)
  )(implicit ev: Schema[R, A => V1], ft: FieldType): (__Field, V => Step[R]) = {
  val args = ev.arguments
  ...
}
```

paulpdaniels: Ah yes, nice

core/src/main/scala/caliban/schema/Schema.scala
paulpdaniels: Done

paulpdaniels:
Comments addressed @ghostdogpr

ghostdogpr (approved):

ghostdogpr:
Looks good, thank you!

PR #692, opened 2020-12-17
Add queryExecution parameter

ghostdogpr:
This PR introduces a new setting called `queryExecution`, which allows changing the default behavior of running effectful fields in parallel for queries.

Possible values are:
- sequential
- parallel (default)
- batched (sequential for arbitrary effects but parallel for batched queries, provides better perf if you use mostly batched queries)


paulpdaniels:
LGTM should we also add benchmarks for the different strategies?

paulpdaniels (approved):

ghostdogpr:
> LGTM should we also add benchmarks for the different strategies?

Would be nice if Yulaw could contribute it since I just user their benchmarks.

YulawOne:
LGTM for me too, I have tested locally and got really good results, where batched and sequential are showing almost the same results, which is impressive. Let's merge this PR and I will contribute benchmarks later as a separate PR. Thank you @ghostdogpr.

PR #699, opened 2020-12-24
Fix error handling logic

### LINKED ISSUE #698: Partial result for nested ZQuery
Let's say I have following case:
```scala
case class User(name: String, friends: Zquery[Any, Throwable, List[String]])
case class Queries(
  user: UserArgs => ZQuery[Any, Throwable, User]
)
```
and I submit a request with following query: 
```graphql
{
 user1: user(id1)
 user2: user(id2)
}
```
and assuming that fetching `friends` for second user fails I'll get:
```graphql
data: {
   "user1": {
         "name": "John",
         "friends": ["Tom", "Emily"]
    },
    "user2": {
          "name": "Jack",
          "friends": null
     }
}
errors: [
{
      "message": "fetching friends failed",
      "path": ["user2"]
}]
```
but let's say that I want `friends` field to be `non-nullable` and I change `User` definition to be:
```scala
case class User(name: String, friends: Zquery[Any, Nothing, List[String]])
case class Queries(
  user: UserArgs => ZQuery[Any, Throwable, User]
)
```
then according to the specification which says:

> If the field which experienced an error was declared as Non-Null, the null result will bubble up to the next nullable field. In that case, the path for the error should include the full path to the result field where the error occurred, even if that field is not present in the response.

[example](http://spec.graphql.org/draft/#example-08b62)
I expect result of request presented above to be: 
```graphql
data: {
   "user1": {
         "name": "John",
         "friends": ["Tom", "Emily"]
    },
    "user2": null
}
errors: [
{
      "message": "fetching friends failed",
      "path": ["user2"]
}]
```
but what I get is: 
```graphql
data: null
errors: [
{
      "message": "fetching friends failed",
      "path": ["user2"]
}]
```
Failure in nested query is propagated as a failure of the whole request.

Can I achieve my expected response with Caliban?

ghostdogpr:
Closes #698

### DIFF
--- a/core/src/main/scala/caliban/execution/Executor.scala
+++ b/core/src/main/scala/caliban/execution/Executor.scala
@@ -1,18 +1,17 @@
 package caliban.execution
 
+import scala.annotation.tailrec
+import scala.collection.mutable.ArrayBuffer
 import caliban.CalibanError.ExecutionError
 import caliban.ResponseValue._
 import caliban.Value._
+import caliban._
 import caliban.parsing.adt._
 import caliban.schema.Step._
-import caliban.schema.{ ReducedStep, Step }
+import caliban.schema.{ ReducedStep, Step, Types }
 import caliban.wrappers.Wrapper.FieldWrapper
-import caliban._
 import zio._
-import zio.query.{ UQuery, ZQuery }
-
-import scala.annotation.tailrec
-import scala.collection.mutable.ArrayBuffer
+import zio.query.ZQuery
 
 object Executor {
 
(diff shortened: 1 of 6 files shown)


paulpdaniels:
Can we also add a test for the error returned? I think the spec has a specific requirement for the location info (it seemed like in the issue we were returning the parent location info so not sure if that gets fixed as well here)

paulpdaniels (approved):

ghostdogpr:
@paulpdaniels Added a test, it returns `user2/friend` as expected.

PR #701, opened 2020-12-31
Implement federation tracing

paulpdaniels:
Federated tracing is somewhat different from the standard tracing model and the two aren't interchangeable within a federated graph. This means that when providing an implementing service for a federated graph we need to use the tracing protocol specific to it. 

This PR adds the the tracing wrapper to the federation package.

TODO:

- [x] Update docs
- [ ] Update to 0.11.0 if it becomes stable


paulpdaniels:
@ghostdogpr could I get your 👀 on this one? I don't think the second item will materialize soon so I think we could probably go ahead with getting this in since the dependency will only be in the federation project.

build.sbt
```diff
@@ -338,6 +338,14 @@ lazy val federation = project
       "dev.zio" %% "zio-test-sbt" % zioVersion % Test,
       compilerPlugin("com.olegpy" %% "better-monadic-for" % "0.3.1")
     ),
+    PB.targets in Compile := Seq(
+      scalapb.gen(grpc = false) -> (sourceManaged in Compile).value / "scalapb"
+    ),
+    libraryDependencies ++= Seq(
+      "com.thesamet.scalapb" %% "scalapb-runtime" % scalapb.compiler.Version.scalapbVersion % "protobuf"
+//      "com.thesamet.scalapb.common-protos" %% "proto-google-common-protos-scalapb_1.0" % "1.17.0-0" % "protobuf",
```
ghostdogpr, line 346: Can be removed?

federation/src/main/scala/caliban/federation/tracing/ApolloFederatedTracing.scala
```diff
@@ -0,0 +1,145 @@
+package caliban.federation.tracing
+
+import caliban.CalibanError.ExecutionError
+import caliban.{ GraphQLRequest, ResponseValue }
+import caliban.ResponseValue.ObjectValue
+import caliban.Value.StringValue
+import caliban.wrappers.Wrapper.{ EffectfulWrapper, FieldWrapper, OverallWrapper }
+import zio.{ clock, Ref, ZIO }
```
ghostdogpr, line 8: minor: sort imports

ghostdogpr (approved):
Just a couple minor comments and it's good to go.

PR #726, opened 2021-01-23
handle chunked websocket messages in akka http adapter

### LINKED ISSUE #725: Akka http adapter does not handle chunked websocket messages
Currently, the akka http websocket adapter only handles messages that are received in a single chunk.  While this is often the case, for subscriptions that have a large amount variables or large query itself - the underlying message may be chunked.  With an application that we are designing we are hitting this case rather frequently due to the size of the subscription start message (as it can have a large amount of variables we use to apply server side filtering criteria).  I will open a PR with a change that allows the adapter to handle both strict (non-chunked) as well as chunked messages.

Futher details from Akka HTTP docs - https://doc.akka.io/docs/akka-http/current/server-side/websocket-support.html#server-websocket-support 

> The data of a message is provided as a stream because WebSocket messages do not have a predefined size and could (in theory) be infinitely long. However, only one message can be open per direction of the WebSocket connection, so that many application level protocols will want to make use of the delineation into (small) messages to transport single application-level data units like “one event” or “one chat message”.
>
> Many messages are small enough to be sent or received in one go. As an opportunity for optimization, the model provides the notion of a “strict” message to represent cases where a whole message was received in one go. Strict messages are represented with the Strict subclass for each kind of message which contains data as a strict, i.e. non-streamed, ByteString or String.
> When receiving data from the network connection the WebSocket implementation tries to create a strict message whenever possible, i.e. when the complete data was received in one chunk. However, the actual chunking of messages over a network connection and through the various streaming abstraction layers is not deterministic from the perspective of the application. Therefore, application code must be able to handle both streamed and strict messages and not expect certain messages to be strict. (Particularly, note that tests against localhost will behave differently than tests against remote peers where data is received over a physical network connection.)

moonkev:
Fixes #725 

Notes: 
* akka-http-circe and akka-http-play-json version 1.35.3 are pulling in akka-serialization-jackson version 2.6.10 which is not compatible with the explicit akka verrsion 2.6.11.  This causes a runtime exception when trying to run the akka example.  As 1.35.3 is the latest, I had to explicitly include akka-serialization-jackson.  Should be able to remove when the json libs catch up.
* The chunking assembly requires a timeout.  I added this as a parameter to the makeWebSocketService.  Set it to 5 seconds, as I figured that was a reasonable timeout, but happy to change the default.  I can't imagine many cases though where it would take longer.

### DIFF
--- a/adapters/akka-http/src/main/scala/caliban/AkkaHttpAdapter.scala
+++ b/adapters/akka-http/src/main/scala/caliban/AkkaHttpAdapter.scala
@@ -6,7 +6,7 @@ import akka.http.scaladsl.model.ws.{ Message, TextMessage }
 import akka.http.scaladsl.server.Directives.{ complete, extractRequestContext }
 import akka.http.scaladsl.server.{ RequestContext, Route }
 import akka.http.scaladsl.unmarshalling.FromEntityUnmarshaller
-import akka.stream.scaladsl.{ Flow, Sink, Source, SourceQueueWithComplete }
+import akka.stream.scaladsl.{ Flow, Keep, Sink, Source, SourceQueueWithComplete }
 import akka.stream.{ Materializer, OverflowStrategy, QueueOfferResult }
 import caliban.AkkaHttpAdapter.{ `application/graphql`, ContextWrapper }
 import caliban.ResponseValue.{ ObjectValue, StreamValue }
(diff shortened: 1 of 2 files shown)


paulpdaniels (approved):
LGTM

ghostdogpr (approved):
This is nice, thank you!

moonkev:
Many thanks for the review @ghostdogpr and @paulpdaniels

PR #728, opened 2021-01-24
Fix rename on a union

### LINKED ISSUE #715: `Schema.rename` on a union and inline fragments
Given the following schema:
```scala
sealed trait Union
object Union {
  case class Child(field: String) extends Union
}

case class Obj(
  union: Union,
)
case class Query(
  test: Obj,
)

object Schema {
  implicit val schemaUnionChild: Schema[Any, Union.Child] = Schema.gen[Union.Child].rename("UnionChild")
  implicit val schemaTestUnion: Schema[Any, Union] = Schema.gen[Union]
  implicit val schemaQuery: Schema[Any, Query] = Schema.gen[Query]
}
```

A query:
```graphql
query {
  test {
    union {
      __typename
      ... on UnionChild {
        field
      }
    }
  }
}
```

Produces the expected result:
```json
{
  "test": {
    "union": {
      "__typename": "UnionChild",
      "field": "f"
    }
  }
}
```

Renaming the `Union` using the `rename`  method on `Schema`:
```scala
implicit val schemaTestUnion: Schema[Any, Union] = Schema.gen[Union].rename("UnionRenamed")
```
gives the following:
```json
{
  "test": {
    "union": {
      "__typename": "UnionRenamed"
    }
  }
}
```

Using the `@GQLName` annotation produces the expected result.

ghostdogpr:
Closes #715

### DIFF
--- a/core/src/test/scala/caliban/execution/ExecutionSpec.scala
+++ b/core/src/test/scala/caliban/execution/ExecutionSpec.scala
@@ -450,6 +450,70 @@ object ExecutionSpec extends DefaultRunnableSpec {
 
         assertM(interpreter.flatMap(_.execute(query)).map(_.data.toString))(equalTo("""{"i":{"id":"ok"}}"""))
       },
+      testM("rename on a union child") {
+        sealed trait Union
+        object Union {
+          case class Child(field: String) extends Union
+        }
+        case class Obj(union: Union)
+        case class Query(test: Obj)
+
+        object Schemas {
+          implicit val schemaUnionChild: Schema[Any, Union.Child] = Schema.gen[Union.Child].rename("UnionChild")
+          implicit val schemaTestUnion: Schema[Any, Union]        = Schema.gen[Union]
+          implicit val schemaQuery: Schema[Any, Query]            = Schema.gen[Query]
+        }
+        import Schemas._
+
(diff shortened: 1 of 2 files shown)


core/src/test/scala/caliban/execution/ExecutionSpec.scala
```diff
@@ -450,6 +450,70 @@ object ExecutionSpec extends DefaultRunnableSpec {
...
+                   ... on UnionChild {
+                     field
+                   }
+                 }
+               }
+             }""")
+
+        assertM(interpreter.flatMap(_.execute(query)).map(_.data.toString))(
+          equalTo("""{"test":{"union":{"__typename":"UnionChild","field":"f"}}}""".stripMargin)
+        )
```
paulpdaniels, line 515: I think should also verify that the parent got renamed.

ghostdogpr: The parent will never be returned by `__typename`, right?

paulpdaniels: It would show up in the introspection / schema though I think.

ghostdogpr: Right. This isn't affected by this PR though so I'll merge this one (finally 😄)

PR #741, opened 2021-02-02
Implement IntrospectionWrapper

macchiatow:


core/src/main/scala/caliban/GraphQL.scala
```diff
@@ -70,15 +70,17 @@ trait GraphQL[-R] { self =>
           schema.subscription.map(_.opType),
           additionalDirectives
         )
-      lazy val introspectionRootSchema: RootSchema[Any] = Introspector.introspect(rootType)
-      lazy val introspectionRootType: RootType          = RootType(introspectionRootSchema.query.opType, None, None)
+
+      lazy val introWrappers = wrappers.collect { case w: IntrospectionWrapper[R] => w }
```
paulpdaniels, line 74: It doesn't seem like `Introspection` actually needs to be variant.

macchiatow: won't compile otherwise because of 
`def introspect[R](rootType: RootType, introWrappers: List[IntrospectionWrapper[R]]): (RootSchema[Any], ZIO[R, ExecutionError, RootSchema[Any]])`

core/src/main/scala/caliban/introspection/Introspector.scala
```diff
@@ -32,8 +37,22 @@ object Introspector {
   /**
    * Generates a schema for introspecting the given type.
    */
-  def introspect(rootType: RootType): RootSchema[Any] = {
-    val types = rootType.types.updated("Boolean", Types.boolean).values.toList.sortBy(_.name.getOrElse(""))
+  def introspect[R](
+    rootType: RootType,
+    introWrappers: List[IntrospectionWrapper[R]]
```
paulpdaniels, line 42: Can you add a default argument (Nil) here since this is part of the public interface?

core/src/main/scala/caliban/wrappers/Wrapper.scala
```diff
@@ -71,6 +73,10 @@ object Wrapper {
     wrapPureValues: Boolean = false
   ) extends Wrapper[R]
 
+  case class IntrospectionWrapper[R](
```
ghostdogpr, line 76: Can you add scaladoc like for the other wrappers?

core/src/main/scala/caliban/GraphQL.scala
```diff
@@ -89,16 +91,16 @@ trait GraphQL[-R] { self =>
...
               wrap((request: GraphQLRequest) =>
                 (for {
                   doc   <- wrap(Parser.parseQuery)(parsingWrappers, request.query.getOrElse(""))
                   intro = Introspector.isIntrospection(doc)
                   _ <- IO.when(intro && !enableIntrospection) {
                         IO.fail(CalibanError.ValidationError("Introspection is disabled", ""))
                       }
-                  typeToValidate  = if (intro) introspectionRootType else rootType
-                  schemaToExecute = if (intro) introspectionRootSchema else schema
+                  typeToValidate  <- if (intro) introspectionRootType else ZIO.succeed(rootType)
```
ghostdogpr, line 102: We don't need to re-create `introspectionRootType` and `introExecSchema` at every request, it can be done in `def interpreter` in the outer effect.

ghostdogpr: And if you do this, there is no need for splitting `introCheckSchema` and `introExecSchema` I think.

ghostdogpr (approved):
A few suggestions, but this is looking great!

2 more things:
- Can you rebase to the latest master? I merged a PR that updates scalafmt and that causes many conflicts.
- Can you add some tests?

Thanks!

ghostdogpr (changes requested):
Ah I didn't mean to approve just yet, updating the status 😆

core/src/main/scala/caliban/GraphQL.scala
macchiatow: this part look really ugly agreed, I have to do this because of signature of check method: override def check(query: String): IO[CalibanError, Unit], meaning check and execute can't use same effect, as I need R at execute. I'm open for suggestions. check won't be able to run Wrapper[R]

macchiatow: I will make a work-around. in case of ::check I will pass Any as env, and Nil for wrappers. this will allow me avoid splitting

macchiatow:
@ghostdogpr @paulpdaniels all review points addressed, plz see. but such I mess created by rebase :)

ghostdogpr:
Hmm, can you rebase to the latest master? The files changed show many files you didn't edit so we can't see your actual changes.

PR #751, opened 2021-02-10
Fix Schema Comparison for Description

kpbochenek:
Comparison result for Descriptions was reverted :)


ghostdogpr (approved):
Whoops, good catch!

PR #753, opened 2021-02-11
 Introduce IntrospectionWrapper (#753)

macchiatow:


macchiatow:
@ghostdogpr ok now rebased :)

core/src/main/scala/caliban/introspection/Introspector.scala
```diff
@@ -44,8 +63,15 @@ object Introspector {
       ),
       args => types.find(_.name.contains(args.name)).get
     )
-    val introspectionSchema = Schema.gen[__Introspection]
-    RootSchema(Operation(introspectionSchema.toType_(), introspectionSchema.resolve(resolver)), None, None)
+
+    for {
+      wrappedInto <- wrap(ZIO.succeed(resolver))(introWrappers)
```
ghostdogpr, line 68: If you move this line to inside `RootSchema`, like `QueryStep(ZQuery.fromEffect(wrap(ZIO.succeed(resolver))(introWrappers).flatMap(wrappedInto => introspectionSchema.resolve(wrappedInto))`, then `introspect` would not need to return an effect. Unless I missed something, I think it can work?

ghostdogpr (commented):
Looking good, I have one suggestion to simplify further. Also can you mention it in the docs (middleware.md)? Thanks!

core/src/main/scala/caliban/introspection/Introspector.scala
macchiatow: seems you were right, fixed

macchiatow: @ghostdogpr

ghostdogpr: It looks good 👍
I didn't have the time to merge and cut a new release this weekend, I'll do it some time this week.

ghostdogpr (approved):

PR #784, opened 2021-03-01
jh/#778 mutation null filter

### LINKED ISSUE #778: Mutation contains null values if None is provided
Hi again,

after generating a schema and using it to connect a client to an existing GraphQL endpoint I stumbled over the following situation and I don't know if this is a bug or a feature and I just use it wrong. 

**Background:**
I generated a schema using graphQL introspection. As a result also the encoder `encode(value: T)` methods are generated in the schema class. 
When I now built a Mutation using this schema and don't provide data for all fields, the fields I don't provide data for stay `None` in scala. As a result, the created mutation, contains `null` values which leads to ` "Cannot convert undefined or null to object"` issues when executing the query. 

**Mutation Creation Code:**
```scala
    Mutation.createEntry(
      Some(EntryCreateInput(
        title = title,
        subTitle = subtitle,
        content = content,
        url = Some(url),
        `type` = Some(EntryTypeType.url),
        publishDate = Some(ZonedDateTime.now().toLocalDate.toString) // todo
      ))
    )(Entry.view()(CloudinaryImage_File.view(),
      Tag.view(Language.view, CloudinaryImage_File.view()),
      _QueryMeta.view,
      Language.view,
      Source.view(GeoLocation.view(LocationGoogle.view)))
    )
```

**Invalid Mutation:**
```
mutation {
  createEntry(
    data: {
      publishDate: null
      title: "<Dummy>"
      subTitle: "<Dummy>"
      image: null
      content: "<Dummy>"
      summary: null
      url: "<Dummy>"
      tags: null
      language: null
      source: null
      hasBeenTagged: null
      type: url
    }
  ) {...}
```

**Workaround/Solution:**
A workaround for us is to adapt the generated schema encoder by adding `.filterNot(_._2.equals(__NullValue))` after the generation of each list. 

**Resulting Valid Mutation:**
```
 mutation {
  createEntry(
    data: {
      title: "<Dummy>"
      subTitle: "<Dummy>"
      content: "<Dummy>"
      url: "<Dummy>"
      type: url
    }
  ) {...}
```

**Question:**
Do I just use the client mutation builder wrong or is this the intended behavior? Or is this also a bug?
If this is a bug, adding `.filterNot(_._2.equals(__NullValue))` in `ClientWriter:421` would fix this. 

Thanks in advance for any response.

johanneshiry:
This is a first draft for a filter option when generating a mutation. Currently untested. If this approach is acceptable I would fix the broken tests and could write tests accordingly (if required). 

Resolves #778

### DIFF
--- a/client/src/main/scala/caliban/client/Argument.scala
+++ b/client/src/main/scala/caliban/client/Argument.scala
@@ -9,17 +9,37 @@ import caliban.client.__Value.__NullValue
 case class Argument[+A](name: String, value: A)(implicit encoder: ArgEncoder[A]) {
   def toGraphQL(
     useVariables: Boolean,
-    variables: Map[String, (__Value, String)]
+    variables: Map[String, (__Value, String)],
+    filterNullValues: Boolean
   ): (String, Map[String, (__Value, String)]) =
     encoder.encode(value) match {
-      case `__NullValue` => ("", variables)
-      case v             =>
-        if (useVariables) {
-          val variableName = Argument.generateVariableName(name, v, variables)
-          (s"$name:$$$variableName", variables.updated(variableName, (v, encoder.formatTypeName)))
-        } else {
-          (s"$name:${v.toString}", variables)
-        }
+      case `__NullValue`         => ("", variables)
+      case v if filterNullValues =>
+        toGraphQL(filterAllNull(v), useVariables, variables)
+      case v                     =>
+        toGraphQL(v, useVariables, variables)
+    }
+
+  private def filterAllNull(v: __Value): __Value =
+    v match {
+      case __Value.__ListValue(values)   =>
+        __Value.__ListValue(values.filterNot(_.equals(__NullValue)))
+      case __Value.__ObjectValue(fields) =>
+        __Value.__ObjectValue(fields.filterNot(_._2.equals(__NullValue)))
+      case v                             =>
+        v
+    }
+
+  private def toGraphQL(
+    v: __Value,
+    useVariables: Boolean,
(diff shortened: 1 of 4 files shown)


ghostdogpr:
The approach seems good 👍 
Adding some tests would be nice.

johanneshiry:
Will write the tests asap. 

Another question considering the `null` filter for explicitly provided fields only (AFAIK you mentioned that there has been such a request as well?). I think with the proposed approach this could be easily done by providing a `List[String]` with the field names that should be filtered if they are `None` or `null` respectively. 

Second questions: is there any place where I can/should document the new possibility to filter `null` values?

ghostdogpr:
> Another question considering the `null` filter for explicitly provided fields only (AFAIK you mentioned that there has been such a request as well?). I think with the proposed approach this could be easily done by providing a `List[String]` with the field names that should be filtered if they are `None` or `null` respectively.

Hmm, that's an idea. It's not super elegant though 😆 
Maybe let's leave it out of this PR and see if people actually need it?

> Second questions: is there any place where I can/should document the new possibility to filter `null` values?

Yep, here: https://github.com/ghostdogpr/caliban/blob/master/vuepress/docs/docs/client.md#request-execution

johanneshiry:
Never said it's elegant :D - but you're right if there is no need for this atm, we can safely skip this here. At least from our side we're happy if no `null` values are inside a mutation. 

Perfect, thank you. Will add a short note in the documentation then as well.

ghostdogpr:
Closing in favor of #923

PR #808, opened 2021-03-20
Fix schema introspection `__type` type

guymers:
According to section 4.5 of the GraphQL specification, the `__type`
meta field returns an optional `__Type`.


ghostdogpr (approved):
Woops, you are right, and glad to get rid of this ugly `.get`!

PR #810, opened 2021-03-21
add scalar mappings

### LINKED ISSUE #453: Codegen: support custom scalar mapping
We should add the ability to do type to scalar mappings when running the client code generation.
If one of the mappings is not present the generation would fail. This would be useful in scenarios where we want to create a CI tool that builds the client after every server release.

```bash
// Encode the map similar to a CURL request, where the classes are fully specified.
// built in types have default handling (could think of another way to support alternative decoders)
// Multiple declarations of `--scalarMappings` would create leftward merge
--scalarMappings='{"JSON": "io.circe.Json", "Other": "Int"}'
```

We could also allow passing in a filename for `scalarMappings` which would look for types with an annotations perhaps i.e.

```scala
@ScalarType("Json")
case class JsonAdt(...)

// Note the required decoder so that we could derive a `ScalarDecoder`
object JsonAdt {
 implicit val decoder: Decoder[JsonAdt] = ???
}


```

nikodemin:
Resolves #453

* sealed trait Codegen.GenType instead of passing writer lambda (see Codegen)
* use implicits in SchemaWriter and ClientWriter
* add additional command line option --scalarMappings
* update client.md and schema.md

### DIFF
--- a/tools/src/main/scala/caliban/tools/Codegen.scala
+++ b/tools/src/main/scala/caliban/tools/Codegen.scala
@@ -9,9 +9,11 @@ object Codegen {
 
   sealed trait GenType
 
-  object Schema extends GenType
+  object GenType {
+    object Schema extends GenType
 
-  object Client extends GenType
+    object Client extends GenType
+  }
 
   def generate(
     arguments: Options,
@@ -27,10 +29,10 @@ object Codegen {
     for {
       schema    <- loader.load
       code       = genType match {
-                     case Schema =>
+                     case GenType.Schema =>
                        SchemaWriter.write(schema, packageName, effect, arguments.imports)(ScalarMappings(scalarMappings))
-                     case Client =>
-                       ClientWriter.write(schema, genView, objectName, packageName, arguments.imports)(
+                     case GenType.Client =>
+                       ClientWriter.write(schema, objectName, packageName, genView, arguments.imports)(
                          ScalarMappings(scalarMappings)
                        )
                    }
--- a/tools/src/main/scala/caliban/tools/Codegen.scala
+++ b/tools/src/main/scala/caliban/tools/Codegen.scala
@@ -27,9 +27,12 @@ object Codegen {
     for {
       schema    <- loader.load
       code       = genType match {
-                     case Schema => SchemaWriter.write(schema, packageName, effect)(ScalarMappings(scalarMappings))
+                     case Schema =>
+                       SchemaWriter.write(schema, packageName, effect, arguments.imports)(ScalarMappings(scalarMappings))
                      case Client =>
-                       ClientWriter.write(schema, genView, objectName, packageName)(ScalarMappings(scalarMappings))
+                       ClientWriter.write(schema, genView, objectName, packageName, arguments.imports)(
+                         ScalarMappings(scalarMappings)
+                       )
                    }
       formatted <- Formatter.format(code, arguments.fmtPath)
       _         <- Task(new PrintWriter(new File(arguments.toPath)))
--- a/tools/src/main/scala/caliban/tools/Codegen.scala
+++ b/tools/src/main/scala/caliban/tools/Codegen.scala
@@ -2,24 +2,35 @@ package caliban.tools
 
 import java.io.{ File, PrintWriter }
 
-import caliban.parsing.adt.Document
+import caliban.tools.implicits.ScalarMappings
 import zio.{ Task, UIO }
 
 object Codegen {
 
+  sealed trait GenType
+
+  object Schema extends GenType
+
+  object Client extends GenType
+
   def generate(
     arguments: Options,
-    writer: (Document, String, Option[String], Boolean, String) => String
+    genType: GenType
   ): Task[Unit] = {
-    val s           = ".*/scala/(.*)/(.*).scala".r.findFirstMatchIn(arguments.toPath)
-    val packageName = arguments.packageName.orElse(s.map(_.group(1).split("/").mkString(".")))
-    val objectName  = s.map(_.group(2)).getOrElse("Client")
-    val effect      = arguments.effect.getOrElse("zio.UIO")
-    val genView     = arguments.genView.getOrElse(false)
-    val loader      = getSchemaLoader(arguments.schemaPath, arguments.headers)
+    val s              = ".*/scala/(.*)/(.*).scala".r.findFirstMatchIn(arguments.toPath)
+    val packageName    = arguments.packageName.orElse(s.map(_.group(1).split("/").mkString(".")))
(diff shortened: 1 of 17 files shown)


ghostdogpr:
Glad to see this being implemented!

It's a little hard to review because your PR modifies formatting and imports in a bunch of unrelated places ^^ It would be nice if you could clean up so that the PR is only about adding the feature. You can use `fmt` in sbt for formatting.

nikodemin:
@ghostdogpr Sorry for inconvenience! I fixed my code, now tests are passing. Also i added one additional feature: additional imports in generated code. I think it will be useful for custom and unsupported types. So you could f.e. define your implicit implementation for ScalarDecoder[OffsetDateTime] and add this import during schema generation. I think this should work when dealing with random type, but I'm not so deep in internals, so please correct me if I'm wrong

ghostdogpr:
No problem 😄 sounds good, I'll try to review in details during the weekend!

tools/src/main/scala/caliban/tools/Codegen.scala
```diff
@@ -2,24 +2,38 @@ package caliban.tools
...
 
-import caliban.parsing.adt.Document
+import caliban.tools.implicits.ScalarMappings
 import zio.{ Task, UIO }
 
 object Codegen {
 
+  sealed trait GenType
+
+  object Schema extends GenType
```
ghostdogpr, line 12: I usually put these inside `object GenType` so that we can call them `GenType.Schema`

tools/src/main/scala/caliban/tools/IntrospectionClient.scala
```diff
@@ -12,8 +12,7 @@ import caliban.parsing.adt.Definition.TypeSystemDefinition.{ DirectiveDefinition
 import caliban.parsing.adt.Type.{ ListType, NamedType }
 import caliban.parsing.adt.{ Directive, Document, Type }
 import sttp.client3._
-import sttp.client3.asynchttpclient.zio.AsyncHttpClientZioBackend
-import sttp.client3.asynchttpclient.zio._
+import sttp.client3.asynchttpclient.zio.{ AsyncHttpClientZioBackend, _ }
```
ghostdogpr, line 15: We can simply remove `AsyncHttpClientZioBackend` I think?

tools/src/main/scala/caliban/tools/SchemaWriter.scala
```diff
@@ -3,15 +3,17 @@ package caliban.tools
...
 import caliban.parsing.adt.Type.{ ListType, NamedType }
 import caliban.parsing.adt.{ Document, Type }
+import caliban.tools.implicits.Implicits._
+import caliban.tools.implicits.ScalarMappings
 
 object SchemaWriter {
 
   def write(
     schema: Document,
-    objectName: String = "",
```
ghostdogpr, line 11: Nice that we don't have to mirror Client and Schema anymore.

tools/src/test/scala/caliban/tools/SchemaWriterSpec.scala
```diff
@@ -174,9 +183,6 @@ object SchemaWriterSpec extends DefaultRunnableSpec {
           )
         )
       },
-      testM("empty schema test") {
```
ghostdogpr, line 177: Any reason to remove that one?

vuepress/docs/docs/schema.md
```diff
@@ -242,6 +242,9 @@ The package of the generated code is derived from the folder of `outputPath`. Th
 
 By default, each Query and Mutation will be wrapped into a `zio.UIO` effect. This can be overridden by providing an alternative effect with the `--effect` option.
 
+If you want to force a mapping between a GraphQL type and a Scala class (such as scalars), you can use the
+`--scalarMappings` option. Also you can add imports for example for your ScalarDecoder implicits by providing `--imports` option.
```
ghostdogpr, line 246: Can you adapt the last sentence? Because there is no `ScalarDecoder` in the server.

ghostdogpr (commented):
Looks good ! I just have a few minor comments.

tools/src/main/scala/caliban/tools/IntrospectionClient.scala
nikodemin: No, it is used on 26 line

tools/src/main/scala/caliban/tools/SchemaWriter.scala
nikodemin: If you mean this writer lambda propagation, like it was before,  I agree)

tools/src/test/scala/caliban/tools/SchemaWriterSpec.scala
nikodemin: I removed this test, because on different OS line separators differs, so it's failed on windows. Also I find this test not so useful, but if you want I can return it with corrected line separator

nikodemin:
@ghostdogpr Thanks for review!
I resolved comments, but there is one open question left about deleted test for empty schema. Please say should I return it or it's redundant

tools/src/main/scala/caliban/tools/IntrospectionClient.scala
ghostdogpr: But since we import `_`, it's not needed. Right?

tools/src/test/scala/caliban/tools/SchemaWriterSpec.scala
ghostdogpr: Fine with me 👍

tools/src/main/scala/caliban/tools/IntrospectionClient.scala
nikodemin: No, it's used for 29-30 lines. If it was unused, then after optimizing imports it should fade away

ghostdogpr: What I mean is that it's already imported by `_` so there is no need to specify it. Just try to remove it and only keep `_`, you'll see.

tools/src/test/scala/caliban/tools/SchemaWriterSpec.scala
nikodemin: @ghostdogpr I returned corrected test

tools/src/main/scala/caliban/tools/IntrospectionClient.scala
nikodemin: Sorry, doesn't noticed that)

ghostdogpr (approved):
Looks all good! Thanks for the contribution!

PR #845, opened 2021-04-12
Add rendering description for argument fields

kpbochenek:
Because rendered schema never contained description for argument fields a 'schema comparison' tool always complained that there are differences.
Should be okay now.

maybe for multiline descriptions for arguments it should do something similar to types(parameter per line), e.g.:
```
fetch(
  nameLike: String!,
  "is user active currently"
  active: Boolean!
  """ multi
  line
  desc
  """
  multi: String
): Boolean!
```

just a tip if someone got moment to improve it even further :)

### DIFF
--- a/core/src/main/scala/caliban/Rendering.scala
+++ b/core/src/main/scala/caliban/Rendering.scala
@@ -85,9 +85,10 @@ object Rendering {
       case _                       => ""
     }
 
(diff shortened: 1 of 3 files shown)


ghostdogpr (approved):
Thanks!

PR #847, opened 2021-04-14
Scala 3 Support

### LINKED ISSUE #631: Support for Scala 3

ghostdogpr:
Resolves #631

- [x] setup Scala 3 compilation
- [x] split the version-specific code into 2.x and 3.x folders
- [x] implement parser on Scala 3
  - [x] enable excluded tests that required the parser to run
- [x] implement derivation on Scala 3
  - [x] `SubscriptionSchema`
  - [x] `ArgBuilder`
  - [x] `Schema`
- [x] implement `gqldoc` macro on Scala 3
- [x] enable tests on Scala 3
- [x] enable Scala 3 on the client
- [x] update docs on wrappers (the syntax had to change a little bit)
- [x] document cases where the derivation doesn't work: value classes, opaque types, nested sealed traits


ghostdogpr:
Decided to not wait for Magnolia and implement derivation myself using Scala 3 derivation features. `SubscriptionSchema` and `ArgBuilder` are done, only the big one `Schema` left.

PR #849, opened 2021-04-16
gqldoc macro for Scala 3

javimartinez:


ghostdogpr (approved):
Awesome, thanks! It looks simpler than with Scala 2.

PR #857, opened 2021-04-22
Improve the Ergonomics of the Examples

jeffwilde:
There are many excellent examples included with Caliban. I find myself coming back to them as simple reference for setting up the various Scala HTTP libraries. I do find though, that I run into difficulties caused by the examples sharing the same root `caliban` package as the main library. 

Copy-pasting and moving the example code into a different package in a user's app will introduce many compile errors that are not easily corrected without manual editing, and while reading the code it is easy to confuse yourself about if a particular import is coming from elsewhere in the examples, or from the Caliban library itself.

For example, [these two imports](https://github.com/ghostdogpr/caliban/blob/a8041585b84e65b99f1b686eb87872ec955216d6/examples/src/main/scala/caliban/tapir/ExampleApp.scala#L3-L4) are hard for a new user to differentiate as being from the example vs from the library. And in [some cases](https://github.com/ghostdogpr/caliban/blob/a8041585b84e65b99f1b686eb87872ec955216d6/examples/src/main/scala/caliban/play/ExampleApp.scala#L3) the imports from the examples and from the Caliban library are mixed into the same import statement. Further, since the examples live in the same package as the core Caliban functionality, some imports from the Caliban library are not specified, which hides them from people learning the library. For example, this use of [`GraphQL`](https://github.com/ghostdogpr/caliban/blob/a8041585b84e65b99f1b686eb87872ec955216d6/examples/src/main/scala/caliban/ExampleApi.scala#L34) and this use of [`RootResolver`](https://github.com/ghostdogpr/caliban/blob/a8041585b84e65b99f1b686eb87872ec955216d6/examples/src/main/scala/caliban/ExampleApi.scala#L36) are both unspecified imports from the `caliban` package.

To allow easy package renaming, and to make the examples easier to understand, this PR moves all the example code into a separate package and updates the example docs with the latest dependencies from the main build.


ghostdogpr (approved):
This is great, thank you!

PR #866, opened 2021-05-05
Reduce runtime required for interpreter creation in cats interop

willtrnr:
I noticed that the `interpreterAsync[F]` requires a `Runtime[R]` but doesn't actually need it to produce the interpreter.

This PR lowers that constraint to a `Runtime[Any]` to allow providing layers later to the interpreter,


ghostdogpr (commented):
Good point! Can you apply the same thing to the Monix interop as well? It's exactly the same code. Thanks!

willtrnr:
All done, I applied the same changes to Monix as well.

ghostdogpr (approved):
Awesome, thank you!

PR #880, opened 2021-05-14
Fix incorrect error returned when using default function Schema and input argument contains field named "value"

### LINKED ISSUE #879: Incorrect error returned when using default function Schema and input argument contains field named "value"
I've prepared testcase that covers this issue here: https://github.com/mszczygiel/caliban/commit/548b54df20da354a6e92749d78e38f5f6e8ad21a

Returned error is `Can't build an Int from input "value"` which is incorrect, and this is due to arbitrary name chosen in https://github.com/ghostdogpr/caliban/blob/master/core/src/main/scala/caliban/schema/Schema.scala#L360

Workaround is to provide my own `functionSchema`, but possibly proper error handling could be built into caliban itself? Maybe `"value"` could be provided from some `ArbitraryName` implicit, or even be optional and fail in case it's not provided and argument is not wrapped in a case class.

mszczygiel:
Fixes #879

### DIFF
--- a/core/src/main/scala/caliban/schema/Schema.scala
+++ b/core/src/main/scala/caliban/schema/Schema.scala
@@ -341,25 +341,43 @@ trait GenericSchema[R] extends SchemaDerivation[R] with TemporalSchema {
     ev2: Schema[RB, B]
   ): Schema[RA with RB, A => B]                                                                                        =
     new Schema[RA with RB, A => B] {
-      override def arguments: List[__InputValue] = {
-        val t = ev1.toType_(true)
-        t.inputFields.getOrElse(t.kind match {
-          case __TypeKind.SCALAR | __TypeKind.ENUM | __TypeKind.LIST =>
-            // argument was not wrapped in a case class, give it an arbitrary name
-            List(__InputValue("value", None, () => if (ev1.optional) t else makeNonNull(t), None))
-          case _                                                     => Nil
-        })
-      }
+      private val inputType                                                  = ev1.toType_(true)
+      private val unwrappedArgumentName                                      = "value"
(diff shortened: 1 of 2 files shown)


ghostdogpr:
Weirdly the CI didn't trigger, can you try login to https://app.circleci.com/ ?

core/src/main/scala/caliban/schema/Schema.scala
```diff
@@ -341,25 +341,43 @@ trait GenericSchema[R] extends SchemaDerivation[R] with TemporalSchema {
...
-      override def arguments: List[__InputValue] = {
-        val t = ev1.toType_(true)
-        t.inputFields.getOrElse(t.kind match {
-          case __TypeKind.SCALAR | __TypeKind.ENUM | __TypeKind.LIST =>
-            // argument was not wrapped in a case class, give it an arbitrary name
-            List(__InputValue("value", None, () => if (ev1.optional) t else makeNonNull(t), None))
-          case _                                                     => Nil
-        })
-      }
+      private val inputType                                                  = ev1.toType_(true)
```
ghostdogpr, line 344: Can you make this one lazy? It might cause issues with recursive types if we call `toType_` while building the schema.

ghostdogpr (approved):
Ah, the CI was triggered this time 👍 
LGTM, will merge once CI is finished.

mszczygiel:
Yeah, I needed to login. Thanks!

ghostdogpr:
Thanks for the contribution! I will generate a snapshot right away if you need it. I'm planning to do a new release soon with Scala 3 final, just waiting for all dependencies to upgrade.

PR #891, opened 2021-05-19
cats-effect 3 and http4s 0.23.0

iRevive:
This PR brings support of cats-effect **3.2.1** and http4s **0.23**.


.circleci/config.yml
```diff
@@ -21,7 +21,8 @@ jobs:
       - checkout
       - restore_cache:
           key: sbtcache
-      - run: sbt ++2.12.13! core/test http4s/compile akkaHttp/compile finch/compile play/test zioHttp/compile examples/compile catsInterop/compile benchmarks/compile tools/test codegenSbt/test clientJVM/test monixInterop/compile tapirInterop/test federation/test codegenSbt/scripted
+      #- run: sbt ++2.12.13! core/test http4s/compile akkaHttp/compile finch/compile play/test zioHttp/compile examples/compile catsInterop/compile benchmarks/compile tools/test codegenSbt/test clientJVM/test monixInterop/compile tapirInterop/test federation/test codegenSbt/scripted
```
iRevive, line 24: Temporary disabled tests for monix and finch modules

adapters/http4s/src/main/scala/caliban/Http4sAdapter.scala
```diff
@@ -315,18 +325,22 @@ object Http4sAdapter {
       route(req.mapK(to)).mapK(from).map(_.mapK(from))
     }
 
-  private def wrapRoute[F[_]: Effect](route: HttpRoutes[Task])(implicit runtime: Runtime[Any]): HttpRoutes[F] = {
-    val toF: Task ~> F    = λ[Task ~> F](_.toIO.to[F])
-    val toTask: F ~> Task = λ[F ~> Task](_.toIO.to[Task])
+  private def wrapRoute[F[_]: Async](
+    route: HttpRoutes[Task]
+  )(implicit dispatcher: Dispatcher[F], runtime: Runtime[Any]): HttpRoutes[F] = {
```
iRevive, line 330: I would prefer passing a dispatcher explicitly.  
But in `zio-interop-cats` the dispatcher provided implicitly everywhere, so I'm not sure which side should we stay on.

interop/cats/src/main/scala/caliban/interop/cats/implicits/package.scala
```diff
@@ -37,6 +38,6 @@ package object implicits {
       CatsInterop.interpreterAsync(underlying)
   }
 
-  implicit def catsEffectSchema[F[_]: Effect, R, A](implicit ev: Schema[R, A]): Schema[R, F[A]] =
+  implicit def catsEffectSchema[F[_]: Async, R, A](implicit F: Dispatcher[F], ev: Schema[R, A]): Schema[R, F[A]] =
```
ghostdogpr, line 41: `Async` can be removed if I understood properly?

ghostdogpr (commented):
I was waiting for a stable http4s release to start, good that you got ahead of me 😄 
I need to dig into CE3, I imagine there is no way out of using Futures?

interop/cats/src/main/scala/caliban/interop/cats/implicits/package.scala
iRevive: Yep. Done

iRevive:
@ghostdogpr a stable version of http4s has been release. I will update the PR today

iRevive:
> I need to dig into CE3, I imagine there is no way out of using Futures?

Yes, in CE3 futures the only way to materialize the effect.

ghostdogpr:
To not be blocked by finch and monix, we could let them depend on CE2 and upgrade only http4s and cats interop. I believe they are independent. Wdyt?

iRevive:
Sounds good to me. I will adjust the dependencies.

iRevive:
How should we deal with the `examples` project? We should keep either `http4s + catsInterop` or `finch + monix`. 
Otherwise, we will have dependencies conflict. 

As an option, I can make a separate project (e.g. `examples-2`) and move `finch` and `monix` examples there.

ghostdogpr:
Ah, right, didn't think of that 😢 
Considering monix and finch are much less used than cats-effect/http4s, maybe we just comment out the example code for those 2. So people who need it will still find the code, and we don't bother splitting the examples project. And we can uncomment when they are available. 

Maybe just add a comment at the top of the file to explain why it's commented out.

adapters/http4s/src/main/scala/caliban/Http4sAdapter.scala
```diff
@@ -327,7 +324,7 @@ object Http4sAdapter {
       } yield response
     }
 
-  def makeWebSocketService[R, E](
+  def makeWebSocketService[R <: Has[_] with Clock with Blocking, E](
```
iRevive, line 327: Unfortunately, the compilation fails without ` <: Has[_] with Clock with Blocking` in Scala 3.

But compiles fine in 2.13.6.

iRevive:
I've updated the docs and the definition of the build.

adapters/http4s/src/main/scala/caliban/Http4sAdapter.scala
ghostdogpr: Where does the constraint come from?

ghostdogpr: I tried your branch and was able to make it work: change the queues to be of `Task` instead of `RIO[R, *]` and it will solve the problem. You also need to put `.provideLayer(Clock.live)` back.

There's still an error but I think it's something else.

ghostdogpr: I think I made it work, will open a PR to your repo.

ghostdogpr: I was able to push to your branch directly, you can have a look

adapters/http4s/src/main/scala/caliban/Http4sAdapter.scala
```diff
@@ -366,14 +366,15 @@ object Http4sAdapter {
             for {
               msg    <- RIO.fromEither(decode[Json](text))
               msgType = msg.hcursor.downField("type").success.flatMap(_.value.asString).getOrElse("")
-              _      <- RIO.whenCase(msgType) {
+              _      <- RIO.whenCase[R, String](msgType) {
```
iRevive, line 369: I guess it didn't compile in Scala 3 because explicit types were missing.

iRevive:
@ghostdogpr thank you for the improvements. Looks much better now

ghostdogpr:
Can you resolve the conflict in build.sbt? Then I can merge it. Thanks!

adapters/http4s/src/main/scala/caliban/Http4sAdapter.scala
ghostdogpr: Yeah weirdly this one seems necessary with Scala 3.

PR #906, opened 2021-05-29
Fix argument types when using aliases

### LINKED ISSUE #905: Caliban Client Codegen: Not using ID type in method signature for top-level queries
Consider the following GraphQL schema:

```graphql
type Query {
    testQuery(testId: ID!): TestType
}

type TestType {
    name: String
}
```

Running caliban codegen on it will result in this Scala:

```scala
import caliban.client.FieldBuilder._
import caliban.client.SelectionBuilder._
import caliban.client._
import caliban.client.Operations._

object GQLTest {

  type TestType

  object TestType {
    def name: SelectionBuilder[TestType, Option[String]] = Field("name", OptionOf(Scalar()))
  }

  type Query = RootQuery

  object Query {

    def testQuery[A](testId: String)(innerSelection: SelectionBuilder[TestType, A]): SelectionBuilder[RootQuery, Option[A]] =
      Field("testQuery", OptionOf(Obj(innerSelection)), arguments = List(Argument("testId", testId)))
  }

}
```

The problem is that the type of `testId` in the method signature to `testQuery` is a `String`. This fails typechecking when making the call to the server. The server is expecting `testId` to be an `ID`.

ghostdogpr:
Closes #905

### DIFF
--- a/client-laminext/src/test/scala/caliban/client/laminext/Client.scala
+++ b/client-laminext/src/test/scala/caliban/client/laminext/Client.scala
@@ -21,13 +21,10 @@ object Client {
       case __StringValue("MARS")  => Right(Origin.MARS)
       case other                  => Left(DecodingError(s"Can't build Origin from input $other"))
     }
-    implicit val encoder: ArgEncoder[Origin]    = new ArgEncoder[Origin] {
-      override def encode(value: Origin): __Value = value match {
-        case Origin.BELT  => __EnumValue("BELT")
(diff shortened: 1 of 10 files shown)


paulpdaniels:
LGTM I guess we'll need another major bump for this

paulpdaniels (approved):

paulpdaniels:
I actually just hit this problem on one of our schemas as well so this is good timing

PR #918, opened 2021-06-07
Fully qualifying FieldBuilder.Field in ClientWriter

### LINKED ISSUE #917: Attempting to render a client for a specification with `Field` as a member name generates uncompilable code
First off, thank you for this project -- integration has been largely straightforward for someone with very limited experience to GraphQL prior to now.

One issue I'm facing is with the following specification:
```graphql
enum Field {
  foo
  bar
  baz
}

type Foobar {
  qux: Int!
}
```

generates the following:

```scala
package com.example

import caliban.client.CalibanClientError.DecodingError
import caliban.client.FieldBuilder._
import caliban.client.SelectionBuilder._
import caliban.client._
import caliban.client.__Value._

object Example {

  sealed trait Field extends scala.Product with scala.Serializable
  object Field {
    case object foo extends Field
    case object bar extends Field
    case object baz extends Field

    implicit val decoder: ScalarDecoder[Field] = {
      case __StringValue("foo") => Right(Field.foo)
      case __StringValue("bar") => Right(Field.bar)
      case __StringValue("baz") => Right(Field.baz)
      case other                => Left(DecodingError(s"Can't build Field from input $other"))
    }
    implicit val encoder: ArgEncoder[Field] = {
      case Field.foo => __EnumValue("foo")
      case Field.bar => __EnumValue("bar")
      case Field.baz => __EnumValue("baz")
    }
  }

  type Foobar
  object Foobar {
    def qux: SelectionBuilder[Foobar, Int] = Field("qux", Scalar())
  }

}
```

which yields the following when I attempt to compile:
```
sbt:calibanbug> compile
parsed config (v2.0.1): .scalafmt.conf
[info] compiling 1 Scala source to .../target/scala-2.13/classes ...
[error] .../target/scala-2.13/src_managed/main/sbt-caliban-ext/Example.scala:32:51: com.example.Example.Field.type does not take parameters
[error]     def qux: SelectionBuilder[Foobar, Int] = Field("qux", Scalar())
[error]                                                   ^
```

There are two options that I see:
- Either `Field` from caliban should be fully qualified (preferred, but difficult)
- ... or the user's supplied `Field` should be noconflicted, similar to `reservedKeywords`/`caseClassReservedFields`

I've got a branch that applies the prefix, I'd just appreciate guidance on where to integrate compilation regression tests, as this isn't effectively tested without running the classes through scalac

blast-hardcheese:
Resolves #917 

Removing FieldBuilder._ is acceptable, as `Field` seems to be the only
used member.

### DIFF
--- a/codegen-sbt/src/sbt-test/codegen/test-compile/project/schema.graphql
+++ b/codegen-sbt/src/sbt-test/codegen/test-compile/project/schema.graphql
@@ -60,6 +60,13 @@ enum Origin {
     BELT
 }
 
+# Regression test for a conflict with SelectionBuilder.Field (#917)
+enum Field {
+    name
+    nicknames
+    origin
+}
+
 # Union
 union Role = Captain | Pilot
 type Captain {
--- a/tools/src/main/scala/caliban/tools/ClientWriter.scala
+++ b/tools/src/main/scala/caliban/tools/ClientWriter.scala
@@ -89,7 +89,6 @@ object ClientWriter {
           |""".stripMargin
       else ""}${if (objects.nonEmpty || queries.nonEmpty || mutations.nonEmpty || subscriptions.nonEmpty)
         """import caliban.client.FieldBuilder._
-          |import caliban.client.SelectionBuilder._
           |""".stripMargin
       else
         ""}${if (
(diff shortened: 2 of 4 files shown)


ghostdogpr:
Just replied to you in the issue regarding testing ^^

blast-hardcheese:
@ghostdogpr Added a regression test to `scripted`

blast-hardcheese:
Thanks again

ghostdogpr (approved):
Looks good, thanks!

PR #923, opened 2021-06-13
[Client] Add the option to drop null values from input objects

### LINKED ISSUE #778: Mutation contains null values if None is provided

ghostdogpr:
Closes #778

2 options: 
- pass `dropInputNullValues` when building the request to drop all of them
- pass a custom `ArgEncoder` when building the query to customize the behavior per input object type

### DIFF
--- a/client/src/main/scala/caliban/client/Selection.scala
+++ b/client/src/main/scala/caliban/client/Selection.scala
@@ -17,12 +17,12 @@ object Selection {
   case class Directive(name: String, arguments: List[Argument[_]] = Nil) {
     def toGraphQL(
       useVariables: Boolean,
-      dropInputNullValues: Boolean,
+      dropNullInputValues: Boolean,
       variables: Map[String, (__Value, String)]
     ): (String, Map[String, (__Value, String)]) = {
       val (newArgs, newVariables) = arguments.foldLeft((List.empty[String], variables)) {
         case ((args, variables), arg) =>
-          val (arg2, variables2) = arg.toGraphQL(useVariables, dropInputNullValues, variables)
+          val (arg2, variables2) = arg.toGraphQL(useVariables, dropNullInputValues, variables)
           (arg2 :: args, variables2)
       }
       (s"@$name(${newArgs.reverse.mkString(",")})", newVariables)
--- a/client/src/main/scala/caliban/client/Selection.scala
+++ b/client/src/main/scala/caliban/client/Selection.scala
@@ -17,11 +17,12 @@ object Selection {
   case class Directive(name: String, arguments: List[Argument[_]] = Nil) {
     def toGraphQL(
       useVariables: Boolean,
+      dropInputNullValues: Boolean,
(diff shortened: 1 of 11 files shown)


client/src/main/scala/caliban/client/Selection.scala
```diff
@@ -17,11 +17,12 @@ object Selection {
   case class Directive(name: String, arguments: List[Argument[_]] = Nil) {
     def toGraphQL(
       useVariables: Boolean,
+      dropInputNullValues: Boolean,
```
paulpdaniels, line 20: Nit: I think it would read better as `dropNullInputValues`

paulpdaniels (approved):

client/src/main/scala/caliban/client/Selection.scala
ghostdogpr: Done~

PR #925, opened 2021-06-13
Fix description of temporal schemas

### LINKED ISSUE #822: Incorrectly rendered description for ZonedDateTime
Hi,

When I use a `ZonedDateTime` in my schema I get the following description for it when it is rendered:
```
"A date-time with a time-zone in the ISO-8601 calendar system using the format (ParseCaseSensitive(false)(ParseCaseSensitive(false)(Value(Year,4,10,EXCEEDS_PAD)'-'Value(MonthOfYear,2)'-'Value(DayOfMonth,2))'T'(Value(HourOfDay,2)':'Value(MinuteOfHour,2)[':'Value(SecondOfMinute,2)[Fraction(NanoOfSecond,0,9,DecimalPoint)]]))ParseStrict(false)Offset(+HH:MM:ss,'Z')ParseStrict(true))['['ParseCaseSensitive(true)ZoneRegionId()']']"
scalar ZonedDateTime
```

ghostdogpr:
Closes #822

### DIFF
--- a/core/src/main/scala/caliban/schema/Schema.scala
+++ b/core/src/main/scala/caliban/schema/Schema.scala
@@ -471,6 +471,8 @@ trait TemporalSchema {
       PureStep(format(value))
   }
 
+  lazy val sampleDate: ZonedDateTime = ZonedDateTime.ofInstant(Instant.EPOCH, ZoneId.systemDefault())
+
   def temporalSchema[A <: Temporal](name: String, description: Option[String] = None)(
     f: A => ResponseValue
   ): Schema[Any, A] = new TemporalSchema[A](name, description) {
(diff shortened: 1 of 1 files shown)


core/src/main/scala/caliban/schema/Schema.scala
```diff
@@ -500,7 +502,9 @@ trait TemporalSchema {
   def localDateTimeSchemaWithFormatter(formatter: DateTimeFormatter): Schema[Any, LocalDateTime] =
     temporalSchemaWithFormatter(
       "LocalDateTime",
-      Some(s"A date-time without a time-zone in the ISO-8601 calendar system in the format of $formatter")
+      Some(
+        s"A date-time without a time-zone in the ISO-8601 calendar system in the format of ${formatter.format(sampleDate)}"
```
paulpdaniels, line 506: Clever! 🥇

paulpdaniels (approved):

core/src/main/scala/caliban/schema/Schema.scala
```diff
@@ -471,6 +471,8 @@ trait TemporalSchema {
       PureStep(format(value))
   }
 
+  lazy val sampleDate: ZonedDateTime = ZonedDateTime.ofInstant(Instant.EPOCH, ZoneId.systemDefault())
```
paulpdaniels, line 474: I think we should use the `ZoneOffset.UTC` instead of `systemDefault()` here since it may result in different behaviors in two schemas generated on different computers.

ghostdogpr: Good point, I changed.

PR #929, opened 2021-06-16
Add support for distinguishing explicit and implicit null in ArgBuilders

stephendavidmarsh:
This PR adds support for distinguishing between explicit null, where a `null` was sent to an input, from implicit null, where no value was provided for an input field. It uses an approach inspired by zio-json's `JsonDecoder`, adding a new `buildMissing` method to `ArgBuilder` that by default calls `build` with `NullValue`. The new method is called when a field is missing. This change should be backwards compatible, with no change in behavior for existing `ArgBuilder` instances.

New `ArgBuilder` instances can override the `buildMissing` method to customize their behavior when a value is missing and provide different behavior from when `null` is supplied.

### DIFF
--- a/core/src/test/scala-2/caliban/Scala2SpecificSpec.scala
+++ b/core/src/test/scala-2/caliban/Scala2SpecificSpec.scala
@@ -50,7 +50,7 @@ object Scala2SpecificSpec extends DefaultRunnableSpec {
         case object NullNullable          extends Nullable[Nothing]
         case object MissingNullable       extends Nullable[Nothing]
 
-        implicit def nullableArgBuilder[A](implicit ev: ArgBuilder[A]) = new ArgBuilder[Nullable[A]] {
+        implicit def nullableArgBuilder[A](implicit ev: ArgBuilder[A]): ArgBuilder[Nullable[A]] = new ArgBuilder[Nullable[A]] {
           def build(input: InputValue): Either[ExecutionError, Nullable[A]] = input match {
             case NullValue => Right(NullNullable)
             case _         => ev.build(input).map(SomeNullable(_))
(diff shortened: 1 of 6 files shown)


core/src/test/scala-2/caliban/Scala2SpecificSpec.scala
```diff
@@ -39,6 +43,30 @@ object Scala2SpecificSpec extends DefaultRunnableSpec {
             """{"events":[{"organizationId":7,"title":"Frida Kahlo exhibition"}],"painters":[{"name":"Claude Monet","movement":"Impressionism"}]}"""
           )
         )
+      },
```
ghostdogpr, line 46: If the test is same for Scala 2 and Scala 3, you can simply add it in the `scala` folder, no need to duplicate it. These files are just for Scala2 or Scala3 specific features (like Scala 3 enums).

stephendavidmarsh: For some reason I though it had to be there for it to run for each version - I'll move the test.

core/src/test/scala-3/caliban/Scala3SpecificSpec.scala
```diff
@@ -1,7 +1,11 @@
 package caliban
 
+import caliban.CalibanError.ExecutionError
```
ghostdogpr, line 3: Can you clean these as well?

stephendavidmarsh: Oops, thought I had gotten all of them.

ghostdogpr (approved):
Thanks for the contribution!

PR #930, opened 2021-06-19
Permit scalarMappings to be used with enumerations

blast-hardcheese:
There's a problem in the latest release of caliban, wherein if you attempt to use `scalarMapping` to remap an `enum`, it just interpolates the fully qualified type in the name position, which is syntacticly invalid.

Given: `--scalarMappings ExampleStreamType:com.example.myserver.constants.ExampleStreamType`, I get:

```scala
sealed trait com.example.myserver.constants.ExampleStreamType extends scala.Product with scala.Serializable
object com.example.myserver.constants.ExampleStreamType {
  case object FOO extends com.example.myserver.constants.ExampleStreamType
  case object BAR extends com.example.myserver.constants.ExampleStreamType
  ...

  implicit val decoder: ScalarDecoder[com.example.myserver.constants.ExampleStreamType] = {
    case __StringValue("FOO") => Right(com.example.myserver.constants.ExampleStreamType.FOO)
    case __StringValue("BAR") => Right(com.example.myserver.constants.ExampleStreamType.BAR)
    ...
    case other => Left(DecodingError(s"Can't build ExampleStreamType from input $other"))
  }
  implicit val encoder: ArgEncoder[ExampleStreamType] = new ArgEncoder[ExampleStreamType] {
    override def encode(value: com.example.myserver.constants.ExampleStreamType): __Value = value match {
      case ExampleStreamType.FOO => __EnumValue("FOO")
      case ExampleStreamType.BAR => __EnumValue("BAR")
      ...
    }
    override def typeName: String = "com.example.myserver.constants.ExampleStreamType"
  }
}
```

This PR addresses this issue by suppressing emitting the enumeration type if the enumeration name is in the `scalarMappings` map.

### DIFF
--- a/tools/src/main/scala/caliban/tools/ClientWriter.scala
+++ b/tools/src/main/scala/caliban/tools/ClientWriter.scala
@@ -59,7 +59,10 @@ object ClientWriter {
 
     val inputs = schema.inputObjectTypeDefinitions.map(writeInputObject).mkString("\n")
 
-    val enums = schema.enumTypeDefinitions.map(writeEnum).mkString("\n")
+    val enums = schema.enumTypeDefinitions
+      .filter(e => !scalarMappings.scalarMap.exists(_.contains(e.name)))
+      .map(writeEnum)
+      .mkString("\n")
 
     val scalars = schema.scalarTypeDefinitions
       .filterNot(s => isScalarSupported(s.name))
--- a/tools/src/test/scala/caliban/tools/ClientWriterSpec.scala
+++ b/tools/src/test/scala/caliban/tools/ClientWriterSpec.scala
@@ -243,6 +243,74 @@ object Client {
           )
         )
       },
+      testM("scalar mapped enum") {
+        val schema =
+          """
+             enum Origin {
+               EARTH
+               MARS
+               BELT
+             }
+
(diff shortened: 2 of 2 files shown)


blast-hardcheese:
(I say "in the latest release of caliban", it has actually been the case since I started using it a few weeks ago. I've got a workaround by mapping caliban types to my domain types by hand during field selection, but I just got around to taking a crack at fixing this properly.)

ghostdogpr:
Yeah I think the feature was always intended to work with scalars only.

Don't you miss the `ArgEncoder` for the mapped enum then? You have to add it on your own (which is pretty easy). Anyway, no problem adding this since it's not affecting anyone who don't make use of it.

blast-hardcheese:
Yeah, I've got `ArgEncoder` and `ScalarDecoder` being brought in via `--implicit`

ghostdogpr (approved):
LGTM! Thanks!

blast-hardcheese:
Perhaps a final thought: I think of an enumeration as being a scalar, which is why I thought of doing this in the first place -- I guess you could abuse types by mapping individual "scalar" string fields to full JSON objects, if you so chose, but I'm certainly not attempting that here.

PR #933, opened 2021-06-21
sbt plugin rewrite

blast-hardcheese:
This PR adds a reimagined workflow for `caliban-codegen-sbt`, automatically generating clients from files placed in `src/main/graphql`

I included what I presume is what's required for updating the docs, if anything more needs to be done there I'm happy to follow up.


blast-hardcheese:
One curious consideration is how sbt scripted tests are able to interpolate the current `scalaBinaryVersion` into paths, so at some point in the future the `target/scala-2.12/...` in `auto-generate`'s `test` may end up being somewhat of an issue. and we may need to alter the tests to just refer to auto-generated elements at compile-time instead of direct `exists` checks.

ghostdogpr:
Actually the scripted tests don't really need to be duplicated, they slow down the build quite a lot already and we expect the same code to be generated in both cases. I think a simple one would be enough.

blast-hardcheese:
Yeah, that is quite surprising. Up from 13 minutes to 20 minutes certainly seems quite excessive. I wonder whether the penalty really is due to starting up an entire environment from scratch, so more gains would be made by merging the two existing tests into one, especially since it's more about existing of files instead of `build.sbt` settings more significantly

blast-hardcheese:
Also, maybe more broadly -- it may not be necessary to run scalafmt for these automatically generated source files, as they are never intended to be committed. At least on my machine, the startup penalty there is nontrivial, so it would benefit me personally. I don't believe it would _significantly_ impact the tests here, though maybe a little bit.

blast-hardcheese:
One breaking change here is renaming `CodegenPlugin` to `CalibanPlugin`. The motivation is that an unscoped `CodegenPlugin` doesn't make sense in a larger project with multiple, collaborative code generators.

If you feel strongly about this, I can try to add a `@deprecated` alias to maintain the old name. Alternately, I can add migration notes or something

ghostdogpr:
> Yeah, that is quite surprising. Up from 13 minutes to 20 minutes certainly seems quite excessive. I wonder whether the penalty really is due to starting up an entire environment from scratch, so more gains would be made by merging the two existing tests into one, especially since it's more about existing of files instead of `build.sbt` settings more significantly

Yeah that would be nice to merge them if you know how to. Also I think the gitlab one slow so no need to run it twice.

ghostdogpr:
> Also, maybe more broadly -- it may not be necessary to run scalafmt for these automatically generated source files, as they are never intended to be committed. At least on my machine, the startup penalty there is nontrivial, so it would benefit me personally. I don't believe it would _significantly_ impact the tests here, though maybe a little bit.

Maybe make it an optional param, that is true by default with the explicit command and false by default with the new way?

ghostdogpr:
> One breaking change here is renaming `CodegenPlugin` to `CalibanPlugin`. The motivation is that an unscoped `CodegenPlugin` doesn't make sense in a larger project with multiple, collaborative code generators.
> 
> If you feel strongly about this, I can try to add a `@deprecated` alias to maintain the old name. Alternately, I can add migration notes or something

Yeah the old name was unfortunate. Deprecating it would be nicer though, to ease the migration.

codegen-sbt/src/sbt-test/codegen/test-compile/test
```diff
@@ -1,14 +1,13 @@
...
-> calibanGenClient project/schema.graphql src/main/scala/Client.scala
-$ exists src/main/scala/Client.scala
 > calibanGenClient project/schema-to-check-name-uniqueness.graphql src/main/scala/client/ClientNameUniqueness.scala --packageName client
 $ exists src/main/scala/client/ClientNameUniqueness.scala
 > calibanGenClient project/gitlab-schema.graphql src/main/scala/client/ClientGitLab.scala --packageName client
 $ exists src/main/scala/client/ClientGitLab.scala
 
 > compile
+
+$ exists target/scala-2.12/src_managed/main/caliban-codegen-sbt/caliban/Client.scala
```
ghostdogpr, line 13: To make scripted tests faster, how about using ` --genView true` in `test-compile` and getting rid of `gen-view-test-compile`? I don't think we need both, as `--genView` is just adding more code.

vuepress/docs/docs/schema.md
```diff
@@ -227,25 +227,42 @@ addSbtPlugin("com.github.ghostdogpr" % "caliban-codegen-sbt" % "1.0.1")
...
+      )
+```
 
-By default, each Query and Mutation will be wrapped into a `zio.UIO` effect. This can be overridden by providing an alternative effect with the `--effect` option.
+The settings available on the `cs` (`CalibanSettings`) builder are:
+```
+  def scalafmtPath(path: String): CalibanSettings
+  def headers(pairs: (String,String)*): CalibanSettings
+  def packageName(name: String): CalibanSettings
+  def genView(value: Boolean): CalibanSettings
```
ghostdogpr, line 260: This setting doesn't exist on the server codegen.

vuepress/docs/docs/client.md
```diff
@@ -32,16 +32,53 @@ addSbtPlugin("com.github.ghostdogpr" % "caliban-codegen-sbt" % "1.0.1")
 
 And enable it in your `build.sbt` file:
 ```scala
-enablePlugins(CodegenPlugin)
+enablePlugins(CalibanPlugin)
 ```
 
-Then call the `calibanGenClient` sbt command.
+At this point, the `caliban` command will cause any files in `src/main/graphql` to be translated into a Caliban-generated client library. This happens automatically any time you `compile`.
```
ghostdogpr, line 38: Is this configurable? What if you want to use an API URL?

blast-hardcheese: Not possible, with how the plugin is currently set up. I think for that, adding `calibanSetting(url: URL)(...)` would work.

vuepress/docs/docs/schema.md
blast-hardcheese: I'm sorry, I don't understand this one. This plugin just runs the client codegen, it seems as though this setting is used there

blast-hardcheese:
> Maybe make it an optional param, that is true by default with the explicit command and false by default with the new way?

I'll time it to see whether it actually is slow, then open a follow-up PR

blast-hardcheese:
OK, out of curiosity I commented out scalafmt. Scripted test reduced in time from 70 seconds to 42 seconds. For auto-generated clients, we could delay scalafmt until the very end (since once scalafmt is fully initialized subsequent runs are fast), but the challenge here is `calibanGenClient` isn't really compatible with that.

blast-hardcheese:
I must admit, I don't fully understand the `url` generation workflow, so this is just my best guess based on my current experience in order to further the discussion.

What will happen is `caliban` or `compile` will trigger client generation, which will hit the URL and then the generated file will be cached until a full `clean` or just `caliban:clean`, at which point subsequent `compile` will re-fetch the schema from the URL. This strikes a reasonable balance between utility and convenience, offsetting the cost of actually fetching the URL.

vuepress/docs/docs/client.md
blast-hardcheese: OK, I've made this change. To your original question, the path is configurable via `calibanSources`

vuepress/docs/docs/schema.md
blast-hardcheese: Aaah, I see. I can revert this one

codegen-sbt/src/sbt-test/codegen/test-compile/src/main/graphql/schema.graphql
```diff
@@ -46,7 +46,7 @@ type Character {
     # union
     role: Role
 
-    # Deprecated field + comment
+    # Deperecated field + comment
```
ghostdogpr, line 49:
```suggestion
    # Deprecated field + comment
```

vuepress/docs/docs/client.md
```diff
@@ -32,16 +32,70 @@ addSbtPlugin("com.github.ghostdogpr" % "caliban-codegen-sbt" % "1.0.1")
 
 And enable it in your `build.sbt` file:
 ```scala
-enablePlugins(CodegenPlugin)
+enablePlugins(CalibanPlugin)
 ```
 
-Then call the `calibanGenClient` sbt command.
+### File generation
```
ghostdogpr, line 38: To disambiguate, how about `From a schema file` and below `From a server URL`

codegen-sbt/src/sbt-test/codegen/test-compile/test
```diff
@@ -1,14 +1,24 @@
...
 > calibanGenClient project/gitlab-schema.graphql src/main/scala/client/ClientGitLab.scala --packageName client
 $ exists src/main/scala/client/ClientGitLab.scala
 
+$ mkdir src/main/scala/genview
+$ mkdir src/main/scala/genview/client
+
+> calibanGenClient project/schema-to-check-name-uniqueness.graphql src/main/scala/genview/client/ClientNameUniqueness.scala --packageName genview.client --genView true
+$ exists src/main/scala/genview/client/ClientNameUniqueness.scala
+$ exec sh verify.sh StarshipView ./src/main/scala/genview/client/ClientNameUniqueness.scala
+> calibanGenClient project/gitlab-schema.graphql src/main/scala/genview/client/ClientGitLab.scala --packageName genview.client --genView true
```
ghostdogpr, line 17: Let's keep only this gitlab one and remove the one without `--genView`, I think it's enough.

vuepress/docs/docs/client.md
```diff
@@ -32,16 +32,70 @@ addSbtPlugin("com.github.ghostdogpr" % "caliban-codegen-sbt" % "1.0.1")
...
+  def scalarMapping(mapping: (String,String)*): CalibanSettings
+  def imports(values: String*): CalibanSettings
+
+  // Only defined for `url` settings, for use in supplying extra headers when fetching the schema itself
+  def headers(pairs: (String,String)*): CalibanSettings
+```
+
+### `calibanGenClient`
+
+Should you prefer or otherwise need the previous style of manual client codegen, supplied by the `calibanGenClient` function, that command is still available. Documentation for that command is as follows:
```
ghostdogpr, line 90: Maybe no need to refer to the "previous style" which is unclear for newcomers, but instead just say something like "If you prefer to generate the client explicitly rather than automatically, ..."

vuepress/docs/docs/client.md
```diff
@@ -32,16 +32,70 @@ addSbtPlugin("com.github.ghostdogpr" % "caliban-codegen-sbt" % "1.0.1")
...
+
 ```scala
 calibanGenClient schemaPath outputPath [--scalafmtPath path] [--headers name:value,name2:value2] [--genView true|false] [--scalarMappings gqlType:f.q.d.n.Type,gqlType2:f.q.d.n.Type2] [--imports a.b.c._,c.d.E]
 
 calibanGenClient project/schema.graphql src/main/client/Client.scala --genView true  
 ```
 This command will generate a Scala file in `outputPath` containing helper functions for all the types defined in the provided GraphQL schema defined at `schemaPath`.
+If you need to disable generating clients from `src/main/graphql`, please include `Compile / caliban / calibanGenerator := Seq.empty` in your project settings.
```
ghostdogpr, line 98: If you don't have a `src/main/graphql` folder or if it's empty, nothing happens, is that right?

vuepress/docs/docs/client.md
```diff
@@ -32,16 +32,70 @@ addSbtPlugin("com.github.ghostdogpr" % "caliban-codegen-sbt" % "1.0.1")
...
+          cs.clientName("ExampleServiceClient")
+            .packageName("com.example.graphql.client")
+      )
+```
+
+### `calibanSetting` config parameters
+
+The settings available on the `cs` (`CalibanSettings`) builder are:
+```
+  def scalafmtPath(path: String): CalibanSettings
```
ghostdogpr, line 77: Would be nice to add a small comment on each of them. They're explained in the `calibanGenClient` which is secondary, that's bit weird.

ghostdogpr (commented):
Looks great! I left a few remarks and questions, mostly around docs.

vuepress/docs/docs/client.md
blast-hardcheese: Correct

ghostdogpr (approved):
Looking good, thanks!

I will merge but prevent the docs generation until the next release.

PR #934, opened 2021-06-23
Exposing the set of members of emitted enumerations

blast-hardcheese:
This is needed for more dynamic mapping of known values in enumerations. Example:

```scala
val (notInterested, interested) = categories.values.partition(user.preferredCateogry.contains _)
```

Again, I'm still onboarding onto GraphQL workflows, so if this PR doesn't make sense, thank you for your patience.

### DIFF
--- a/tools/src/main/scala/caliban/tools/ClientWriter.scala
+++ b/tools/src/main/scala/caliban/tools/ClientWriter.scala
@@ -380,6 +380,10 @@ object ClientWriter {
       .map(v => s"""case ${typedef.name}.${safeEnumValue(v.enumValue)} => __EnumValue("${v.enumValue}")""")
       .mkString("\n")}
           }
+
+          val values: Vector[${enumName}] = Vector(${typedef.enumValuesDefinition
+      .map(v => safeEnumValue(v.enumValue))
+      .mkString(", ")})
         }
        """
   }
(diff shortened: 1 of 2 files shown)


tools/src/main/scala/caliban/tools/ClientWriter.scala
```diff
@@ -380,6 +380,10 @@ object ClientWriter {
       .map(v => s"""case ${typedef.name}.${safeEnumValue(v.enumValue)} => __EnumValue("${v.enumValue}")""")
       .mkString("\n")}
           }
+
+          val values: Vector[${enumName}] = Vector(${typedef.enumValuesDefinition
```
ghostdogpr, line 384: How about a `Set` instead of `Vector`?

blast-hardcheese: It's possible to go from `Vector` to `Set` but not the other way around, I believe this is why existing enumeration implementations expose their members as sequences (scala enumerations and scalapb being two examples)

blast-hardcheese: I agree with you that people _shouldn't_ depend on the ordering of members of an enumeration, but preventing them from doing so at all is just waiting for someone to come and ask for that feature, then breaking all consumers who are already using whatever implementation we decide on here.

ghostdogpr: Yeah, I don't think it's right to expect any kind of order but I don't have strong feelings about it. Fine to merge like this.

PR #936, opened 2021-06-23
Exposing a distinct "value" parameter for enumeration members, to avoid relying on default toString()

blast-hardcheese:
- `def value: String` on containing enumeration could be a `val value: String` if you prefer. I think having a weak reference there is better for some reason, but I can't remember why or if this is true.
- This is motivated by having been bit by refactors that change a member from a discrete type to an `Option[A]` or `List[A]` or otherwise, but relying on the default `.toString()` doesn't flag this as an error. This also provides an alternative for those using wartremover's `ToString` wart, which prevents relying on the default `toString` as well.

This was able to be merged without conflict after applying #934, another reason for not changing too much about emitted classes


ghostdogpr (approved):
LGTM!

PR #939, opened 2021-06-26
Add metadata federation resolver

paulpdaniels:
Adds a variation to the federation `EntityResolver` that allows the user to access the `Field` meta info for the top-level entity.


ghostdogpr (approved):
Nice!

Issue #218, opened 2020-02-15
Derive operations schema from traits


Krever:
Right now deriving schema for operations (queries, mutations, subscriptions) requires defining a case class in form of 
```
case class MyOpInput(p1: Int, p2: String)
case class Foo(myOp: MyOpInput => String)
```
which is slightly inconvenient because you need to define `MyOpInput` for each operation.
Instead, we could derive the schema from trait like this:
```
trait Foo {
  def myOp(p1: Int, p2: String): String
}
```
so we can scrap this one layer of boilerplate.
I think there are no technical limitations in general but it would require custom macro (or even macro annotation) instead of magnolia-based derivation.

ghostdogpr:
The idea is nice but as you mentioned, Magnolia doesn’t allow doing this. If anyone has macro skills (I have zero) and want to give it a try, let us know 😄

Krever:
Unfortunately, the amount of my macro experience is just the same. In case either of us or anyone else would like to give it a shot in the future I link below some examples of similar macros

1. cats-tagless derives FunctorK, so something that is able to transform all methods returning `F[_]` into methods returning `G[_]` when provided `F ~> G`: https://github.com/typelevel/cats-tagless/blob/master/macros/src/main/scala/cats/tagless/DeriveMacros.scala
2. autowire provides an RPC layer based on traits so something rather similar to what Caliban needs: https://github.com/lihaoyi/autowire/blob/master/autowire/shared/src/main/scala/autowire/Macros.scala

I understand nothing from this code though :)

argson67:
Ask and ye shall receive :-)

ghostdogpr:
Addressed by https://github.com/zio/caliban-deriving

PR #940, opened 2021-06-29
Derivation macros

vigoo:
Continuation of https://github.com/ghostdogpr/caliban/pull/247 
Implements https://github.com/ghostdogpr/caliban/issues/218


ghostdogpr:
Hi @vigoo !
The master branch has quite a lot of changes compared to yours.
I am a little skeptical regarding this one, because it adds a lot of macro code which is difficult to maintain (and need to be rewritten for Scala 3), for a benefit that is not super obvious 🤔  What's your thinking behind this?

vigoo:
Hi @ghostdogpr !

(I actually did not want to open the PR here yet only in the fork ;)

This feature is important to a customer and I was asked to try to revive it. So my plan for today/tomorrow is to:

- Make the branch up-to-date
- See what's missing and try to make it feature complete
- See if it can be done with Scala 3 macros too and try to reach feature parity

As I only forked it so far and did not take much look I can't tell more yet.

ghostdogpr:
Gotcha! We can discuss later where this should live then.

vigoo:
### Status update

The Scala 2.x implementation seems to be quite usable. I haven't started the Scala 3 implementation yet but it looks completely doable.

I kept the original behavior for product types and tried to address the comments from the original issue. The version in this PR currently supports:

**Product types**
For any `case class` the macro generates a GraphQL `type`:
- taking all the case class parameters as input fields
- all public paramterless methods and fields are exposed as fields
- all public methods with parameters are exposed as functions

**Sum types**
For sum types (`sealed trait` with a set of case class / case objects implementations) if the base trait contains any methods, it will become a GraphQL `interface` and each constuctor a `type`. 
Everything else works exactly as for product types as mentioned above.

If the base trait has no members, but the constructors have parameters, it becomes a GraphQL `union`. Members of constructors are still traited the same way as mentioned above.

If all the constructors are `case object`s then it becomes a GraphQL `enum`.

**Arbitrary traits**
If the derivation is invoked on a simple `trait` (not sealed) then it works as if it were a concrete parameterless case class. This means a schema derived for an arbitrary trait can be used to serve any implementation of it.

**Environment**
I added support for the environment (`-R`) parameter of `ZIO` and `ZQuery` members. The environment must be defined by the caller of the derivation and it is used for all implicit `Schema` search inside the derivation. This is in sync with how the magnolia-based derivation works.

Addition to this the macro also collects the environments it finds and if the provided `R` is not a subtype of the detected one it emits a warning.

**Coexisting with the magnolia based auto-derivation**
The macro summons the implicit `Schema`s for the types it finds, which by default is satisfied by the existing magnolia based auto-derivation. Replacing the default derivation with this new derivation macro must be done type-by-type with a semi-auto derivation style, invoking the macro by hand. 

I removed the annotation macro because it is not supported in Scala 3.

For `R=Any` schemas putting the derivation call in the companion object correctly replaces the schema for the given type and works together with the magnolia derivation.

For other environments, the derived implicits must be placed directly in the object extending `GenericSchema[R]`; then the `RootResolver` invoked from within the object will use the proper instances as desired.

I added a full example in the `examples` module to demonstrate this.

**Missing**
- Scala 3 version of the macro
- Mutations and subscriptions are not supported

**Placement**
The `derivation` module is completely independent so if we don't want to have it here it could be released outside the `caliban` project. The only exception is that I moved the single new attribute to `core` as it was recommended in the original PR. (but that can be undone if needed)

build.sbt
```diff
@@ -438,6 +440,31 @@ lazy val federation = project
     scalacOptions += "-Ywarn-unused:-locals"
   )
 
+lazy val derivation = project
```
ghostdogpr, line 443: To run the tests for it, it needs to be added to .circleci/config.yml

derivation/src/main/scala-2/caliban/derivation/DerivationMacros.scala
```diff
@@ -0,0 +1,423 @@
...
+
+    val subclassInOut =
+      subclasses.map { subclass =>
+        val inputs  = inputFields(subclass.typeSignature)
+        val outputs = outputFields(subclass.typeSignature)
+
+        (subclass, (inputs, outputs))
+      }.toMap
+
+    val isEnum      = outputs.isEmpty && subclassInOut.forall { case (_, (in, out)) => in.isEmpty && out.isEmpty }
```
ghostdogpr, line 356: Recently was added the ability to use union instead of enum using the `GQLUnion` (and the `fixEmptyUnionObject` dirty trick).

vigoo:
We decided to keep this in a separate repo for now

vigoo:
Continuation: https://github.com/zio/caliban-deriving

PR #941, opened 2021-06-29
Handle GQLInterface and GQLUnion annotations on enum-like traits

### LINKED ISSUE #928: Force a sealed trait to not be an enum
Currently, if all children of a sealed trait are objects, it is rendered as an enum in the schema. It would be nice to be able to add an annotation to render it as a `union`. Otherwise, if the trait is later extended with another class that has some fields, the schema changes in an incompatible way.

rtimush:
Closes #928

### DIFF
--- a/core/src/main/scala-3/caliban/schema/SchemaDerivation.scala
+++ b/core/src/main/scala-3/caliban/schema/SchemaDerivation.scala
@@ -48,9 +48,18 @@ trait SchemaDerivation[R] {
             true
           case _ => false
         }
+        lazy val isInterface = annotations.exists {
+          case GQLInterface() => true
+          case _              => false
+        }
+        lazy val isUnion = annotations.exists {
+          case GQLUnion() => true
+          case _          => false
+        }
+
         new Schema[R, A] {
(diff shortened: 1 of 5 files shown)


core/src/main/scala-3/caliban/schema/SchemaDerivation.scala
```diff
@@ -48,9 +48,18 @@ trait SchemaDerivation[R] {
             true
           case _ => false
         }
+        lazy val isInterface = annotations.exists { case GQLInterface() =>
+          true
```
ghostdogpr, line 52: Is it Scalafmt that forces this weird formatting? If you format it properly by hand, does the fmt check fail?

I should upgrade scalafmt to the latest version, not sure if it will help.

ghostdogpr (approved):
Looks good, just a question about weird formatting with Scala 3

core/src/main/scala-3/caliban/schema/SchemaDerivation.scala
rtimush: Updated. I just thought that scalafmt would fix the formatting and forgot to check if it really did.

ghostdogpr (approved):

PR #942, opened 2021-06-30
Fix string escaping problem

mingyuchi:
since there are quite a bit of works needed to escape string arguments so that it can be used in graphql queries + mutations, I made it to use circe's json string encoders and added tests for that.
It's only done for the client side only for now since there are no circe dependency in the server side. In the future, we can probably make our own escapers


ghostdogpr:
Great, thanks!

PR #948, opened 2021-07-05
Allow union types to reuse the same member

### LINKED ISSUE #826: Validation Error when the same interface is used in multiple unions
Hello here,

I'm trying to do something and I end up with

ValidationError Error: Type 'ForbiddenRejection' is defined multiple times (OBJECT in io....schema.Types.InitiateCreditTransfersPayload.ForbiddenRejection, OBJECT in io....schema.Types.RefundPayload.ForbiddenRejection).

This error happens when I create the interpreter `federate(partnerGql).interpreter` from a GraphQL generated with caliban-codegen-sbt.

The reason with this happens is that the generation generated two sealed trait with the same case class name.

A simplified version of the graphql schema looks like this:

```graphql
interface Rejection {
  message: String!
}

type ForbiddenRejection implements Rejection {
  message: String!
}

type Query {
  query(id: ID!): QueryPayload!
  query2(id: ID!): Query2Payload!
}

type QuerySuccessPayload {
  id: ID!
}

type Query2SuccessPayload {
  id: ID!
}

union QueryPayload = QuerySuccessPayload | ForbiddenRejection
union Query2Payload = Query2SuccessPayload | ForbiddenRejection
``` 

For now the only solution I have is to never do this.

Am I doing something wrong ?

AlixBa:
closes #826

### DIFF
--- a/tools/src/main/scala/caliban/tools/SchemaWriter.scala
+++ b/tools/src/main/scala/caliban/tools/SchemaWriter.scala
@@ -24,7 +24,7 @@ object SchemaWriter {
       .map(union => (union, union.memberTypes.flatMap(schema.objectTypeDefinition)))
       .toMap
 
-    val unions = unionTypes.map { case (union, objects) => writeUnion(union, objects) }.mkString("\n")
+    val unions = writeUnions(unionTypes)
 
     val objects = schema.objectTypeDefinitions
       .filterNot(obj =>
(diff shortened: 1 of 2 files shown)


ghostdogpr (commented):
It looks great! The only thing I'm wondering about is: should we keep it inside an object when there is only 1? That way it would limit breaking changes with existing code and it makes things a bit cleaner in the normal case. What do you think?

AlixBa:
Tried it.
I've kept the same old behavior for "not reused union members" and got "reused union members" out of the Union object using multiple sealed trait extend.

ghostdogpr (approved):
Thanks!

PR #951, opened 2021-07-06
Allow for Abstract Effect Type in Schema Gen

LaurenceWarne:
Hi! This PR allows the effect type of generated the generated `Queries` and `Mutations` classes to be abstract, ie:

```scala
case class Mutations[F[_]](
    foo: FooArgs => F[Foo],
    bar: BarArgs => F[Bar]
)
```

To do this, you can use the `--abstractEffectType` argument (hopefully I've added doc in the right place for this).

Currently the `effect` parameter will be used as the abstract type value so to get the above example you would have to `--abstractEffectType true --effect F`, though I'm unsure if this is best since `effect` defaults to the concrete type `zio.UIO` (so you would get `Mutations[UIO[_]]` without specifying `F` :fearful: ).  I was thinking maybe ignore the value of `effect` if `--abstractEffectType` is set and always use `F`, what do you think?

Let me know what you think... or if this is just a terrible idea :slightly_smiling_face:

### DIFF
--- a/tools/src/main/scala/caliban/tools/SchemaWriter.scala
+++ b/tools/src/main/scala/caliban/tools/SchemaWriter.scala
@@ -12,7 +12,8 @@ object SchemaWriter {
     schema: Document,
     packageName: Option[String] = None,
     effect: String = "zio.UIO",
-    imports: Option[List[String]] = None
+    imports: Option[List[String]] = None,
+    isEffectTypeAbstract: Boolean = false
   )(implicit scalarMappings: ScalarMappings): String = {
     val schemaDef = schema.schemaDefinition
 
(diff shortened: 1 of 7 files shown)


tools/src/main/scala/caliban/tools/SchemaWriter.scala
```diff
@@ -103,17 +104,19 @@ object SchemaWriter {
...
-       |${writeDescription(op.description)}case class ${op.name}(
+       |${writeDescription(op.description)}case class ${op.name}$typeParamOrEmpty(
        |${op.fields.map(c => writeRootField(c, op, effect)).mkString(",\n")}
        |)""".stripMargin
 
+  }
   def writeSubscriptionField(field: FieldDefinition, od: ObjectTypeDefinition)(implicit
     scalarMappings: ScalarMappings
-  ): String =
+  ): String                                                       =
```
LaurenceWarne, line 119: `sbt fmt` did this :D

LaurenceWarne:
Unsure why CI is failing due to a time out, `sbt ++2.12.13! codegenSbt/scripted` passes locally for me?

ghostdogpr:
For some reason the 2.12 compilation of `core` times out (10 min) since a few days ago, not sure why. It takes a couple minutes max locally, weird. I'll review your PR tomorrow and restart the CI job if needed.

ghostdogpr:
I merged https://github.com/ghostdogpr/caliban/pull/952 which seems to improve it, can you rebase so that you have it too?

LaurenceWarne:
Done :+1:

ghostdogpr:
Looks good! How about using `F` as a default when `--abstractEffectType` is true and `--effect` is not given? And keep `UIO` when it's false or missing?

LaurenceWarne:
> Looks good! How about using `F` as a default when `--abstractEffectType` is true and `--effect` is not given? And keep `UIO` when it's false or missing?

Yeah that makes sense, I've made a few changes for this :+1: 

`Codegen.scala` does not appear to have any unit tests, so I've not added a test for `if abstractEffectType is true set effect to F else set to UIO`, though I've tested the new behaviour on a local project - do you think this is a problem?

ghostdogpr (approved):
Great, thanks for the contribution!

PR #953, opened 2021-07-06
Clarify when deriving schemas explicitly in the context of having ZIO environment

ithinkicancode:


ghostdogpr:
Thanks! I will merge it when I do the release so that I don't trigger the docs update which contain stuff about unreleased features.

ghostdogpr (approved):

PR #956, opened 2021-07-08
Add Play framework-related regex to codegen tool

### LINKED ISSUE #950: Client Codegen doesn't handle the name properly with play projects.
Currently the client codegen uses a regex based system to extract the name of the client from the file name (or falls back to `Client` if it can't find it). The regex however relies on the presence of `/scala` in the path name which breaks with `play` (and any custom directory structures).

Work around is to make your client projects part of a separate module if using play (one that follows the correct directory structure). But given that play is pretty widely used it seems like we could support it out of the box since it is just about adding a `|play` to the regex.

Issue illustrated here: https://scastie.scala-lang.org/ucJFJg63QHqsgHHDl7AibQ

yonlugoh:
Expanded the regex which previously accepted only path names with `/scala` to also accept `|play.*` or `|app` (The issue enclosed suggested to only add `play` but the scastie example provided has `/app` in the path name instead of `play`)

closes #950

### DIFF
--- a/codegen-sbt/src/sbt-test/codegen/test-compile/test
+++ b/codegen-sbt/src/sbt-test/codegen/test-compile/test
@@ -13,11 +13,11 @@ $ exists src/main/scala/client/ClientNameUniqueness.scala
 
 > calibanGenClient project/schema-to-check-name-uniqueness.graphql app/com/caliban/client/ClientPlayFramework.scala --packageName client 
 $ exists app/com/caliban/client/ClientPlayFramework.scala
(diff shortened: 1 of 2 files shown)


ghostdogpr (approved):
LGTM, thanks! @paulpdaniels wanna have a look?

paulpdaniels:
Yeah I mistyped. `/app` is the top level route for play apps. So I don't believe we need the "play" variant just the "app" one.

paulpdaniels (approved):

PR #959, opened 2021-07-08
Automatically regenerate when configs change

### LINKED ISSUE #958: sbt plugin does not automatically evict cached files when CalibanSettings changes
As a workaround `caliban:clean` will clean _just_ the caliban-generated files, but I've learned a little more about how sbt caching works and am working on a fix.

blast-hardcheese:
Resolves #958

- Takes all available settings
- Renders them to equivalent CLI args
- Uses that as a tracked input for whether to run the codegen

### DIFF
--- a/codegen-sbt/src/main/scala/caliban/codegen/CalibanSourceGenerator.scala
+++ b/codegen-sbt/src/main/scala/caliban/codegen/CalibanSourceGenerator.scala
@@ -22,6 +22,51 @@ object CalibanSourceGenerator {
     interimPath.getParent.resolve(scalaName).toFile
   }
 
+  def collectSettingsFor(fileSettings: Seq[CalibanFileSettings], source: File): CalibanFileSettings =
+    // Supply a default packageName.
+    // If we do not, `src_managed.main.caliban-codegen-sbt` will be used,
+    // which is not only terrible, but invalid.
+    CalibanSettings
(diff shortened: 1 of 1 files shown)


codegen-sbt/src/main/scala/caliban/codegen/CalibanSourceGenerator.scala
```diff
@@ -11,6 +11,27 @@ object CalibanSourceGenerator {
   import zio._
   import zio.console._
 
+  import sjsonnew.{ :*:, LList, LNil }
```
ghostdogpr, line 14: Is this a dependency of sbt?

ghostdogpr (approved):
Well, I will trust you on this 😆

codegen-sbt/src/main/scala/caliban/codegen/CalibanSourceGenerator.scala
blast-hardcheese: Yeah, built in already. There is one downside about this approach, where if sbt breaks bincompat, it means a new release will need to be cut (or newer versions of the plugin will not be compatible with some older versions of sbt).

I haven't found a good solution for this, and this seems to be the only way to implement this functionality, so I'm at a loss other than to just raise awareness about it.

PR #961, opened 2021-07-09
Support view generation for big types.

alexdupre:
i.e. types with more than 22 fields / function args / selection args.

The limit has been overcome by encapsulating the arguments inside new
case classes (that haven't the 22 function arguments limit) when needed.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -172,7 +172,7 @@ lazy val codegenSbt = project
   .settings(
     scriptedLaunchOpts := {
       scriptedLaunchOpts.value ++
-        Seq("-Xmx1024M", "-Dplugin.version=" + version.value)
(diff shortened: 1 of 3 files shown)


ghostdogpr:
Can you check the scripted tests? Some code that was previously compiling is not compiling anymore. Also it would be nice to add a new scripted test for this use case (since it verifies that it actually compiles).

alexdupre:
As I wrote in discord, the scripted test is failing for a memory limit of the CI environment. It runs fine on my machine, so I guess it needs a higher `-Xss` parameter. I can try increasing it, do you know what's the current default value?

The fact that now is failing while it wasn't before is because the scripted tests already exploit this new feature, so a new test is not needed there.

ghostdogpr:
Gitlab is good to test the most extreme cases 😆

ghostdogpr:
I'll review everything in #964 so let's close this one

PR #963, opened 2021-07-09
Add support to split the single client file into multiple files.

alexdupre:
Two new options have been added:
--splitFiles to split the client code generation into multiple files
--enableFmt  to disable formatting the generated files with scalafmt

The huge gitlab schema compiled in the scripted tests has been converted
to use these new options.


docs/docs/client.html
```diff
@@ -64,10 +64,12 @@
   def genView(value: Boolean): CalibanSettings                  // Provide a case class and helper method to select all fields on an object (default: false)
   def scalarMapping(mapping: (String,String)*): CalibanSettings // A mapping from GraphQL scalar types to JVM types, as unknown scalar types are represented as String by default.
   def imports(values: String*): CalibanSettings                 // A list of imports to be added to the top of a generated client
+  def splitFiles(value: Boolean): CalibanSettings               // Split single client object into multiple files (default: false)
```
ghostdogpr, line 67: Ah, to modify the docs it's not directly in the html, look at the .md files in the `vuepress` folder.

alexdupre: Fixed, thanks

blast-hardcheese (changes requested):
Needs https://github.com/alexdupre/caliban/pull/1 otherwise the sbt plugin will error if `splitFiles(true)` is used.

One question as well

ghostdogpr:
I'll review everything in #964 so let's close this one

PR #965, opened 2021-07-09
Update Scala 3 version

ghostdogpr:


ghostdogpr:
```
[error] -- Error: /Users/pierre/GIT/caliban/core/src/main/scala-3/caliban/introspection/IntrospectionDerivation.scala:11:66 
[error] 11 |  implicit lazy val typeSchema: Schema[Any, __Type]             = Schema.derived
[error]    |                                                                  ^^^^^^^^^^^^^^
[error]    |           given instance gen is declared as erased, but is in fact used
```

Weird. We're not even using `erased` anywhere.

PR #976, opened 2021-07-22
Support growing enums in the client code generator

rtimush:
In GraphQL, adding new values to Enums is not considered a breaking change. It would be nice to be able to handle such cases of schema evolution in the generated clients. This pull request adds a new configuration option to the client code generator (`--extensibleEnums`). When this option is specified, all unknown enum values will be mapped to an `__Unknown` case class, instead of producing a DecodingError.

There is at least one GraphQL library that uses the same approach: https://opensource.expediagroup.com/graphql-kotlin/docs/client/client-features#default-enum-values.


ghostdogpr (approved):
Awesome! 

Could you add a word in the docs (in the vuepress folder). It's better in a separate PR so that I merge it after the next release.

PR #985, opened 2021-07-30
Interpolate additionalImports on splitFiles(true)

blast-hardcheese:
`--imports` was not considered when generating classes when using `--splitFiles`


ghostdogpr (approved):
Nice 👍

PR #989, opened 2021-07-31
Add union type redirect

paulpdaniels:
Adds the ability to redirect a type union to to a different type. This allows you to reuse types between unions like:

```scala
sealed trait MyUnion
object MyUnion {
  @GQLValueType
  case class Redirect(value: UserProfile) extends MyUnion
  case class Error(msg: String) extends MyUnion
}

// Doesn't need to be part of the sealed trait anymore so can be reused for other types.
case class UserProfile(id: String, email: String)
```

producing a graphql union with the `Redirect` type removed and the `UserProfile` treated as part of the union instead.

```graphql
type UserProfile {
  id: String!
  email: String!
}

type Error {
  msg: String!
}


union MyUnion = UserProfile | Error

```

### DIFF
--- a/core/src/main/scala-2/caliban/schema/SchemaDerivation.scala
+++ b/core/src/main/scala-2/caliban/schema/SchemaDerivation.scala
@@ -22,9 +22,16 @@ trait SchemaDerivation[R] extends LowPriorityDerivedSchema {
 
   type Typeclass[T] = Schema[R, T]
 
+  def isValueType[T](ctx: ReadOnlyCaseClass[Typeclass, T]): Boolean =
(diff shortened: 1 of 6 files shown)


ghostdogpr:
Sorry, got a couple hectic days at work, will try to review tomorrow!

ghostdogpr (commented):
Looks good!

Can you add some docs (I don't mind if you do it in another PR so I can merge it when I release)?

Also, can you fix the example in the PR, it doesn't seem complete (no annotation and the type names don't match). Could be useful if someone stumble upon it.

paulpdaniels:
@ghostdogpr done!

PR #992, opened 2021-08-04
add clientName option to calibanGenClient

pbuszka:
Added a new option which defines name of the scala object used for generated client. The current file name semantics is not reliable when generating client to an external location which has no [scala|play|app] in the directory name. packageName works well for it's part but if the file path does not contain [scala|play|app] it will always fallback use Client name even though the file name is Foo.scala.

An explciit option makes it more reliable and independent of the specific file path.

I chose clientName to stay consistent as it is already used in CalibanSettings

### DIFF
--- a/codegen-sbt/src/main/scala/caliban/codegen/CalibanCli.scala
+++ b/codegen-sbt/src/main/scala/caliban/codegen/CalibanCli.scala
@@ -69,7 +69,7 @@ object CalibanCli {
 
   private val genClientHelpMsg =
     s"""
-       |calibanGenClient schemaPath outputPath [--scalafmtPath path] [--headers name:value,name2:value2] [--packageName name] [--genView true|false]
+       |calibanGenClient schemaPath outputPath [--scalafmtPath path] [--headers name:value,name2:value2] [--packageName name] [--clientName name] [--genView true|false]
        |
        |This command will create a Scala file in `outputPath` containing client code for all the
        |typed defined in the provided GraphQL schema defined at `schemaPath`. Instead of a path,
--- a/tools/src/main/scala/caliban/tools/Codegen.scala
+++ b/tools/src/main/scala/caliban/tools/Codegen.scala
@@ -21,7 +21,7 @@ object Codegen {
   ): Task[List[File]] = {
     val s                  = ".*/[scala|play.*|app][^/]*/(.*)/(.*).scala".r.findFirstMatchIn(arguments.toPath)
     val packageName        = arguments.packageName.orElse(s.map(_.group(1).split("/").mkString(".")))
-    val objectName         = s.map(_.group(2)).getOrElse("Client")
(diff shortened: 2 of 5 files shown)


ghostdogpr (commented):
Looks good! How about modifying the scripted tests to verify this? https://github.com/ghostdogpr/caliban/tree/master/codegen-sbt

ghostdogpr:
By the way, there is a `clientName` option when using `CalibanSourceGenerator` (in `CalibanSettings`), it would be nice to pass this down when those settings are converted to `Options` (that's in `CalibanSourceGenerator.renderArgs`). cc @blast-hardcheese

blast-hardcheese:
`--clientName` conflicts with `--splitFiles` (`clientName` is not used if `--splitFiles true`), imo we should emit an error if both of these are specified

blast-hardcheese:
> By the way, there is a `clientName` option when using `CalibanSourceGenerator` (in `CalibanSettings`), it would be nice to pass this down when those settings are converted to `Options` (that's in `CalibanSourceGenerator.renderArgs`). cc @blast-hardcheese

as far as passing through `--clientName` in codegen-sbt, I think this is all that's required:

```diff
diff --git a/codegen-sbt/src/main/scala/caliban/codegen/CalibanSourceGenerator.scala b/codegen-sbt/src/main/scala/caliban/codegen/CalibanSourceGenerator.scala
index 140f1ce..786c698 100644
--- a/codegen-sbt/src/main/scala/caliban/codegen/CalibanSourceGenerator.scala
+++ b/codegen-sbt/src/main/scala/caliban/codegen/CalibanSourceGenerator.scala
@@ -71,6 +71,7 @@ object CalibanSourceGenerator {
       if (values.nonEmpty) {
         opt :: values.mkString(",") :: Nil
       } else Nil
+    val clientName                                                         = singleOpt("--clientName", settings.clientName)
     val scalafmtPath                                                       = singleOpt("--scalafmtPath", settings.scalafmtPath)
     val headers                                                            = pairList("--headers", settings.headers)
     val packageName                                                        = singleOpt(
@@ -96,7 +97,7 @@ object CalibanSourceGenerator {
       "--extensibleEnums",
       settings.extensibleEnums.map(_.toString())
     ) // NB: Presuming zio-config can read toString'd booleans
-    scalafmtPath ++ headers ++ packageName ++ genView ++ scalarMappings ++ imports ++ splitFiles ++ enableFmt ++ extensibleEnums
+    clientName ++ scalafmtPath ++ headers ++ packageName ++ genView ++ scalarMappings ++ imports ++ splitFiles ++ enableFmt ++ extensibleEnums
   }
 
   def apply(
```

blast-hardcheese:
at some point it would be nice to not have to round-trip through CLI args, I just haven't had the time to remove the boilerplate.

ghostdogpr:
Merged within https://github.com/ghostdogpr/caliban/pull/1003

PR #1002, opened 2021-08-08
Adding cli driver

blast-hardcheese:
Looking for feedback on if this is a useful development tool, or if it's already provided by something else I haven't seen yet.

Usage:

```
cli ./codegen-sbt/src/sbt-test/codegen/test-compile/src/main/graphql/schema.graphql /tmp/foop/src/main/scala/example/Schema.scala --splitFiles true
```

I also wasn't able to find where in zio-config something like `--help` would be able to be registered, presumably that would only be satisfied by something like `zio-cli` once that reaches maturity. For this reason, I could drop the second commit that exposes underlying errors, as it's unrelated to the primary goal of providing a `cli` driver at the sbt console to aide in iteration.

### DIFF
--- a/cli/src/main/scala/caliban/cli/Launch.scala
+++ b/cli/src/main/scala/caliban/cli/Launch.scala
@@ -2,13 +2,34 @@ package caliban.codegen
 
 import _root_.caliban.tools.Codegen.GenType
 import _root_.caliban.tools._
+import java.nio.file.{ Path, Paths }
 import zio._
 import zio.config.{ read, ConfigDescriptor, ConfigSource, ReadError }
 import zio.console._
 
+case class MissingGenType(args: List[String]) extends NoSuchElementException {
+  override def toString() = s"MissingGenType(${args.mkString(" ")})"
+}
+
 object Launch extends App {
+  val extractType: List[String] => IO[NoSuchElementException, (GenType, List[String])] = {
+    case "client" +: rest => UIO((GenType.Client, rest))
+    case "schema" +: rest => UIO((GenType.Schema, rest))
+    case args             => ZIO.fail(MissingGenType(args))
+  }
+
+  def mkDirs(path: Path): Task[Unit] =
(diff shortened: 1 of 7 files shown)


ghostdogpr:
Not sure if I would use it but why not, it doesn't hurt 😄

cli/src/main/scala/caliban/cli/Launch.scala
```diff
@@ -0,0 +1,14 @@
...
+import _root_.caliban.tools.Codegen.GenType
+import _root_.caliban.tools._
+import zio._
+import zio.config.{ read, ConfigDescriptor, ConfigSource, ReadError }
+import zio.console._
+
+object Launch extends App {
+  def run(args: List[String]) = (for {
+    opts  <- ZIO.fromEither(Options.fromArgs(args))
+    files <- Codegen.generate(opts, GenType.Client)
```
paulpdaniels, line 12: So this wouldn't support server gen?

blast-hardcheese: Resolved, thanks for taking a look.

```
sbt> cli {client|schema} [opt [opt...]]
```

cli/src/main/scala/caliban/cli/Launch.scala
```diff
@@ -0,0 +1,35 @@
...
+
+object Launch extends App {
+  val extractType: List[String] => IO[NoSuchElementException, (GenType, List[String])] = {
+    case "client" +: rest => UIO((GenType.Client, rest))
+    case "schema" +: rest => UIO((GenType.Schema, rest))
+    case args             => ZIO.fail(MissingGenType(args))
+  }
+
+  def mkDirs(path: Path): Task[Unit] =
+    Task(path.toFile.mkdirs).as(())
```
paulpdaniels, line 22: Hint: `.unit`

paulpdaniels (approved):

PR #1003, opened 2021-08-08
Caliban settings directly to options

blast-hardcheese:
Basing this on #992 to reduce merge conflicts, but just implementing [my comment from that issue](https://github.com/ghostdogpr/caliban/pull/992#issuecomment-894833065) wherein it would be more ideal to just emit the `Options` direction out of `CalibanSettings` instead of round-tripping through the CLI parser unnecessarily.


ghostdogpr (approved):
Looks good, thanks!

PR #1011, opened 2021-08-23
Support Short schemas and args

Fluxx:
I ran into a situation where I had a `Short` type that I needed a schema for, but Caliban didn't have one. This PR adds one which is implemented as a Scalar, along with an `ArgBuilder` for it. Few notes:

* My editor automatically trims training whitespace on save, so there are some whitespace changes. I am happy to remove them if you would like.
* I didn't see any specific tests for schemas so I didn't add any, but if you point me in the right direction I can add them.
* Let me know how to update docs. I think I only need to add the row to the table on [the schema page](https://ghostdogpr.github.io/caliban/docs/schema.html#enums-unions-interfaces), but I wasn't sure if the docs HTML checked in was automatically generated or not.

### DIFF
--- a/client/src/main/scala/caliban/client/ArgEncoder.scala
+++ b/client/src/main/scala/caliban/client/ArgEncoder.scala
@@ -11,7 +11,7 @@ import scala.annotation.implicitNotFound
  */
 @implicitNotFound(
   """Cannot find an ArgEncoder for type ${A}.
-     
+
 Caliban needs it to know how to encode arguments of type ${A}.
 """
 )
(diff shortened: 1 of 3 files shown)


ghostdogpr:
Looks good! Can you check the Scala 3 test that failed? Maybe it was using an implicit conversion which is no longer allowed?

Fluxx:
I looked into it and you were correct. There does not appear to be an implicit short -> big decimal conversion in Scala3. I converted the value to an int first, which then uses the existing int -> big decimal conversion.

ghostdogpr (approved):
LGTM, thanks!

PR #1012, opened 2021-08-24
sort trace nodes for reliability

paulpdaniels:
Closes #1009 

Update ZIO and then try to make the test less flaky by sorting the nodes.


ghostdogpr (approved):
Looks like it's green!

PR #1028, opened 2021-09-03
CompileTimeCalibanClientPlugin: couple of sbt plugins allowing, at compile time, to generate Caliban client code from Caliban `GraphQL` server code

guizmaii:
> [Sponsor message]
> This work has been sponsored by [Conduktor](https://conduktor.io/)
>
> [Conduktor](https://conduktor.io/) is the best Apache Kafka Desktop Client available today.
> It's the all-in-one friendly interface to work with the Kafka ecosystem. 
> Develop and manage Apache Kafka with confidence thanks to [Conduktor](https://conduktor.io/) 
>
> ♥️
> [/Sponsor message]

Demo project for the POC: https://github.com/guizmaii/poc_compile_time_caliban_client_generation


codegen-sbt/src/main/scala/caliban/codegen/CompileTimeCalibanPlugin.scala
```diff
@@ -0,0 +1,154 @@
...
+       |
+       |You need to configure the reference of the class containing the call to `CompileTime.generateClient`:
+       |
+       |```
+       |Compile / ctCaliban / ctCalibanGeneratorAppRef := Some("com.example.GeneratorApp")
+       |```
+       |
+       |See documentation for more details: (TODO Add link)
+       |
+       |""".stripMargin
```
guizmaii, line 54: TODO @guizmaii: Fix this doc

codegen-sbt/src/sbt-test/compiletime-codegen/simple/project/plugins.sbt
```diff
@@ -0,0 +1,13 @@
+import sbt.librarymanagement.Resolver
+
+resolvers += Resolver.mavenLocal
+
+addSbtPlugin("com.github.sbt"            % "sbt-native-packager" % "1.9.4")
```
ghostdogpr, line 5: Can you remove the dependencies that are not strictly necessary for this test? More dependencies make it harder to maintain especially when scala/sbt/etc versions change.

tools/src/main/scala/caliban/tools/compiletime/CompileTime.scala
```diff
@@ -0,0 +1,44 @@
...
+import zio.{ ExitCode, Task, URIO, ZEnv }
+
+object CompileTime {
+
+  def generateClient[R](args: List[String], api: GraphQL[R]): URIO[ZEnv, ExitCode] =
+    args match {
+      case toPath :: packageName :: clientName :: _ =>
+        (
+          for {
+            _      <- Task.unit
```
ghostdogpr, line 15: You can simplify this slightly, removing `Task.unit` and apply `exitCode` to `generate`.

codegen-sbt/src/sbt-test/compiletime-codegen/simple/project/Libraries.scala
```diff
@@ -0,0 +1,21 @@
+import sbt._
+
+object Libraries {
+
+  val zioVersion     = "1.0.11"
+  val calibanVersion = "1.1.1"
```
ghostdogpr, line 6: Is there a way to depend on the current caliban instead?

codegen-sbt/src/sbt-test/compiletime-codegen/simple/project/Libraries.scala
```diff
@@ -0,0 +1,21 @@
...
+  val sttpVersion    = "3.3.14"
+
+  val zio     = "dev.zio" %% "zio"         % zioVersion
```
ghostdogpr, line 9: You can remove dependencies that are transitive from caliban.

codegen-sbt/src/sbt-test/compiletime-codegen/simple/project/Libraries.scala
```diff
@@ -0,0 +1,21 @@
...
+  val prelude = "dev.zio" %% "zio-prelude" % "1.0.0-RC6"
```
ghostdogpr, line 10: Also not stricly needed for this POC, I'd prefer not using it.

codegen-sbt/src/sbt-test/compiletime-codegen/simple/project/BuildHelper.scala
```diff
@@ -0,0 +1,34 @@
+import com.typesafe.sbt.packager.Keys.{dockerBaseImage => _, dockerExposedPorts => _, dockerRepository => _, dockerUpdateLatest => _}
+import sbt.Keys._
+import sbt.{Compile, _}
+
+object BuildHelper {
+
+  val commonSettings = Seq(
+    libraryDependencies += compilerPlugin("org.typelevel" %% "kind-projector" % "0.13.0" cross CrossVersion.full),
```
ghostdogpr, line 8: Would be nice to strip everything that is not necessary.

codegen-sbt/src/main/scala/caliban/codegen/CompileTimeCalibanPlugin.scala
```diff
@@ -0,0 +1,223 @@
...
+ */
+object CompileTimeCalibanServerPlugin extends AutoPlugin {
+  override def requires = plugins.JvmPlugin
+  override def trigger  = noTrigger
+
+  override lazy val projectSettings: Seq[Def.Setting[_]] =
+    Seq(
+      libraryDependencies ++= Seq(
+        "com.github.ghostdogpr" %% "caliban-tools" % BuildInfo.version % Compile,
+        "dev.zio"               %% "zio"           % "1.0.11"          % Compile
```
ghostdogpr, line 23: This is not needed

codegen-sbt/src/main/scala/caliban/codegen/CompileTimeCalibanPlugin.scala
```diff
@@ -0,0 +1,223 @@
...
+                    caliban.codegen.BuildInfo.version,
+                    zio.BuildInfo.version,
+                    serverProject.id,
+                    (ctCaliban / ctCalibanGeneratorAppRef).value,
+                    (ctCaliban / ctCalibanClientName).value,
+                    (ctCaliban / ctCalibanPackageName).value
+                  )
+                )
+
+              log.debug(s"ctCaliban - cachedSettings: $cachedSettings")
```
ghostdogpr, line 175: Can be removed?

codegen-sbt/src/sbt-test/compiletime-codegen/simple/modules/server/src/main/scala/poc/generator/CalibanClientGenerator.scala
```diff
@@ -0,0 +1,10 @@
+package poc.generator
+
+import caliban.tools.compiletime.CompileTime
+import io.guizmaii.poc.caliban.server.GraphQLApi
+import zio.{ExitCode, URIO}
+
+private[generator] object CalibanClientGenerator extends zio.App {
+  override def run(args: List[String]): URIO[zio.ZEnv, ExitCode] =
+    CompileTime.generateClient(args, GraphQLApi.api)
```
ghostdogpr, line 9: Since the user needs to write that code, how about passing all arguments by Scala code? It's actually much simpler to use than doing it from sbt.

ghostdogpr (commented):
Did a quick review. First, this is great!

My comments are mostly about stripping down the test from anything that is not strictly necessary. We can link to your POC in the docs for a more complex example, the test should be more about catching bug and regressions.

Do you plan to add the other options in this PR?

Would be nice to have a review from @blast-hardcheese for the sbt part as he knows it better than me.

codegen-sbt/src/main/scala/caliban/codegen/CompileTimeCalibanPlugin.scala
guizmaii: Yes. Done

tools/src/main/scala/caliban/tools/compiletime/CompileTime.scala
guizmaii: Yup! Done. Thanks :)

guizmaii:
> Do you plan to add the other options in this PR?

Yes, I just did 🎉

codegen-sbt/src/sbt-test/compiletime-codegen/simple/project/Libraries.scala
guizmaii: Yep. I changed the code.

codegen-sbt/src/sbt-test/compiletime-codegen/simple/modules/server/src/main/scala/poc/generator/CalibanClientGenerator.scala
guizmaii: Hum. I didn't think about this possibility and yes, it's way simpler. Let me check 🤔

codegen-sbt/src/main/scala/caliban/codegen/CompileTimeCalibanPlugin.scala
guizmaii: Already brought by `caliban-tools`? 🤔

ghostdogpr: Yeah caliban-tools depends on both client and core (server) and core depends on zio.

codegen-sbt/src/sbt-test/compiletime-codegen/simple/modules/server/src/main/scala/poc/generator/CalibanClientGenerator.scala
guizmaii: Done :)

codegen-sbt/src/sbt-test/compiletime-codegen/client-and-package/modules/client/src/main/scala/io/guizmaii/poc/caliban/client/Client.scala
```diff
@@ -0,0 +1,32 @@
+package io.guizmaii.poc.caliban.client
```
blast-hardcheese, line 1: Was this left over from when you moved the test repo over into the scripted roots?

codegen-sbt/src/sbt-test/compiletime-codegen/client-and-package/modules/server/src/main/scala/poc/generator/CalibanClientGenerator.scala
```diff
@@ -0,0 +1,18 @@
+package poc.generator
+
+import caliban.tools.compiletime.CompileTime
+import caliban.tools.compiletime.CompileTime.GenerateClientsSettings
+import io.guizmaii.poc.caliban.server.GraphQLApi
+import zio.{ExitCode, URIO}
+
+private[generator] object CalibanClientGenerator extends zio.App {
```
blast-hardcheese, line 8: From the user's perspective, what value would they get from writing this file by hand?

I would think that it would be possible to generate this file based on sbt plugin configurations, which would have the advantage of permitting future Caliban alterations to this generated file if needs change in the future.

A `settingKey` with the fully-qualified path to `:= "io.guizmaii.poc.caliban.server.GraphQLApi.api"`,
(or `Seq` of them if generating multiple clients with the same `CalibanCommonSettings` makes sense),
or maybe dynamically generating the object name here based on the ClientName

blast-hardcheese, line 8: Does the `private[generator]` prevent this from being detected by sbt's `mainClass` detector?

My concern is because this is provided by the server, now `server/run` in sbt will give you the `Multiple main classes detected!` message, which will no doubt break flow for the user.

codegen-sbt/src/sbt-test/compiletime-codegen/default-values/.scalafmt.conf
```diff
@@ -0,0 +1,25 @@
...
+trailingCommas = preserve
+
+rewrite.rules = [RedundantBraces, SortModifiers]
+
+rewrite.sortModifiers.order = [
+  "implicit", "override", "private", "protected", "final", "sealed", "abstract", "lazy"
+]
+rewrite.redundantBraces.generalExpressions = false
+rewriteTokens = {
+  "⇒": "=>"
```
blast-hardcheese, line 22: fancy

tools/src/main/scala/caliban/tools/compiletime/CompileTime.scala
```diff
@@ -0,0 +1,78 @@
...
+  def generateClient[R](args: List[String])(settings: GenerateClientsSettings, api: GraphQL[R]): URIO[ZEnv, ExitCode] =
+    args match {
+      case baseDir :: _ =>
+        (for {
+          toPath <- effectBlocking {
+                      val directory = s"$baseDir/src/main/scala/${packagePath(settings.packageName)}"
+
+                      createDirectories(directory)
+
+                      s"$directory/${settings.clientName}.scala"
```
blast-hardcheese, line 55: This looks like it doesn't support `splitFiles`, do you need help with that?

tools/src/main/scala/caliban/tools/compiletime/CompileTime.scala
```diff
@@ -0,0 +1,78 @@
...
+                        settings.toCalibanCommonSettings.toOptions(schemaPath = "", toPath = toPath),
+                        GenType.Client
+                      )
+          _      <- effectBlocking {
+                      val metadataDir = s"$baseDir/target/ctCaliban"
+
+                      createDirectories(metadataDir)
+
+                      Files.writeString(
+                        new File(s"$metadataDir/metadata").toPath,
```
blast-hardcheese, line 69: This looks like it wouldn't support multiple configurations of this client generator, since they'd all just overwrite each other.

Based on my understanding (and personal use!) of the other caliban plugin, I would expect multiple generators to be supported, but does that not make sense here?

tools/src/main/scala/caliban/tools/compiletime/Utils.scala
```diff
@@ -0,0 +1,17 @@
+package caliban.tools.compiletime
+
+import java.io.File
+import java.nio.file.{ FileAlreadyExistsException, Files }
+
+private[compiletime] object Utils {
+
+  def packagePath(packageName: String): String = packageName.replaceAll("\\.", "/")
```
blast-hardcheese, line 8: My gut is saying `"/"` -> `java.io.File.separator` but I've never _explicitly_ tried to support other platforms, so I don't know if internally to the JVM `"/"` gets translated to the platform-native path separator

blast-hardcheese (changes requested):
Some minor questions, but the big ones are:
- Can we codegen the boilerplate, or are we in a chicken-and-egg situation where we don't run soon enough?
- It looks like multiple distinct Caliban query roots isn't supported. Since the other plugin supports this kind of concept, would it be too much to consider supporting that here? Even if it's too much to address in this PR, I'd like to make sure when that gets added in the future, it will come with no additional work to the user

This is a great concept though, and something I've actually implemented in my own project (we use Sangria as the query engine but Caliban as the client generator, I've wired up some sbt goo in order to do something similar to what you've done here, this is very interesting to see!)

tools/src/main/scala/caliban/tools/compiletime/Utils.scala
guizmaii: Fixed. Thanks!

tools/src/main/scala/caliban/tools/compiletime/CompileTime.scala
guizmaii: > but does that not make sense here?

Yes, I think so. Because `CompileTime.generateClient` returns an IO, it can be chained to generate multiple clients:

```scala
CompileTime.generateClient(args)(settings, GraphQLApi_0.api) *> 
  CompileTime.generateClient(args)(settings, GraphQLApi_1.api)
```

We should support this use IMO so we need to fix that. I didn't think about this problem. Thanks!

I'll try to propose a solution.

tools/src/main/scala/caliban/tools/compiletime/CompileTime.scala
guizmaii: True. I don't know what `splitFiles` does. If you have an idea on how to fix this problem, go for it. Otherwise, I'll try to propose a solution.

ghostdogpr: It generates the code is multiple files instead of a big one.

codegen-sbt/src/sbt-test/compiletime-codegen/client-and-package/modules/client/src/main/scala/io/guizmaii/poc/caliban/client/Client.scala
guizmaii: Yes, I tried to keep the test projects as close as my POC project so it's easy for me to backport things from my POC  project which I use to test my code.

codegen-sbt/src/sbt-test/compiletime-codegen/client-and-package/modules/server/src/main/scala/poc/generator/CalibanClientGenerator.scala
guizmaii: I agree. I thought about generating this file but wasn't sure if it was possible.

We could require an sbt configuration which would take a sequence of a fully-qualified path coupled with a `CalibanCommonSettings `:

```
lazy val ctCalinanConf: settingKey[Seq(String, CalibanCommonSettings)]("...")

// usage:
ctCalinanConf += ("io.guizmaii.poc.caliban.server.GraphQLApi.api",  CalibanCommonSettings(clientName = "CalibanClient"))
```

I already wrote a version where I was passing all the `CalibanCommonSettings` to the `App` args. So this is not a problem.

The problem is that I don't know how to generate this code and that we might be in a chicken-and-egg situation.

To fight this chicken-and-egg problem, one potential solution could be to also "generate" an sbt module that would contain this generated "generator" and the "caliban-client" module would depend on this generated project. 🤔
That would gives us this dependency structure:

```
client -> caliban-client -> generated-hidden-project -> server
```

Where `->` means "dependsOn".

On the paper, it could work I think 🤔 That would be great, I think. Let's see if sbt allows us to do that.

guizmaii: > Does the private[generator] prevent this from being detected by sbt's mainClass detector?

I don't think so. Good catch!

The solution presented here could solve this problem: https://github.com/ghostdogpr/caliban/pull/1028#discussion_r704905293

guizmaii:
Thanks for your review @blast-hardcheese! ♥️

I think that most problem you discovered will be fixed by the solution presented here: https://github.com/ghostdogpr/caliban/pull/1028#discussion_r704905293

I'll see if I'm able to make it work!

codegen-sbt/src/sbt-test/compiletime-codegen/client-and-package/modules/server/src/main/scala/poc/generator/CalibanClientGenerator.scala
blast-hardcheese: I _think_ you could generate this via `CompileTimeCalibanServerPlugin`, using the same technique you have in the `CompileTimeCalibanClientPlugin` via `(Compile / sourceGenerators)`. The code would be templated, with the component parts interpolated based on the settings defined in the server config.

This would have a knock-on effect, as the client would then be able to refer to those scoped settings, so no need to define the FQCN's outside of the project that contains it, which is nice.

guizmaii: I'm sorry but I'm not sure to understand the second part of your comment 🤔

blast-hardcheese: So, what I'm thinking is

```diff
-                  val generatorAppRefValue: String = (ctCaliban / ctCalibanGeneratorAppRef).value
+                  val generatorAppRefValue: String = (serverProject / ctCaliban / ctCalibanGeneratorAppRef).value
 
                   log.debug(s"ctCaliban - baseDirValue: $baseDirValue")
                   log.debug(s"ctCaliban - generatorAppRef: $generatorAppRefValue")
 
                   Def.task {
                     log.info(s"ctCaliban - Starting to generate...")
 
                     (serverProject / runMain)
                       .toTask(s" $generatorAppRefValue $baseDirValue")
                       .value
```

By defining those settings in the `serverProject`'s scope, it means that all the explicit FQCN's are located in the same project in which they are defined. This is good for grepping as well as for the server declaring its intent to expose a certain set of its API outwards, instead of the clients reaching inwards to poke at the goo. Good hygiene, imho.

### Scope creep 👇 

I don't think this is even worth discussing too deeply here, but by defining that setting inside the `serverProject` scope, it might be possible to have that setting depend on the output of `Compile`, which could actually let us walk the emitted classes looking for top-level fields that have compatible types, finding exposed APIs automatically, kind of like how `run` (as opposed to `runMain`) works. This would get us closer to the ideal of less complex configuration out of the box, with the option to override the derived setting with an explicit one  that lets the user reduce the scope down to generating only some clients.

guizmaii: Ok, I understand the second part of your comment, now. #slow 😅

Let me check because my idea seems to be impossible to achieve because it seems impossible to modify the `caliban-client` module dependencies (ie. adding dynamically a `.dependsOn(generated-hidden-project)`) in the plugin code.

sbt seems to only give access to a read-only reference to the current module (named `ResolvedProject`) via `thisProject`

blast-hardcheese: I don't know what `generated-hidden-project` is for, could you elaborate?

guizmaii: @blast-hardcheese Your idea seems to work 🎉 

I have a first POC working 🎉 

See: https://github.com/guizmaii/caliban/pull/4

blast-hardcheese: Will look tomorrow, that's great to hear!

tools/src/main/scala/caliban/tools/compiletime/CompileTime.scala
guizmaii: In the new version, the `splitFiles` works fine. I'll write some test about it.

guizmaii:
Closed in favour of https://github.com/ghostdogpr/caliban/pull/1037

PR #1032, opened 2021-09-06
add then to reservedKeywords

### LINKED ISSUE #1031: type with `then` field cause `deprecated` warning in `caliban-sbt`
Using `caliban-sbt` on a schema with type like this:

```
type Order {
 then: String
}
```

would cause `then is a reserved word (since 2.10.0); usage as an identifier is deprecated` on the generated code.

is there a way to fix this?

Thanks

RhnSharma:
Fixes #1031
Hi @ghostdogpr, how does this look?
Let me know if this needs any changes.
Thanks


ghostdogpr (approved):
Looks good, thanks!

PR #1033, opened 2021-09-06
add http linking to an example of Caliban using play's route file

rlavolee:
You'll find in the following example how to use Caliban with Play's route file.
I hope it helps in any ways.


ghostdogpr (approved):
Thanks!

Issue #7, opened 2019-09-12
Implement missing query validations


ghostdogpr:
Missing validations:
- [x] https://graphql.github.io/graphql-spec/June2018/#sec-Field-Selection-Merging
- [x] https://graphql.github.io/graphql-spec/June2018/#sec-Fragment-spread-is-possible
- [x] https://graphql.github.io/graphql-spec/June2018/#sec-Values-of-Correct-Type (while doing this, `ArgBuilder` error type should be changed to `ValidationError`)
- [x] https://graphql.github.io/graphql-spec/June2018/#sec-Input-Object-Field-Names
- [x] https://graphql.github.io/graphql-spec/June2018/#sec-Input-Object-Field-Uniqueness
- [x] https://graphql.github.io/graphql-spec/June2018/#sec-Input-Object-Required-Fields
- [x] https://graphql.github.io/graphql-spec/June2018/#sec-Directives-Are-In-Valid-Locations
- [x] https://graphql.github.io/graphql-spec/June2018/#sec-Directives-Are-Unique-Per-Location
- [x] https://graphql.github.io/graphql-spec/June2018/#sec-All-Variable-Usages-are-Allowed

paulpdaniels:
For field selection merging, there is a paper that details a faster implementation: https://tech.xing.com/graphql-overlapping-fields-can-be-merged-fast-ea6e92e0a01

And a reference implementation that was added to sangria: https://github.com/sangria-graphql-org/sangria/pull/12

PR #1042, opened 2021-09-11
Add test cases for bugs with mutation/variable/case-class.

edlott:
I found an issue with the following scenario:
- Invoke a mutation
- Use a variable
- Make the type of the variable a case-class type

I use this to edit a multi-field record.  The 'update' method takes a record parameter (input) and returns the same thing after it's been saved to the database.

Caliban is having trouble with the variable's type.  If I set it to the type name listed in the schema, it fails.  If I lie and set it to any primitive type, it works.

This pull request contains test I added to reproduce this bug.

### DIFF
--- a/core/src/test/scala/caliban/TestUtils.scala
+++ b/core/src/test/scala/caliban/TestUtils.scala
@@ -76,6 +76,16 @@ object TestUtils {
     implicit val schema: Schema[Any, Character] = Schema.gen[Character]
   }
 
+  case class StudioRequest(
+    actor: String,
+    reason: String
+  )
+
+  case class StudioResponse(
+    request: StudioRequest,
+    status: String
+  )
+
   val characters = List(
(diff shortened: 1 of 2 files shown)


ghostdogpr:
The error is actually correct: `Variables can only be input types. Objects, unions, and interfaces cannot be used as inputs.`. In fact, `StudioRequest` is an object type and `StudioRequestInput` is an input type (the suffix `Input` is added to the input type to make sure the name doesn't clash with the object type, see `GQLInputType` docs if you want to change it). You can see it if you call `render` or if you introspect your schema. If you change `StudioRequest!` by `StudioRequestInput!`, it works.

The reason the second test is passing is that the validation for that is not implemented, but this is a known issue (the last item in https://github.com/ghostdogpr/caliban/issues/7)

I didn't know about Cas Anvar 😭

ghostdogpr:
Closing this one then, let me know if you need more details.

PR #1061, opened 2021-09-24
Fix for Scala 2.13 and 3

guizmaii:


build.sbt
```diff
@@ -28,6 +28,7 @@ val zioHttpVersion         = "1.0.0.0-RC17"
 
 inThisBuild(
   List(
+    version := "1.1.167",
```
ghostdogpr, line 31: What's that?

codegen-sbt/src/main/scala/caliban/codegen/CompileTimeCalibanPlugin.scala
```diff
@@ -150,7 +150,13 @@ object CompileTimeCalibanServerPlugin extends AutoPlugin {
 
   override lazy val projectSettings: Seq[Def.Setting[_]] =
     Seq(
-      libraryDependencies += "com.github.ghostdogpr" %% "caliban-tools" % caliban.codegen.BuildInfo.version % Compile,
+      libraryDependencies += {
+        CrossVersion.partialVersion((thisProject / scalaVersion).value) match {
```
ghostdogpr, line 154: It didn't pick up the version properly?

build.sbt
guizmaii: Sorry that was WIP

codegen-sbt/src/main/scala/caliban/codegen/CompileTimeCalibanPlugin.scala
guizmaii: It's using the Scala version of sbt apparently. At runtime, our tests were crashing. We'll add some tests to prove that it working now.

PR #1063, opened 2021-10-03
Add request handler for play websocket requests

easel:
PlayAdapter now includes `def handleWebSocketRequestHeader: RequestHeader => ZIO[R, Result, Unit]` 
and always passes websocket requests through webSocketOrResult to allow for authentication. Play auth example
now authenticates websocket requests.


ghostdogpr:
Hi! What's the jitpack file doing? It seems to cause the CI to fail and I am not sure what it is for.

easel:
@ghostdogpr I've mostly run aground attempting to solve the problem of passing the authentication token into the subscription provider. The root problem is that the pattern in use for play and akka http relies on a fiber ref, and as things stand, passing that fiber ref into the stream doesn't work. 

I'm not familiar enough with the guts of zio streams to know if there's a way to force them to somehow instantiate and run on the same the same fiber as they are instantiated on, but the few naive things I've tried have run ground.

The brute force approach is to just allow a `A => IO[Auth, ZStream[Auth, E, T]]` as member of the `Subscription`. I've partially traced through the schema generation and execution code to make this work but haven't quite made it yet.

All that being said, I think this PR as it stands is an improvement on the current state (no more misspelled and unused methods and at least you can enforce security on subscriptions) so perhaps it's best to merge it and deal with passing context into subscriptions elsewhere. I'm happy either way.

easel:
Ok, I've dropped the `RIO[ZStream]`. Should we consider deprecating the `RIO[ZStream]` subscription schema as well?

ghostdogpr:
Yeah let's drop `RIO[ZStream]` as well. 

Okay for adding a new PR with the Scala 3 failing test, I can take a look at it. Thanks!

ghostdogpr:
Minor: can you clean and sort imports?

easel:
Imports should be cleaned up now.

easel:
@ghostdogpr I considered trying to merge them as well. I can say that for my use case the implementations are almost identical. I guess the challenge is that the semantics aren't quite the same -- for websockets failures are on the error channel, while request wrapper just passes through a (potentially modified) request.

That said, if we changed, perhaps `RequestWrapper` would be better off with an explicit error channel?  This is arguably a better situation from a security perspective because it means that the request wrapper can fail out a request without cooperating with some downstream code to do it.

I can experiment with some options -- I think it would be unfortunate to change the signature of `RequestWrapper` such that it broke everybodies code. 

If we can introduce a `RequestWrapperOrError` under `RequestWrapper` with a `wrapRequestOrFail(requestHeader: RequestHeader): ZIO[R, Result, Result]` and provide a default implementation of the new `wrapRequestOrFail` method that lifts the result to the error channel if status code doesn't begin with 200 or something, legacy code would work on websockets "for free". 

At some point we could also deprecate the one-parameter `RequestWrapper`, and then flip the `makePost`, `makeGet` etc methods that use it over to the new semantics where the graphql code doesn't get run if the wrapper errors out, bringing them in line with the websocket semantics.

Is that what you had in mind?

ghostdogpr:
I like your last idea. Deprecate the existing method and expose one with the unified semantics (with a different name for RequestWrapper so that both the old way and the new way can work at the same time?). And if the old way implementation could actually use the new way under the hood, the code stays simple.

ghostdogpr:
@easel I'd love your feedback on this ongoing PR: https://github.com/ghostdogpr/caliban/pull/1125
I'm re-implementing the adapters based on Tapir so that they all share the same code and functionalities. I implemented something called "request interceptors" which can do what `ContextWrapper` and `RequestWrapper` used to do in Play and Akka adapters. If you could have a look at this, it would be awesome.

PR #1074, opened 2021-10-08
feat: Validate field merging (5.3.2).

frekw:
Hey!

I took a stab at implementing validating field selection merging ([5.3.2](https://spec.graphql.org/June2018/#sec-Field-Selection-Merging)) since I noticed that Apollo Gateway was rejecting some queries Caliban considered valid.


frekw:
Hm, Scala 3 seems to be failing. I'll look into it later this weekend.

PR #1086, opened 2021-10-13
Fix error handling in laminext integration

### LINKED ISSUE #1083: Don't try to decode the response in laminext module when ok is false
See https://github.com/ghostdogpr/caliban/blob/b5eb43b95626435ba8ef0def88b5f1acf7c06296/client-laminext/src/main/scala/caliban/client/laminext/package.scala#L50

ghostdogpr:
Closes #1083


mingyuchi:
naisu

PR #1100, opened 2021-10-17
Add test for custom R in schema (derivation fails on scala 3)

easel:
Per earlier discussion, this test passes on everything but scala 3, implying a functionality gap in schema derivation.

### DIFF
--- a/vuepress/docs/docs/schema.md
+++ b/vuepress/docs/docs/schema.md
@@ -200,6 +200,7 @@ If you require a ZIO environment, you will need to have the content of `caliban.
 ```scala
 object schema extends GenericSchema[MyEnv]
 import schema._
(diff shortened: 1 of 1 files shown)


ghostdogpr:
I just had a look and this is actually normal: in Scala 3, wildcard imports don't import givens (the replacement for implicits).
So when you do `import schema._`, it doesn't import the `gen` method which is a `given`.
If you simply do:
```scala
object schema extends GenericSchema[Auth]
import schema._
import schema.gen
```
it works as expected. If the code was scala 3 only, you could import `schema.given` but since it cross-compiles you have to import `gen`.

easel:
Thanks for chasing that down! With the change, the test passes. I think I added the change and a comment to the custom schema docs, as well. Let me know if you think it's helpful.

PR #1103, opened 2021-10-19
[Client] Allow query to specify every subtype of an interface

AlixBa:
Rationale: We would like compilation to fail if there a new enum entry and we don't implement a selection builder

### DIFF
--- a/tools/src/main/scala/caliban/tools/ClientWriter.scala
+++ b/tools/src/main/scala/caliban/tools/ClientWriter.scala
@@ -109,7 +109,12 @@ object ClientWriter {
       s"""$description${deprecated}def $safeName$typeParam$args$innerSelection$implicits: SelectionBuilder[$typeName, $outputType] = _root_.caliban.client.SelectionBuilder.Field("$name", $builder$argBuilder)"""
     }
 
-    def collectFieldInfo(field: FieldDefinition, typeName: String, optionalUnion: Boolean): FieldInfo = {
+    def collectFieldInfo(
+      field: FieldDefinition,
+      typeName: String,
+      optionalUnion: Boolean,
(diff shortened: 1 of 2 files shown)


ghostdogpr:
So, I've recently implemented [this PR](https://github.com/ghostdogpr/caliban/pull/1099) for unions. Basically it generates 2 methods instead of 1, one with all union subtypes optional and the other where they are mandatory. How about doing the same for interfaces? It would be consistent and would prevent an additional configuration parameter. Also you would be free to use a mix of them in your code, which is more flexible than a single conguration. What do you think?

AlixBa:
I'll have something by the end of the day 🤞

ghostdogpr:
@AlixBa yeah, that's right.

And it would be nice to add `onInterfaceName: SelectionBuilder[InterfaceName, A],` to select common fields from the interface (and generate that code, it doesn't exist currently). I guess it makes sense only in `methodOption`, not on the other one?

ghostdogpr:
The test error comes from this code generated in `writeView`:
```scala
case (field @ FieldTypeInfo(_, _, _, interfaceTypes, _, _, Some(_)), fieldName) if interfaceTypes.nonEmpty =>
          val tpe = genericSelectionFieldTypesMap(field)
          interfaceTypes.map(intType => s"${fieldName}On$intType: Option[SelectionBuilder[$intType, $tpe]] = None")
```
I guess it doesn't need to be an `Option` anymore (like for unions just a few lines above)?

You can run `scripted` in sbt to test it.

AlixBa:
Ups, sorry @ghostdogpr 
It is, indeed

ghostdogpr (approved):
LGTM, thanks!

PR #1116, opened 2021-10-29
Fix package name extraction in code generation

pavlosgi:
The old regex doesn't work properly on paths such as "src/main/scala/company/service/infra/graphql/model.scala". This returns "infra/graphql" as the package name. This is because of [] square brackets to match on one of scala, play or app. I replaced that with a non capturing regex group and that works as expected returning "company/service/infra/graphql" as the package name

### DIFF
--- a/tools/src/test/scala/caliban/tools/CodegenSpec.scala
+++ b/tools/src/test/scala/caliban/tools/CodegenSpec.scala
@@ -1,0 +1,89 @@
+package caliban.tools
+
+import Codegen.getPackageAndObjectName
+import zio.test.Assertion._
+import zio.test._
+import zio.test.environment.TestEnvironment
+
+object CodegenSpec extends DefaultRunnableSpec {
+
+  override def spec: ZSpec[TestEnvironment, Any] =
+    suite("CodegenSpec")(
+      test("Package name is empty or default if path doesn't follow correct format") {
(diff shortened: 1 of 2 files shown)


ghostdogpr:
It looks good! Do you mind adding a few unit tests to capture the different cases? Ideally with the 2 lines moved to a function.

pavlosgi:
Done

tools/src/test/scala/caliban/tools/CodegenSpec.scala
```diff
@@ -0,0 +1,89 @@
...
+import zio.test.environment.TestEnvironment
+
+object CodegenSpec extends DefaultRunnableSpec {
+
+  override def spec: ZSpec[TestEnvironment, Any] =
+    suite("CodegenSpec")(
+      test("Package name is empty or default if path doesn't follow correct format") {
+        assert(packageAndObject("a/b/c", None, None))(
+          equalTo(None -> "Client")
+        )
```
ghostdogpr, line 15: You need to join the `assert` with `&&`, otherwise only the last one will be executed (it's `assert` from zio-test, not the regular scala assert, it's a pure function that returns a test result).

pavlosgi: 👍

ghostdogpr (approved):
Thanks!

PR #1125, opened 2021-11-04
Use Tapir to implement adapters

ghostdogpr:
Closes #1117
Closes #841

- [x] tapir adapter
- [x] migration of http4s http adapter
- [x] migration of http4s ws adapter
- [x] migration of zio-http http adapter
- [x] ~~migration of zio-http ws adapter (currently not supported by tapir)~~ <-- will keep existing code for now
- [x] migration of akka-http http adapter
- [x] migration of akka-http ws adapter
- [x] migration of play http adapter
- [x] migration of play ws adapter (need https://github.com/softwaremill/tapir/pull/1585)
- [x] support for custom json backend
- [x] support for uploads
- [x] support for request interceptors (formerly context wrappers)
- [x] create test suite for adapters
- [x] decommission finch adapter
- [x] https://github.com/softwaremill/tapir/issues/1566
- [x] https://github.com/softwaremill/tapir/issues/1583
- [x] https://github.com/softwaremill/tapir/pull/1601
- [x] wait for an official tapir release? (using a milestone release now)
- [x] polishing
- [ ] docs

PR #1129, opened 2021-11-07
Add auth header to context for websockets in akka-http adapter

alterationx10:
The idea is to pass in a function on `makeWebSocketService` that can parse the String payload of the `connection_init` message. If it finds it, update a Ref of the context that was created.

Then on the `start` message, it will get the current context from the Ref, and pass that into `startSubscription`. At that point, the ContextWrapper passed into `makeWebSocketService` can parse for an auth header, just as the corresponding `makeHttpService` would. 

This seemed to work well for me, but i'm not sure if this part is "safe", as I haven't worked with `Ref`s much, yet.
```
          val ctxRef          = runtime.unsafeRun(Ref.make(ctx))
```
I assume that's still scope per request that comes in, but would like to confirm it.

### DIFF
--- a/adapters/akka-http/src/main/scala/caliban/AkkaHttpAdapter.scala
+++ b/adapters/akka-http/src/main/scala/caliban/AkkaHttpAdapter.scala
@@ -2,6 +2,7 @@ package caliban
 
 import akka.http.scaladsl.model.MediaTypes.`application/json`
 import akka.http.scaladsl.model._
+import akka.http.scaladsl.model.headers.Authorization
 import akka.http.scaladsl.model.ws.{ Message, TextMessage }
 import akka.http.scaladsl.server.Directives.{ complete, extractRequestContext }
 import akka.http.scaladsl.server.{ RequestContext, Route }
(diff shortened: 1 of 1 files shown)


adapters/akka-http/src/main/scala/caliban/AkkaHttpAdapter.scala
```diff
@@ -166,14 +167,17 @@ trait AkkaHttpAdapter {
...
   def makeWebSocketService[R, E](
     interpreter: GraphQLInterpreter[R, E],
     skipValidation: Boolean = false,
     enableIntrospection: Boolean = true,
     keepAliveTime: Option[Duration] = None,
     contextWrapper: ContextWrapper[R, GraphQLResponse[E]] = ContextWrapper.empty,
     queryExecution: QueryExecution = QueryExecution.Parallel,
-    wsChunkTimeout: Duration = 5.seconds
+    wsChunkTimeout: Duration = 5.seconds,
+    onInitAuthHeader: String => RIO[R, Option[Authorization]] = emptyAuth(_: String)
```
alterationx10, line 180: I tried to pass inline a `_ => ZIO.succeed(Option.empty[Authorization])` here, but the compiler didn't like something about inferring the String

ghostdogpr:
As we discussed on Discord, while this PR is okay, this feature is already supported as part of https://github.com/ghostdogpr/caliban/pull/1125 which I hope to get it in the next release.

PR #1133, opened 2021-11-10
Correctly validate optional list in inputs

darl:
Didn't read graphql spec, but this seems like correct behaviour.


ghostdogpr:
Your fix is correct in the case of a List but I think we should go even further.

In the spec (http://spec.graphql.org/June2018/#sec-Scalars): `For all types below, with the exception of Non‐Null, if the explicit value null is provided, then the result of input coercion is null.`.  So I think other types need the same treatment. @frekw right?

frekw (approved):
The change itself looks good. But I think we should do the same for `INPUT_OBJECT` and `ENUM` as well?

frekw:
Ah you beat me to it @ghostdogpr :)

darl:
So case with `// handle item as the first item in the list` is incorrect?

darl:
Nevermind. I got it. Will fix soon

ghostdogpr (approved):
Perfect, thanks!

PR #1138, opened 2021-11-12
[Close #1119] Add extension methods for __Type: list, nonNull

### LINKED ISSUE #1119: Idea: Add extension functions to `Types` functions which take `underlying` so they can be used post-fix
This being able to let you optionally write:

```scala
Types.string.nonNull
```

Rather than:
```scala
Types.makeNonNull(Types.string)
```

https://github.com/ghostdogpr/caliban/blob/8b523050d610bb3072120e6e2321fadeeb380ab6/core/src/main/scala/caliban/schema/Types.scala#L23-L25

Something like:

```scala
extension (type: __Type)
	def list(): __Type = __Type(__TypeKind.LIST, ofType = Some(type)) 
	def nonNull(): __Type = __Type(__TypeKind.NON_NULL, ofType = Some(type))
```

GavinRay97:
Closes #1119

### DIFF
--- a/core/src/main/scala/caliban/introspection/adt/__Type.scala
+++ b/core/src/main/scala/caliban/introspection/adt/__Type.scala
@@ -102,3 +102,10 @@ case class __Type(
       case _                   => true
     }
 }
+
+object __Type {
+  implicit class __TypeExtensions(val it: __Type) {
+    def list: __Type    = __Type(__TypeKind.LIST, ofType = Some(it))
+    def nonNull: __Type = __Type(__TypeKind.NON_NULL, ofType = Some(it))
+  }
+}


ghostdogpr:
How about using this new functions in existing code to simplify it?

ghostdogpr:
Tip: add "Closes #1119" in the PR description, so that it's automatically closed when the PR is merged.

GavinRay97:
> How about using this new functions in existing code to simplify it?

Good idea, will do this when have a moment

core/src/main/scala/caliban/introspection/adt/__Type.scala
```diff
@@ -102,3 +102,10 @@ case class __Type(
       case _                   => true
     }
 }
+
+object __Type {
+  implicit class __TypeExtensions(val it: __Type) {
+    def list: __Type    = __Type(__TypeKind.LIST, ofType = Some(it))
+    def nonNull: __Type = __Type(__TypeKind.NON_NULL, ofType = Some(it))
+  }
+}
```
paulpdaniels, line 111: Just curious, why do we need to make these into type extensions instead of just putting them on the `__Type` class? Is it a bincompat thing?

ghostdogpr: Oh yeah, didn't realize that 😆

GavinRay97: 🤦 Aye, good point

ghostdogpr:
Done in https://github.com/ghostdogpr/caliban/pull/1188

PR #1140, opened 2021-11-14
Bug fix: adds the support for fields starting with capital letters

several27:
The codegen generates semantically invalid scala code when Graphql fields types start with a capital letter. 

E.g., for a schema: 

```graphql
type Aspect {
    Name: String!
    Value: String!
}
```

Caliban codegen generates the following view: 

```scala
...

def view: ViewSelection = (Name, Value).map {
  case (Name, Value) ⇒ AspectView(Name, Value)
}
...
```

This code is invalid because Scala accepts only variable names starting with a lower case for pattern matching. 

This PR fixes this behavior by decapitalizing the variable name, so that the generated code is: 

```scala
...

def view: ViewSelection = (Name, Value).map {
  case (name$, value$) ⇒ AspectView(name$, value$)
}
...
```

The PR fixes the bug by adding the `$` symbol to the end of the variable name and decapitalizes for the following cases: 

- when a variable name is a keyword 
- when a variable name contains an underscore 
- (new case) when a variable name starts with a capital letter

Would appreciate any help reviewing this and pushing it whenever possible (currently, this bug prevents us from using **this excellent library**)!. 

Thanks

### DIFF
--- a/tools/src/main/scala/caliban/tools/ClientWriter.scala
+++ b/tools/src/main/scala/caliban/tools/ClientWriter.scala
@@ -597,7 +597,7 @@ object ClientWriter {
         """
 
   def safeUnapplyName(name: String): String =
-    if (reservedKeywords.contains(name) || name.endsWith("_")) s"$name$$"
+    if (reservedKeywords.contains(name) || name.endsWith("_") || name.charAt(0).isUpper) s"$$$name"
     else name
 
   def safeName(name: String): String =
--- a/tools/src/main/scala/caliban/tools/ClientWriter.scala
+++ b/tools/src/main/scala/caliban/tools/ClientWriter.scala
@@ -597,8 +597,12 @@ object ClientWriter {
         """
 
   def safeUnapplyName(name: String): String =
-    if (reservedKeywords.contains(name) || name.endsWith("_") || name.charAt(0).isUpper) s"$$$name"
+    if (reservedKeywords.contains(name) || name.endsWith("_") || name.charAt(0).isUpper) s"${decapitalize(name)}$$"
     else name
+
+  private def decapitalize(name: String): String = if (name.length > 1 && name.charAt(0).isUpper)
+    name.charAt(0).toLower + name.substring(1)
+  else name
 
   def safeName(name: String): String =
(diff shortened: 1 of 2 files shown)


tools/src/main/scala/caliban/tools/ClientWriter.scala
```diff
@@ -597,9 +597,17 @@ object ClientWriter {
         """
 
   def safeUnapplyName(name: String): String =
-    if (reservedKeywords.contains(name) || name.endsWith("_")) s"$name$$"
+    if (reservedKeywords.contains(name) || name.endsWith("_") || name.charAt(0).isUpper) s"${decapitalize(name)}$$"
```
ghostdogpr, line 600: Can you check for size, so that if `name` is empty, it doesn't throw an exception?

tools/src/test/scala/caliban/tools/ClientWriterViewSpec.scala
```diff
@@ -10,10 +10,10 @@ import zio.test.environment.TestEnvironment
 
 object ClientWriterViewSpec extends DefaultRunnableSpec {
 
-  val gen: String => RIO[Blocking, String] = (schema: String) =>
+  val gen: String ⇒ RIO[Blocking, String] = (schema: String) ⇒
```
ghostdogpr, line 13: Can you undo these?

ghostdogpr (commented):
Looks good! Just some minor comments and I can merge it.

tools/src/test/scala/caliban/tools/ClientWriterViewSpec.scala
several27: Ah yes, accidentally used a different formatter - should be good now :)

tools/src/main/scala/caliban/tools/ClientWriter.scala
several27: Was assuming that name can't be empty, as it's a graphql field, but no problem made the change!

several27:
> Looks good! Just some minor comments and I can merge it.

Please give it another check. Thank you!

ghostdogpr (approved):
Thanks! A snapshot will be generated after I merge so you are able to use this right away.

Issue #1130, opened 2021-11-08
Add Scala 3 keywords to the list of reserved keywords


ghostdogpr:
In ClientWriter.scala 

See the list here: https://dotty.epfl.ch/docs/internals/syntax.html#regular-keywords

jgoday:
Hi @ghostdogpr, I started to take a look at caliban, playing with both server and generating client code.
I have made a PR for this issue: #1145.

ghostdogpr:
Closed by https://github.com/ghostdogpr/caliban/pull/1145

PR #1145, opened 2021-11-15
Scala3 reserved keyworkds in tools/ClientWriter

jgoday:
See #1130.

Add scala3 reserved and soft keywords to tools/src/main/scala/caliban/tools/ClientWriter.scala,
to safe generate client code from graphql schemas.

Some scala3 operators are not included, because they are not considered valid names in a graphql schema
```
:         =         <-        =>        <:        >:        #             @         =>>       ?=>
```

[https://dotty.epfl.ch/docs/internals/syntax.html#regular-keywords](https://dotty.epfl.ch/docs/internals/syntax.html#regular-keywords)


ghostdogpr (commented):
Looks good to me, thanks!

PR #1146, opened 2021-11-16
Adding a new required param to an input field is not considered a breaking change

kormbrek-rally:
I do not wish to merge this code. It is just a demonstration of an issue I came across when trying to use the schema comparison functionality in caliban-tools.


ghostdogpr:
Indeed, `FieldAdded` is hardcoded to breaking = false. It should be possible to check if the field is optional instead. Will look into it when i have time.

ghostdogpr:
Addressed by https://github.com/ghostdogpr/caliban/pull/1147

PR #1149, opened 2021-11-17
validation: allow Ints passed to Float arguments

darl:
Maybe translation similar to #1136 would be better. 
But I don't know reasons behind #1136 implementation.

### DIFF
--- a/core/src/main/scala/caliban/validation/ValueValidator.scala
+++ b/core/src/main/scala/caliban/validation/ValueValidator.scala
@@ -138,8 +138,8 @@ object ValueValidator {
         }
       case "Float"   =>
         argValue match {
-          case _: Value.FloatValue | NullValue => IO.unit
-          case t                               => failValidation(s"$errorContext has invalid type $t", "Expected 'Float'")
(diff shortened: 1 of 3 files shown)


ghostdogpr:
The reason behind the transformation of `StringValue` into `EnumValue` is so that when you use `Field` metadata, you get a value of a type that matches the schema. Basically it's coercing "early" so that you don't even need to do it in the ArgBuilder. I guess we should apply the same logic for this case. @frekw right? What do you think?

ghostdogpr:
@darl I opened https://github.com/ghostdogpr/caliban/pull/1150

frekw:
@ghostdogpr yeah I think doing it like in #1150 is the only way to do it since we're basically paving over limitations in JSON. If I read the spec correctly, coercion should only be applied when we're dealing with variables so doing it like this probably runs the risk of accepting invalid queries for `Content-Type: application/graphql`.

ghostdogpr:
@frekw shouldn't it be applied to arguments too? http://spec.graphql.org/June2018/#sec-Coercing-Field-Arguments

ghostdogpr:
Answering myself: not for enums, but for float yes. Will handle it in my other PR.

PR #1150, opened 2021-11-17
Coerce int values into float when expected

ghostdogpr:
Also did a bit of unrelated cleanup.

### DIFF
--- a/core/src/test/scala/caliban/validation/InputObjectSpec.scala
+++ b/core/src/test/scala/caliban/validation/InputObjectSpec.scala
@@ -24,7 +24,7 @@ object InputObjectSpec extends DefaultRunnableSpec {
           int <- gql.interpreter
           res <- int.execute(query)
         } yield assert(res.errors.headOption)(
-          isSome((isSubtype[CalibanError.ValidationError](anything)))
+          isSome(isSubtype[CalibanError.ValidationError](anything))
         )
       },
       testM("fails if a non-null field on an input object is null from variables") {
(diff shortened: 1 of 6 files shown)


core/src/test/scala/caliban/validation/InputObjectSpec.scala
```diff
@@ -139,6 +139,34 @@ object InputObjectSpec extends DefaultRunnableSpec {
...
+
+        for {
+          int <- gql.interpreter
+          res <- int.executeRequest(
+                   GraphQLRequest(
+                     query = Some(query),
+                     variables = Some(
+                       Map(
+                         "input" -> InputValue.ObjectValue(
+                           Map("float" -> Value.IntValue(42))
```
frekw, line 163: Hm, this strikes me as a bit odd since JSON only supports floats 🤔 I would probably have expected the coercion to be the other way around...

But perhaps we have some smartness in the parser to parse numbers without decimals as ints so we need to deal with both cases?

ghostdogpr: The parser would parse into a `FloatValue` only if there is a dot or a `e` for exponent.

ghostdogpr: In scala 2. The scala 3 one checks for int before float so the result is the same.

ghostdogpr: According to the spec, float should not be coerced into int.

PR #1156, opened 2021-11-19
Support multiple setting sets per file

kubukoz:
This would be useful in situations where you want to generate both a server and a client from a single schema file, with potentially different settings.

Update: Added some more changes:

- separated the scripted tests so there's one for the direct `calibanGenClient` calls and one for the setting-based generation
- added scripted test for multiple set of settings per file
- mentioned the usage in the docs
- mentioned how to cherry-pick schemas for generation


ghostdogpr (commented):
Looks good! Would be nice to test it with scripted tests.

codegen-sbt/src/sbt-test/codegen/gen-client-task/test
```diff
@@ -0,0 +1,34 @@
+$ absent src/main/scala/client/ClientNameUniqueness.scala
```
kubukoz, line 1: Moved from below

codegen-sbt/src/sbt-test/codegen/gen-client-task/verify.sh
```diff
@@ -0,0 +1,9 @@
+#!/usr/bin/env bash
```
kubukoz, line 1: Copy-pasted from the other test

codegen-sbt/src/sbt-test/codegen/test-compile/build.sbt
```diff
@@ -1,20 +1,24 @@
...
+  .enablePlugins(CalibanPlugin)
   .settings(
     libraryDependencies ++= Seq(
       "com.github.ghostdogpr" %% "caliban-client" % Version.pluginVersion
     ),
     Compile / caliban / calibanSettings ++= Seq(
-      calibanSetting(file("src/main/graphql/schema.graphql"))(  // Explicitly constrain to disambiguate
-        cs =>
-          cs.clientName("Client")
+      calibanSetting(file("src/main/graphql/schema.graphql"))( // Explicitly constrain to disambiguate
```
kubukoz, line 11: tbh I don't get what this "constrains", is that comment up to date?

codegen-sbt/src/sbt-test/codegen/test-compile/build.sbt
```diff
@@ -1,20 +1,24 @@
+import _root_.caliban.tools.Codegen
+
 lazy val root = project
   .in(file("."))
-  .enablePlugins(CodegenPlugin)  // Intentionally maintain the deprecated name
+  .enablePlugins(CalibanPlugin)
```
kubukoz, line 5: Using the deprecated name in the other test. This way we test both.

codegen-sbt/src/sbt-test/codegen/test-compile/build.sbt
```diff
@@ -1,20 +1,24 @@
...
-      calibanSetting(file("src/main/graphql/schema.graphql"))(  // Explicitly constrain to disambiguate
-        cs =>
-          cs.clientName("Client")
+      calibanSetting(file("src/main/graphql/schema.graphql"))( // Explicitly constrain to disambiguate
+        _.clientName("Client")
+      ),
+      // Another entry for the same file, which will cause another generator to run
+      calibanSetting(file("src/main/graphql/schema.graphql"))(
+        _.genType(Codegen.GenType.Schema)
+          .scalarMapping("Json" -> "String")
```
kubukoz, line 17: avoids circe dependency

codegen-sbt/src/sbt-test/codegen/test-compile/build.sbt
```diff
@@ -1,20 +1,24 @@
...
+          .effect("scala.util.Try")
```
kubukoz, line 18: avoids zio dependency

vuepress/docs/docs/schema.md
```diff
@@ -292,6 +292,9 @@ You can also indicate that the effect type is abstract via `--abstractEffectType
 If you want to force a mapping between a GraphQL type and a Scala class (such as scalars), you can use the
 `--scalarMappings` option. Also you can add additional imports by providing `--imports` option.
 
+Since Caliban 1.2.4, you can generate schemas using an sbt `sourceGenerator`, which means your schemas will be generated every time you compile (or when you import your build into [Metals](https://scalameta.org/metals/)).
```
kubukoz, line 295: This would probably be better written as a normal section like in the Client docs, but I didn't want to spend much time working on these docs for now.

kubukoz:
@ghostdogpr updated with a test and some other changes (see note in PR description) :)

cc @guizmaii I'd appreciate some eyes on this

vuepress/docs/docs/schema.md
ghostdogpr: It will be from 1.3.0 (I don't think it was part of 1.2.4).

ghostdogpr (approved):

codegen-sbt/src/main/scala/caliban/codegen/CalibanSourceGenerator.scala
```diff
@@ -45,17 +45,19 @@ object CalibanSourceGenerator {
...
     val defaults: CalibanCommonSettings = CalibanCommonSettings.empty.copy(packageName = Some("caliban"))
 
-    CalibanFileSettings(
-      file = source,
-      settings = fileSettings.collect { case needle if source.toPath.endsWith(needle.file.toPath) => needle }
-        .foldLeft[CalibanCommonSettings](defaults) { case (acc, next) => acc.combine(next.settings) }
-    )
+    val matchingSettingSets = fileSettings.collect {
+      case needle if source.toPath.endsWith(needle.file.toPath) => needle.settings
+    }.map(defaults.combine(_))
```
guizmaii, line 56: Are we sure this `.map(defaults.combine(_))` is equivalent to the previous `.foldLeft[CalibanCommonSettings](defaults) { case (acc, next) => acc.combine(next.settings) }` 🤔

Or is it an expected change?

codegen-sbt/src/sbt-test/codegen/test-compile/build.sbt
blast-hardcheese: Certainly not the most helpful comment, as I don't understand what needed to be disambiguated here -- perhaps back when it was possible (or intended) to star-glob multiple files? I could see trying to disambiguate between the few `.graphql` test specifications we had at that time.

codegen-sbt/src/main/scala/caliban/codegen/CalibanSourceGenerator.scala
```diff
@@ -94,25 +96,31 @@ object CalibanSourceGenerator {
...
+      val generateFromURLs = ZIO.foreach(urlSettings)(setting =>
+        generateUrlSource(setting.url, setting.settings).catchAll {
+          case Some(reason) =>
+            putStrLn(reason.toString) *> putStrLn(reason.getStackTrace.mkString("\n")).as(List.empty)
+          case None         => ZIO.succeed(List.empty)
+        }
+      )
+
+      Runtime.default.unsafeRun {
+        ZIO.mapN(generateFromFiles, generateFromURLs)((_ ++ _)).map(_.flatten)
```
guizmaii, line 122: Can `parMapN` speed up things in this case? 🤔

codegen-sbt/src/main/scala/caliban/codegen/CalibanSourceGenerator.scala
blast-hardcheese, line 56: This is the intended change in this PR, from what I can tell. I was `foldLeft`-ing in the beginning to permit tiered configuration, but that was before the `cs` pattern was discovered. I don't think it's useful to accumulate anymore, if the previous implementation was desired, it is possible to define a generic `CalibanSettings => CalibanSettings` and apply that before you chain the rest of your config.

vuepress/docs/docs/client.md
```diff
@@ -69,6 +69,20 @@ In order to supply more configuration options to the code generator, you can use
       )
 ```
 
+The path where the generator will look for schemas can be customized by overriding the `calibanSources` settings:
+
+```scala
+Compile / caliban / calibanSources := file("caliban")
+```
+
+If you want to only include certain files in a directory, you can override that as well with an explicit `caliban / sources` entry:
```
blast-hardcheese, line 78: This isn't actually true, since `calibanSources` is a `File`, not a `List[File]` -- the comment doesn't scale and will trip up users

blast-hardcheese (approved):
👍 from me, good addition.

codegen-sbt/src/main/scala/caliban/codegen/CalibanSourceGenerator.scala
kubukoz: IME they're so fast there's no need for this. I'd rather have sbt manage parallelism on a higher level (more tasks at once or whatever), but I'm fine with changing this if you prefer

guizmaii: I don't have an opinion. Was just asking 🙂

vuepress/docs/docs/schema.md
kubukoz: true, I'll update that

codegen-sbt/src/main/scala/caliban/codegen/CalibanSourceGenerator.scala
kubukoz: I'm not sure what that'd look like, can you provide some pseudocode?

codegen-sbt/src/main/scala/caliban/codegen/CalibanSourceGenerator.scala
kubukoz: cool, I'd leave it as it is for now then :)

vuepress/docs/docs/client.md
kubukoz: Updated, hopefully it's correct now

ghostdogpr:
LGTM, I will merge!

Issue #1159, opened 2021-11-20
Implement additions from 2021 spec update 


ghostdogpr:
New spec: https://spec.graphql.org/October2021/
Changelog and diff: https://github.com/graphql/graphql-spec/blob/main/changelogs/October2021.md
- [x] Custom Scalar Specification URLs
- [x] Allow interfaces to implement other interfaces
- [x] Repeatable directives
- [x] Allow directives on variable definitions
- [x] `__typename` is not valid at subscription root
- [ ] Schema descriptions

ghostdogpr:
Working on `Custom Scalar Specification URLs`.

nukky94:
Hi, any updates on the remaining tasks, espacially implemeting `Repeatable directives`?
Thanks

ghostdogpr:
I don't think anyone has been working on it so far. Wanna give it a shot? Or you need it?

nukky94:
Yes, need it, I got a parsing error when trying to build the client for a schema which contains repeatable directives.

PR #1163, opened 2021-11-21
Spec 2021: __typename is not valid at subscription root

ghostdogpr:
Part of #1159


frekw (approved):
Took me a minute to figure out what was going on until I noticed the `.flip`. 😅 Looks good!

PR #1169, opened 2021-11-24
feat(client): provide a way to select interfaces fields

### LINKED ISSUE #237: Caliban client: better helper for interfaces
Currently for each field that returns an interface, we generate a function that takes an optional `SelectionBuilder` for each type that implements the interface. If we want to select fields from the interface (common to all types), we still need to provide a `SelectionBuilder` for each type.

We need something more user-friendly for that particular use case.

AlixBa:
Closes #237

### DIFF
--- a/client/src/test/scala/caliban/client/SelectionBuilderSpec.scala
+++ b/client/src/test/scala/caliban/client/SelectionBuilderSpec.scala
@@ -348,6 +348,50 @@ object SelectionBuilderSpec extends DefaultRunnableSpec {
           assert(query.toGraphQL().query)(
             equalTo("""query{addCharacter(character:{name:"name",nicknames:[]}){name}}""")
           )
+        },
+        test("interface common fields and subtype fields combination") {
+          import caliban.client.FieldBuilder._
+          import caliban.client._
+
+          type Order = String
+          object Order {
+            def name: SelectionBuilder[Order, String] = _root_.caliban.client.SelectionBuilder.Field("name", Scalar())
+          }
+
+          type Ascending = String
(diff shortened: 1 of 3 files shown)


AlixBa:
@ghostdogpr is the 2nd commit close enough to what you had in mind for #237?

ghostdogpr:
@AlixBa what I was thinking is a little different. Let's say that for each subtype you want to get the common fields from the interface, plus an extra field for one of the subtypes. To do that, you have to repeat the interface fields for each subtypes. So I was thinking about adding an extra parameter to the `option` function, that takes an optional `SelectionBuilder` of the interface. That way, for my example, you only need to pass this one and the `SelectionBuilder` for the subtype that needs an extra field. Does that make sense?

By the way I merged your other PR, so you can rebase to get rid of the conflicts.

ghostdogpr:
Ah, what I suggested doesn't work 😆 
```scala
def parentOption[A](
  onParent: Option[SelectionBuilder[Parent, A] = None,
  onSubType1: Option[SelectionBuilder[SubType1, A]] = None,
  onSubType2: Option[SelectionBuilder[SubType2, A]] = None
)
```
There is no way to merge the 2 A from the interface and the subtype.

So I think what you did is good. In my example, we can combine `parentInterface` and `parentOption` with `~` to select the common fields and the subtype ones.

AlixBa:
@ghostdogpr I've added one. Not sure it is the right place/what you were expecting though ^^

ghostdogpr:
Yep, that's good! Thanks!

PR #1171, opened 2021-11-25
Custom Scalar Specification URLs

ghostdogpr:
Part of #1159


core/src/main/scala/caliban/schema/Schema.scala
```diff
@@ -239,18 +246,18 @@ trait GenericSchema[R] extends SchemaDerivation[R] with TemporalSchema {
...
+  implicit val booleanSchema: Schema[Any, Boolean]       = scalarSchema("Boolean", None, None, BooleanValue.apply)
+  implicit val stringSchema: Schema[Any, String]         = scalarSchema("String", None, None, StringValue.apply)
+  implicit val uuidSchema: Schema[Any, UUID]             = scalarSchema("ID", None, None, uuid => StringValue(uuid.toString))
+  implicit val shortSchema: Schema[Any, Short]           = scalarSchema("Short", None, None, IntValue(_))
+  implicit val intSchema: Schema[Any, Int]               = scalarSchema("Int", None, None, IntValue(_))
+  implicit val longSchema: Schema[Any, Long]             = scalarSchema("Long", None, None, IntValue(_))
+  implicit val bigIntSchema: Schema[Any, BigInt]         = scalarSchema("BigInt", None, None, IntValue(_))
+  implicit val doubleSchema: Schema[Any, Double]         = scalarSchema("Float", None, None, FloatValue(_))
+  implicit val floatSchema: Schema[Any, Float]           = scalarSchema("Float", None, None, FloatValue(_))
+  implicit val bigDecimalSchema: Schema[Any, BigDecimal] = scalarSchema("BigDecimal", None, None, FloatValue(_))
```
paulpdaniels, line 259: I wonder if we can go ahead and provide urls for any of these. `UUID` for instance does have a specification I believe.

paulpdaniels (approved):

core/src/main/scala/caliban/schema/Schema.scala
ghostdogpr: You mean the official UUID spec, or something specific to GraphQL?

ghostdogpr: Ah, the spec links to `https://tools.ietf.org/html/rfc4122`

ghostdogpr: But we use `ID` for that, which is a standard scalar, and they say that standard scalars shouldn't have that URL.

core/src/main/scala/caliban/Rendering.scala
```diff
@@ -91,6 +94,11 @@ object Rendering {
     case Some(value)            => if (value.contains("\n")) s"""\"\"\"$value\"\"\" """ else s""""$value" """
   }
 
+  private def renderSpecifiedBy(specifiedBy: Option[String]): String = specifiedBy match {
```
frekw, line 97: nit/question: do we prefer to `fold` or `match` in these cases?

ghostdogpr: Good point, will use fold. Done.

core/src/main/scala/caliban/schema/Schema.scala
frekw: But maybe for the types in `TemporalSchema` could link to the relevant RFCs?

PR #1173, opened 2021-11-25
[WIP] Use a tail recursive algorithm for `calculateDepth`

guizmaii:


core/src/main/scala/caliban/wrappers/Wrappers.scala
```diff
@@ -97,23 +97,25 @@ object Wrappers {
...
+    @inline def depth(field: Field): Int = if (field.name.isEmpty) 0 else 1
+
+    @tailrec
+    def doCalculateDepth(acc: Int)(rest: List[Field]): Int =
+      rest match {
+        case Nil    => acc
+        case h :: t => doCalculateDepth(acc + depth(h))(t)
       }
+
+    doCalculateDepth(depth(field))(field.fields)
```
paulpdaniels, line 118: I think this will only go a single level deep no?

PR #1180, opened 2021-11-26
Add cost estimation wrapper

paulpdaniels:
Adds a simple cost estimation wrapper for computing the effective cost of executing a query and rejecting it if it is too expensive. Different from `maxDepth` or `maxFields` because it supports an arbitrary function to determine the cost of the field. Out of the box it provides a simple `cost` directive that can be added to fields or types to specify their overall cost to execute. Note, because it is a "speculative" cost it can't take advantage of the actual return values meaning that it relies on the user to apply the correct multipliers and to determine a good metric for assessing a fields cost.

### DIFF
--- a/core/src/main/scala/caliban/wrappers/CostEstimation.scala
+++ b/core/src/main/scala/caliban/wrappers/CostEstimation.scala
@@ -61,7 +61,7 @@ object CostEstimation {
               values.collect {
                 case StringValue(name) =>
                   f.arguments.get(name).collectFirst {
-                    case i: IntValue   => i.toInt
+                    case i: IntValue   => i.toInt.toDouble
                     case d: FloatValue => d.toDouble
                   }
                 case _                 => None
--- a/core/src/main/scala/caliban/wrappers/CostEstimation.scala
+++ b/core/src/main/scala/caliban/wrappers/CostEstimation.scala
@@ -1,0 +1,124 @@
+package caliban.wrappers
+
+import caliban.CalibanError.ValidationError
+import caliban.{ InputValue, Value }
+import caliban.InputValue.ListValue
+import caliban.Value.{ FloatValue, IntValue, StringValue }
+import caliban.execution.{ ExecutionRequest, Field }
+import caliban.parsing.adt.{ Directive, Document }
+import caliban.schema.Types
+import caliban.wrappers.Wrapper.ValidationWrapper
+import zio.{ IO, URIO, ZIO }
+
+import scala.annotation.tailrec
+
+object CostEstimation {
+
+  final val COST_DIRECTIVE_NAME = "cost"
+
+  object CostDirective {
+    def apply(cost: Double): Directive =
+      Directive(COST_DIRECTIVE_NAME, arguments = Map("weight" -> FloatValue(cost)))
+
+    def apply(cost: Double, multipliers: List[String]): Directive =
+      Directive(
+        COST_DIRECTIVE_NAME,
+        arguments = Map("weight" -> FloatValue(cost), "multipliers" -> ListValue(multipliers.map(StringValue)))
+      )
+  }
+
+  /**
+   * Computes field cost by examining the @cost directive
+   */
+  val costDirective = (f: Field) => {
+    def computeDirectiveCost(directives: List[Directive]) =
+      directives.collectFirst {
+        case d if d.name == COST_DIRECTIVE_NAME =>
+          // TODO should we support string backed formulas?
+          // TODO Support list arguments / multipliers
+          val weight = d.arguments
+            .get("weight")
+            .collect {
+              case i: IntValue   => i.toInt.toDouble
+              case f: FloatValue => f.toDouble
+            }
+            .getOrElse(1.0)
+
+          val multipliers = d.arguments
+            .get("multipliers")
+            .toList
+            .collect { case ListValue(values) =>
+              values.collect {
+                case StringValue(name) =>
+                  f.arguments.get(name).collectFirst {
+                    case i: IntValue   => i.toInt
+                    case d: FloatValue => d.toDouble
+                  }
+                case _                 => None
+              }.flatten
+            }
+            .flatten
+            .sum max 1.0
+
+          weight * multipliers
+      }
+
+    val directiveCost = computeDirectiveCost(f.directives).getOrElse(1.0)
+
+    val typeCost = Types
+      .innerType(f.fieldType)
+      .directives
+      .flatMap(computeDirectiveCost)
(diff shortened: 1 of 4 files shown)


core/src/main/scala/caliban/wrappers/CostEstimation.scala
```diff
@@ -0,0 +1,257 @@
...
+          } yield req
+    }
+
+  private def addCostToExtensions(
+    cost: Ref[Double]
+  )(resp: GraphQLResponse[CalibanError]): UIO[GraphQLResponse[CalibanError]] =
+    cost.get.map { total =>
+      resp.copy(
+        extensions = Some(
+          ObjectValue(
```
frekw, line 217: This is a really nice pattern! Previously I've used HTTP headers to communicate things like rate limits but this is much nicer.

core/src/test/scala/caliban/wrappers/CostEstimationSpec.scala
```diff
@@ -0,0 +1,155 @@
...
+
+object CostEstimationSpec extends DefaultRunnableSpec {
+  case class WithArgs(limit: Int)
+  case class E(value: Int)
+  @GQLDirective(CostDirective(10))
+  case class D(e: E, @GQLDirective(CostDirective(1, List("limit"))) f: List[String])
+  case class Test(
+    a: Int,
+    @GQLDirective(CostDirective(100)) b: UIO[Int],
+    @GQLDirective(CostDirective(5, multipliers = List("limit"))) c: WithArgs => UIO[List[Int]],
```
frekw, line 24: Didn't completely get the multipliers until I saw this, this is really nice.

vuepress/docs/docs/middleware.md
```diff
@@ -154,4 +156,37 @@ a federated graph:
...
+These wrappers are in the `CostEstimation` object which also comes with a special directive that can be used out of the box to instrument your
+schema for cost analysis.
+
+Given a schema in which different fields have different costs to execute, either because they require additional network or computing resources, or database access, or they 
+have some other dependency that makes them expensive to compute. You can add the `CostDirective` to your resolver like so:
+
+```scala
+case class SpokenLineArgs(offset: Int, limit: Int)
+
+@GQLDirective(CostDirective(2))
```
frekw, line 175: Is it possible to specify a directive in `CostEstimation`, like `GQLCost` that desugars to `@GQLDirective(CostDirective(2))`? I'm thinking that if you use this in an API this will be used very frequently in your schema so saving some typing might be worth it.

frekw (commented):
I think it looks really really nice. Both the directives and the multipliers are much nicer than other solutions  I've seen. Great work @paulpdaniels! 👏

vuepress/docs/docs/middleware.md
paulpdaniels: Hmm interesting idea, I'll look into if this is possible.

core/src/test/scala/caliban/wrappers/CostEstimationSpec.scala
paulpdaniels: Yeah I had another idea that I saw ibm uses which is to have the cost field be a string that can be coerced into a formula for computing the cost of the field, i.e. `5 * limit`, but parsing that seemed like it might be a bit difficult (if we allowed floats for instance) so I went for the easy path instead.

ghostdogpr:
(will review tomorrow or thursday, when work gets quieter 😄)

core/src/main/scala/caliban/wrappers/CostEstimation.scala
```diff
@@ -0,0 +1,257 @@
...
+
+  /**
+   * A more powerful version of [[queryCost]] which allows the field cost computation to return an effect instead of a plain value
+   * when computing the cost estimate.
+   * @param f The field cost estimate function
+   *
+   * @see queryCostZIOWith for a more powerful version
+   */
+  def queryCostZIO[R](f: Field => URIO[R, Double]): Wrapper[R] = EffectfulWrapper(
+    Ref.make(0.0).map { cost =>
```
paulpdaniels, line 120: I can probably extract this into a helper function actually

core/src/test/scala/caliban/wrappers/CostEstimationSpec.scala
```diff
@@ -0,0 +1,155 @@
...
+import caliban.wrappers.CostEstimation.CostDirective
+import zio.console.putStr
+import zio.test.environment.TestConsole
+import zio.test.{ assertTrue, DefaultRunnableSpec, ZSpec }
+import zio.{ Ref, UIO }
+
+object CostEstimationSpec extends DefaultRunnableSpec {
+  case class WithArgs(limit: Int)
+  case class E(value: Int)
+  @GQLDirective(CostDirective(10))
```
ghostdogpr, line 19: How about using the nice `apply` here too?

vuepress/docs/docs/middleware.md
ghostdogpr: Also use the new apply in the docs

vuepress/docs/docs/middleware.md
```diff
@@ -154,4 +156,37 @@ a federated graph:
...
+  @GQLDirective(CostDirective(100, multipliers = List("limit")))
+  spokenLines: SpokenLineArgs => UIO[List[String]]
+)
+
+case class Query(
+  @GQLDirective(CostDirective(5)) characters: UIO[List[Character]],
+)
+```
+
+In the above example we have provided a couple different examples. For instance, we can add the directive to both types and
```
ghostdogpr, line 188: Can you add an example of using the wrapper? (so that they can see they have to use `costDirective`).

core/src/main/scala/caliban/wrappers/CostEstimation.scala
```diff
@@ -0,0 +1,268 @@
...
+  }
+
+  /**
+   * Computes field cost by examining the @cost directive. This can be used in conjunction with [[queryCost]] or [[maxCost]]
+   * In order to compute the estimated cost of executing a query.
+   *
+   * @note This will be executed *before* the actual resolvers are called, which allows you to stop potentially expensive queries
+   *       from being run, but may also require more work on the developer part to determine the correct heuristic for estimating field cost.
+   */
+  val costDirective = (f: Field) => {
```
ghostdogpr, line 43: Can you add the return type?

ghostdogpr (commented):
This is great 👍 

I have some minor comments. What about the remaining TODOs?

ghostdogpr (approved):
Awesome!

PR #1186, opened 2021-12-02
Schema Gen: Add Option for Preserving Input Names

LaurenceWarne:
Hi!  This PR adds the `--preserveInputNames` option to the schema gen cli which will prompt the case classes of Input types to be generated with the `@GQLInputName("$name")` where `name` is the name of the Input type.

The motivation is that I would like the derived schema to be as close as possible to the `.gql` file used as input to the schema gen, so I will often name my input types like `FooInput`, but then this is changed in the derived schema to `FooInputInput` :slightly_smiling_face:

Thanks

### DIFF
--- a/codegen-sbt/src/main/scala/caliban/codegen/CalibanCli.scala
+++ b/codegen-sbt/src/main/scala/caliban/codegen/CalibanCli.scala
@@ -52,7 +52,7 @@ object CalibanCli {
 
   private val genSchemaHelpMsg =
     s"""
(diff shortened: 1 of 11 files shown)


ghostdogpr:
Looks great. How about making it the default actually? I think most users expect it to work that way.

LaurenceWarne:
> Looks great. How about making it the default actually? I think most users expect it to work that way.

Yeah that's what I originally thought, but would this break existing builds?  For example, someone using `FooArgs` in their schema file and then having their client use `FooArgsInput` - though this is a little peculiar I admit :smile: 

Happy to make it the default if you think it's not a big deal.

ghostdogpr:
Maybe let's keep it this way and I might change it in a future major version. Don't really want to break on a minor version 😄

LaurenceWarne:
Thanks!

PR #1195, opened 2021-12-09
Respect variables with no value provided

guymers:
When resolving a value for a variable, if a value has not been provided
and it has no default then do not include it instead of setting it to
`null`.

### DIFF
--- a/core/src/main/scala/caliban/execution/Field.scala
+++ b/core/src/main/scala/caliban/execution/Field.scala
@@ -97,20 +97,21 @@ object Field {
     variableDefinitions: List[VariableDefinition],
     variableValues: Map[String, InputValue]
   ): Map[String, InputValue] = {
-    def resolveVariable(value: InputValue): InputValue =
+    def resolveVariable(value: InputValue): Option[InputValue] =
       value match {
-        case InputValue.ListValue(values)   => InputValue.ListValue(values.map(resolveVariable))
+        case InputValue.ListValue(values)   =>
+          Some(InputValue.ListValue(values.flatMap(resolveVariable)))
         case InputValue.ObjectValue(fields) =>
-          InputValue.ObjectValue(fields.map { case (k, v) => k -> resolveVariable(v) })
+          Some(InputValue.ObjectValue(fields.flatMap { case (k, v) => resolveVariable(v).map(k -> _) }))
         case InputValue.VariableValue(name) =>
-          (for {
-            definition  <- variableDefinitions.find(_.name == name)
-            defaultValue = definition.defaultValue getOrElse NullValue
-            value        = variableValues.getOrElse(name, defaultValue)
-          } yield value) getOrElse NullValue
-        case value: Value                   => value
+          for {
+            definition <- variableDefinitions.find(_.name == name)
+            value      <- variableValues.get(name).orElse(definition.defaultValue)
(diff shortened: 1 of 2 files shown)


core/src/main/scala/caliban/execution/Field.scala
```diff
@@ -97,20 +97,21 @@ object Field {
...
+        case InputValue.ListValue(values)   =>
+          Some(InputValue.ListValue(values.flatMap(resolveVariable)))
         case InputValue.ObjectValue(fields) =>
-          InputValue.ObjectValue(fields.map { case (k, v) => k -> resolveVariable(v) })
+          Some(InputValue.ObjectValue(fields.flatMap { case (k, v) => resolveVariable(v).map(k -> _) }))
         case InputValue.VariableValue(name) =>
-          (for {
-            definition  <- variableDefinitions.find(_.name == name)
-            defaultValue = definition.defaultValue getOrElse NullValue
-            value        = variableValues.getOrElse(name, defaultValue)
```
frekw, line 109: nit 💅 : `getOrElse`

core/src/main/scala/caliban/execution/Field.scala
```diff
@@ -97,20 +97,21 @@ object Field {
...
-          } yield value) getOrElse NullValue
-        case value: Value                   => value
+          for {
+            definition <- variableDefinitions.find(_.name == name)
+            value      <- variableValues.get(name).orElse(definition.defaultValue)
```
frekw, line 109: nit 💅 : getOrElse

frekw (approved):
LGTM! ⭐

core/src/main/scala/caliban/execution/Field.scala
guymers: `definition.defaultValue` is an `Option` so `getOrElse` is not applicable?

ghostdogpr: Yeah I think your right

ghostdogpr:
Thanks!

core/src/main/scala/caliban/execution/Field.scala
frekw: But so is `variableValues.get(name)` since you're getting it from a map, right?

ghostdogpr: @frekw `getOrElse` expects an `A`, but `defaultValue` is an `Option[A]`. It's `orElse`, not `getOrElse`.

frekw: I meant `getOrElse` from Map :) https://scastie.scala-lang.org/jywLcoZARNG8KOrI7LDHqw

frekw: But in either case, doesn't matter much.

ghostdogpr: No, `b` is of type `Any` if you do that :P

https://scastie.scala-lang.org/YjZ63y0XRXSxAMGDC6K65Q

frekw: 😬

frekw: And also 🤦  I guess

PR #1198, opened 2021-12-10
[POC] Update zio-http to v1.0.0.0-RC18

guizmaii:


ghostdogpr:
I don't think we can do that without updating tapir first.

PR #1199, opened 2021-12-11
Fix fields merging and improve performance

ghostdogpr:
Fields were not properly merged during execution. I moved the merging to the validation phase and added some performance optimizations.

Before:
```
[info] Benchmark                               Mode  Cnt      Score      Error  Units
[info] GraphQLBenchmarks.fragmentsCaliban     thrpt    5     27.739 ±    2.741  ops/s
[info] GraphQLBenchmarks.fragmentsSangriaNew  thrpt    5     24.269 ±    4.216  ops/s
[info] GraphQLBenchmarks.fragmentsSangriaOld  thrpt    5      0.165 ±    0.011  ops/s
[info] GraphQLBenchmarks.introspectCaliban    thrpt    5   1585.284 ±  183.560  ops/s
[info] GraphQLBenchmarks.introspectSangria    thrpt    5    499.669 ±   19.512  ops/s
[info] GraphQLBenchmarks.simpleCaliban        thrpt    5  24539.809 ± 1188.386  ops/s
[info] GraphQLBenchmarks.simpleSangria        thrpt    5   7551.214 ± 1423.378  ops/s
```
After:
```
[info] Benchmark                               Mode  Cnt      Score     Error  Units
[info] GraphQLBenchmarks.fragmentsCaliban     thrpt    5     44.638 ±   1.523  ops/s
[info] GraphQLBenchmarks.fragmentsSangriaNew  thrpt    5     28.994 ±  16.953  ops/s
[info] GraphQLBenchmarks.fragmentsSangriaOld  thrpt    5      0.155 ±   0.039  ops/s
[info] GraphQLBenchmarks.introspectCaliban    thrpt    5   1782.754 ± 309.681  ops/s
[info] GraphQLBenchmarks.introspectSangria    thrpt    5    500.839 ±  15.303  ops/s
[info] GraphQLBenchmarks.simpleCaliban        thrpt    5  26091.734 ± 305.640  ops/s
[info] GraphQLBenchmarks.simpleSangria        thrpt    5   7831.910 ± 139.665  ops/s
```

### DIFF
--- a/core/src/main/scala/caliban/execution/Executor.scala
+++ b/core/src/main/scala/caliban/execution/Executor.scala
@@ -1,7 +1,6 @@
 package caliban.execution
 
 import scala.annotation.tailrec
-import scala.collection.mutable.ArrayBuffer
 import caliban.CalibanError.ExecutionError
 import caliban.ResponseValue._
 import caliban.Value._
@@ -52,7 +51,7 @@ object Executor {
           value match {
             case EnumValue(v) =>
               // special case of an hybrid union containing case objects, those should return an object instead of a string
-              val obj = mergeFields(currentField, v).collectFirst {
+              val obj = filterFields(currentField, v).collectFirst {
                 case f: Field if f.name == "__typename" =>
                   ObjectValue(List(f.alias.getOrElse(f.name) -> StringValue(v)))
                 case f: Field if f.name == "_"          =>
@@ -71,8 +70,8 @@ object Executor {
             Types.listOf(currentField.fieldType).fold(false)(_.isNullable)
           )
         case ObjectStep(objectName, fields) =>
-          val mergedFields = mergeFields(currentField, objectName)
-          val items        = mergedFields.map {
+          val filteredFields = filterFields(currentField, objectName)
+          val items          = filteredFields.map {
             case f @ Field(name @ "__typename", _, _, alias, _, _, _, _, directives) =>
               (alias.getOrElse(name), PureStep(StringValue(objectName)), fieldInfo(f, path, directives))
             case f @ Field(name, _, _, alias, _, _, args, _, directives)             =>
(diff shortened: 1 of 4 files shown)


core/src/main/scala/caliban/execution/Executor.scala
```diff
@@ -173,28 +172,8 @@ object Executor {
...
 
-  private[caliban] def mergeFields(field: Field, typeName: String): List[Field] = {
-    // ugly mutable code but it's worth it for the speed ;)
-    val array = ArrayBuffer.empty[Field]
-    val map   = collection.mutable.Map.empty[String, Int]
-
-    field.fields.foreach { field =>
-      if (field.condition.forall(_.contains(typeName))) {
-        val name = field.alias.getOrElse(field.name)
-        map.get(name) match {
```
ghostdogpr, line 184: `map` was never updated 🤦‍♂️

core/src/test/scala/caliban/execution/ExecutionSpec.scala
```diff
@@ -871,6 +870,9 @@ object ExecutionSpec extends DefaultRunnableSpec {
             |    ... on Human {
             |      height
             |    }
+            |    ... on Human {
```
ghostdogpr, line 873: Added this intentionally to make sure the result has the field only once.

core/src/main/scala/caliban/execution/Executor.scala
frekw: Good catch!

frekw (approved):
I don't know the execution code in detail but I think it looks sensible.

The performance optimizations in the validation also make the code easier to read so a real win-win 🤝

core/src/main/scala/caliban/execution/Field.scala
```diff
@@ -35,7 +36,33 @@ object Field {
     rootType: RootType
   ): Field = {
     def loop(selectionSet: List[Selection], fieldType: __Type): Field = {
-      val fieldList = List.newBuilder[Field]
+      val fieldList  = ArrayBuffer.empty[Field]
```
paulpdaniels, line 39: Should we give this a size hint as well? I'd expect that to provide a minor improvement in 99% of cases where people don't duplicate their fields (and really only penalizes us in the pathological case).

paulpdaniels (approved):

core/src/main/scala/caliban/execution/Field.scala
ghostdogpr: But we don't know the size without going through fragments recursively.

PR #1203, opened 2021-12-13
Fix zio-http authentication example

guizmaii:


examples/src/main/scala/example/ziohttp/AuthExampleApp.scala
```diff
@@ -52,7 +61,7 @@ object Auth {
               case _                              => None
             })
             .orElseFail(CalibanError.ExecutionError("Unable to decode payload"))
-            .flatMap(user => ZIO.service[Auth].flatMap(_.setUser(user).orDie))
+            .flatMap(auth.setUser)
```
guizmaii, line 64: It was not the correct `Auth` instance used

examples/src/main/scala/example/ziohttp/AuthExampleApp.scala
```diff
@@ -7,39 +7,48 @@ import caliban.interop.tapir.{ StreamTransformer, WebSocketHooks }
...
+          def setUser(name: String): UIO[Unit]      = ref.set(name)
+        }
       }
-    }
-    .toLayer
+      .toLayer
 
   object WebSockets {
-    val wsSession = Http.fromEffect(Ref.make[String]("unknown"))
+    private val wsSession = Http.fromEffect(Ref.make[String](null))
```
ghostdogpr, line 42: How about `Option[String]` and no null 😄

guizmaii: `Option` creates a useless allocation. In terms of performances, `null` is better solution here

ghostdogpr: It's insignificant compared to everything that happens when handling a query. I think it's better that examples look cleaner.

guizmaii: Changed for Option

frekw (approved):
Nice!

ghostdogpr:
Thanks!

PR #1212, opened 2021-12-17
Flatten SelectionBuilder zipping

ghostdogpr:
(Not to be merged until 1.4.0 as it breaks some code)


client/src/main/scala/caliban/client/Zippable.scala
```diff
@@ -0,0 +1,401 @@
+package caliban.client
+
+trait Zippable[-A, -B] {
```
paulpdaniels, line 3: I think we need the source attribution for ZIO.

ghostdogpr: Yep, added!

paulpdaniels (approved):

PR #1219, opened 2021-12-20
Expose federation service schema

Fluxx:
The Apollo federation spec requires a federated subgraph expose a `_service: _Service` field on the subgraph's query type. This [`_Service` type](https://www.apollographql.com/docs/federation/federation-spec/#type-_service) has a `sdl: String!` field, that is the SDL of the service's schema. This "service schema" is the schema of the subgraph + any entity types the subgraph resolves. It does _not_ include any of the extra federation types (i.e. `_Any`) or fields (i.e. `_service`, `_entities`). 

I needed to expose the service schema (to work with Apollo validation), but was unable to see a way to do that. Currently, the `federate` helper returns a `GraphQL` object, which only has a `render` method on it that renders the full graph schema, including the extra federation types.

This PR makes an attempt to expose that service schema by introducing a new `FederatedGraphQL` type that the `federate` helper returns. `FederatedGraphQL` has a `service: GraphQL` value on it, which is the "service schema."

The easiest way I saw to do this was to inline creation of the `FederatedGraphQL` instance for the two `federate` methods, delegating most methods to the underlying `GraphQL` object. I had to adjust the field visibility in order to access those underlying fields.

This was just my first stab at this, so happy to alter approach or make other changes.

### DIFF
--- a/core/src/main/scala/caliban/GraphQL.scala
+++ b/core/src/main/scala/caliban/GraphQL.scala
@@ -171,7 +171,7 @@ trait GraphQL[-R] { self =>
    */
   final def combine[R1 <: R](that: GraphQL[R1]): GraphQL[R1] =
     new GraphQL[R1] {
-      override private[caliban] val schemaBuilder: RootSchemaBuilder[R1]              = self.schemaBuilder |+| that.schemaBuilder
+      override private[caliban] val schemaBuilder: RootSchemaBuilder[R1]    = self.schemaBuilder |+| that.schemaBuilder
       override private[caliban] val wrappers: List[Wrapper[R1]]             = self.wrappers ++ that.wrappers
       override private[caliban] val additionalDirectives: List[__Directive] =
         self.additionalDirectives ++ that.additionalDirectives
--- a/core/src/main/scala/caliban/GraphQL.scala
+++ b/core/src/main/scala/caliban/GraphQL.scala
@@ -22,9 +22,9 @@ import zio.{ IO, URIO }
  */
 trait GraphQL[-R] { self =>
 
-  protected val schemaBuilder: RootSchemaBuilder[R]
-  protected val wrappers: List[Wrapper[R]]
-  protected val additionalDirectives: List[__Directive]
+  private[caliban] val schemaBuilder: RootSchemaBuilder[R]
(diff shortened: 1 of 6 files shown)


paulpdaniels:
So one easy way you can do supply the "validation" schema is to just preserve the api *before* federating and use the `.withAdditionalTypes(Types.collectTypes()` on the `EntityResolver#toType` that will give you a properly augmented schema for apollo schema checks.

I do agree with the general idea (that we should be able to render the federated schema _without_ the extra types) but I think it would be more generalize if we could do this by augmenting the `render` method to accept some sort of filtering.

Fluxx:
> So one easy way you can do supply the "validation" schema is to just preserve the api _before_ federating and use the `.withAdditionalTypes(Types.collectTypes()` on the `EntityResolver#toType` that will give you a properly augmented schema for apollo schema checks.

Totally. This is actually what we're doing now internally to access the service schema.

> I do agree with the general idea (that we should be able to render the federated schema _without_ the extra types) but I think it would be more generalize if we could do this by augmenting the `render` method to accept some sort of filtering.

Oh that is a good idea. We don't currently preserve the federated schema without the extra type anywhere in the `GraphQL` type, so we'd have to add it (I guess similar to how this PR does), but making the API `render()` sounds good to me.

Fluxx:
@paulpdaniels I rebased off master and formatted the code. Let me know if you think I should do the `render()` with arguments idea - happy do make those changes.

paulpdaniels:
@Fluxx yeah I would personally prefer modifying `render` since that could be used in other use cases as well.

I'm not sure exactly what the interface would need to look like, either wrapping each type in some sort of sealed hierarchy so that we could do:

```scala
.render {
case AField(field) =>  
}
```

or maybe just encode them all as separate parameters to the overload:

```scala
.render(
  field = (f: Field) => true,
  directive = (_: Directive) => true
)
```

I think ideally we need some sort of visitor pattern so that we could handle a variety of use cases.

Fluxx:
Hey @paulpdaniels,

Wanted to let you know I still plan to address this. I was out of town the end of last week and am just now getting caught up. I should be able to get to this this week I think.

ghostdogpr:

PR #1220, opened 2021-12-20
Do not render emplty field list

Fluxx:
According to [the GraphQL object spec](https://spec.graphql.org/June2018/#sec-Objects), for a given `type Foo` the fields definition for that type is optional. As far as I can tell, if you want to not specify a fields list, you need to also omit the wrapping `{...}`. i.e:

```graphql
// This is invalid
type Foo {
}

// This is valid
type Foo
```

Currently, Caliban always prints schema with the wrapping `{...}`, even if there are no fields for the type, which I believe is invalid. I tested this against our schema in Apollo and saw the below. Sensitive parts omitted, but the important part is the AST parse failure.

```
$ echo 'schema {query: Q} type Q {}' | rover subgraph check [...]
Checking the proposed schema for subgraph [...]
error: Could not parse partial schema as AST.

$ echo 'schema {query: Q} type Q' | rover subgraph check [...]
Checking the proposed schema for [...]
Compared 108 schema changes against 25 operations
```

### DIFF
--- a/core/src/main/scala/caliban/GraphQL.scala
+++ b/core/src/main/scala/caliban/GraphQL.scala
@@ -32,15 +32,23 @@ trait GraphQL[-R] { self =>
   /**
    * Returns a string that renders the API types into the GraphQL format.
    */
-  final def render: String =
-    s"""schema {
-       |${schemaBuilder.query.flatMap(_.opType.name).fold("")(n => s"  query: $n\n")}${schemaBuilder.mutation
-      .flatMap(_.opType.name)
-      .fold("")(n => s"  mutation: $n\n")}${schemaBuilder.subscription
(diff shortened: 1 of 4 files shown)


ghostdogpr:
Looks okay! The CI failed because of formatting, can you run `fmt` in sbt?

Fluxx:
Will do.

Also, we ran into a similar problem where an empty `schema` also causes issues, where both `schema` and `schema {}` are invalid. I will see if I can make that change as part of this PR as well.

Fluxx:
Ran `fmt` and also added some behavior to handle empty schemas. Please take a look!

ghostdogpr (approved):
Thanks!

PR #1223, opened 2021-12-22
Root query: generate `{x}Option` method for fields with an optional interface type

iRevive:
`ClientWriter` does not handle optional interfaces properly of the root query.  

I fixed the processing of the root query in the scope of the PR. If the output type of a field is an interface, the `ClientWriter` will generate two methods: `def {x}(...)` and `def {x}Option(...)`. 

### Schema 1 (without root schema)

```diff
type Queries {
  node(id: ID!): Node
}
interface Node {
  id: ID!
}
type NodeA implements Node {
  id: ID!
  a: String
}
type NodeB implements Node {
  id: ID!
  b: Int
}
```

`ClientWriter` generates two methods: `node(id: ID)(...)` and `nodeOption(id: ID)(...)`. 

### Schema 2 (with top-level `schema` definition)

```diff
+ schema {
+  query: Queries
+}
type Queries {
  node(id: ID!): Node
}
interface Node {
  id: ID!
}
type NodeA implements Node {
  id: ID!
  a: String
}
type NodeB implements Node {
  id: ID!
  b: Int
}
```

`ClientWriter` generates only one method: `node(id: ID)(...)`.

### DIFF
--- a/tools/src/main/scala/caliban/tools/ClientWriter.scala
+++ b/tools/src/main/scala/caliban/tools/ClientWriter.scala
@@ -383,7 +383,7 @@ object ClientWriter {
               field,
               "_root_.caliban.client.Operations.RootMutation",
               optionalUnion = false,
-              optionalInterface = true,
+              optionalInterface = false,
               commonInterface = false
             )
           )
--- a/tools/src/main/scala/caliban/tools/ClientWriter.scala
+++ b/tools/src/main/scala/caliban/tools/ClientWriter.scala
@@ -88,6 +88,15 @@ object ClientWriter {
       safeTypeName(name) -> op
     }.toMap
 
+    val knownInterfaceTypes = typesMap.collect { case (key, _: InterfaceTypeDefinition) => key }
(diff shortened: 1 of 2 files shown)


ghostdogpr (commented):
Looks good! Shouldn't the same thing be applied to mutations and subscriptions?

iRevive:
Good point. I guess mutations and subscriptions should be updated as well. I will push an update soon.

iRevive:
Done

tools/src/main/scala/caliban/tools/ClientWriter.scala
```diff
@@ -330,16 +359,36 @@ object ClientWriter {
...
+            )
           )
-        )
+        } else {
+          Vector(
+            writeField(
+              field,
+              "_root_.caliban.client.Operations.RootMutation",
+              optionalUnion = false,
+              optionalInterface = true,
```
ghostdogpr, line 386: Should be false I think?

iRevive: Fixed, thanks!

ghostdogpr (approved):
Thanks!

PR #1246, opened 2022-01-09
Add contextual interop between cats-effect and ZIO

iRevive:
## Summary 

Currently, it's hard to share contextual details between ZIO and cats-effect code. 

In my particular scenario, I need to solve two problems:
1) Share `TraceId` between cats-effect and ZIO effects to keep the logs relevant
2) Propagate `Span` from cats-effect to ZIO to build a valid trace 

## Solution

The contextual effect (e.g. `Kleisli[IO, Env, *]`) can be partially aligned with `RIO[Env, *]`.

Utility type classes can solve the described issues: `ToEffect`, `FromEffect`, and `CatsInterop` - a combination of the first two.

The instance available in the implicit scope defines the interop behavior between ZIO and cats-effect.

## Notes

- The changes are binary breaking
- The transition for end-users should be seamless since `ToEffect` and `FromEffect` are derived from `Async` and `Dispatcher` implicitly.
- ~~Perhaps, a better place for this change is https://github.com/zio/interop-cats.~~ 
- `examples/ContextualCatsInterop.scala` - a minimal example.


ghostdogpr:
Will have a detailed look during the weekend 🙏

iRevive:
Thanks! It's mostly proof of concept, hence some polishing may be required.

interop/cats/src/main/scala/caliban/interop/cats/FromEffect.scala
```diff
@@ -0,0 +1,49 @@
+package caliban.interop.cats
+
+import cats.~>
+import cats.effect.Async
+import cats.effect.std.Dispatcher
+import zio.{ RIO, Runtime }
+
+trait ToEffect[F[_], R] {
```
ghostdogpr, line 8: The file is named `FromEffect.scala`, maybe rename of split this into 2 files?

ghostdogpr (commented):
Looks great to me! Could you add an example for http4s (a mix between existing `ExampleAppF` and `AuthExampleApp`, `AuthExampleAppF`) that would use this mechanism?

Also update the docs about cats interop?

interop/cats/src/main/scala/caliban/interop/cats/FromEffect.scala
iRevive: I split this into 2 files and updated docs.

iRevive:
> Could you add an example for http4s (a mix between existing ExampleAppF and AuthExampleApp, AuthExampleAppF) that would use this mechanism?

Done 

> Also update the docs about cats interop?

Done

ghostdogpr (approved):
Looks great to me. I think I'd rather wait until 1.4.0 because of the breaking changes (similarly to https://github.com/ghostdogpr/caliban/pull/1212) so I'll keep this open. Do you need it quickly?

iRevive:
@ghostdogpr no, this update is not urgent. 1.4.0 sounds like a good plan. 

Perhaps we can experiment with a release candidate and test this particular update on public projects?

ghostdogpr:
I need to figure out how I do this, with also the new zio 2.0 branch going on. I'll have a week-long vacation the first week of February, I'll probably finish some stuff and cut releases around then.

ghostdogpr:
I'll do a 1.3.3 release today, then let's merge this as well as other 1.4.x stuff 😄

ghostdogpr:
Would you mind moving the docs changes to a separate PR? So that I merge the code now but the docs at the next release.

iRevive:
@ghostdogpr done. I made a separate PR https://github.com/ghostdogpr/caliban/pull/1279

PR #1247, opened 2022-01-10
Fix compilation of projects using the ConpileTimeCaliban sbt plugin and the scalac `-Xfatal-warnings` flag

guizmaii:


ghostdogpr (approved):
Thanks!

PR #1254, opened 2022-01-12
use fully-qualified name for Vector in generated code

cornerman:
This is to avoid name-clashes with types from the graphql schema, i.e.,
if a type is called Vector.

Example:
enum ImageType {
  Raster
  Vector
}

### DIFF
--- a/tools/src/main/scala/caliban/tools/ClientWriter.scala
+++ b/tools/src/main/scala/caliban/tools/ClientWriter.scala
@@ -703,7 +703,7 @@ object ClientWriter {
             ${encoderCases.mkString("\n")}
           }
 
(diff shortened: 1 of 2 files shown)


cornerman:
We encountered some edge cases with the line length - regarding the scalafmt config. Therefore, we had to add a line-break in one of the test cases. Does it make sense to pass the expected-code through the formatter as well?

ghostdogpr (approved):

ghostdogpr:
> We encountered some edge cases with the line length - regarding the scalafmt config. Therefore, we had to add a line-break in one of the test cases. Does it make sense to pass the expected-code through the formatter as well?

Hmm, I think that's fine as it is. Might be difficult to find differences if we call the formatter on the expected code.

PR #1264, opened 2022-01-17
[WIP] CompileTime Plugin: Add Scala version to the generation cache

### LINKED ISSUE #1261: CompileTime Plugin does not support cross-building nor changing scala version
Test case is simple:

1. go to codegen-sbt/src/sbt-test/compiletime-codegen/test-compile
2. launch sbt
3. force Scala to 2.13.7 with `++ 2.13.7!`
4. compile. It works
5. Go back to the normal Scala version with `++ 2.12.14`
6. compile and it does not work because it does not generated the client files

If done the other way around, 2.12.14 works but not 2.13.7.

guizmaii:
TODO @guizmaii: Add scripted to tests to validate that changing the Scala version of the project triggers a re-generation of the Caliban client


ghostdogpr:

guizmaii:
See #1415

PR #1266, opened 2022-01-18
support uuid as arg encoder

smiklos:
Added support for UUID to the ArgEncoder typeclass hierarchy. 

With this clients don't need to turn their UUID to String and can use UUID values directly

### DIFF
--- a/client/src/main/scala/caliban/client/ArgEncoder.scala
+++ b/client/src/main/scala/caliban/client/ArgEncoder.scala
@@ -1,9 +1,18 @@
 package caliban.client
 
-import caliban.client.__Value.{ __BooleanValue, __ListValue, __NullValue, __NumberValue, __ObjectValue, __StringValue }
+import caliban.client.__Value.{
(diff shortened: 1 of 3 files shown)


client/src/main/scala/caliban/client/ArgEncoder.scala
```diff
@@ -42,6 +51,8 @@ object ArgEncoder {
 
   implicit val unit: ArgEncoder[Unit] = (_: Unit) => __ObjectValue(Nil)
 
+  implicit val uuid: ArgEncoder[UUID] = (value: UUID) => __UUIDValue(value)
```
ghostdogpr, line 54: You could just use `__StringValue` here, no need to add a new primitive `__UUIDValue`.

smiklos: Done

ghostdogpr (approved):
Thanks!

PR #1268, opened 2022-01-20
Configure mdoc and fix documentation

iRevive:
I found some outdated sections in the documentation. Several examples were not compiling.

https://scalameta.org/mdoc should prevent such a scenario.

~I'm not familiar with Circle CI, I hope I added a new job correctly.~

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -451,8 +451,10 @@ lazy val docs = project
     scalaVersion       := scala213,
     crossScalaVersions := Seq(scala213),
     name               := "caliban-docs",
-    mdocIn             := file(".") / "vuepress" / "docs",
+    mdocIn             := (ThisBuild / baseDirectory).value / "vuepress" / "docs",
+    run / fork         := true,
     scalacOptions -= "-Xfatal-warnings",
+    scalacOptions += "-Wunused:imports",
     libraryDependencies ++= Seq(
       "com.softwaremill.sttp.client3" %% "async-http-client-backend-zio" % sttpVersion,
       "io.circe"                      %% "circe-generic"                 % circeVersion
--- a/build.sbt
+++ b/build.sbt
@@ -443,6 +443,23 @@ lazy val federation = project
     )
   )
 
+lazy val docs = project
+  .in(file("mdoc"))
+  .enablePlugins(MdocPlugin)
+  .settings(commonSettings)
+  .settings(
+    scalaVersion       := scala213,
+    crossScalaVersions := Seq(scala213),
+    name               := "caliban-docs",
+    mdocIn             := file(".") / "vuepress" / "docs",
+    scalacOptions -= "-Xfatal-warnings",
+    libraryDependencies ++= Seq(
+      "com.softwaremill.sttp.client3" %% "async-http-client-backend-zio" % sttpVersion,
+      "io.circe"                      %% "circe-generic"                 % circeVersion
+    )
+  )
+  .dependsOn(core, catsInterop, tapirInterop, http4s, tools)
+
 lazy val commonSettings = Def.settings(
   scalacOptions ++= Seq(
     "-deprecation",
--- a/vuepress/docs/docs/README.md
+++ b/vuepress/docs/docs/README.md
@@ -37,16 +37,16 @@ Creating a GraphQL API with Caliban is as simple as creating a case class. Indee
 
 Let's say we have a class `Character` and 2 functions: `getCharacters` and `getCharacter`:
 
-```scala
+```scala mdoc:silent
 case class Character(name: String, age: Int)
 
-def getCharacters: List[Character] = ???
+def getCharacters: List[Character] = Nil
 def getCharacter(name: String): Option[Character] = ???
 ```
 
 Let's create a case class named `Queries` that will represent our API, with 2 fields named and modeled after the functions we want to expose (a _record of functions_). We then create a value of this class that calls our actual functions. This is our resolver.
 
-```scala
+```scala mdoc:silent
 // schema
 case class CharacterName(name: String)
 case class Queries(characters: List[Character],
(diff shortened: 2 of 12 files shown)


ghostdogpr:
Great idea! Will check in details this weekend

vuepress/docs/docs/README.md
```diff
@@ -37,16 +37,16 @@ Creating a GraphQL API with Caliban is as simple as creating a case class. Indee
 
 Let's say we have a class `Character` and 2 functions: `getCharacters` and `getCharacter`:
 
-```scala
+```scala mdoc:silent
 case class Character(name: String, age: Int)
 
-def getCharacters: List[Character] = ???
+def getCharacters: List[Character] = Nil
```
ghostdogpr, line 43: Why can't this one be `???` and the next line can? This compiles, right?

vuepress/docs/docs/README.md
```diff
@@ -139,20 +140,24 @@ You can use `.rename` to change the names of the generated root types.
...
 
 Similarly, subscriptions are passed as the third argument to `RootResolver`:
 
-```scala
+```scala mdoc:nest:silent
+import zio.stream.ZStream
+
 case class Subscriptions(deletedCharacter: ZStream[Any, Nothing, Character])
-val subscriptions = Subscriptions(???)
+val subscriptions = Subscriptions(ZStream())
```
ghostdogpr, line 160: Doesn't it compile with `???`?

vuepress/docs/docs/interop.md
```diff
@@ -111,8 +111,19 @@ The conversion rules are the following:
 - there should be an implicit `Schema` for both the input and the output types and an implicit `ArgBuilder` for the input type (see the [dedicated docs](schema.md))
 
 Let's look at an example. Imagine we have the following Tapir endpoint:
-```scala
-val addBook: Endpoint[(Book, String), Nothing, Unit, Nothing] =
+```scala mdoc:silent:reset
+import caliban._
```
ghostdogpr, line 115: Is it mandatory to have all imports visible? It goes a little against the flow of the tutorial. How about adding needed imports in each snippet progressively?

ghostdogpr (commented):
This is awesome, there was a lot of not up-to-date stuff 😱

vuepress/docs/docs/README.md
iRevive: Yes, unfortunately. 

mdoc compiles and **evaluates** the snippet. 

There is a modifier [compile-only](https://scalameta.org/mdoc/docs/modifiers.html#compile-only), but the methods/variables will not be available in the sections below.

vuepress/docs/docs/README.md
iRevive: Since it's the last section, we can use `compile-only` trick

vuepress/docs/docs/interop.md
iRevive: Makes sense. Done

vuepress/docs/docs/interop.md
```diff
@@ -126,23 +133,30 @@ val addBook: Endpoint[(Book, String), Nothing, Unit, Nothing] =
...
-If you want to reuse `bookAddLogic` for both GraphQL and regular HTTP, you can turn your `Endpoint` into a `ServerEndpoint` by calling `.serverLogic`:
-```scala
-val addBookEndpoint: ServerEndpoint[(Book, String), Nothing, Unit, Nothing, UIO] =
-  addBook.serverLogic[UIO] { case (book, token) => bookAddLogic(book, token).either }
+If you want to reuse `bookAddLogic` for both GraphQL and regular HTTP, you can turn your `Endpoint` into a `ServerEndpoint` by calling `.serverLogicSuccess`:
+```scala mdoc:silent
+import sttp.tapir.server.ServerEndpoint
+
+val addBookEndpoint: ServerEndpoint.Full[Unit, Unit, (Book, String), Nothing, Unit, Any, UIO] =
+  addBook.serverLogicSuccess[UIO] { case (book, token) => bookAddLogic(book, token) }
```
iRevive, line 156: If we want to keep using `addBook.serverLogic` in the example, the effect type should be changed (since `UIO` cannot fail).

[tapir/ExampleApp](https://github.com/ghostdogpr/caliban/blob/master/examples/src/main/scala/example/tapir/ExampleApp.scala) uses `IO[String, Unit]` for demonstration purposes.

ghostdogpr (approved):
Thanks!

build.sbt
```diff
@@ -443,6 +443,25 @@ lazy val federation = project
...
+  .in(file("mdoc"))
+  .enablePlugins(MdocPlugin)
+  .settings(commonSettings)
+  .settings(
+    scalaVersion       := scala213,
+    crossScalaVersions := Seq(scala213),
+    name               := "caliban-docs",
+    mdocIn             := (ThisBuild / baseDirectory).value / "vuepress" / "docs",
+    run / fork         := true,
+    scalacOptions -= "-Xfatal-warnings",
```
iRevive, line 456: There is only one warning, actually:

```
warning: middleware.md:133:55: The outer reference in this type test cannot be checked at run time.
  case err @ ExecutionError(_, _, _, Some(exampleError: ExampleAppEncodableError), _) =>
```

But I cannot find a simple way to sort it out.

ghostdogpr: Yeah I saw it too. That's good enough 🤷‍♂️

ghostdogpr:
@iRevive I just noticed that with this, the project fails to load in IntelliJ. I have to comment the `docs` block in build.sbt for it to load properly.

It looks like it's looking for a snapshot version:
```
[warn] 	Note: Unresolved dependencies path:
[error] stack trace is suppressed; run last docs / update for the full output
[error] stack trace is suppressed; run last docs / ssExtractDependencies for the full output
[error] (docs / update) sbt.librarymanagement.ResolveException: Error downloading com.github.ghostdogpr:caliban-http4s_2.13:1.3.2+21-9a980ebd-SNAPSHOT
...
```

Any idea why?

iRevive:
@ghostdogpr 

hm, I can assume it's caused by `crossScalaVersion` and `scalaVersion` configurations in `doc` module.

I will investigate it locally

PR #1270, opened 2022-01-24
Render directives at the top of the schema 

Fluxx:
Any custom `__Directives` added to the schema as part of the `GraphQL.graphQL` method do not render when the `graph.render()` method is called. This PR adds support for that.

### DIFF
--- a/core/src/main/scala/caliban/Rendering.scala
+++ b/core/src/main/scala/caliban/Rendering.scala
@@ -79,6 +79,44 @@ object Rendering {
       }
       .mkString("\n\n")
 
+  def renderDirectives(directives: List[__Directive]): String =
+    directives.map(renderDirective).mkString("\n")
+
+  private def renderDirective(directive: __Directive): String = {
+    val inputs             = directive.args match {
+      case i if i.nonEmpty => s"""(${i.map(renderInputValue).mkString(", ")})"""
+      case _               => ""
+    }
+    val locationStrings    = directive.locations.map {
+      case __DirectiveLocation.QUERY                  => "QUERY"
+      case __DirectiveLocation.MUTATION               => "MUTATION"
+      case __DirectiveLocation.SUBSCRIPTION           => "SUBSCRIPTION"
+      case __DirectiveLocation.FIELD                  => "FIELD"
+      case __DirectiveLocation.FRAGMENT_DEFINITION    => "FRAGMENT_DEFINITION"
+      case __DirectiveLocation.FRAGMENT_SPREAD        => "FRAGMENT_SPREAD"
+      case __DirectiveLocation.INLINE_FRAGMENT        => "INLINE_FRAGMENT"
+      case __DirectiveLocation.SCHEMA                 => "SCHEMA"
+      case __DirectiveLocation.SCALAR                 => "SCALAR"
+      case __DirectiveLocation.OBJECT                 => "OBJECT"
+      case __DirectiveLocation.FIELD_DEFINITION       => "FIELD_DEFINITION"
+      case __DirectiveLocation.ARGUMENT_DEFINITION    => "ARGUMENT_DEFINITION"
+      case __DirectiveLocation.INTERFACE              => "INTERFACE"
+      case __DirectiveLocation.UNION                  => "UNION"
+      case __DirectiveLocation.ENUM                   => "ENUM"
+      case __DirectiveLocation.ENUM_VALUE             => "ENUM_VALUE"
+      case __DirectiveLocation.INPUT_OBJECT           => "INPUT_OBJECT"
+      case __DirectiveLocation.INPUT_FIELD_DEFINITION => "INPUT_FIELD_DEFINITION"
+    }
+    val directiveLocations = locationStrings.mkString(" | ")
+
+    val body = s"""directive @${directive.name}${inputs} repeatable on ${directiveLocations}""".stripMargin
+
+    renderDescription(directive.description) match {
+      case ""        => body
+      case something => something + body
+    }
+  }
+
   private def renderInterfaces(t: __Type): String =
     t.interfaces()
       .fold("")(_.flatMap(_.name) match {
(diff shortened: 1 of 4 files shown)


core/src/test/scala/caliban/RenderingSpec.scala
```diff
@@ -11,8 +11,11 @@ object RenderingSpec extends DefaultRunnableSpec {
   override def spec: ZSpec[TestEnvironment, Any] =
     suite("rendering")(
       test("it should render directives") {
-        assert(graphQL(resolver).render.trim)(
-          equalTo("""schema {
+        assert(graphQL(resolver, directives = List(Directives.Test)).render.trim)(
+          equalTo(""""Test directive"
```
Fluxx, line 15: I've been playing around with graphql-java, and one thing they do for their schema render is they render the built-in directives (`@skip`, `@include`, `@deprecated`, `@specifiedBy`) by default in every schema, even if they are not specified. Did we have any interest in doing this? I think [they may be optional according to the spec](https://spec.graphql.org/October2021/#sel-FAHnBPLCAACCcooU)? But might be a nice thing to do?

core/src/main/scala/caliban/Rendering.scala
```diff
@@ -79,6 +79,44 @@ object Rendering {
...
+      case __DirectiveLocation.SCALAR                 => "SCALAR"
+      case __DirectiveLocation.OBJECT                 => "OBJECT"
+      case __DirectiveLocation.FIELD_DEFINITION       => "FIELD_DEFINITION"
+      case __DirectiveLocation.ARGUMENT_DEFINITION    => "ARGUMENT_DEFINITION"
+      case __DirectiveLocation.INTERFACE              => "INTERFACE"
+      case __DirectiveLocation.UNION                  => "UNION"
+      case __DirectiveLocation.ENUM                   => "ENUM"
+      case __DirectiveLocation.ENUM_VALUE             => "ENUM_VALUE"
+      case __DirectiveLocation.INPUT_OBJECT           => "INPUT_OBJECT"
+      case __DirectiveLocation.INPUT_FIELD_DEFINITION => "INPUT_FIELD_DEFINITION"
```
Fluxx, line 108: This is a copy paste to [the `__DirectiveLocation` code](https://github.com/ghostdogpr/caliban/blob/master/core/src/main/scala/caliban/introspection/adt/__DirectiveLocation.scala). Wasn't sure if we wanted to somehow encode the string version of the location in the `case object`?

core/src/test/scala/caliban/RenderingSpec.scala
ghostdogpr: I don't have a strong opinion about it. It's a little polluting in simple cases but could be useful in some. @paulpdaniels @frekw any opinions?

core/src/test/scala/caliban/TestUtils.scala
```diff
@@ -140,6 +141,23 @@ object TestUtils {
...
+      args = List(
+        __InputValue(
+          "foo",
+          None,
+          () => Types.int,
+          None,
+          None
+        )
+      )
+    ),
```
ghostdogpr, line 158: This comma makes the build fail

core/src/main/scala/caliban/Rendering.scala
```diff
@@ -79,6 +79,44 @@ object Rendering {
...
+    }
+    val directiveLocations = locationStrings.mkString(" | ")
+
+    val body = s"""directive @${directive.name}${inputs} repeatable on ${directiveLocations}""".stripMargin
```
ghostdogpr, line 112: Should we add `repeatable: Boolean` to directives with default to true?

Fluxx: Oh shoot I meant to ask about this. Right now it's hard-coded as `repeatable`. I didn't look at what we did for this with the introspection API but it would be good to allow it to be set as non-repeatable.

core/src/test/scala/caliban/TestUtils.scala
Fluxx: Saw that, will fix.

core/src/main/scala/caliban/Rendering.scala
ghostdogpr: Ah this keyword is new in the 2021 spec, which is why it was not there before. Would be nice to add (otherwise, don't include the keyword).

core/src/test/scala/caliban/RenderingSpec.scala
ghostdogpr: It might be annoying for the tests if it's always included. Maybe an optional parameter?

core/src/main/scala/caliban/Rendering.scala
Fluxx: Cool, let me see what I can do about adding it.

core/src/test/scala/caliban/RenderingSpec.scala
Fluxx: > It might be annoying for the tests if it's always included. Maybe an optional parameter to `render`?

I have another PR https://github.com/ghostdogpr/caliban/pull/1219 that I'm gonna work on updating the `render()` API. I'll look at adding this functionality when I tackle that I think if that sounds good?

core/src/main/scala/caliban/Rendering.scala
ghostdogpr: On the top of my head, it should be added in the parser, in the `Directive` and `__Directive` classes, and maybe a validation rule should be added to check there is only one given if not repeatable.

If that's too much, it can be done in another PR 😄

core/src/test/scala/caliban/RenderingSpec.scala
ghostdogpr: Sure!

core/src/main/scala/caliban/Rendering.scala
Fluxx: Cool. Yeah I will remove the `repeatable` statement for now, and then do another PR to add `repeatable`.

Fluxx:
@ghostdogpr removed the `repeatable` statement. Please take a look.

I'll start on the second PR to add `repeatable` back right now, so it should be a fast follow.

ghostdogpr (approved):
Thanks!

PR #1274, opened 2022-01-26
docs module: do not enforce a specific version of Scala

iRevive:
Follow up to https://github.com/ghostdogpr/caliban/pull/1268#issuecomment-1021769185


ghostdogpr:
Thanks!

PR #1275, opened 2022-01-27
Add missing directive rendering for derrived types

Fluxx:
Adds rendering of directives for unions, enums, interfaces, and input objects whose schemas are automatically derived. Doing so required adding `directives` arguments to the various helper methods in `Types`, and the extracting the directives out inside the scala 2 and 3 `SchemaDerrivation` traits to pass in to those new arguments.

I also updated the test schema used in `RenderingSpec` to add directives to the various parts of the schema that were missing them.

### DIFF
--- a/core/src/main/scala-2/caliban/schema/SchemaDerivation.scala
+++ b/core/src/main/scala-2/caliban/schema/SchemaDerivation.scala
@@ -142,7 +142,7 @@ trait SchemaDerivation[R] extends LowPriorityDerivedSchema {
           getDescription(ctx),
           subtypes.map { case (t, _) => fixEmptyUnionObject(t) },
           Some(ctx.typeName.full),
-          directives = Some(getDirectives(ctx.annotations))
+          Some(getDirectives(ctx.annotations))
         )
       else {
         val impl         = subtypes.map(_._1.copy(interfaces = () => Some(List(toType(isInput, isSubscription)))))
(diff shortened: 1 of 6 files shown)


paulpdaniels:
I think there is technically one more place that we are missing which is the `schema` level directive. i.e.

```scala
schema @directive {
  query: Query
}
```

I think we could possibly modify the `RootResolver` or add them to the `GraphQL` construction wdyt?

Fluxx:
> I think we could possibly modify the `RootResolver` or add them to the `GraphQL` construction wdyt?

I'd probably say add a `directives` argument to `RootResolver`. It'll then have to pass them to the `RootSchemaBuilder` built in `GraphQL.graphql` so that they're available to `render()` so it can add them there.

ghostdogpr (approved):

ghostdogpr:
For schema directives, can't we just pass them to `graphQL(...)` in the `directives` parameter?

Fluxx:
> For schema directives, can't we just pass them to `graphQL(...)` in the `directives` parameter?

So per https://github.com/ghostdogpr/caliban/pull/1270, the directives passed there now rendered in the full schema SDL (as well as being part of the introspection schema as the were previous to that PR). We could also apply those to the `schema` type itself in the SDL, but I think that's a bit presumptuous and I as a backend author would want a way to control the directives attached to the `schema` type in the SDL vs just being declared in the schema. The closest place that exists to defining the `schema` for an author is the `RootResolver`, at least in my mind.

I'll also say, long-term I would probably argue that we should collect all directives used by all types in the SDL, and render them in the schema automatically (since they are used in the schema) vs using them in the schema + also having to remember to pass them in to `Graphql.graphQL()` so they show at the top.

Fluxx:
Did you mean to merge this?

PR #1280, opened 2022-01-31
Add schema reporting

paulpdaniels:
Adds the schema reporting protocol specified here: https://www.apollographql.com/docs/studio/schema/schema-reporting-protocol/. Which allows the server to report it's schema to apollo at start up time.


reporting/src/main/scala/caliban/reporting/ReportingDaemon.scala
```diff
@@ -0,0 +1,82 @@
...
+                                case ReportingError.RetryableError(innerThrowable) =>
+                                  ZIO.debug(
+                                    s"Schema reporting encountered an error: ${innerThrowable.getMessage}, retrying in 20 seconds"
+                                  ) *> loop(false).delay(20.seconds)
+                              },
+                              resp => loop(resp.withCoreSchema).delay(resp.in)
+                            )
+                        loop(false)
+                      }
+                      .forkManaged
```
paulpdaniels, line 57: I'm debating if this or a "registration" based approach would be better, where we build the daemon and then register the schemas later. It takes a bit more complex bookkeeping though so I'm not sure if would actually be beneficial since most consumers would know their graphs upfront.

reporting/src/main/scala/caliban/reporting/ReportingDaemon.scala
```diff
@@ -0,0 +1,82 @@
...
+                                case ReportingError.ClientError(error)             =>
+                                  ZIO.debug(
+                                    s"Schema reporting for ${ref.graphRef} failed because of a client error ${error.getMessage}. This is likely a defect, will not retry"
+                                  )
+                                case ReportingError.RetryableError(innerThrowable) =>
+                                  ZIO.debug(
+                                    s"Schema reporting encountered an error: ${innerThrowable.getMessage}, retrying in 20 seconds"
+                                  ) *> loop(false).delay(20.seconds)
+                              },
+                              resp => loop(resp.withCoreSchema).delay(resp.in)
```
paulpdaniels, line 53: So in the protocol this reporting continues indefinitely, regardless of if the schema will actually change after startup. I suppose since the schema is static currently, we could bail after getting a successful response, or I could change the `coreSchema` value to be a `ZIO`, that would allow consumers of this to build dynamic schemas that could potentially update at runtime.

reporting/src/main/scala/caliban/reporting/ReportingDaemon.scala
ghostdogpr: I know there are a few users who don't use fixed schemas but build it dynamically, but this is probably a large minority.

reporting/src/main/scala/caliban/reporting/SchemaReporter.scala
```diff
@@ -0,0 +1,98 @@
...
+}
+
+object SchemaReporter {
+
+  def make(accessToken: String) = (for {
+    client <- ZIO.service[SttpClient.Service]
+  } yield new SchemaReporter {
+    import caliban.reporting.client.{ Mutation, ReportSchemaError, ReportSchemaResponse, SchemaReport }
+
+    private final val REPORTING_URL = "https://schema-reporting.api.apollographql.com"
```
ghostdogpr, line 23: Is this always this URL?

reporting/src/test/scala/caliban/reporting/MockSchemaReporter.scala
```diff
@@ -0,0 +1,20 @@
+package caliban.reporting
+
+import zio.test.mock
```
ghostdogpr, line 3: This module was recently removed from ZIO core, I think because it is not as well maintained as the rest. I wonder if this might bite us later. Is it hard to do without?

paulpdaniels: Probably not that hard, it’s just a single method that needs to be mocked

reporting/src/main/scala/caliban/reporting/SchemaReporter.scala
paulpdaniels: Yeah it’s a fixed URL, I checked some of the other implementions and they all hard code it

vuepress/docs/.vuepress/config.js
```diff
@@ -59,7 +59,8 @@ module.exports = {
               'adapters',
               'interop',
               'federation',
-              'relay-connections'
+              'relay-connections',
+              'reporting'
```
ghostdogpr, line 63: The file is named schema-reporting. It makes the docs step of the build fail.

ghostdogpr:
The release also failed on Scala 3, not sure why the tests didn't catch it

ghostdogpr:
Ah, because the new module wasn't added to the CI config

ghostdogpr:
Fixed it in master

Issue #1252, opened 2022-01-12
Parsing Error for missing empty block


cornerman:
We are getting a Parsing Error when generating client code for a graphql schema with the sbt plugin.

Our graphql schema contains the following line:
```
input foobar
```

And we get the following error:
```
Parsing Error: Position 5143:1, found ""
caliban.parsing.Parser$.$anonfun$parseQuery$7(Parser.scala:23)
zio.ZIO$.$anonfun$fail$1(ZIO.scala:2898)
zio.internal.FiberContext.evaluateNow(FiberContext.scala:413)
zio.Runtime.unsafeRunWith(Runtime.scala:221)
zio.Runtime.unsafeRunSync(Runtime.scala:84)
zio.Runtime.unsafeRunSync$(Runtime.scala:81)
zio.Runtime$$anon$3.unsafeRunSync(Runtime.scala:288)
zio.Runtime.unsafeRun(Runtime.scala:59)
zio.Runtime.unsafeRun$(Runtime.scala:58)
zio.Runtime$$anon$3.unsafeRun(Runtime.scala:288)
caliban.codegen.CalibanSourceGenerator$.generateSources$1(CalibanSourceGenerator.scala:122)
caliban.codegen.CalibanSourceGenerator$.$anonfun$apply$35(CalibanSourceGenerator.scala:131)
caliban.codegen.CalibanSourceGenerator$.$anonfun$apply$35$adapted(CalibanSourceGenerator.scala:130)
sbt.util.Tracked$.$anonfun$outputChangedW$1(Tracked.scala:128)
caliban.codegen.CalibanSourceGenerator$.apply(CalibanSourceGenerator.scala:144)
caliban.codegen.CalibanPlugin$.$anonfun$calibanScopedSettings$3(CalibanPlugin.scala:40)
scala.Function1.$anonfun$compose$1(Function1.scala:49)
sbt.internal.util.$tilde$greater.$anonfun$$u2219$1(TypeFunctions.scala:62)
sbt.std.Transform$$anon$4.work(Transform.scala:68)
sbt.Execute.$anonfun$submit$2(Execute.scala:282)
sbt.internal.util.ErrorHandling$.wideConvert(ErrorHandling.scala:23)
sbt.Execute.work(Execute.scala:291)
sbt.Execute.$anonfun$submit$1(Execute.scala:282)
sbt.ConcurrentRestrictions$$anon$4.$anonfun$submitValid$1(ConcurrentRestrictions.scala:265)
[...]

PR #1282, opened 2022-02-01
Throwing an exception in codegen for empty enums

### LINKED ISSUE #1253: Broken code generation for empty enums
We are working with a hasura-generated schema, which somehow contains empty enums like this one:
```
enum my_empty_enum {}
```

But when compiling the generated code, we get this compile error:
```
[error]      type mismatch;
[error]        scala.Unit|caliban.client.ArgEncoder[example.hasura.unauthorized.graphql.client.HasuraUnauthorized.my_empty_enum]
[error]      L523:     implicit val encoder: ArgEncoder[my_empty_enum]    = {}
```

I am not sure what an empty enum is supposed to mean. It is only empty when accessing hasura unauthorized, but has values when accessing it authorized.

Should this be supported or is that an invalid schema?

jyoo980:
* Caliban currently generates code for empty enums.
  * The generated code, however, fails to compile.
  * Although Caliban should successfully parse an empty enum, it should _not_
    generate code for one at all.
  * In this case, I throw an exception when code generation is attempted for
    an empty enum.

See #1253 for the full discussion on this.

Also closes #1253

### DIFF
--- a/tools/src/main/scala/caliban/tools/ClientWriter.scala
+++ b/tools/src/main/scala/caliban/tools/ClientWriter.scala
@@ -661,35 +661,34 @@ object ClientWriter {
         s"$fieldName.fold(__NullValue: __Value)(value => ${writeInputValue(ListType(ofType, nonNull = true), "value", typeName)})"
     }
 
-    def writeEnum(
-      typedef: EnumTypeDefinition,
-      extensibleEnums: Boolean
-    ): String = {
+    def writeEnum(typedef: EnumTypeDefinition, extensibleEnums: Boolean): String =
+      typedef.enumValuesDefinition match {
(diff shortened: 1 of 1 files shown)


jyoo980:
Ideally I'd like to add some tests for this behaviour. I don't see any code that executes failure cases in `ClientWriterSpec.scala`, would this be a good place to add them?

ghostdogpr:
> Ideally I'd like to add some tests for this behaviour. I don't see any code that executes failure cases in `ClientWriterSpec.scala`, would this be a good place to add them?

Yep that's okay! Note that you need to use `enum MyEnum {}` syntax, since `enum MyEnum` without braces will fail at parsing (https://github.com/ghostdogpr/caliban/issues/1252).

ghostdogpr:

PR #1295, opened 2022-02-15
Add mapType method to Schema

Fluxx:
Adds a new `Schema.mapType` method to support building a new schema by mapping over the current schema's `__Type`. I've found behavior like this handy by taking a base `Schema` produced by derivation, and then altering it slightly.

### DIFF
--- a/core/src/main/scala/caliban/schema/Schema.scala
+++ b/core/src/main/scala/caliban/schema/Schema.scala
@@ -86,6 +86,18 @@ trait Schema[-R, T] { self =>
   }
 
   /**
+   * Maps over this schema's [[__Type]], converting it to another [[__Type]].
+   *
+   * @param op The operation to run on this schemas [[__Type]]
+   */
+  def mapType(op: __Type => __Type): Schema[R, T] = new Schema[R, T] {
+    override def resolve(value: T): Step[R] = self.resolve(value)
+    override def toType(isInput: Boolean, isSubscription: Boolean): __Type = {
+      op(self.toType_(isInput, isSubscription))
+    }
+  }
+
+  /**
    * Changes the name of the generated graphql type.
    * @param name new name for the type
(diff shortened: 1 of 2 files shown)


ghostdogpr (approved):
It's a little dangerous though, because if you change the name it might not match what `resolve` is doing. Maybe mention it in the scaladoc?

paulpdaniels:
Perhaps we should also allow it to map over the resolve? Like a `dimap` function? Then you could rewrite `rename` in terms of it.

ghostdogpr:
> Perhaps we should also allow it to map over the resolve? Like a `dimap` function? Then you could rewrite `rename` in terms of it.

I like that

Fluxx:
> Perhaps we should also allow it to map over the resolve? Like a `dimap` function? Then you could rewrite `rename` in terms of it.

Happy to make this change, but I am not familiar with a `dimap` function. Googling it, not 100% sure I get it (my functional skills are not that great), but it looks like we'd do something like this?

```scala
def map(mapType: __Type => __Type, mapResove: Step[R] => Step[R]): Schema[R, T]
```

And the `map` method would just do both the type and resolve mapping at the same time?

ghostdogpr:
That's the idea, but looking at `rename` this wouldn't be enough, since we need the `isInput` parameter, and the `self.toType` is used in the resolver as well. Maybe we keep it simple and offer a simple `mapType` as you did, and use `new Schema[R, T] { }` for more complex cases? 🤔

paulpdaniels:
@ghostdogpr could we encode the `isSubscription` `isInput` on the `__Type` itself? Would that break anything? Then you could map over the resulting `__Type`

ghostdogpr:
Not really. First, `__Type` is defined by the spec and used directly in the introspection schema. And then multiple fields would be affected, that would be quite inconvenient. Might hurt caching too.

ghostdogpr:

PR #1302, opened 2022-02-23
Make sure relay validation errors gets propagated to the user

SvenW:
If this connection is used right now and a user doesn't pass in any args or invalid args the only error we get back is:
```
"errors": [
    {
      "message": "Effect failure",
      "locations": [
        ...
      ],
      "path": [
       ...
      ]
    }
  ]

```
With this we make sure to propagate the actual error to the user so they understand what they're doing wrong


ghostdogpr (approved):
Makes sense, this actually runs during execution, not validation phase.

PR #1304, opened 2022-02-24
Handle Long as String to be able to process all the range of values

mdulac:
Handle the long values between 9007199254740991 and Long.MAX_VALUE 9223372036854775807 (and same for the negative ones)

Max Integer value in JS is 9007199254740991, and we need (in my project) to cover all the range of Scala Long.
One solution could be that Long ArgBuilder handle values as String, then JS can send us bigger values.

It's the behavior of Circe when you try to parse a long field from a String value.

This is the topic of my PR :)

### DIFF
--- a/core/src/main/scala/caliban/schema/ArgBuilder.scala
+++ b/core/src/main/scala/caliban/schema/ArgBuilder.scala
@@ -7,9 +7,9 @@ import caliban.parsing.Parser
 import caliban.uploads.Upload
 import zio.Chunk
 
+import java.time._
 import java.time.format.DateTimeFormatter
 import java.time.temporal.Temporal
-import java.time._
 import java.util.UUID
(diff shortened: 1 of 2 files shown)


ghostdogpr:
For reference, here's what the spec says for `Int`:

> When expected as an input type, only integer input values are accepted. All other input values, including strings with numeric content, must raise a query error indicating an incorrect type. If the integer input value represents a value less than -231 or greater than or equal to 231, a query error should be raised.

> Numeric integer values larger than 32‐bit should either use String or a custom‐defined Scalar type, as not all platforms and transports support encoding integer numbers larger than 32‐bit.

While it's explicitly forbidden for `Int` scalar, `Long` not being an official scalar, I think it makes sense to allow `String` for it.

ghostdogpr (approved):

Issue #1305, opened 2022-02-25
Subscription doesn't close when used with zstream aggregator


gaspb:
Hi,
Encountered an issue with subscription, I'm not sure whether it's related to ZIO or Caliban but I wasn't able to reproduce with just ZIO 

In ExampleService, use 
```scala
  def deletedEvents: ZStream[ExampleService, Nothing, String] =
    ZStream.accessStream(_.get.deletedEvents
      .groupedWithin(2, 1.second).provideSomeLayer[ExampleService](zio.clock.Clock.live)
      .flattenChunks
      .ensuring(ZIO.debug("completed !")))
```

When closing the socket from the frontend ( `{"id": x, "type": "stop"}` ), completion event is not emitted and the stream never closes 
(tested on master & version 1.3.3 with http4s)

ghostdogpr:
Hmm, if I remove `groupedWithin` and `flattenChunks`, I am getting the interruption properly. If I change `groupedWithin` into `grouped(2).flattenChunks`, I am getting it too. So I am guessing something becomes uninterruptible with `groupedWithin` or `provideSomeLayer` 🤔

ghostdogpr:
I eliminated `provideSomeLayer` by moving the Clock dependency on the schema, and the problem is the same, so it comes from `groupedWithin`.

ghostdogpr:
I think I found a solution: https://github.com/ghostdogpr/caliban/pull/1308

gaspb:
just tested, I can confirm it works :+1: thanks for the fast fix !

ghostdogpr:
Thanks! Just merged, it will generate a snapshot.

PR #1308, opened 2022-02-25
Make sure the subscription stream is interruptible

ghostdogpr:
Closes https://github.com/ghostdogpr/caliban/issues/1305


gaspb:
good catch, thanks :pray:

PR #1310, opened 2022-02-28
Add forward and backward pagination

dojusa:
Often APIs support only one way of pagination for connections, either [forward or backward](https://relay.dev/graphql/connections.htm#sec-Forward-pagination-arguments). 

The proposed change is to introduce the aforementioned types of paginations, so APIs can easily support the desired pagination, avoiding clients to miss use their connection.

* ForwardPaginationArgs which accepts only `first` and `after`
* BackwardPaginationArgs which accepts only `last` and `before`

### DIFF
--- a/core/src/main/scala/caliban/relay/PaginationArgs.scala
+++ b/core/src/main/scala/caliban/relay/PaginationArgs.scala
@@ -15,7 +15,10 @@ object Pagination {
   def apply[C: Cursor](
     args: ForwardPaginationArgs[C]
   ): ZIO[Any, CalibanError, Pagination[C]] =
-    validatePositivePresence(args.first, "first")
+    (args.first match {
+      case None    => ZIO.fail(s"first cannot be empty")
+      case Some(a) => validatePositive("first", a).map(First(_))
+    })
       .validate(args.after match {
         case None    => ZIO.succeed(NoCursor)
         case Some(x) => ZIO.fromEither(Cursor[C].decode(x)).map(After(_))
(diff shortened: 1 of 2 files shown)


ghostdogpr (approved):
LGTM, thanks!

Since I've never used Relay before, just waiting on @frekw's review but seems good to go.

core/src/main/scala/caliban/relay/PaginationArgs.scala
```diff
@@ -87,4 +117,22 @@ abstract class PaginationArgs[C: Cursor] { self =>
   )
 }
 
+abstract class ForwardPaginationArgs[C: Cursor] { self =>
+  val first: Option[Int]
+  val after: Option[String]
```
frekw, line 122: I think this should be `Option[C]`, no?

core/src/main/scala/caliban/relay/PaginationArgs.scala
```diff
@@ -87,4 +117,22 @@ abstract class PaginationArgs[C: Cursor] { self =>
...
+
+  def toPagination: ZIO[Any, CalibanError, Pagination[C]] = Pagination(
+    self
+  )
+}
+
+abstract class BackwardPaginationArgs[C: Cursor] { self =>
+  val last: Option[Int]
+  val before: Option[String]
```
frekw, line 131: Same here

core/src/main/scala/caliban/relay/PaginationArgs.scala
frekw: No, sorry, disregard that.

frekw (approved):
Nice! Thanks @dojusa 🙏

PR #1313, opened 2022-02-28
Fix custom directive parsing in Scala 3 impl

### LINKED ISSUE #1312: Scala 3 query parser impl fails on custom directives
Hello there

Whenever I try to parse a schema that contains a custom directive definition, parser would fail:

```scala
import caliban.parsing.Parser
import zio.Runtime

Runtime.default.unsafeRun(caliban.parsing.Parser.parseQuery("directive @test on FIELD_DEFINITION")) 
//Parsing Error: Parsing error at offset 24, expected: InRange(24,|,|) 
```

It works just fine in Scala 2 (codegen in sbt), yet fails in Scala 3. I guess it's due to different parser implementations.

Here is a scastie, one can switch the Scala version to verify:
https://scastie.scala-lang.org/j6L2yuJtQSKF1eo5m5REdQ

sergeykolbasov:
Closes #1312 

Test is rather minimalistic. To properly cover it all one would have to write round-trip property tests `Document -> String -> Parser -> Document`

### DIFF
--- a/core/src/main/scala-3/caliban/parsing/Parser.scala
+++ b/core/src/main/scala-3/caliban/parsing/Parser.scala
@@ -509,6 +509,7 @@ object Parser {
         P.string("QUERY").as(ExecutableDirectiveLocation.QUERY),
         P.string("MUTATION").as(ExecutableDirectiveLocation.MUTATION),
         P.string("SUBSCRIPTION").as(ExecutableDirectiveLocation.SUBSCRIPTION),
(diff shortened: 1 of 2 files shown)


ghostdogpr:
The CI wasn't triggered, can you try login to https://app.circleci.com/ and force push?

ghostdogpr:
Never mind, I will merge and fix it afterwards if there's any problem. Looks fine.

PR #1337, opened 2022-04-08
Fix parsing of empty comment lines in Scala 3

sergeykolbasov:
Usually it pops up with GraphiQL

```
# Welcome to GraphiQL
#
# GraphiQL is an in-browser tool for writing, validating, and
# testing GraphQL queries.
#
# Type queries into this side of the screen, and you will see intelligent
# typeaheads aware of the current GraphQL type schema and live syntax and
# validation errors highlighted within the text.
#
# GraphQL queries typically start with a "{" character. Lines that start
# with a # are ignored.
#
# An example GraphQL query might look like:
#
#     {
#       field(arg: "value") {
#         subField
#       }
#     }
#
# Keyboard shortcuts:
#
#  Prettify Query:  Shift-Ctrl-P (or press the prettify button above)
#
#     Merge Query:  Shift-Ctrl-M (or press the merge button above)
#
#       Run Query:  Ctrl-Enter (or press the play button above)
#
#   Auto Complete:  Ctrl-Space (or just start typing)
#
```

And then if you send a query after those comments, it breaks the Caliban parser as it expects _something_ following the `#` before newline, while it's perfectly fine to have just a line break (according to spec).

It works fine in Scala 2, but is broken in Scala 3.


ghostdogpr:
Thanks!

PR #1344, opened 2022-04-15
Add federation V2

paulpdaniels:
https://www.apollographql.com/docs/federation/federated-types/federated-directives

Adds the new directives that were added (and adds an argument to one that was modified).  I tried to consolidate the two under a single `FederationSupport` type that handles both (since there weren't any real changes to the resolution path).

Also adds directive support for the top level schema, which is required to tell the gateway that the subgraph is a v2 graph.

I'll see if I can improve the example in a follow up, but I'll need to rethink the current example to try and shoehorn in more examples using the new directives.


core/src/main/scala/caliban/Rendering.scala
```diff
@@ -81,6 +81,18 @@ object Rendering {
       }
       .mkString("\n\n")
 
+  def renderSchemaDirectives(directives: List[Directive]): String =
```
ghostdogpr, line 84: Can you add test for that?

ghostdogpr (commented):
Looks good! One test is failing with a nasty error.

paulpdaniels:
Don't merge yet, I'm working on the compatibility checks, and they are failing I think I implemented the link directive incorrectly

core/src/main/scala/caliban/Rendering.scala
paulpdaniels: Done in `RenderingSpec`

ghostdogpr:
Can you make a PR to update the docs?

paulpdaniels:
Yep sorry I misunderstood I thought I needed to do that after release

ghostdogpr:
That's also okay 😄 but if it's already there I can push it right after the release. Either way is okay 👍

PR #1353, opened 2022-05-02
Fix rendering of object arguments to directives

joheriks:
Fixes issue with `Option` being stringified. E.g., the following directive
```scala
Directive(
  name="dir",
  arguments = Map(
    "object" -> InputValue.ObjectValue(Map(
      "key1" -> Value.StringValue("value1"),
      "key2" -> Value.StringValue("value2")))))
```       
was rendered as:
```graphql
@dir(object: {Some(key1: "value1"),Some(key2: "value2")})
```


ghostdogpr (approved):
Good catch!

PR #1354, opened 2022-05-03
add support for graphql-transport-ws

paulpdaniels:
Adds support for the new graphql subscription [variant](https://github.com/enisdenjo/graphql-ws/blob/master/PROTOCOL.md). Splits the handling of the protocol into a separate type so that it can be implemented in a single place rather than in the adapters + zio-http.

There is probably even more refactoring that can be done here to get a smaller LOC profile, but I got as far as the weekend would allow so figured this would be a good checkin that we can continue to enhance.

### DIFF
--- a/adapters/akka-http/src/test/scala/caliban/AkkaHttpAdapterSpec.scala
+++ b/adapters/akka-http/src/test/scala/caliban/AkkaHttpAdapterSpec.scala
@@ -42,7 +42,7 @@ object AkkaHttpAdapterSpec extends DefaultRunnableSpec {
                      }
       _           <- ZIO.fromFuture { _ =>
                        implicit val s: ActorSystem = system
-                       Http()(system).newServerAt("localhost", 8088).bind(route)
+                       Http().newServerAt("localhost", 8088).bind(route)
                      }
                        .toManaged(server => ZIO.fromFuture(_ => server.unbind()).ignore)
       _           <- clock.sleep(3 seconds).toManaged_
--- a/adapters/akka-http/src/test/scala/caliban/AkkaHttpAdapterSpec.scala
+++ b/adapters/akka-http/src/test/scala/caliban/AkkaHttpAdapterSpec.scala
@@ -3,6 +3,7 @@ package caliban
 import akka.actor.ActorSystem
 import akka.http.scaladsl.Http
 import akka.http.scaladsl.server.Directives._
+import akka.stream.Materializer
 import caliban.interop.tapir.TestData.sampleCharacters
 import caliban.interop.tapir.TestService.TestService
 import caliban.interop.tapir.{ FakeAuthorizationInterceptor, TapirAdapterSpec, TestApi, TestService }
@@ -13,45 +14,44 @@ import zio._
 import zio.clock.Clock
 import zio.console.Console
 import zio.duration._
-import zio.internal.Platform
 import zio.random.Random
 import zio.test.{ DefaultRunnableSpec, TestFailure, ZSpec }
 
-import scala.concurrent.ExecutionContextExecutor
 import scala.language.postfixOps
 
 object AkkaHttpAdapterSpec extends DefaultRunnableSpec {
 
-  implicit val system: ActorSystem                                                            = ActorSystem()
-  implicit val executionContext: ExecutionContextExecutor                                     = system.dispatcher
-  implicit val runtime: Runtime[TestService with Console with Clock with Random with Uploads] =
-    Runtime.unsafeFromLayer(
-      TestService.make(sampleCharacters) ++ Console.live ++ Clock.live ++ Random.live ++ Uploads.empty,
-      Platform.default
-    )
-
-  val apiLayer: ZLayer[zio.ZEnv, Throwable, Has[Unit]] =
+  val apiLayer: ZLayer[zio.ZEnv, Throwable, TestService] =
     (for {
+      runtime     <- ZManaged.runtime[TestService with Console with Clock with Random with Uploads]
+      system      <- ZManaged.make(Task.effectTotal(ActorSystem()))(sys => ZIO.fromFuture(_ => sys.terminate()).ignore)
+      ec           = system.dispatcher
+      mat          = Materializer(system)
+      service     <- ZManaged.service[TestService.Service]
       interpreter <- TestApi.api.interpreter.toManaged_
       route        = path("api" / "graphql") {
-                       AkkaHttpAdapter.makeHttpService(interpreter, requestInterceptor = FakeAuthorizationInterceptor.bearer)
-                     } ~ path("upload" / "graphql") {
-                       AkkaHttpAdapter.makeHttpUploadService(interpreter)
-                     } ~ path("ws" / "graphql") {
-                       AkkaHttpAdapter.makeWebSocketService(interpreter)
-                     }
-      _           <- ZIO
-                       .fromFuture(_ => Http().newServerAt("localhost", 8088).bind(route))
-                       .toManaged(server =>
-                         ZIO.fromFuture(_ => server.unbind()).ignore *> ZIO.fromFuture(_ => system.terminate()).ignore
+                       AkkaHttpAdapter.makeHttpService(interpreter, requestInterceptor = FakeAuthorizationInterceptor.bearer)(
+                         runtime,
+                         implicitly,
+                         implicitly
                        )
+                     } ~ path("upload" / "graphql") {
+                       AkkaHttpAdapter.makeHttpUploadService(interpreter)(runtime, implicitly, implicitly, implicitly)
+                     } ~ path("ws" / "graphql") {
+                       AkkaHttpAdapter.makeWebSocketService(interpreter)(ec, runtime, mat, implicitly, implicitly)
+                     }
+      _           <- ZIO.fromFuture { _ =>
+                       implicit val s: ActorSystem = system
+                       Http()(system).newServerAt("localhost", 8088).bind(route)
+                     }
+                       .toManaged(server => ZIO.fromFuture(_ => server.unbind()).ignore)
       _           <- clock.sleep(3 seconds).toManaged_
-    } yield ())
+    } yield service)
       .provideCustomLayer(TestService.make(sampleCharacters) ++ Uploads.empty ++ Clock.live)
       .toLayer
 
   def spec: ZSpec[ZEnv, Any] = {
-    val suite: ZSpec[Has[Unit], Throwable] =
+    val suite: ZSpec[TestService, Throwable] =
       TapirAdapterSpec.makeSuite(
         "AkkaHttpAdapterSpec",
         uri"http://localhost:8088/api/graphql",
--- a/examples/src/main/scala/example/akkahttp/ExampleApp.scala
+++ b/examples/src/main/scala/example/akkahttp/ExampleApp.scala
@@ -18,9 +18,9 @@ import zio.Runtime
 
 object ExampleApp extends App {
 
-  implicit val system: ActorSystem                                      = ActorSystem()
-  implicit val executionContext: ExecutionContextExecutor               = system.dispatcher
-  implicit val runtime: Runtime[ExampleService with Console with Clock] =
+  implicit val system: ActorSystem                                              = ActorSystem()
+  implicit val executionContext: ExecutionContextExecutor                       = system.dispatcher
+  implicit val runtime: Runtime.Managed[ExampleService with Console with Clock] =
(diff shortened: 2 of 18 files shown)


guizmaii:
Hey @paulpdaniels,

What are these new variants? Where can I read about them, please?

ghostdogpr:
> Hey @paulpdaniels,
> 
> What are these new variants? Where can I read about them, please?

https://github.com/apollographql/subscriptions-transport-ws/blob/master/PROTOCOL.md was deprecated in favor of https://github.com/enisdenjo/graphql-ws/blob/master/PROTOCOL.md

interop/tapir/src/main/scala/caliban/interop/tapir/TapirAdapter.scala
```diff
@@ -230,18 +231,14 @@ object TapirAdapter {
     makeHttpUploadEndpoint.serverLogic(logic)
   }
 
-  def makeWebSocketEndpoint[R, E](implicit
+  def makeWebSocketEndpoint(implicit
     inputCodec: JsonCodec[GraphQLWSInput],
     outputCodec: JsonCodec[GraphQLWSOutput]
-  ): PublicEndpoint[ServerRequest, TapirResponse, CalibanPipe, ZioStreams with WebSockets] = {
-    val protocolHeader = Header("Sec-WebSocket-Protocol", "graphql-ws")
+  ): PublicEndpoint[ServerRequest, TapirResponse, CalibanPipe, ZioStreams with WebSockets] =
```
ghostdogpr, line 237: Do we make it configurable or do we support both? It looks like you supported both in the Tapir Adapter but made it configurable in zio-http?

interop/tapir/src/test/scala/caliban/interop/tapir/ws/ProtocolSpec.scala
```diff
@@ -0,0 +1,100 @@
...
+                     interpreter,
+                     false,
+                     true,
+                     None,
+                     QueryExecution.Parallel,
+                     WebSocketHooks.empty
+                   )
+  } yield pipe
+
+  override def spec: ZSpec[_root_.zio.test.environment.TestEnvironment, Any] = suite("Websocket Protocol")(
```
ghostdogpr, line 30: How about integrating this into the test suite in `TapirAdapterSpec`? It has the advantage of being ran for each http adapter.

interop/tapir/src/main/scala/caliban/interop/tapir/TapirAdapter.scala
paulpdaniels: Yeah there was a subtle difference in how the two were used. Since the zio-http doesn't get the request context, so I pushed it up to the user. I guess I could force them to pass the request?

paulpdaniels: There's also an issue that I think I might still need to address, where certain errors need to immediately kill the connection with an error.

ghostdogpr: If you can I guess it's best if there's nothing to configure.

interop/tapir/src/test/scala/caliban/interop/tapir/ws/ProtocolSpec.scala
paulpdaniels: I was actually hoping to make this test be the de facto test that runs agnostic of the adapter.

ghostdogpr: It's go for testing the internal logic, but each http lib has its quirks so I think it's also useful to run it together. I found several bugs like that (in Caliban and in Tapir).

paulpdaniels:
Only the akka adapter seems to fail 🤔

ghostdogpr:
I'm traveling this weekend and Monday but I'll take a look in the coming week!

interop/tapir/src/test/scala/caliban/interop/tapir/TestService.scala
```diff
@@ -89,6 +94,8 @@ object TestService {
 
       def deletedEvents: ZStream[Any, Nothing, String] =
         ZStream.unwrapManaged(subscribers.subscribe.map(ZStream.fromQueue(_)))
+
+      override def reset: UIO[Unit] = characters.set(initial)
```
ghostdogpr, line 98:
```suggestion
      def reset: UIO[Unit] = characters.set(initial)
```
just for consistency

adapters/akka-http/src/test/scala/caliban/AkkaHttpAdapterSpec.scala
```diff
@@ -13,45 +14,44 @@ import zio._
...
+                       AkkaHttpAdapter.makeWebSocketService(interpreter)(ec, runtime, mat, implicitly, implicitly)
                      }
-      _           <- ZIO
-                       .fromFuture(_ => Http().newServerAt("localhost", 8088).bind(route))
-                       .toManaged(server =>
-                         ZIO.fromFuture(_ => server.unbind()).ignore *> ZIO.fromFuture(_ => system.terminate()).ignore
-                       )
+      _           <- ZIO.fromFuture { _ =>
+                       implicit val s: ActorSystem = system
+                       Http()(system).newServerAt("localhost", 8088).bind(route)
```
ghostdogpr, line 45: Do we need the implicit if we pass it explicitly?

interop/tapir/src/main/scala/caliban/interop/tapir/TapirAdapter.scala
```diff
@@ -230,18 +232,38 @@ object TapirAdapter {
     makeHttpUploadEndpoint.serverLogic(logic)
   }
 
-  def makeWebSocketEndpoint[R, E](implicit
+  /**
+   * A codec which expects only text and close frames (all other frames cause a decoding error). Close frames correspond to a `Left`,
+   * while text frames are handled using the given `stringCodec` and wrapped with a `Right`
+   * @return
+   */
+  private implicit def textOrCloseWebSocketFrameEither[A, CF <: CodecFormat](implicit
```
ghostdogpr, line 240: We previously relied on a function in Tapir (if I remember properly), what's different in this one?

interop/tapir/src/main/scala/caliban/interop/tapir/TapirAdapter.scala
```diff
@@ -254,77 +276,22 @@ object TapirAdapter {
...
-                                 .provide(env)
-                                 .forkDaemon
-                             )(_.interrupt) *> ZStream.fromQueueWithShutdown(output)
-                         }
-      } yield pipe
-
+  ): ServerEndpoint[ZioWebSockets, RIO[R, *]] =
     makeWebSocketEndpoint.serverLogic[RIO[R, *]](serverRequest =>
-      requestInterceptor(serverRequest)(io).catchAll(ZIO.left(_))
+      requestInterceptor(serverRequest._1)(
```
ghostdogpr, line 281: How about `{ case (something, somethingElse) =>` instead of `(serverRequest =>` so that it reads more easily?

interop/tapir/src/main/scala/caliban/interop/tapir/WebSocketHooks.scala
```diff
@@ -12,6 +12,8 @@ trait WebSocketHooks[-R, +E] { self =>
   def beforeInit: Option[InputValue => ZIO[R, E, Any]] = None
   def afterInit: Option[ZIO[R, E, Any]]                = None
   def onMessage: Option[StreamTransformer[R, E]]       = None
+  def onPong: Option[InputValue => ZIO[R, E, Any]]     = None
```
ghostdogpr, line 15: Nice!

interop/tapir/src/main/scala/caliban/interop/tapir/WebSocketHooks.scala
```diff
@@ -40,6 +42,13 @@ trait WebSocketHooks[-R, +E] { self =>
             })
           case _                    => None
         }
+
+      override def onPong: Option[InputValue => ZIO[R2, E2, Any]] = (self.onPong, other.onPong) match {
```
ghostdogpr, line 46: How about `onAck` here and below?

interop/tapir/src/test/scala/caliban/interop/tapir/TapirAdapterSpec.scala
```diff
@@ -159,45 +161,111 @@ object TapirAdapterSpec {
...
+            io.map { messages =>
+              assertTrue(messages.head.map(_.`type`) == Right(Ops.ConnectionAck)) &&
+              assertTrue(
+                messages(1).map(_.payload.get.toString) == Right("""{"data":{"characterDeleted":"Amos Burton"}}""")
+              ) &&
+              assertTrue(messages(2).map(_.`type`) == Right(Ops.Complete))
+            }
           }
-        }
+        ) @@ TestAspect.sequential
```
ghostdogpr, line 263: Do we need it if we have it below?

interop/tapir/src/main/scala/caliban/interop/tapir/ws/Protocol.scala
```diff
@@ -0,0 +1,362 @@
...
+    queryExecution: QueryExecution,
+    webSocketHooks: WebSocketHooks[R, E]
+  ): URIO[R, CalibanPipe]
+
+}
+
+object Protocol {
+
+  def fromName(name: String): Protocol = name match {
+    case "graphql-transport-ws" => GraphQLWS
```
ghostdogpr, line 33: Reuse the constant below?

ghostdogpr (commented):
Looks very nice!

In addition to unit tests, did you also test with a client such as Altair? Just wondering if I should do additional manual tests just to make sure there is no regression.

interop/tapir/src/test/scala/caliban/interop/tapir/TapirAdapterSpec.scala
paulpdaniels: I don't think so, I was trying to debug the tests and probably added it, but the issue turned out to be that the `ActorSystem` wasn't being shutdown in between tests.

interop/tapir/src/main/scala/caliban/interop/tapir/TapirAdapter.scala
paulpdaniels: So before it didn't handle any other frames IIRC, and tapir has another method that uses an option instead and converts the `None` into a standard close frame. This is needed in order to handle the websocket error codes that are specified in the transport specification

adapters/akka-http/src/test/scala/caliban/AkkaHttpAdapterSpec.scala
paulpdaniels: Nope, I'll remove

interop/tapir/src/main/scala/caliban/interop/tapir/TapirAdapter.scala
paulpdaniels: 👍

paulpdaniels:
@ghostdogpr I just tried with the zio-http/akka and it works (it supports both formats simultaneously!). Http4s had what seems like a small compatibilty issue (sent back a null payload when it shouldn't), so I patched that.

examples/src/main/scala/example/akkahttp/ExampleApp.scala
```diff
@@ -18,9 +18,9 @@ import zio.Runtime
 
 object ExampleApp extends App {
 
-  implicit val system: ActorSystem                                      = ActorSystem()
-  implicit val executionContext: ExecutionContextExecutor               = system.dispatcher
-  implicit val runtime: Runtime[ExampleService with Console with Clock] =
+  implicit val system: ActorSystem                                              = ActorSystem()
+  implicit val executionContext: ExecutionContextExecutor                       = system.dispatcher
+  implicit val runtime: Runtime.Managed[ExampleService with Console with Clock] =
```
ghostdogpr, line 23: Any reason to change that?

ghostdogpr (approved):
LGTM to me! Just a tiny remark but good to merge for me.

examples/src/main/scala/example/akkahttp/ExampleApp.scala
paulpdaniels: Yeah, the `shutdown` method is only available on the `Runtime.Managed`, and I am terminating it now when the server shuts down

PR #1362, opened 2022-05-19
Fix cache poisoning

paulpdaniels:
Fixes a potential security vulnerability in the ApolloPersistedQuery wrapper that allowed for cache poisoning.

### DIFF
--- a/core/src/main/scala/caliban/wrappers/ApolloPersistedQueries.scala
+++ b/core/src/main/scala/caliban/wrappers/ApolloPersistedQueries.scala
@@ -6,6 +6,9 @@ import caliban.wrappers.Wrapper.OverallWrapper
 import caliban.{ CalibanError, GraphQLRequest, GraphQLResponse, InputValue }
 import zio.{ Has, Layer, Ref, UIO, ZIO }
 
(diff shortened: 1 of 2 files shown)


frekw (approved):
LGTM!

core/src/main/scala/caliban/wrappers/ApolloPersistedQueries.scala
```diff
@@ -6,6 +6,9 @@ import caliban.wrappers.Wrapper.OverallWrapper
 import caliban.{ CalibanError, GraphQLRequest, GraphQLResponse, InputValue }
 import zio.{ Has, Layer, Ref, UIO, ZIO }
 
+import java.nio.charset.StandardCharsets
+import javax.xml.bind.DatatypeConverter
```
ghostdogpr, line 10: Hmm it doesn't find this in the CI, is that specific to some JVMs?

paulpdaniels: Fixed, I thought javax.xml would be safe but it wasn't too hard to write my own.

Issue #1351, opened 2022-04-25
Generated views do not compile if a type has more than 22 fields


kormbrek-rally:
The way views are created, if a type is more than 22 fields, the generated view (and thus the client) does not compile.

In caliban/codegen-sbt/src/sbt-test/codegen/test-compile/src/main/graphql/genview/schema.graphql, add
```
type Big {
    field0: String
    field1: String
    field2: String
    field3: String
    field4: String
    field5: String
    field6: String
    field7: String
    field8: String
    field9: String
    field10: String
    field11: String
    field12: String
    field13: String
    field14: String
    field15: String
    field16: String
    field17: String
    field18: String
    field19: String
    field20: String
    field21: String
    field22: String
    field23: String
}
```
`cd codegen-sbt/src/sbt-test/codegen/test-compile`
`sbt -Dplugin.version=1.4.0`
type in `test` and hit Enter

```
[error] /.../caliban/codegen-sbt/src/sbt-test/codegen/test-compile/target/scala-2.12/src_managed/main/caliban-codegen-sbt/genview/Client.scala:222:16: too many elements for tuple: 24, allowed: 22
[error]           case (
[error]                ^
[error] one error found
[error] (Compile / compileIncremental) Compilation failed
```

We are likely going to restructure the data to avoid the limit of 22, but I thought this was worth reporting.

ollyw:
In the short term, would there be any value in not generating views with more than 22, so that the generated code still compiles? If you have a project with many views, and only one has more than 22 it is a bit frustrating. The generator could just leave a comment in the generated code to say why.

Alternatively if you have an idea of what data structure is preferable, perhaps I could take a go at making the change

ghostdogpr:
That would be a good quick fix and better than the current state!

A fix for the longer term would be to create nested tuples when there are more than 22 fields.

PR #1363, opened 2022-05-19
Support generated views with more than 22 fields

ollyw:
Fixes issues with client view generation where the number of fields is larger than the Scala tuple limit of 22, as described in https://github.com/ghostdogpr/caliban/issues/1351.

This is my first Caliban PR, so any guidance welcome :-)

### DIFF
--- a/tools/src/test/scala/caliban/tools/ClientWriterViewSpec.scala
+++ b/tools/src/test/scala/caliban/tools/ClientWriterViewSpec.scala
@@ -789,7 +789,7 @@ object Client {
       User21Selection,
       User22Selection,
       User23Selection
-    ] = (user1(selectionArgs.user1Selection) ~ user2(selectionArgs.user2Selection) ~ user3(
+    ] = (((user1(selectionArgs.user1Selection) ~ user2(selectionArgs.user2Selection) ~ user3(
       selectionArgs.user3Selection
     ) ~ user4(selectionArgs.user4Selection) ~ user5(selectionArgs.user5Selection) ~ user6(
       selectionArgs.user6Selection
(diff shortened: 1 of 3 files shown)


tools/src/test/scala/caliban/tools/ClientWriterViewSpec.scala
```diff
@@ -789,7 +789,7 @@ object Client {
       User21Selection,
       User22Selection,
       User23Selection
-    ] = (user1(selectionArgs.user1Selection) ~ user2(selectionArgs.user2Selection) ~ user3(
+    ] = (((user1(selectionArgs.user1Selection) ~ user2(selectionArgs.user2Selection) ~ user3(
```
ghostdogpr, line 792: Oh so we had a test for it but that produces code that didn't compile? 😆 That sounds like a good use case to add a scripted test, to make sure the generated code compiles. Can you add this test case to the scripted suite? For example here https://github.com/ghostdogpr/caliban/tree/master/codegen-sbt/src/sbt-test/codegen/test-compile/src/main/graphql/genview

ghostdogpr (commented):
Looks good, just have a suggestion to strengthen the tests

tools/src/test/scala/caliban/tools/ClientWriterViewSpec.scala
ollyw: Yeah, it was a surprise to me also that there was existing code to handle large types. At the start I wrote a separate test, ran it and it *and* the existing test failed :laughing: . That is how I found out! I tested the changes manually in the project I am using Caliban and it worked.

I have added the type to the codegen-set genview test. See what you think

ghostdogpr:
Looks okay but I just noticed the CI is not triggered. Can you try login to https://app.circleci.com/ and force push?

ollyw:
Logged in and CI passes

ghostdogpr:
Thanks!

PR #1370, opened 2022-05-27
Properly validate floating point numbers on input

moonkev:
It looks like a change in the latest release to add input validation has a small bug for Float input types.  If you input an integer, it will properly coerce to a floating point, however if you enter a whole number with anything after the decimal point other than zero, it will fail on validation.  

For example, if you have a variable in a query alias named f of type `Float!` and provided an input of 99.1234, it would fail with `Variable 'f' with value 99.1324 cannot be coerced into Float`.  It seems when you enter a float, they are coming to the validation code as FloatValue.BigDecimalNumber, which is not checked in the pattern match for Float input validation.  This fix adds a check for any FloatValue and passes them on as is.


frekw (approved):
Yikes what a silly mistake by me 🤦‍♂️

Thanks!

ghostdogpr (approved):

PR #1371, opened 2022-05-30
Improve persisted query performance by caching the document

paulpdaniels:
This PR improves the performance of the `ApolloPersistedQuery` wrapper by caching the parsed `Document` instead of the unparsed string. The advantage of this approach is that in the happy path we get to skip subsequent parsing steps and simply return the `Document`

### DIFF
--- a/core/src/main/scala/caliban/wrappers/ApolloPersistedQueries.scala
+++ b/core/src/main/scala/caliban/wrappers/ApolloPersistedQueries.scala
@@ -62,7 +62,7 @@ object ApolloPersistedQueries {
                   case Some(doc) => docVar.set(Some(Right(doc))) as request
                   case None      =>
                     request.query match {
-                      case Some(value) if checkHash(hash, value) => docVar.set(Some(Left(hash))) as request
+                      case Some(value) if checkHash(hash, value) => docVar.set(Some(Left(hash))).as(request)
                       case Some(_)                               => ZIO.fail(ValidationError("Provided sha does not match any query", ""))
                       case None                                  => ZIO.fail(ValidationError("PersistedQueryNotFound", ""))
                     }
--- a/core/src/main/scala/caliban/wrappers/ApolloPersistedQueries.scala
+++ b/core/src/main/scala/caliban/wrappers/ApolloPersistedQueries.scala
@@ -2,7 +2,8 @@ package caliban.wrappers
 
 import caliban.CalibanError.ValidationError
 import caliban.Value.{ NullValue, StringValue }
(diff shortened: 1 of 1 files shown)


core/src/main/scala/caliban/wrappers/ApolloPersistedQueries.scala
```diff
@@ -42,13 +57,12 @@ object ApolloPersistedQueries {
           readHash(request) match {
             case Some(hash) =>
               ZIO
-                .accessM[ApolloPersistence](_.get.get(hash))
+                .serviceWith[Service](_.get(hash))
                 .flatMap {
-                  case Some(query) => UIO(request.copy(query = Some(query)))
-                  case None        =>
+                  case Some(doc) => docVar.set(Some(Right(doc))) as request
```
frekw, line 62: Do we prefer `as x` over `.as(x)`?

frekw:
There is probably a great reason for this that I'm missing, but why are we passing in the `dicVar` rather than just accessing the service directly (since we need to use the service in others code paths anyhow)?

I also think this might be susceptible to DoS attacks by sending ever growing (but correct) queries, should we have some sort of LRU expiry in place (possibly via zio-cache)?

frekw:
Nvm I just grokked the docVar, but I think it needs to be a FiberRef, otherwise parallel requests will overwrite each other's value, no?

core/src/main/scala/caliban/wrappers/ApolloPersistedQueries.scala
paulpdaniels: I'm not sure, I don't think we have a style guide for it, but I'm fine with either I know the infix can get a little out of hand sometimes.

ghostdogpr: I tend to prefer `.as(x)` and `.` in general except when there's a symbol alias.

ghostdogpr (approved):

frekw:
Oh right, the `EffectfulWrapper` will create a unique ref per request. My bad! 🙂

I also think we can postpone the DoS stuff and just make a note of it since it's orthogonal to the changes in this PR.

frekw (approved):

core/src/main/scala/caliban/wrappers/ApolloPersistedQueries.scala
frekw: @ghostdogpr that's usually what I do as well but I don't really have any strong opinions around it so I'm fine with either

ghostdogpr (approved):
LGTM!

PR #1372, opened 2022-05-31
Escape quotes in description strings

joheriks:
Escape `"` with `\"` in normally quoted strings and `"""`  with `\"""` in
triple-quoted strings. It previously rendered syntactically invalid GraphQL 
when a single-line description contained `"` or a multi-line description
contained `"""`.
https://spec.graphql.org/October2021/#sec-String-Value


ghostdogpr:
Thanks!

PR #1373, opened 2022-05-31
Escape special characters in normally quoted strings

joheriks:
Follow-up to #1372: also escape other special characters, since 
direct rendering of these (e.g. `\b`) may produce syntactically
incorrect GraphQL.


ghostdogpr (approved):
Thanks again!

PR #1375, opened 2022-06-01
Correct field types for interface fields on union

### LINKED ISSUE #1374: Field types are `String` for interface fields selected on a union
Given the follow schema (extended from the one in `FieldSpec`):

```
union Union = A | B

interface Interface { id: String!  inner: Inner! }

type A implements Interface { id: String!  inner: Inner! }
type B implements Interface { id: String!  inner: Inner!  count: Int! }
type C implements Interface { id: String!  inner: Inner! }

type Inner { num: Int! }

type Queries {
  union: Union!
  interface: Interface!
}
```

Fetching on the interface returns the expected field types:
```
{
  interface {
    ... on Interface {
      id
      inner { num }
    }
    ... on B {
      ... on Interface { id }
      inner { num }
      count
    }
  }
}
```
```
{
  interface (Interface!) {
    id (String!)
    inner (Inner!) {
      num (Int!)
    }
    id (String!)
    inner (Inner!) {
      num (Int!)
    }
    count (Int!)
  }  
}
```

Fetching on the union returns the following field types:
```
{
  union {
    ... on Interface {
      id
      inner { num }
    }
    ... on B {
      ... on Interface { id }
      inner { num }
      count
    }
  }
}
```
```
{
  union (Union!) {
    id (String)
    inner (String) {
      num (String)
    }
    id (String!)
    inner (Inner!) {
      num (Int!)
    }
    count (Int!)
  }  
}
```

Note that everything being fetched on `Interface` at the union level says it is a `String`.

Here is a basic test case https://gist.github.com/guymers/df3eb754ca556cd59b8c2f87740ec62b

guymers:
Ensure that when an interface fragment is used on a union that the
fields its selected have the correct type instead of always being
`String`.

Closes #1374

### DIFF
--- a/core/src/main/scala/caliban/execution/Field.scala
+++ b/core/src/main/scala/caliban/execution/Field.scala
@@ -130,8 +130,10 @@ object Field {
             fragments
               .get(name)
               .foreach { f =>
-                val t =
-                  innerType.possibleTypes.flatMap(_.find(_.name.contains(f.typeCondition.name))).getOrElse(fieldType)
+                val t = innerType.possibleTypes
+                  .flatMap(_.find(_.name.contains(f.typeCondition.name)))
+                  .orElse(rootType.types.get(f.typeCondition.name))
+                  .getOrElse(fieldType)
                 loop(f.selectionSet, t).fields
                   .map(field =>
                     if (field._condition.isDefined) field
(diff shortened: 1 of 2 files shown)


ghostdogpr (commented):
Can you try login to https://app.circleci.com/ and force push? The CI wasn't triggered for some reason.

core/src/main/scala/caliban/execution/Field.scala
```diff
@@ -152,6 +154,7 @@ object Field {
           if (checkDirectives(resolvedDirectives)) {
             val t     = innerType.possibleTypes
               .flatMap(_.find(_.name.exists(typeCondition.map(_.name).contains)))
+              .orElse(typeCondition.flatMap(typeName => rootType.types.get(typeName.name)))
```
guymers, line 157: With this, is there a need for searching through `possibleTypes`?

ghostdogpr: No, `rootTypes.types` should return all types however nested they are so the interface should be found.

ghostdogpr (approved):

ghostdogpr:
Thanks!

PR #1376, opened 2022-06-02
Coerce non-null values into lists of size one

guymers:
As per the specification, if a value passed as an input to a list type is
not a list and is not null then it should be coerced into a list of size
one.

### DIFF
--- a/core/src/main/scala/caliban/parsing/VariablesCoercer.scala
+++ b/core/src/main/scala/caliban/parsing/VariablesCoercer.scala
@@ -24,7 +24,7 @@ object VariablesCoercer {
     rootType: RootType
   ): IO[ValidationError, GraphQLRequest] = {
     val variableDefinitions    = doc.operationDefinitions.flatMap(_.variableDefinitions)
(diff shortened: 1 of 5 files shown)


frekw (approved):
Looks great, thanks! 🙏

guymers:
I can't get `listListInt: ListListIntArg => Option[List[Option[List[Option[Int]]]]]` to compile on Scala 3 so I have made the tests that need it only run on Scala 2.

frekw:
@guymers what compilation error are you seeing on Scala 3? I saw that you replaced `genMacro` since it's not available for Scala 3, but can't see what failed for you after doing so :)

ghostdogpr:
Maybe ran into this? https://github.com/lampepfl/dotty/issues/13044

guymers:
Yeah its that. Defining a schema and arg builder for `Option[List[Option[Int]]]` allowed `listInt` to compile, but defining them doesn't help for `Option[List[Option[List[Option[Int]]]]]`.

ghostdogpr:
Thanks for the fix and adding all the test cases!

PR #1377, opened 2022-06-02
Client auto-gen Option conflict

alterationx10:
When a schema has a  `type Option`, and anything that return a nullable (i.e. `Option[String]` - the resulting code won't compile due to conflict with scala.Option.

In `ClientWriter.scala`, I have updated the appropriate places to print `scala.Option` instead of `Option`

Due to `writeType(field.ofType).replace(fieldType, typeLetter)` possibly replacing multiple instances of the word `Option`, i.e. `"scala.Option[Option]" => "scala.A[A]"` I've made an ugly "replaceLast" as `writeType(field.ofType).reverse.replaceFirst(fieldType.reverse, typeLetter).reverse` if the fieldType == Option

I've adjusted the existing tests in `ClientWriterSpec.scala`, and also added a test to try and return different possible nullable/required values of `type Option`


ghostdogpr:
Can you try login to https://app.circleci.com/ and force push? The CI wasn't triggered for some reason.

Also, I think it would be good to add a scripted test for this use case, because they verify that the generated code actually compiles (if we had such a use case with Option we would have known it generates code that doesn't compile). You can add it for example here: https://github.com/ghostdogpr/caliban/blob/master/codegen-sbt/src/sbt-test/codegen/test-compile/src/main/graphql/schema.graphql

alterationx10:
I missed ClientWriterViewSpec 🤦 . I'll fix that, and get a scripted test in too!

alterationx10:
I've added a scripted `test-compile-option`, which is mostly a copy of `test-compile`, but with a different schema. Wasn't sure if I should just update the schema in `test-compile`, but happy to collapse it there if desired.

Also, good thing a scripted test was added, because `SchemaWriter` also needed to print `scala.Option` 😅

codegen-sbt/src/sbt-test/codegen/test-compile-option/build.sbt
```diff
@@ -0,0 +1,24 @@
+import _root_.caliban.tools.Codegen
```
ghostdogpr, line 1: I think you can simply add the test case to the existing schema, to keep things simpler 😄

tools/src/main/scala/caliban/tools/ClientWriter.scala
```diff
@@ -232,12 +232,21 @@ object ClientWriter {
...
-            s"[$typeLetter]",
-            s"(innerSelection: SelectionBuilder[$fieldType, $typeLetter])",
-            writeType(field.ofType).replace(fieldType, typeLetter),
-            writeTypeBuilder(field.ofType, "Obj(innerSelection)")
-          )
+          if (fieldType == "Option") {
+            (
+              s"[$typeLetter]",
+              s"(innerSelection: SelectionBuilder[$fieldType, $typeLetter])",
+              writeType(field.ofType).reverse.replaceFirst(fieldType.reverse, typeLetter).reverse,
```
ghostdogpr, line 239: How about moving this line to a separation function, with the `if (fieldType == "Option")` in that function, so that the 3 other lines are not duplicated?

ghostdogpr (commented):
Looks great! Added a couple comments.

codegen-sbt/src/sbt-test/codegen/test-compile-option/build.sbt
alterationx10: 👍 I moved in into the existing compile test

tools/src/main/scala/caliban/tools/ClientWriter.scala
alterationx10: :100: I added a `safeFieldTypeReplace` method to handle this

ghostdogpr (approved):
Thanks!

PR #1378, opened 2022-06-10
Deferred execution

paulpdaniels:
Adds experimental support for the `@defer` and the `@stream` based on the current [RFC](https://github.com/graphql/graphql-wg/blob/main/rfcs/DeferStream.md) and supercedes #1148  

You can read more at the provided links but the tldr for it is that when enabled the client can make requests such as

```graphql
{
  characters {
    name
    ... @defer(label:"moreinfo") {
    	nicknames @stream(initialCount:1)
        connections(by:Origin) {
          name
        }
    }
  }
}
```

And the executor will move deferred and streaming components into a separate execution chain that will be delivered _after_ the primary payload. To do this the server adapter needs to detect the presence of deferred fields and switch the response from `application/json` to `multipart/mixed` (optionally we could also look at supporting `text/eventstream` or via websockets). The primary payload is then sent with the deferred fields not included, but has a couple extra top-level fields indicating that there is more data coming. The subsequent payloads contain data as well as a `path` which tells the client how to patch in the data.

Because this behavior is still experimental (very few clients support it) it is still opt-in, and because using it requires the transport to be changed it is much safer to force servers to opt-in to this behavior, especially if they roll their own adapters.

For testing I forked this lib: https://github.com/n1ru4l/graphql-bleeding-edge-playground and modified some values locally so I could test against a running server, though I still have a task to fix all the adapters because in order to support the streaming behavior the response types needed to be updated to use binary streams directly instead of the higher level `GraphQLResponse`.

- [ ] Fix Http4s encoding
- [ ] Test all the adapters
- [ ] Update the docs

Further reading:

The working group repository for the defer/stream specification
https://github.com/robrichard/defer-stream-wg

### DIFF
--- a/adapters/http4s/src/main/scala/caliban/Http4sAdapter.scala
+++ b/adapters/http4s/src/main/scala/caliban/Http4sAdapter.scala
@@ -50,15 +50,16 @@ object Http4sAdapter {
     queryExecution: QueryExecution = QueryExecution.Parallel,
     requestInterceptor: RequestInterceptor[R] = RequestInterceptor.empty
   )(implicit interop: ToEffect[F, R]): HttpRoutes[F] = {
-    val endpoints                                       = TapirAdapter.makeHttpService[R, E](
+    val endpoints = TapirAdapter.makeHttpService[R, E](
       interpreter,
       skipValidation,
       enableIntrospection,
       queryExecution,
       requestInterceptor
     )
-    val endpointsF: List[ServerEndpoint[ZioStreams, F]] = endpoints.map(convertHttpEndpointToF[F, R, E])
-    Http4sServerInterpreter().toRoutes(endpointsF)
+//    val endpointsF: List[ServerEndpoint[ZioStreams, F]] = endpoints.map(convertHttpEndpointToF[F, R, E])
+//    Http4sServerInterpreter().toRoutes(endpointsF)
+    ???
   }
 
   def makeHttpUploadService[R <: Has[_] with Random, E](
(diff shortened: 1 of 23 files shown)


core/src/main/scala/caliban/execution/Executor.scala
```diff
@@ -227,4 +323,18 @@ object Executor {
...
+      val l = List.newBuilder[A1]
+      val r = List.newBuilder[A2]
+      list.foreach { x =>
+        f(x) match {
+          case Left(x1)  => l += x1
+          case Right(x2) => r += x2
+        }
+      }
+      (l.result(), r.result())
+    }
```
paulpdaniels, line 338: Needed because 2.12 doesn't have `partitionMap`

adapters/http4s/src/main/scala/caliban/Http4sAdapter.scala
```diff
@@ -195,21 +197,21 @@ object Http4sAdapter {
...
+//      endpoint.INPUT,
+//      endpoint.ERROR_OUTPUT,
+//      endpoint.OUTPUT,
+//      Fs2Streams[F],
+//      F
+//    ](
+//      endpoint.endpoint,
+//      _ => a => interop.toEffect(endpoint.securityLogic(zioMonadError)(a)),
+//      _ => u => req => interop.toEffect(endpoint.logic(zioMonadError)(u)(req))
+//    )
```
paulpdaniels, line 214: TODO get this working

ghostdogpr:

PR #1381, opened 2022-06-19
Set api mappings for ZIO

guymers:
ZIO does not set `apiURL` so `autoAPIMappings` will not be able to
generate an entry in `apiMappings`.


ghostdogpr (approved):
Thanks!

PR #1248, opened 2022-01-10
Tools scala3 zio r17

pmeheut:
Attempted solution to: https://github.com/ghostdogpr/caliban/issues/1239

Used https://github.com/scalameta/scalafmt/issues/3035#issuecomment-1007187349 suggestion to load scalafmt dynamically, allowing to generate a Scala 2.12/2.13 & 3 of caliban-tools and using the sbt-codegen plugin in Scala 3.

Issue #1239, opened 2022-01-06
CompileTime Plugin is not usable with Scala 3


pmeheut:
Hi,
I have a simple Scala 3.1.0 cross Scala.js application based on [https://github.com/adpi2/scala3-full-stack-example](https://github.com/adpi2/scala3-full-stack-example)
I tried to replace the REST api by GraphQL but it has proved to be a minefield.

I need to use Scala 3 versions of Circe, Caliban to derive codecs and schema of my code. This works fine.
But as soon as I tried to use the akka-http-adapter, I got a lof of conflicts with 2.13 versions of libraries.

So I switched to ZHttp as this a Scala 3 http server. I made it work.

But then, I wanted to write the client code so I used `caliban-codegen-sbt` but of course, back to 2.13 libraries dependencies and nothing works. 
I've tried to exclude dependencies, to write subprojects in 2.13 but each time, I find another problem.
I even look to the sbt plugin source code to see if it could be easily recompiled in Scala 3. No luck because of the macros of course.

It should not be this hard. So back to my title question: is there a way to write a full stack scala 3 jvm/scala 3.js application with Caliban and automatic schema generation, etc?

ghostdogpr:
Akka HTTP doesn't provide artifacts for Scala 3 yet (some other parts of Akka like Akka Actors do, but not Akka HTTP yet). [They say](https://doc.akka.io/docs/akka/current/project/scala3.html#using-2-13-artifacts-in-scala-3) it works with 2.13 artifacts but I haven't tried it. For native scala 3 support, zio http and http4s are recommended.

Sbt plugins can only be compiled with Scala 2.12 (sbt only supports that), but you can use it regardless of your Scala version. The code generated should work with any Scala version. Is that not the case?

ghostdogpr:
I've just tested it in a small Scala 3 project, it generates client code well.

pmeheut:
I've tried with a very simple project. It works fine in 2.13.7. When I change scalaVersion to 3.1.0, I cannot resolve `com.github.ghostdogpr:caliban-tools_3:1.3.1`.
I do not use it explicitly and the message occurs as soon as I activate the `CompileTimeCalibanServerPlugin`.
I've tried to force it with `CrossVersion.for3Use2_13` but it changes nothing.

ghostdogpr:
I only tried `CalibanPlugin`, let me try the compile time one.

ghostdogpr:
[...]

PR #1415, opened 2022-07-23
tools and codegen-sbt compiles in scala3 making compile-time plugin available

pmeheut:
This is a solution to #1239 and a Caliban 2 version of PR #1248.
The tests compile. I've also used it in a real project and it works for its scope.

Except for the Scala 3 changes, the problem was that the plugin generates the cached files in streams, without the usual scala version based directory structure.
It messed with +compile.
So I changed that.

In tools, a new problem is that the current zio.config.magnolia scala 3 implementation has a very different interface. So I had to add files specific to 2.12, 2.13 and 3.


tools/src/main/scala-2.12/caliban/tools/DescriptorUtils.scala
```diff
@@ -0,0 +1,9 @@
+package caliban.tools
```
ghostdogpr: Since the code is the same between 2.12 and 2.13, I think you can place it into a `scala-2` folder instead of duplicating it.

codegen-sbt/src/sbt-test/compiletime-codegen/test-compile/modules/potatoes-clients/src/main/scala/poc/caliban/client/generated/potatoes/Color.scala
```diff
@@ -0,0 +1,33 @@
+package poc.caliban.client.generated.potatoes
```
ghostdogpr: Why adding these files?

ghostdogpr (commented):
Can you run `fmt` to format the code and make the CI go to the next steps?
Also can you resolve the merge conflicts?

tools/src/main/scala-2.12/caliban/tools/DescriptorUtils.scala
pmeheut: Done.

codegen-sbt/src/sbt-test/compiletime-codegen/test-compile/modules/potatoes-clients/src/main/scala/poc/caliban/client/generated/potatoes/Color.scala
pmeheut: I removed the files added by mistake. I ran fmt before creating the PR and ran it again: no changes.

pmeheut: But I do not know why we have a conflict in build.sbt. I only added a scalafmtVersion in the middle of the others, sorted alphabetically.

ghostdogpr: I fixed the conflict. The formatting issue is just a line break missing in `DescriptorUtils` at the end of the file with scala 3. I guess you don't see it because you're on scala 2 by default.

ghostdogpr: Fixed 😄

ghostdogpr: Ah it still fails, I guess it's more than the line break. You can run `fmt` after switching to scala 3 (`+3.1.3`)

ghostdogpr:
I guess we also need to modify `.circleci/config.yml` to make sure the scripted tests are working with scala 2.13 and 3?

codegen-sbt/src/sbt-test/compiletime-codegen/test-compile/modules/potatoes-clients/src/main/scala/poc/caliban/client/generated/potatoes/Color.scala
pmeheut: Done.

pmeheut:
About .circleci/config.yml, I guess  so but I'm not familiar with this part.

ghostdogpr:
> About .circleci/config.yml, I guess so but I'm not familiar with this part.

So basically [here](https://github.com/ghostdogpr/caliban/blob/e2c13143f19c96339bd17c8741fa17766c0d7b2d/.circleci/config.yml#L46) we create a build step that runs `sbt ++2.12.16 codegenSbt/scripted`, which runs the scripted tests under scala 2.12. We probably don't need 2.13 but we need something similar for scala 3.

pmeheut:
Ok. It does not work even for 2.13.8 right now because the test script is written with 2.12 hard dependencies. I'll have a look at it but because I'm gonna need to understand it first, it may take a few days (or months if my real work catches me up).

codegen-sbt/src/main/scala/caliban/codegen/CalibanPlugin.scala
```diff
@@ -49,4 +50,7 @@ object CalibanPlugin extends AutoPlugin {
       Compile / sourceGenerators += (Compile / caliban).taskValue,
       Test / sourceGenerators += (Test / caliban).taskValue
     )
+
+  def cacheDirectory(baseCacheDirectory: File, scalaVersion: String): File =
+    Paths.get(baseCacheDirectory.getAbsolutePath).resolve(s"scala-${scalaVersion}").toFile
```
guizmaii, line 55: Nice trick! 🙂

codegen-sbt/src/main/scala/caliban/codegen/CompileTimeCalibanPlugin.scala
```diff
@@ -394,7 +394,7 @@ private[caliban] object Functions {
     generateSources: Def.Initialize[Task[Seq[File]]]
   ): Def.Initialize[Task[Seq[File]]] =
     Def.taskDyn {
-      val cacheDirectory = streams.value.cacheDirectory
+      val cacheDirectory = CalibanPlugin.cacheDirectory(streams.value.cacheDirectory, scalaVersion.value)
```
guizmaii, line 397: +1

pmeheut:
It can be done and I've been able to pass the test manually but it is not simple as far as I can tell. We have several problems:
1) the scripted ``test`` file contains reference to the scalaVersion. So we would need to duplicate each entry for 3.1.3 and change them each time with upgrade the scala 3 compiler version because this file cannot be parameterized
2) when running with ++3.1.3, we do not generate the 2.12 jar nor the plugin we need as far as I can tell. So sbt ends with an error

To run it, I need to fix the version in the main build.sbt file, (i.e. ``version := "sbt-test-SNAPSHOT"``), build and publish the jars for 2.12 and then I can run the scripted test for 3.1.3 after editing the test file

The process can be automated and changing temporarily ``build.sbt`` and ``test`` files during a CI test should be ok.

This leads to two questions:
- is it worth it? The scala3 code is the basically the same as the scala212 code. So it would be as efficient to test only the difference, i.e. ``DescriptorUtils.scala``
- is there a better, simpler solution, something I missed?

ghostdogpr:
Hmm, maybe a simpler alternative for now could be just to add `tools/test` to the `test3_jdk` CI step to run tests on scala 3, and add a test that covers `DescriptorUtils` if it's not already there. It wouldn't test the whole thing but that's acceptable to me.

pmeheut:
Ok, I'll write such tests based on the use we make of it in `Options.scala``.

guizmaii:
Guys, I was successfully running the scripted tests on all the Scala versions thanks to this change in the CI: https://github.com/ghostdogpr/caliban/pull/1264/files#diff-78a8a19706dbd2a4425dd72bdab0502ed7a2cef16365ab7030a5a0588927bf47R53-R56

IIRC, The issue is that when we're publishing Caliban locally, it generates 3 different version numbers: 1 per version of Scala. The fix is to force the version in the `build.sbt` before to publish locally and to use this forced version number in the test projects

(Sorry if my response is a bit out of the blue, I only quickly read your conversation. Too quickly maybe)

pmeheut:
`The fix is to force the version in the build.sbt before to publish locally and to use this forced version number in the test projects`
Yes, this is what I've done to test the scala3 modifications and what I suggested above. I'll see if I can implement it without spending to much time on it because if we go this route, we also need to edit the ``test``file from inside sbt to replace 2.12 by the current scalaVersion value.

ghostdogpr:
Since this is very useful and I will do a release, I will merge the PR already. With existing tests we know it doesn't break 2.x support, so it can only be better than before 😄 Feel free to add tests later if you have time.

PR #1426, opened 2022-07-30
Tests for "tools and codegen-sbt compiles in scala3", #1415

pmeheut:
Because the test file contains 2.12 references, I wrote a sbt command, ``codegenScriptedScala3``, to edit it before testing in scala 3.
It also builds and publishLocal caliban scala212 and scala3 version with a fixed version (i.e. "test-codegen-sbt-compile-scala3") because this is needed.
Once the test is done, we revert test to 2.12.
The command reads scala3 exact version from build.sbt so there is no changes needed here when we upgrade to a more recent scala compiler.
Everything is implemented in an auto-plugin so changes are minimal.

I've added the test to .circleci/config.yml and tested it as best as I could.


.circleci/config.yml
```diff
@@ -179,6 +193,14 @@ workflows:
           filters:
             tags:
               only: /^v(0|[1-9]\d*)\.(0|[1-9]\d*)\.(0|[1-9]\d*)(?:-((?:0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*)(?:\.(?:0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*))*))?(?:\+([0-9a-zA-Z-]+(?:\.[0-9a-zA-Z-]+)*))?$/
+      - scripted3_jdk:
+          requires:
+            - lint
+          filters:
+            branches:
+              only: series/2.x
```
ghostdogpr: I think you need to remove this filter. Because of it, it's not running on your PR and won't be on other PRs. The 1.x branch won't have this change in ci config so it won't run it anyway.

pmeheut: Done

ghostdogpr (approved):
Thanks!

PR #1435, opened 2022-08-09
[Series/2.x] Make `RequestInterceptor.empty` a `val`, so that we don't instantiate many of them

guizmaii:


ghostdogpr:
Thanks~ (x2)

PR #1437, opened 2022-08-11
Entity resolvers should support partial responses and errors combined

SvenW:
Right now I can't find a way for entity resolvers to return both partial responses and the errors to the entities that wasn't resolved.

So either I get 
```
{
  "data": null,
  "errors": [
    {
      "message": "Not found James",
      "locations": [
        {
          "line": 2,
          "column": 3
        }
      ],
      "path": [
        "_entities",
        1
      ]
    }
  ]
}
```
or I can get
```
{
  "data": {
    "_entities": [
      {
        "__typename": "Character",
        "name": "James Holden"
      },
      null
    ]
  }
}
```
by just capturing all errors.
I opened this PR just to get some guidance and to see if you have any input, thanks.

### DIFF
--- a/federation/src/main/scala/caliban/federation/FederationSupport.scala
+++ b/federation/src/main/scala/caliban/federation/FederationSupport.scala
@@ -60,6 +60,7 @@ abstract class FederationSupport(
     import genericSchema._
 
     implicit val entitySchema: Schema[R, _Entity] = new Schema[R, _Entity] {
+      override def optional: Boolean                                         = true
       override def toType(isInput: Boolean, isSubscription: Boolean): __Type =
         __Type(
           __TypeKind.UNION,
--- a/federation/src/test/scala/caliban/federation/FederationV1Spec.scala
+++ b/federation/src/test/scala/caliban/federation/FederationV1Spec.scala
@@ -30,7 +30,10 @@ object FederationV1Spec extends ZIOSpecDefault {
   case class OrphanArgs(name: String)
 
   val entityResolver =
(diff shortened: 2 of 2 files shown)


paulpdaniels:
I'll look into this, AFAICR it shouldn't be doing anything special wrt to partial resolution, I'm wondering if we even support partial array failures in general 🤔

SvenW:
So I took a little bit of time yesterday just to see what happens here. It seems that for EntityResolvers we end up here https://github.com/ghostdogpr/caliban/blob/series/2.x/core/src/main/scala/caliban/execution/Executor.scala#L146. Shouldn't these types of steps be liststeps? I can of course solve the issue with some sort of hack here and use `handleError`. Do you have any pointers?

SvenW:
> So I took a little bit of time yesterday just to see what happens here. It seems that for EntityResolvers we end up here https://github.com/ghostdogpr/caliban/blob/series/2.x/core/src/main/scala/caliban/execution/Executor.scala#L146. Shouldn't these types of steps be liststeps? I can of course solve the issue with some sort of hack here and use `handleError`. Do you have any pointers?

I looked at the spec for `_Entity` and noticed it was `NON_NULL` which caused the Executor to handle it in a faulty way where the items in a `ListStep` couldn't be nullable thus returning only the errors instead of a partially resolved response.

frekw (approved):
LGTM 👍

ghostdogpr:
LGTM! @paulpdaniels wanna double check?

paulpdaniels:
Easy fix nice!

paulpdaniels (approved):

PR #1458, opened 2022-09-04
fix ping/pong handler in tapir websocket protocol

rolang:
Hi, I believe I found a bug in the protocol while investigating an issue we have with our websocket server using the [graphql-ws](https://github.com/enisdenjo/graphql-ws) client with [keepAlive](https://github.com/enisdenjo/graphql-ws/blob/6358a8fd794edba24e676a33e01c740170243006/docs/interfaces/client.ClientOptions.md#keepalive) config (see config description). Currently It causes the server to close the websocket after it receives a ping message instead of responding with pong.

This update would make it respond with a "pong" message to "ping" by default, or after a successfully executed "onPing" hook if defined.

### DIFF
--- a/interop/tapir/src/main/scala/caliban/interop/tapir/WebSocketHooks.scala
+++ b/interop/tapir/src/main/scala/caliban/interop/tapir/WebSocketHooks.scala
@@ -9,12 +9,12 @@ trait StreamTransformer[-R, +E] {
 }
 
 trait WebSocketHooks[-R, +E] { self =>
-  def beforeInit: Option[InputValue => ZIO[R, E, Any]]     = None
-  def afterInit: Option[ZIO[R, E, Any]]                    = None
-  def onMessage: Option[StreamTransformer[R, E]]           = None
-  def onPong: Option[InputValue => ZIO[R, E, Any]]         = None
-  def onPing: Option[Option[InputValue] => ZIO[R, E, Any]] = None
-  def onAck: Option[ZIO[R, E, ResponseValue]]              = None
+  def beforeInit: Option[InputValue => ZIO[R, E, Any]]                       = None
+  def afterInit: Option[ZIO[R, E, Any]]                                      = None
+  def onMessage: Option[StreamTransformer[R, E]]                             = None
+  def onPong: Option[InputValue => ZIO[R, E, Any]]                           = None
+  def onPing: Option[Option[InputValue] => ZIO[R, E, Option[ResponseValue]]] = None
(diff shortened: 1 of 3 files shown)


rolang:
Ah, just realised that this would cause issues on receiving the pong message from client when keepAlive config is set server-side. 
I guess we could keep the response to pong like before and just add the ping case.

rolang:
Ok fixed...
Could maybe cover that in tests too.

rolang:
Just made the ping hook accepting an optional payload, because one may want to run it independent from given payload, e.g. to perform something like health checks? ([ping](https://github.com/enisdenjo/graphql-ws/blob/master/PROTOCOL.md#ping)).

ghostdogpr:
Thanks for your work, I will have more time to review this in the weekend.

@paulpdaniels in case you have time to take a look 😄

interop/tapir/src/main/scala/caliban/interop/tapir/WebSocketHooks.scala
```diff
@@ -9,11 +9,12 @@ trait StreamTransformer[-R, +E] {
...
-  def beforeInit: Option[InputValue => ZIO[R, E, Any]] = None
-  def afterInit: Option[ZIO[R, E, Any]]                = None
-  def onMessage: Option[StreamTransformer[R, E]]       = None
-  def onPong: Option[InputValue => ZIO[R, E, Any]]     = None
-  def onAck: Option[ZIO[R, E, ResponseValue]]          = None
+  def beforeInit: Option[InputValue => ZIO[R, E, Any]]     = None
+  def afterInit: Option[ZIO[R, E, Any]]                    = None
+  def onMessage: Option[StreamTransformer[R, E]]           = None
+  def onPong: Option[InputValue => ZIO[R, E, Any]]         = None
+  def onPing: Option[Option[InputValue] => ZIO[R, E, Any]] = None
```
paulpdaniels: Should we make this `Option[Option[InputValue] => ZIO[R, E, Option[ResponseValue]]]`? 

Spec shows that pong can have a payload as well.

https://github.com/enisdenjo/graphql-ws/blob/master/PROTOCOL.md#pong

rolang: Right, makes sense to me, added. :+1:

paulpdaniels (approved):

ghostdogpr (approved):
Thanks!

PR #1471, opened 2022-10-11
feat: Only customize names without Input-suffix

brodin:
If a field is already called `FooInput` it will today be changed to `FooInputInput` if I don't explicitly change provide a `@GQLInputName` annotation, which is not what I expect since I already took precautions to avoid clashes.


frekw:
I would give this a 👍 since it's something that's annoyed me multiple times in the past (I also tend to suffix my inputs with `Input` to make it easier to map GQL -> Scala).

ghostdogpr:
Sounds reasonable to me!

Can you apply the same logic to Scala 2?
You will also need to run `fmt` in sbt to format the code.

kyri-petrou:
Just came across this PR while looking at my own PR and thought to just point out something _just in case_. I really like the idea because I came across this "issue" before, but it will introduce breaking changes to the schemas of users that would be affected by this change, as the input type name will change from `XXInputInput` to `XXInput`.

Not entirely sure what percentage of users this would affect, and of them, how many don't run schema compatibility checks during CI, nor whether this should be of any concern to the library itself. Still, I think it would be good to explicitly warn about this in the release notes

brodin:
> Just came across this PR while looking at my own PR and thought to just point out something _just in case_. I really like the idea because I came across this "issue" before, but it will introduce breaking changes to the schemas of users that would be affected by this change, as the input type name will change from `XXInputInput` to `XXInput`.
> 
> Not entirely sure what percentage of users this would affect, and of them, how many don't run schema compatibility checks during CI, nor whether this should be of any concern to the library itself. Still, I think it would be good to explicitly warn about this in the release notes

Great point @kyri-petrou ! Maybe removing the helper is the only sustainable thing to do? And do a breaking change? This kind of "magic" fixers is maybe not a connivance that it worth having?

frekw:
@kyri-petrou excellent points.

Personally, I think this is a change worth making regardless. The two places where backwards compat breaks are, as you say, CI checks and/or parameterized queries. And if anyone is using parameterized queries I think the chances that they've used the annotations to fix this already is fairly high (because `$foo: FooInputInput`) sticks out like a sore thumb. But _definitely_ something we should highlight prominently in the changelog.

ghostdogpr:
Yeah I can increase the minor version when I release rather than the patch version (so it would be 1.5.0 / 2.1.0) to make sure we emphasize on this change.

core/src/main/scala-3/caliban/schema/SchemaDerivation.scala
```diff
@@ -29,7 +29,11 @@ trait SchemaDerivation[A] {
    * GraphQL needs 2 different types, and they can't have the same name.
    * By default, we add the "Input" suffix after the type name.
```
ghostdogpr: Can you also edit the comment to specify it's not done if the suffix is already present?

brodin: Did my best, but it will not win any nobel literature prize, feedback wanted 🙏

brodin:
> Sounds reasonable to me!
> 
> Can you apply the same logic to Scala 2? You will also need to run `fmt` in sbt to format the code.

It did not allow me to use the very scala-esk pattern matching for scala-2, should I align the two implementations?

ghostdogpr (approved):
Looks good, thanks!
We are preparing a major version 2.1 that will have a few breaking changes so it's a good time to merge this in.

PR #1480, opened 2022-10-27
Deferred execution

paulpdaniels:
Description copied over from #1378 

Update: The spec has been moved forward, defer is now supported out of the box by apollo-client, so I am re-opening this to not forget about it.

Adds experimental support for the `@defer` ~and the `@stream`~ based on the current [RFC](https://github.com/graphql/graphql-wg/blob/main/rfcs/DeferStream.md) and supercedes #1148  

You can read more at the provided links but the tldr for it is that when enabled the client can make requests such as

```graphql
{
  characters {
    name
    ... @defer(label:"moreinfo") {
        # Note that @stream will not be part of this PR but left as an example
    	nicknames @stream(initialCount:1)
        connections(by:Origin) {
          name
        }
    }
  }
}
```

And the executor will move deferred and streaming components into a separate execution chain that will be delivered _after_ the primary payload. To do this the server adapter needs to detect the presence of deferred fields and switch the response from `application/json` to `multipart/mixed` (optionally we could also look at supporting `text/eventstream` or via websockets). The primary payload is then sent with the deferred fields not included, but has a couple extra top-level fields indicating that there is more data coming. The subsequent payloads contain data as well as a `path` which tells the client how to patch in the data.

Because this behavior is still experimental (very few clients support it) it is still opt-in, and because using it requires the transport to be changed it is much safer to force servers to opt-in to this behavior, especially if they roll their own adapters.

For testing I forked this lib: https://github.com/n1ru4l/graphql-bleeding-edge-playground and modified some values locally so I could test against a running server, though I still have a task to fix all the adapters because in order to support the streaming behavior the response types needed to be updated to use binary streams directly instead of the higher level `GraphQLResponse`.

- [x] Fix Http4s encoding
- [x] Test all the adapters
- [ ] Update the docs

Further reading:

The working group repository for the defer/stream specification
https://github.com/robrichard/defer-stream-wg

*Edit:*

There were some changes to the spec that I need to account for: https://github.com/graphql/graphql-spec/pull/742/files

### DIFF
--- a/adapters/akka-http/src/main/scala/caliban/AkkaHttpAdapter.scala
+++ b/adapters/akka-http/src/main/scala/caliban/AkkaHttpAdapter.scala
@@ -9,6 +9,7 @@ import caliban.execution.QueryExecution
 import caliban.interop.tapir.TapirAdapter.{
   zioMonadError,
   CalibanBody,
+  CalibanEndpoint,
   CalibanPipe,
   CalibanResponse,
   TapirResponse,
@@ -45,30 +46,14 @@ class AkkaHttpAdapter private (private val options: AkkaHttpServerOptions)(impli
     requestCodec: JsonCodec[GraphQLRequest],
     responseCodec: JsonCodec[ResponseValue]
   ): Route = {
-    val endpoints: List[ServerEndpoint[ZioStreams, RIO[R, *]]] = TapirAdapter.makeHttpService[R, E](
+    val endpoints: List[CalibanEndpoint[R]] = TapirAdapter.makeHttpService[R, E](
       interpreter,
       skipValidation,
       enableIntrospection,
       queryExecution,
       requestInterceptor
     )
-    akkaInterpreter.toRoute(
-      endpoints.map(endpoint =>
-        convertHttpStreamingEndpoint(
-          endpoint.asInstanceOf[
-            ServerEndpoint.Full[
-              Unit,
-              Unit,
-              (GraphQLRequest, ServerRequest),
-              TapirResponse,
-              CalibanResponse,
-              ZioStreams,
-              RIO[R, *]
-            ]
-          ]
-        )
-      )
-    )
+    akkaInterpreter.toRoute(endpoints.map(convertHttpStreamingEndpoint[R, (GraphQLRequest, ServerRequest)]))
   }
 
   def makeHttpUploadService[R, E](
@@ -91,21 +76,7 @@ class AkkaHttpAdapter private (private val options: AkkaHttpServerOptions)(impli
       queryExecution,
       requestInterceptor
     )
-    akkaInterpreter.toRoute(
-      convertHttpStreamingEndpoint(
-        endpoint.asInstanceOf[
-          ServerEndpoint.Full[
-            Unit,
-            Unit,
-            (GraphQLRequest, ServerRequest),
-            TapirResponse,
-            CalibanResponse,
-            ZioStreams,
-            RIO[R, *]
-          ]
-        ]
-      )
-    )
+    akkaInterpreter.toRoute(convertHttpStreamingEndpoint(endpoint))
   }
 
   def makeWebSocketService[R, E](
@@ -159,24 +130,21 @@ object AkkaHttpAdapter {
 
   type AkkaPipe = Flow[GraphQLWSInput, Either[GraphQLWSClose, GraphQLWSOutput], Any]
 
-  def convertHttpStreamingEndpoint[R](
+  def convertHttpStreamingEndpoint[R, I](
     endpoint: ServerEndpoint.Full[
       Unit,
       Unit,
-      (GraphQLRequest, ServerRequest),
+      I,
       TapirResponse,
       CalibanResponse,
       ZioStreams,
-      RIO[
-        R,
-        *
-      ]
+      RIO[R, *]
     ]
   )(implicit runtime: Runtime[R], mat: Materializer): ServerEndpoint[AkkaStreams, Future] =
     ServerEndpoint[
       Unit,
       Unit,
-      (GraphQLRequest, ServerRequest),
+      I,
       TapirResponse,
       (MediaType, Either[ResponseValue, AkkaStreams.BinaryStream]),
       AkkaStreams,
@@ -185,7 +153,7 @@ object AkkaHttpAdapter {
       endpoint.endpoint
         .asInstanceOf[
           PublicEndpoint[
-            (GraphQLRequest, ServerRequest),
+            I,
             TapirResponse,
             (MediaType, Either[ResponseValue, AkkaStreams.BinaryStream]),
             AkkaStreams
@@ -207,7 +175,6 @@ object AkkaHttpAdapter {
                           .succeed(Source.queue[ByteString](0, OverflowStrategy.fail).preMaterialize())
                           .flatMap { case (queue, source) =>
                             stream
-                              .asInstanceOf[ZStream[Any, Throwable, Byte]]
                               .runForeachChunk(chunk => ZIO.fromFuture(_ => queue.offer(ByteString(chunk.toArray))))
                               .ensuring(ZIO.succeed(queue.complete()))
                               .forkDaemon
--- a/adapters/akka-http/src/main/scala/caliban/AkkaHttpAdapter.scala
+++ b/adapters/akka-http/src/main/scala/caliban/AkkaHttpAdapter.scala
@@ -6,7 +6,14 @@ import akka.stream.{ Materializer, OverflowStrategy }
 import akka.util.ByteString
 import caliban.AkkaHttpAdapter.{ convertHttpStreamingEndpoint, convertWebSocketEndpoint }
 import caliban.execution.QueryExecution
-import caliban.interop.tapir.TapirAdapter.{ zioMonadError, CalibanPipe, TapirResponse, ZioWebSockets }
+import caliban.interop.tapir.TapirAdapter.{
+  zioMonadError,
+  CalibanBody,
+  CalibanPipe,
+  CalibanResponse,
+  TapirResponse,
+  ZioWebSockets
+}
 import caliban.interop.tapir.{ RequestInterceptor, TapirAdapter, WebSocketHooks }
 import sttp.capabilities.WebSockets
 import sttp.capabilities.akka.AkkaStreams
@@ -54,12 +61,9 @@ class AkkaHttpAdapter private (private val options: AkkaHttpServerOptions)(impli
               Unit,
               (GraphQLRequest, ServerRequest),
               TapirResponse,
-              (MediaType, ZioStreams.BinaryStream),
+              CalibanResponse,
               ZioStreams,
-              RIO[
-                R,
-                *
-              ]
+              RIO[R, *]
             ]
           ]
         )
@@ -95,7 +99,7 @@ class AkkaHttpAdapter private (private val options: AkkaHttpServerOptions)(impli
             Unit,
             (GraphQLRequest, ServerRequest),
             TapirResponse,
-            (MediaType, ZioStreams.BinaryStream),
+            CalibanResponse,
             ZioStreams,
             RIO[R, *]
           ]
@@ -161,7 +165,7 @@ object AkkaHttpAdapter {
       Unit,
       (GraphQLRequest, ServerRequest),
       TapirResponse,
-      (MediaType, ZioStreams.BinaryStream),
+      CalibanResponse,
       ZioStreams,
       RIO[
         R,
@@ -174,7 +178,7 @@ object AkkaHttpAdapter {
       Unit,
       (GraphQLRequest, ServerRequest),
       TapirResponse,
-      (MediaType, AkkaStreams.BinaryStream),
+      (MediaType, Either[ResponseValue, AkkaStreams.BinaryStream]),
       AkkaStreams,
       Future
     ](
@@ -183,7 +187,7 @@ object AkkaHttpAdapter {
           PublicEndpoint[
             (GraphQLRequest, ServerRequest),
             TapirResponse,
-            (MediaType, AkkaStreams.BinaryStream),
+            (MediaType, Either[ResponseValue, AkkaStreams.BinaryStream]),
             AkkaStreams
           ]
         ],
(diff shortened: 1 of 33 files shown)


ghostdogpr:
Apologies for the lack of feedback, this is quite big and I need enough time to dive in. Considering taking a day off one of these days just for catching up with various PRs and issues 😄

paulpdaniels:
@ghostdogpr let me push the latest later today, there were some additional payload changes to account for and I'm going to remove the stream directive and add it separately since it has a trickier implementation

interop/tapir/src/main/scala/caliban/interop/tapir/TapirAdapter.scala
```diff
@@ -142,21 +157,64 @@ object TapirAdapter {
...
+          .mapConcat(_.getBytes(StandardCharsets.UTF_8))
+
+        (
+          MediaType.MultipartMixed.copy(otherParameters = Map("boundary" -> "-", "deferSpec" -> "20220824")),
+          encoded
+        )
+
+      case response =>
+        val encodedResponse =
+          ZStream.fromIterable(responseCodec.encode(response.toResponseValue).getBytes(StandardCharsets.UTF_8))
```
kyri-petrou: I was just scrolling through this PR since it's quite interesting (great job by the way!) and this happened to catch my gaze.

For my own understanding, wouldn't this cause the response to be streamed in chunks of 1 byte? If yes, would it be better to use `ZStream.succeed` instead or there shouldn't be any overhead doing this?

guersam: [`ZStream.fromIterable`](https://github.com/zio/zio/blob/d5df88d8860d544e138c4f06972ab2b6314083bc/streams/shared/src/main/scala/zio/stream/ZStream.scala#L4185-L4216) iterates the given `Array[Byte]` and splits it into chunks of the default size (4096 currently.)

Although it's a reasonable size, I'd suggest wrapping the array directly using `Chunk.fromArray` to avoid copy.

`ZStream.fromChunk(Chunk.fromArray(responseCodec.encode(response.toResponseValue).getBytes(StandardCharsets.UTF_8)))`

paulpdaniels: @guersam fixed

adapters/akka-http/src/main/scala/caliban/AkkaHttpAdapter.scala
```diff
@@ -64,7 +87,21 @@ class AkkaHttpAdapter private (private val options: AkkaHttpServerOptions)(impli
       queryExecution,
       requestInterceptor
     )
-    akkaInterpreter.toRoute(TapirAdapter.convertHttpEndpointToFuture(endpoint))
+    akkaInterpreter.toRoute(
+      convertHttpStreamingEndpoint(
+        endpoint.asInstanceOf[
```
ghostdogpr: Shall we change the return type of `TapirAdapter.makeHttpUploadService` instead if casting?

adapters/akka-http/src/main/scala/caliban/AkkaHttpAdapter.scala
```diff
@@ -32,17 +34,37 @@ class AkkaHttpAdapter private (private val options: AkkaHttpServerOptions)(impli
...
       interpreter,
       skipValidation,
       enableIntrospection,
       queryExecution,
       requestInterceptor
     )
-    akkaInterpreter.toRoute(endpoints.map(TapirAdapter.convertHttpEndpointToFuture(_)))
+    akkaInterpreter.toRoute(
+      endpoints.map(endpoint =>
+        convertHttpStreamingEndpoint(
```
ghostdogpr: Is there a performance degradation for using streaming in the case where we don't need it?

build.sbt
```diff
@@ -403,7 +403,7 @@ lazy val examples = project
   .settings(commonSettings)
   .settings(
     publish / skip     := true,
-    run / fork         := true,
+//    run / fork         := true,
```
ghostdogpr: Restore that? Or is there a reason to remove it?

core/src/test/scala/caliban/execution/DeferredExecutionSpec.scala
```diff
@@ -0,0 +1,201 @@
...
+        rest     <- runIncrementalResponses(response)
+      } yield assertTrue(
+        rest.head.toString == """{"character":{"name":"Roberta Draper","nicknames":["Bobbie","Gunny"]}}"""
+      ) && assertTrue(
+        rest.tail.toList.map(_.toString) == List(
+          """{"incremental":[{"data":"Gunny","path":["character","nicknames",1]}],"label":"nicknames","hasNext":true}""",
+          """{"hasNext":false}"""
+        )
+      )
+    } @@ TestAspect.ignore
```
ghostdogpr: This test is ignored

core/src/test/scala/caliban/execution/TestDeferredSchema.scala
```diff
@@ -0,0 +1,78 @@
...
+    RootResolver(
+      Query(
+        character = args =>
+          ZIO
+            .serviceWithZIO[CharacterService](_.characterBy(_.name == args.name))
+            .map(_.headOption.map(character2CharacterZIO))
+      )
+    )
+
+  val interpreter = (graphQL(resolver) @@ DeferSupport.deferSupport @@ DeferSupport.streamSupport).interpreter
```
ghostdogpr: How about `DeferSupport.defer` and `DeferSupport.stream`?

examples/src/main/scala/example/ziohttp/ExampleApp.scala
```diff
@@ -2,12 +2,14 @@ package example.ziohttp
...
 import example.{ ExampleApi, ExampleService }
-
 import caliban.ZHttpAdapter
+import zhttp.http.Middleware.cors
 import zio._
 import zio.stream._
 import zhttp.http._
+import zhttp.http.middleware.Cors.CorsConfig
 import zhttp.service.Server
+import zio.Console
```
ghostdogpr, line 10: Sort imports

federation/src/test/scala/caliban/federation/v2/FederationV2Spec.scala
```diff
@@ -30,7 +30,10 @@ object FederationV2Spec extends ZIOSpecDefault {
 
         for {
           interpreter <- api.interpreter
-          data        <- interpreter.execute(query).map(resp => decode[Json](resp.data.toString)).absolve
+          data        <- interpreter
+                           .execute(query)
+                           .map { resp => println(resp.data.toString); decode[Json](resp.data.toString) }
```
ghostdogpr: Remove the println?

interop/tapir/src/main/scala/caliban/interop/tapir/TapirAdapter.scala
```diff
@@ -1,7 +1,8 @@
 package caliban.interop.tapir
 
-import java.nio.charset.StandardCharsets
+import caliban.ResponseValue.StreamValue
```
ghostdogpr, line 3: Sort imports

core/src/main/scala/caliban/execution/Executor.scala
```diff
@@ -72,20 +76,38 @@ object Executor {
...
+              val field       = fields
+                .get(name)
+                .fold(NullStep: ReducedStep[R])(reduceStep(_, f, args, Left(alias.getOrElse(name)) :: path))
+
+              val info = fieldInfo(f, path, directives)
+
+              fragment.collectFirst {
+                // The defer spec provides some latitude on how we handle responses. Since it is more performant to return
+                // pure fields rather than spin up the defer machinery we return pure fields immediately to the caller.
+                case IsDeferred(label) if featureSet(Feature.Defer) && !field.isPure =>
```
ghostdogpr, line 94: You don't use `Feature.Stream` anywhere, is that still relevant?

adapters/akka-http/src/main/scala/caliban/AkkaHttpAdapter.scala
kyri-petrou: Not necessarily applicable here since it all boils down to the underlying implementation, but streaming can incur a significant overhead. E.g., when Akka gRPC went from one of the lowest performing servers to the best performing one a year ago, it was mostly due to handling unary rpcs without streaming https://www.lightbend.com/blog/akka-grpc-update-delivers-1200-percent-performance-improvement.

I think as long as we don't stream the initial response we're probably safe, but might still be worth it to do a performance comparison just to understand it better. Happy to help with that if needed!

ghostdogpr: I was thinking exactly about that article when I asked 😄

build.sbt
paulpdaniels: Oh I needed it for debugging purposes since intellij won't run the debugger properly when it is forked

core/src/main/scala/caliban/execution/Executor.scala
paulpdaniels: No I can remove it, originally I had both deferred and streaming in this PR but the streaming changes became much more complicated so I meant to split those out. Technically I could keep the directive support and just not actually provide streaming yet, the spec allows for the server to choose _not_ to stream the values if a single value would be more efficient.

adapters/akka-http/src/main/scala/caliban/AkkaHttpAdapter.scala
paulpdaniels: Interesting I can try, though the difficult part will be encoding it in Tapir since it needs to basically be a `Either[TapirResponse, (MediaType, ZStream[Any, Throwable, Byte])]` AFAICT

paulpdaniels: Ok finally was able to get back to this, and make this change. By default now if there is no deferred block then it uses the fast single path.

core/src/main/scala/caliban/execution/Executor.scala
```diff
@@ -158,17 +184,83 @@ object Executor {
...
+        query         = makeQuery(step, errors, deferred)
+        result       <- query.runCache(cache)
+        resultErrors <- errors.get
+        defers       <- deferred.get
+        stream        =
+          (makeDeferStream(defers, cache)
+            .mapChunks(chunk => Chunk.single(GraphQLIncrementalResponse(chunk.toList, hasNext = true))) ++ ZStream
+            .succeed(GraphQLIncrementalResponse.empty))
+            .map(_.toResponseValue)
+            .provideEnvironment(env)
```
kyri-petrou: Doubt this makes much difference, but shall we place this inside the `if (defers.nonEmpty) { ... }` block?

core/src/main/scala/caliban/execution/Fragment.scala
```diff
@@ -0,0 +1,29 @@
...
+
+object IsStream {
+  def unapply(field: Field): Option[(Option[String], Option[Int])] =
+    field.directives.collectFirst { case Directive("stream", args, _) =>
+      (
+        args.get("label").collect { case StringValue(v) => v },
+        args.get("initialCount").collect { case v: IntValue => v.toInt }
+      )
+    }
+}
```
kyri-petrou, line 29: Should this be inside the `Fragment` object?

kyri-petrou (commented):
Awesome work on this! I did a first pass over the changes, looks good to me so far from what I can tell - just some minor comments below

core/src/main/scala/caliban/execution/Fragment.scala
paulpdaniels: No I don't think so. The stream is applied to fields instead of fragments even though it does similar things

core/src/main/scala/caliban/execution/Executor.scala
paulpdaniels: Good point yeah we can reduce the allocation if no deferred are found

adapters/akka-http/src/main/scala/caliban/AkkaHttpAdapter.scala
```diff
@@ -3,14 +3,24 @@ package caliban
 import akka.http.scaladsl.server.Route
 import akka.stream.scaladsl.{ Flow, Sink, Source }
 import akka.stream.{ Materializer, OverflowStrategy }
-import caliban.AkkaHttpAdapter.convertWebSocketEndpoint
+import akka.util.ByteString
+import caliban.AkkaHttpAdapter.{ convertHttpStreamingEndpoint, convertWebSocketEndpoint }
 import caliban.execution.QueryExecution
-import caliban.interop.tapir.TapirAdapter.{ zioMonadError, CalibanPipe, ZioWebSockets }
+import caliban.interop.tapir.TapirAdapter.{
```
ghostdogpr: Minor: shall we just import `_`?

adapters/http4s/src/test/scala/caliban/Http4sAdapterSpec.scala
```diff
@@ -28,6 +28,7 @@ object Http4sAdapterSpec extends ZIOSpecDefault {
     requestCodec: JsonCodec[GraphQLRequest],
     mapCodec: JsonCodec[Map[String, Seq[String]]],
     responseCodec: JsonCodec[GraphQLResponse[CalibanError]],
+    responseValueCodec: JsonCodec[ResponseValue],
```
ghostdogpr, line 30: Do we still need the previous line?

adapters/play/src/main/scala/caliban/PlayAdapter.scala
```diff
@@ -2,14 +2,25 @@ package caliban
 
 import akka.stream.{ Materializer, OverflowStrategy }
 import akka.stream.scaladsl.{ Flow, Sink, Source }
+import akka.util.ByteString
+import caliban.PlayAdapter.convertHttpStreamingEndpoint
 import caliban.execution.QueryExecution
-import caliban.interop.tapir.TapirAdapter.{ zioMonadError, CalibanPipe, ZioWebSockets }
+import caliban.interop.tapir.TapirAdapter.{
```
ghostdogpr: `_`?

interop/tapir/src/test/scala/caliban/interop/tapir/TapirAdapterSpec.scala
```diff
@@ -2,21 +2,22 @@ package caliban.interop.tapir
...
 import sttp.tapir.client.sttp.ws.zio._
 import sttp.tapir.model.{ ConnectionInfo, ServerRequest }
-import zio.stream.ZStream
+import sttp.tapir.{ AttributeKey, DecodeResult }
+import zio.json._
+import zio.stream.{ ZPipeline, ZSink, ZStream }
 import zio.test.TestAspect.before
 import zio.test._
-import zio.{ test => _, _ }
+import zio.{ test => _, ZIO, _ }
```
ghostdogpr: ZIO can be removed

examples/src/main/scala/example/ExampleApi.scala
```diff
@@ -15,39 +16,89 @@ import scala.language.postfixOps
...
+      maxFields(300) @@               // query analyzer that limit query fields
       maxDepth(30) @@                 // query analyzer that limit query depth
       timeout(3 seconds) @@           // wrapper that fails slow queries
       printSlowQueries(500 millis) @@ // wrapper that logs slow queries
       printErrors @@                  // wrapper that logs errors
-      apolloTracing                   // wrapper for https://github.com/apollographql/apollo-tracing
-
+      apolloTracing @@                // wrapper for https://github.com/apollographql/apollo-tracing
+      DeferSupport.defer @@    // wrapper that enables @defer directive support
+      DeferSupport.stream      // wrapper that enables @stream directive support
```
ghostdogpr: If streaming is not supported, how about deleting all references to it? I think it's quite confusing to see it here and there.

examples/src/main/scala/example/ziohttp/ExampleApp.scala
```diff
@@ -2,11 +2,11 @@ package example.ziohttp
 
 import example.ExampleData._
 import example.{ ExampleApi, ExampleService }
-
```
ghostdogpr: Imports shouldn't be changed here

adapters/zio-http/src/main/scala/caliban/ZHttpAdapter.scala
```diff
@@ -24,6 +24,7 @@ object ZHttpAdapter {
   )(implicit
     requestCodec: JsonCodec[GraphQLRequest],
     responseCodec: JsonCodec[GraphQLResponse[E]],
+    responseValueCodec: JsonCodec[ResponseValue],
```
ghostdogpr, line 26: Previous line is not needed?

kyri-petrou (approved):
LGTM

ghostdogpr:
Ah do you mind putting the docs in another PR so that it doesn't get published right away? Thanks!

adapters/play/src/main/scala/caliban/PlayAdapter.scala
```diff
@@ -34,7 +37,8 @@ class PlayAdapter private (private val options: Option[PlayServerOptions]) {
     runtime: Runtime[R],
     materializer: Materializer,
     requestCodec: JsonCodec[GraphQLRequest],
-    responseCodec: JsonCodec[GraphQLResponse[E]]
+    responseCodec: JsonCodec[GraphQLResponse[E]],
+    responseValueCodec: JsonCodec[ResponseValue]
```
ghostdogpr, line 40: Do we need both?

vuepress/docs/docs/optimization.md
```diff
@@ -173,3 +173,43 @@ Queries( (field) => {
...
+      nicknames
+      ... @defer {
+        role
+        origin
+      }
+  }
+}
+```
+
+The server now has the option of deferring splitting the execution of the query and returning the first part of the result immediately.
```
ghostdogpr:
```suggestion
The server now has the option of splitting the execution of the query and returning the first part of the result immediately.
```

vuepress/docs/docs/optimization.md
```diff
@@ -173,3 +173,43 @@ Queries( (field) => {
...
+While sending the fields in the fragment separately.
```
ghostdogpr: Maybe add this to the previous sentence after a comma?

adapters/zio-http/src/main/scala/caliban/ZHttpAdapter.scala
ghostdogpr, line 26: No need for `responseCodec` I think?

adapters/play/src/main/scala/caliban/PlayAdapter.scala
paulpdaniels: Nope removed

paulpdaniels:
@ghostdogpr comments addressed and docs removed!

ghostdogpr (approved):

PR #1513, opened 2022-11-19
Update `zio` to 2.0.4 and `zio-query` to 0.3.2

kyri-petrou:
There seems to be an issue with ZIO and Scala 3 (Scala 2.12/2.13 seems unaffected). Since zio 2.0.3, ZLayers are not backwards, failing with a runtime error:

```sh
Exception in thread "zio-fiber-253" java.lang.Error: Defect in zio.ZEnvironment: Could not find SttpBackend[=λ %A → ZIO[-Any,+Throwable,+A],+{package::WebSockets & ZioStreams}] inside ZEnvironment(SttpBackend[=λ %A → ZIO[-Any,+Throwable,+A],+{package$::WebSockets & ZioStreams}] -> sttp.client3.FollowRedirectsBackend@3c4b5d00)
        at zio.ZEnvironment$$anon$1.get(ZEnvironment.scala:167)
        at zio.ZIO$ServiceWithZIOPartiallyApplied$.apply$extension$$anonfun$13$$anonfun$1(ZIO.scala:5109)
        at caliban.interop.tapir.TapirAdapterSpec.makeSuite.tests.io(TapirAdapterSpec.scala:219)
        at caliban.interop.tapir.TapirAdapterSpec.makeSuite.tests.io(TapirAdapterSpec.scala:257)
        at caliban.interop.tapir.TapirAdapterSpec.makeSuite.tests(TapirAdapterSpec.scala:266)
        at caliban.interop.tapir.TapirAdapterSpec.makeSuite.tests(TapirAdapterSpec.scala:267)
```

The only affected parts are `AsyncHttpClientZioBackend.layer()`, and a workaround I found to work is to use `ZLayer.scoped(AsyncHttpClientZioBackend.scoped())` instead (which is the [exact same](https://github.com/softwaremill/sttp/blob/master/async-http-client-backend/zio/src/main/scala/sttp/client3/asynchttpclient/zio/AsyncHttpClientZioBackend.scala#L140-L145) as what the `layer()` method does).

There's an issue open on ZIO about this (see https://github.com/zio/zio/issues/7506). Do we want to use the workaround or wait for a fix / sttp zio 2.0.3+ release?


ghostdogpr:
We still have the issue on Scala.js with zio, right? For that one maybe better to wait.

Are the examples not affected? If the issue affects end-users I'd rather wait for a fix too.

kyri-petrou:
> We still have the issue on Scala.js with zio, right? For that one maybe better to wait.

I run the JS tests locally and it seems that it wasn't an issue although let's see what CI thinks - might have been just a 2.0.3 problem only?

> Are the examples not affected?

It's a Scala 3 issue only, but I can add a comment in the examples with the workaround if we think that the issue is not a blocker to update the zio version

ghostdogpr:
Got it. If ScalaJS is solved we can merge this 🤞

kyri-petrou:
Okay so running `sbt ++3.2.0 clientJS/test clientLaminextJS/Test/fastLinkJS` locally fails, but we don't have that test in CI. Any reason that we don't test the ScalaJS packages with Scala 3?

kyri-petrou:
Also the test seems to have failed due to some unrelated reason so probably just needs a restart

ghostdogpr:
> Okay so running `sbt ++3.2.0 clientJS/test clientLaminextJS/Test/fastLinkJS` locally fails, but we don't have that test in CI. Any reason that we don't test the ScalaJS packages with Scala 3?

Good question 🤔 I can't remember if there was a reason. So https://github.com/zio/zio/issues/7496 is gone with scala 2?

ghostdogpr:
> Also the test seems to have failed due to some unrelated reason so probably just needs a restart

done

kyri-petrou:
> So https://github.com/zio/zio/issues/7496 is gone with scala 2?

Looking at the CI logs at the time, it seems that [the CI for ScalaJS](https://app.circleci.com/pipelines/github/ghostdogpr/caliban/4270/workflows/f340dc54-db2f-4ab8-9796-de7fda0f0e01/jobs/22852) for Scala 2 failed for an unrelated reason. But when I was attempting to update to zio 2.0.3 I must have accidentally came across it by running the ScalaJS on Scala 3 locally without realising they're not run in CI

kyri-petrou:
In short: The issue was never there with ScalaJS (Scala 2) and it's still there with Scala 3

ghostdogpr:
Checking if it was working before: https://github.com/ghostdogpr/caliban/pull/1514

ghostdogpr:
OK, it worked. Can you rebase to see if the CI actually fails on scala 3?

kyri-petrou:
Emm okay seems that I messed something up with the rebase

kyri-petrou:
But yeah the Scala 3 JS tests are failing

ghostdogpr:
Sad 😢 Let's wait

ghostdogpr:
I wonder why zio-query didn't have that problem on Scala 3 / JS: https://github.com/zio/zio-query/actions/runs/3474449221/jobs/5807621645

kyri-petrou:
Yeah that's fair. I'll close the PR (since I also botched up the commit history) and keep an eye out for a new ZIO release

ghostdogpr:
It's on scala 3.2.1, maybe it makes a difference?

kyri-petrou:
I tried that as well, but didn't work. I think the reason is related to the workaround specified here: https://github.com/zio/zio/issues/7496#issuecomment-1301042967

kyri-petrou:
@ghostdogpr actually, I _think_ the reason zio-query didn't have a problem is because the `VersionSpecific` trait in zio is `private[zio]` (see https://github.com/zio/zio/blob/f472c6c6ff79eba9365886436efad16b8725732c/core/shared/src/main/scala/zio/VersionSpecific.scala#L26). zio-query would be able to access it because their tests are namespaced under `package zio.query`.

For reference, this is the error we're getting:

```
Referring to non-existent method zio.VersionSpecific$$anon$1.apply(java.lang.Object)java.lang.Object
   called from java.util.HashMap.computeIfAbsent(java.lang.Object,java.util.function.Function)java.lang.Object
   called from zio.VersionSpecific.taggedIsSubtype(izumi.reflect.macrortti.LightTypeTag,izumi.reflect.macrortti.LightTypeTag)boolean
   called from zio.package$.taggedIsSubtype(izumi.reflect.macrortti.LightTypeTag,izumi.reflect.macrortti.LightTypeTag)boolean
   called from zio.ZEnvironment$$anon$1.get(izumi.reflect.macrortti.LightTypeTag,zio.Unsafe)java.lang.Object
```

ghostdogpr:
I tracked the change in zio that caused that, commented on the issue

PR #1518, opened 2022-11-21
Fix scalafmt newline issue

### LINKED ISSUE #1510: [CodeGen] Generated files are invalid according to scalafmt
The scalafmt formatting done by `calibanGenClient` is very nearly correct, but has an extra newline at the end of the file that `scalafmtCheck` rejects (at least when using default settings).

I've created an [example project](https://github.com/jfwilson/caliban-newline-issue) that demonstrates the problem, or you can reproduce from scratch by:

1. Create a `.scalafmt.conf` file, e.g.

    ```
    version = 3.6.1
    runner.dialect = scala213
    preset = default
    ```

2. Create minimal `sbt` project with scalafmt and caliban-codegen

    ```
    addSbtPlugin("org.scalameta" % "sbt-scalafmt" % "2.4.6")
    addSbtPlugin("com.github.ghostdogpr" % "caliban-codegen-sbt" % "2.0.1")
    ```

4. Run `calibanGenClient` to generate client scala

    ```
    calibanGenClient project/ClientSchema.graphql src/main/scala/client/ClientSchema.scala
    ```

Expected behaviour:

`scalafmtCheck` passes

Actual:

`scalafmtCheck` fails due to extra newline at the end of the client scala file

I suspect the issue is due to the use of `PrintWriter.println` on already-formatted `objectCode` here: https://github.com/ghostdogpr/caliban/blob/series/2.x/tools/src/main/scala/caliban/tools/Codegen.scala#L69

jfwilson:
PR to resolve https://github.com/ghostdogpr/caliban/issues/1510, where generated, formatted, files were getting an extra newline added causing `scalafmtCheck` to fail.

I've added `scalafmtCheck` commands to the `codegen/gen-client-task` test for coverage, but this has involved adding `scalafmt` config to the test project which could add to maintenance overhead. Happy to be guided as to how you would like the tests, if at all.

The fix in current form does mean that non-formatted files won't have a newline terminator which is not ideal, but I doubt anyone would be paying too much attention to the layout of unformatted files so haven't tried to address this.

### DIFF
--- a/codegen-sbt/src/sbt-test/codegen/gen-client-task/.scalafmt.conf
+++ b/codegen-sbt/src/sbt-test/codegen/gen-client-task/.scalafmt.conf
@@ -1,0 +1,4 @@
+version = "3.1.2"
+
+runner.dialect = scala213
+preset = default
--- a/codegen-sbt/src/sbt-test/codegen/gen-client-task/project/plugins.sbt
+++ b/codegen-sbt/src/sbt-test/codegen/gen-client-task/project/plugins.sbt
@@ -3,3 +3,4 @@ sys.props.get("plugin.version") match {
   case _       => sys.error("""|The system property 'plugin.version' is not defined.
                                |Specify this property using the scriptedLaunchOpts -D.""".stripMargin)
 }
+addSbtPlugin("org.scalameta" % "sbt-scalafmt" % "2.5.0")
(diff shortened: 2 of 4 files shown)


ghostdogpr (approved):
Looks perfect, thanks!

PR #1522, opened 2022-11-27
Compile benchmarks for Scala 3

kyri-petrou:
I was running some performance tests for Scala 2 vs 3 the other day and I thought you might be interested in being able to run the benchmarks for Scala 3 as well

### DIFF
--- a/benchmarks/src/main/scala/caliban/GraphQLBenchmarks.scala
+++ b/benchmarks/src/main/scala/caliban/GraphQLBenchmarks.scala
@@ -8,6 +8,7 @@ import sangria.execution._
 import sangria.macros.derive._
 import sangria.marshalling.circe._
 import sangria.parser.QueryParser
+import caliban.schema.{ Schema => CSchema }
 import sangria.schema._
 import zio.{ Runtime, Task, UIO, Unsafe, ZIO }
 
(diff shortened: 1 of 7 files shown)


kyri-petrou:
I've also updated the Scala versions in the `./sbt` file hoping it's going to make a difference with the flaky test (it failed again)

benchmarks/src/main/scala/caliban/GraphQLBenchmarks.scala
```diff
@@ -361,8 +369,7 @@ class GraphQLBenchmarks {
           new NoUndefinedVariables,
           new NoUnusedFragments,
           new NoUnusedVariables,
-          // new OverlappingFieldsCanBeMerged,
-          new experimental.OverlappingFieldsCanBeMerged,
+          new OverlappingFieldsCanBeMerged,
```
ghostdogpr, line 372: Is the experimental one the new default?

ghostdogpr (approved):

benchmarks/src/main/scala/caliban/GraphQLBenchmarks.scala
kyri-petrou: I believe so, `experimental.OverlappingFieldsCanBeMerged` doesn't exist anymore (was probably removed in Sangria 3.x)

ghostdogpr: Thanks!

PR #1527, opened 2022-12-01
Jsoniter-based JSON parsing

kyri-petrou:
Just putting this here so that you can take a look and let me know what you think

The one issue with the implementation - which I'm happy to hear your thoughts on, is that the JSON parsing is done in a stack recursive way, and to my understanding there's no straightforward way to rewrite it in a stack-safe way which would allow "infinite" recursion. To avoid stack overflow errors, I've currently hard-coded a max recursion depth of 512. This should be much more than enough in most real-life applications (if not all).

However, Caliban does technically allow for objects to be "infinitely" deep (unless the `maxDepth` wrapper is used) so I'm not sure what the right approach is. Perhaps we should check during runtime the maxDepth and which JSON codec is being used, and raise an error if a maxDepth is not set or is larger than the max value? Doesn't sound very elegant but it would at least prevent unexpected errors during runtime post-startup. Happy to hear your thoughts on this.

### TODO

* [x] Add doco
* [x] Decide on max recursion limit


core/src/main/scala/caliban/interop/tapir/tapir.scala
```diff
@@ -9,7 +9,7 @@ import sttp.tapir.{ Schema, SchemaType }
  */
 private[caliban] trait IsTapirSchema[F[_]]
 private[caliban] object IsTapirSchema {
-  implicit val isCirceEncoder: IsTapirSchema[Schema] = null
+  implicit val isTapirSchema: IsTapirSchema[Schema] = null
```
kyri-petrou, line 12: I renamed this because it was confusing whether the variable name meant this would only work with the circe encoder. Happy to change it back if it's an issue to rename it

kyri-petrou:
Aha.. seems that `jsoniter` doesn't work with Java 8 🙃 🥲

ghostdogpr:
How about making a zio-test aspect that only runs the test if jvm version is higher than x? Then we can use it to run that test on java 11 only.

kyri-petrou:
> How about making a zio-test aspect that only runs the test if jvm version is higher than x? Then we can use it to run that test on java 11 only.

Good call. I added a test aspect which can be used to skip tests when the JVM version is < 11

core/src/main/scala/caliban/interop/tapir/tapir.scala
ghostdogpr: Hmm looks like it was a copy paste mistake 😄

core/src/test/scala/caliban/TestUtils.scala
```diff
@@ -563,4 +564,11 @@ object TestUtils {
 
     val resolverEmpty = new RootResolver(Option.empty[Unit], Option.empty[Unit], Option.empty[Unit])
   }
+
+  val skipJdk8: TestAspectPoly = {
+    val jdkVersionArr = System.getProperty("java.version").split('.')
+    val maybeMajor    = jdkVersionArr.head.toInt
```
ghostdogpr: Use `headOption` so that it doesn't throw here. I think by default we just run the test if we can't identity the version.

interop/cats/src/test/scala/caliban/interop/cats/CatsInteropSpec.scala
```diff
@@ -25,7 +25,7 @@ object CatsInteropSpec extends ZIOSpecDefault {
       val inner   = RootContext(SecurityContext(true), LogContext("internal-trace-id"))
 
       def main(inner: RootContext)(implicit runtime: Runtime[RootContext]) =
-        Dispatcher[Effect].use { dispatcher =>
+        Dispatcher.sequential[Effect].use { dispatcher =>
```
ghostdogpr: The previous behavior was `parallel`, any reason to change to `sequential`?

examples/src/main/scala/example/jsoniter/ExampleHttp4sApp.scala
```diff
@@ -0,0 +1,76 @@
+package example.jsoniter
```
ghostdogpr: Regarding examples, I would rather not copy the http4s one which is pretty complex. How about just having one based on zioHttp/ExampleApp? I don't think we even need the auth example.

core/src/main/scala/caliban/interop/jsoniter/jsoniter.scala
```diff
@@ -0,0 +1,333 @@
...
+      ValueJsoniter.responseValueCodec.encodeValue(x.toResponseValue, out)
+    override def nullValue: CalibanError                                          =
+      null.asInstanceOf[CalibanError]
+  }
+}
+
+private[caliban] object GraphQLResponseJsoniter {
+  import com.github.plokhotnyuk.jsoniter_scala.macros._
+
+  private case class GraphQLResponseDTO(data: ResponseValue, errors: Option[List[CalibanError]])
```
paulpdaniels, line 297: We appear to be losing the `extension` mapping here.

core/src/test/scala/caliban/TestUtils.scala
kyri-petrou: Good point. Changed it to pattern matching instead since it makes the code a bit more clear I think

interop/cats/src/test/scala/caliban/interop/cats/CatsInteropSpec.scala
kyri-petrou: Not too sure what I was thinking here 🤦 . Changed it to `parallel`.

examples/src/main/scala/example/jsoniter/ExampleHttp4sApp.scala
kyri-petrou: Yeah that's a good point. Changed it to zioHttp, although I only implemented the `httpRoutes` since the implementation of the websockets is quite lengthy.

On this note, would it be worth it to add methods in the adapters where the codecs can be implicitly provided instead? Or do you recon it's a relatively rare usecase for users to change the default JSON implementation?

core/src/main/scala/caliban/interop/jsoniter/jsoniter.scala
kyri-petrou: This is only for decoding the response (which I believe we don't do anywhere). I followed the implementation from `circe` / `zio-json` where we do the same thing:
https://github.com/ghostdogpr/caliban/blob/ef94d73cadd976a82c949b33fa42813b9484cc32/core/src/main/scala/caliban/interop/circe/circe.scala#L160-L174

Happy  to change it - although if that's the case, should we also change it for the other json parsers?

kyri-petrou:
Test is failing due to an unrelated dependency that failed to resolve - restarting the workflow will likely fix it

examples/src/main/scala/example/jsoniter/ExampleHttp4sApp.scala
ghostdogpr: It's kind of historical, except for Akka the adapters were fixed to a particular Json library. Even if we exposed a method that took a codec instead, the adapters would still depend on a particular Json library so I imagine people wouldn't use them.

ghostdogpr:
> Test is failing due to an unrelated dependency that failed to resolve - restarting the workflow will likely fix it

Restarted it

kyri-petrou:
Happy to mark this as ready for review, although there's one thing we haven't discussed yet:

> The one issue with the implementation - which I'm happy to hear your thoughts on, is that the JSON parsing is done in a stack recursive way, and to my understanding there's no straightforward way to rewrite it in a stack-safe way which would allow "infinite" recursion. To avoid stack overflow errors, I've currently hard-coded a max recursion depth of 512. This should be much more than enough in most real-life applications (if not all)

I couldn't find of a rather elegant way to to add a check to see if the user has a `maxDepth` wrapper. Do we have any other options regarding this or are we happy with just documenting it? Since this codec is not used by default, I'm guessing that users interested in using it will look at the example and test it first

ghostdogpr:
I think documenting it is probably enough (https://ghostdogpr.github.io/caliban/docs/adapters.html#built-in-adapters + in the code).

kyri-petrou:
Thanks. Added comments + a warning box in the limitations toolbox. Not sure if you prefer it outside (or not as a warning box at all) - let me know and I can change it

vuepress/docs/docs/adapters.md
```diff
@@ -55,6 +55,12 @@ That's it! `http4sRoute` is a valid http4s route ready to serve our API.
 ::: tip Limitations
 The zio-http interpreter in tapir does not include multipart and websocket support.
 
-Caliban comes with json encoders and decoders for circe, zio-json and play-json.
+Caliban comes with json encoders and decoders for circe, zio-json, jsoniter and play-json.
 If you use another json library, you will need to create encoders and decoders for it (which is very simple, you can simply look at the existing ones).
+
```
ghostdogpr, line 60: Isn't an enclosing `:::` missing?

kyri-petrou: I thought so too, but if I add one then this appears on the page 
image

ghostdogpr: Interesting, I didn't know it supported nesting like that!

kyri-petrou: Me neither, I only found out cause I accidentally pasted it within the enclosing marker and saw the browser just in time 😅

kyri-petrou:
By the way I managed to rewrite the encoder using tail recursion, although I'm not entirely sure what the performance implications of this might be. I don't guess we have anything in `/benchmarks` that I could reuse to do some performance comparisons for this:

```scala
  private def encodeResponseValueTailRec(out: JsonWriter, root: ResponseValue) = {
    @tailrec
    def inner(value: LazyList[Either[ResponseValue, () => Unit]]): LazyList[Either[ResponseValue, () => Unit]] = {
      var stop   = true
      val buffer = LazyList.newBuilder[Either[ResponseValue, () => Unit]]
      value.foreach {
        case v @ Right(_) => buffer += v
        case Left(v)      =>
          v match {
            case v: Value                     =>
              buffer += Right(() => encodeValue(out)(v))
            case ResponseValue.ListValue(l)   =>
              buffer += Right(() => out.writeArrayStart())
              l.foreach(v => buffer += Left(v))
              buffer += Right(() => out.writeArrayEnd())
              stop = false
            case ResponseValue.ObjectValue(o) =>
              buffer += Right(() => out.writeObjectStart())
              o.foreach { case (k, v) =>
                buffer += Right(() => out.writeKey(k))
                buffer += Left(v)
              }
              buffer += Right(() => out.writeObjectEnd())
              stop = false
            case s: ResponseValue.StreamValue =>
              buffer += Right(() => out.writeVal(s.toString))
          }
      }
      val res    = buffer.result()
      if (stop) res else inner(res)
    }

    inner(LazyList(Left(root))).foreach {
      case Right(op) => op()
      case Left(_)   => throw new IllegalStateException("what the")
    }
  }
```

kyri-petrou:
The experiment with tail recursion above turned out it was performing very badly, so I'll bench it for now. I pushed an extra commit which optimises the encoding of the objects and added benchmarks for the different JSON implementations. Based on the test data as in the commit, these are the results:

```shell
[info] JsonEncodingBenchmarks.circe         thrpt    5     23.035 ±    0.248  ops/s
[info] JsonEncodingBenchmarks.jsoniter      thrpt    5     99.296 ±   19.913  ops/s
[info] JsonEncodingBenchmarks.zio           thrpt    5     38.769 ±    1.446  ops/s
```

I think there's quite a bit of room to improve the circe and zio encoders by changing the pattern matching the same way as in [this commit](https://github.com/ghostdogpr/caliban/pull/1527/commits/a47bd3e8e910afb8e009be2081c2056d9c2b4318), but I can do those changes in a followup PR.

guizmaii:
@kyri-petrou Super interesting results! Could you maybe try to run your benchmark with the change you're suggesting to see the diff?

kyri-petrou:
> @kyri-petrou Super interesting results! Could you maybe try to run your benchmark with the change you're suggesting to see the diff?

Sure thing. These are the results with the optimizations / improvements applied to both the circe and ZIO encoders:

```sh
[info] JsonEncodingBenchmarks.circe     thrpt    5  33.647 ±  0.194  ops/s
[info] JsonEncodingBenchmarks.jsoniter  thrpt    5  98.933 ± 18.321  ops/s
[info] JsonEncodingBenchmarks.zio       thrpt    5  67.650 ±  0.432  ops/s
```

I'm not sure why but `circe` seems to benefit the least from them.  `zio-json` on the other hand improves significantly, but still ~30% slower than jsoniter

guizmaii:
Thanks for that! ❤️

Weird how Circe doesn't profit from it indeed

ghostdogpr:
Separate PR would be welcome for the other json libs improvements! I will merge this one.

ghostdogpr (approved):

PR #1528, opened 2022-12-03
Update http4s to 0.23.16

kyri-petrou:
As part of this update, I've also swapped out Blaze for Ember in the tests / examples. The main reason for this change is that [since 0.23.12](https://github.com/http4s/http4s/releases/tag/v0.23.12), Blaze is no longer published as part of the standard http4s release cycle.

I'm happy to revert the changes to tests / examples, although that means we'd have to have a separate version for blaze and http4s


ghostdogpr (approved):
Thanks!

PR #1531, opened 2022-12-06
Update zio to 2.0.5 and zio-query to 0.3.2

kyri-petrou:
ScalaJS issue seems to be resolved in the new zio release 🎉

The ZLayer issue (https://github.com/zio/zio/issues/7506) seems to still be there for Scala 3 but it only affects the tests and the workaround is fairly simple


ghostdogpr (approved):
Awesome!

PR #1540, opened 2022-12-09
Change client codegen for _ fields

### LINKED ISSUE #1388: Make client codegen work when a field is named _
The generated code doesn't compile when a field is named `_` (example: https://gql.waveapps.com/graphql/public). It's escaped with backticks but we get:
```
[error] 7642 |    def `_`
[error]      |        ^
[error]      |        wildcard invalid as backquoted identifier
```
We need to escape this one another way.

guymers:
A underscore with backticks is not valid in Scala 3 so convert them into `_$` instead.

Closes #1388


ghostdogpr:
LGTM, thanks!

PR #1544, opened 2022-12-15
Fixing rendering of descriptions for multi-line descriptions ending in quote

### LINKED ISSUE #1449: Rendering. renderDescription produces SDL that can't be parsed
I'm having a problem with the parsing/rendering code in caliban. I have a couple of cases where the SDL generated by Rendering can't be parsed by RemoteSchema (or by a bunch of other graphQL tools for that matter).

The following snippet shows the problem.
```scala
import caliban.Rendering
import caliban.tools.{RemoteSchema, SchemaLoader}
import zio.Console._
import zio._

object ReproduceBug extends ZIOAppDefault {
  val tripleQuoteSchema =
      "\"\"\"This is a description \nwith a \"quote at the end\"\n\"\"\"" +
      """
        |type MyType {
        |   aField: String
        |}
        |
        |type Query{someQuery: MyType}
        |""".stripMargin

  def parseSchema(schema: String) = SchemaLoader.fromString(schema).load.flatMap(d => ZIO.fromOption(RemoteSchema.parseRemoteSchema(d)))

  override def run = {
    var topType = """schema{
                    |   query: Query
                    |}
                    |""".stripMargin
    for {
      schema <- parseSchema(topType + tripleQuoteSchema)
      _ <- printLine(s"Schema read ok")
      renderedSchema <- ZIO.attempt(Rendering.renderTypes(schema.types))
      _ <- printLine(s"Rendered schema contains four double quotes on description:\n$renderedSchema")
      parsedRenderedSchema <- parseSchema(topType + renderedSchema).catchAll(_ =>
        printLine("ERROR: Rendered schema can't be parsed")) *> ZIO.succeed()
    } yield ()
  }
}
```

This is how I'm working around the issue.
```scala
  private def renderDescription(description: Option[String], newline: Boolean = true): String = description match
    case None                   => ""
    case Some(value) if newline =>
      if value.contains("\n") then
        renderTripleQuotedString("\n" + value + (if value.endsWith("\n") then "" else "\n")) + "\n"
      else renderString(value) + "\n"
    case Some(value)            =>
      if value.contains("\n") then renderTripleQuotedString(value + (if value.endsWith("\"") then " " else "")) + " "
      else renderString(value) + " "
```
If you think this is the correct fix I can submit a PR. 

Another concern I have as I read the October 2021 spec the grammar production for BlockString should not be greedy. So the parser should deal with BlockStrings that end in `"""`, `""""` and `"""""`. 

All the tools/parsers I tried are greedy since it is a lot easier to implement without looking ahead the incoming tokens. This makes my code necessary to interact with other tools.

camarena:
Even though the spec indicates that parsing triple quotes should not be greedy most of the tools (even ours) are. This PR breaks the rendering of a description that will result in four unescaped quotes.

Fixes #1449

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,4 +39,5 @@ Want to see your company here? [Submit a PR](https://github.com/ghostdogpr/calib
 * [LeadIQ](https://leadiq.com)
 * [Sanjagh.pro](https://sanjagh.pro)
 * [Soundtrack Your Brand](https://www.soundtrackyourbrand.com)
+* [StepZen](https://www.stepzen.com)
 * [Undo](https://www.undo.app)
--- a/core/src/main/scala/caliban/Rendering.scala
+++ b/core/src/main/scala/caliban/Rendering.scala
@@ -171,14 +171,15 @@ object Rendering {
       }
     }
 
-    def nl =
-      if (newline) "\n" else " "
+    def enl = if (newline) "\n" else " "
+
+    def bnl = if(newline) "\n" else ""
 
     description match {
       case None                                   => ""
       case Some(value) if value.exists(_ == '\n') =>
-        s"${renderTripleQuotedString(s"$nl$value")}$nl"
-      case Some(value)                            => s"${renderString(value)}$nl"
+        s"${renderTripleQuotedString(s"$bnl$value")}$enl"
+      case Some(value)                            => s"${renderString(value)}$enl"
     }
   }
 
--- a/core/src/main/scala/caliban/Rendering.scala
+++ b/core/src/main/scala/caliban/Rendering.scala
@@ -149,11 +149,37 @@ object Rendering {
       case _                       => ""
     }
 
-  private def renderDescription(description: Option[String], newline: Boolean = true): String = description match {
-    case None                   => ""
-    case Some(value) if newline =>
-      if (value.contains("\n")) renderTripleQuotedString("\n" + value) + "\n" else renderString(value) + "\n"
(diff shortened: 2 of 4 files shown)


README.md
```diff
@@ -39,4 +39,5 @@ Want to see your company here? [Submit a PR](https://github.com/ghostdogpr/calib
 * [LeadIQ](https://leadiq.com)
 * [Sanjagh.pro](https://sanjagh.pro)
 * [Soundtrack Your Brand](https://www.soundtrackyourbrand.com)
+* [StepZen](https://www.stepzen.com)
```
ghostdogpr, line 42: Nice!

core/src/main/scala/caliban/Rendering.scala
```diff
@@ -149,11 +149,34 @@ object Rendering {
       case _                       => ""
     }
 
-  private def renderDescription(description: Option[String], newline: Boolean = true): String = description match {
-    case None                   => ""
-    case Some(value) if newline =>
-      if (value.contains("\n")) renderTripleQuotedString("\n" + value) + "\n" else renderString(value) + "\n"
-    case Some(value)            => if (value.contains("\n")) renderTripleQuotedString(value) else renderString(value) + " "
+  private def renderDescription(description: Option[String], newline: Boolean = true): String = {
+    // Most of the graphql tools (including ours) are greedy on triple quotes. To be compatible we
```
ghostdogpr:
```suggestion
    // Most of the graphql tools are greedy on triple quotes. To be compatible we
```
Removed this because it doesn't make sense without context

ghostdogpr (approved):
LGTM! Just a small remark

PR #1552, opened 2022-12-23
Make all adapters usable with any json library

### LINKED ISSUE #1548: Make all adapters usable with any json library
Instead of hardcoding the Json library in some adapters, expose the implicit `JsonCodec` that are needed by the underlying `TapirAdapter` (this approach is currently applied to the Akka Http Adapter only, the other 3 are hardcoded).

Update the docs to indicate to users which dependency and import they need depending on the Json library.

The examples project can show various combinations (e.g. play/play-json, zio-http/zio-json, http4s/circe, akka/jsoniter).

kyri-petrou:
Closes #1548

There are 2 things that I'm not sure about:

1. Should we deprecate the `makeXXX` methods instead of "breaking" them, and provide new methods that require the implicits instead? This would make the changes backwards compatible, _but_ we'd have to duplicate the method signatures many times with different names.
2. Should the tests run against all the json implementations?

I gave it a good go with docs, but I feel like they could be improved a bit. Let me know if you have any recommendations on how to improve them


adapters/zio-http/src/main/scala/caliban/ZHttpAdapter.scala
```diff
@@ -60,15 +64,21 @@ object ZHttpAdapter {
...
                    .collectZIO[WebSocketChannelEvent] {
                      case ChannelEvent(ch, UserEventTriggered(HandshakeComplete)) =>
                        out.runForeach(ch.writeAndFlush(_)).race(ch.awaitClose)
                      case ChannelEvent(_, ChannelRead(WebSocketFrame.Text(text))) =>
-                       ZIO.fromEither(decode[GraphQLWSInput](text)).flatMap(queue.offer)
+                       ZIO.fromEither {
+                         inputCodec.decode(text) match {
+                           case DecodeResult.Value(v)    => Right(v)
+                           case DecodeResult.Error(_, e) => Left(e)
+                           case _                        => Left(new Throwable("failed to decode input"))
```
kyri-petrou: Do we need to customise the error message for all the different decoding failures here? There are the following failures that are currently not handled explicitly: `InvalidValue`, `Mismatch`, `Missing`, `Multiple`. The issue with these methods is that they don't wrap around a `Throwable`

ghostdogpr:
> 1. Should we deprecate the `makeXXX` methods instead of "breaking" them, and provide new methods that require the implicits instead? This would make the changes backwards compatible, _but_ we'd have to duplicate the method signatures many times with different names.

I think we should delete them, otherwise the adapters will still depend on some json libraries, what we want is to remove those dependencies.

> 2. Should the tests run against all the json implementations?

Probably not since we already have individual tests, but we can use a different one for each adapter to increase coverage (e.g. play+play-json, http4s+circe, zio-http+zio-json, akka-http+jsoniter)

adapters/zio-http/src/main/scala/caliban/ZHttpAdapter.scala
ghostdogpr: Or maybe `toString` the failure instead?

kyri-petrou:
> but we can use a different one for each adapter to increase coverage (e.g. play+play-json, http4s+circe, zio-http+zio-json, akka-http+jsoniter)

I think that's a good approach. I'll make the change

kyri-petrou:
Interestingly, after changing the `TapirAdapterSpec.makeSuite` method to require all the implicits (both for encoding and decoding) that unearthed some bugs in the json decoders for play-json and zio-json 😅.

The issue is very minimal though since it only affected the decoders for the `GraphQLResponse` value, which we don't decode anywhere in the server logic

kyri-petrou:
>>but we can use a different one for each adapter to increase coverage (e.g. play+play-json, http4s+circe, zio-http+zio->>json, akka-http+jsoniter)
>
>I think that's a good approach. I'll make the change

Forgot that jsoniter works only on JDK11+. So that we can still run the tests for AkkaAdapter against JDK8, I used play-json there and made the Http4sAdapterSpec run against both the circe and jsoniter codecs, ignoring the jsoniter one on JDK8.

I also mentioned that jsoniter works only on JDK11+ in the docs

adapters/zio-http/src/main/scala/caliban/ZHttpAdapter.scala
kyri-petrou: Done

ghostdogpr (commented):
Looks awesome, and some nice cleanups! Before I review the docs, can you also modify the README that contains the Getting Started docs? We need to add the list of optional Json dependencies and mention the required import there. Thanks!

kyri-petrou:
> Looks awesome, and some nice cleanups! Before I review the docs, can you also modify the README that contains the Getting Started docs? We need to add the list of optional Json dependencies and mention the required import there. Thanks!

Ahh good catch, totally forgot about that part! I gave it a go, although I'm not too sure what to do with the tapir version. Should we specify one (e.g., `1.2.2`) or just let the user define it? I wonder if there's a way to make it possible for the users to get the same tapir version that the caliban libraries are using. I tried finding a way via SBT but couldn't figure out one.

ghostdogpr:
Let's put the current version we depend on, so that they can just copy paste the dependencies easily and we know it's gonna work. And if there's a breaking change when we upgrade tapir, let's try to update it at the same time.

kyri-petrou:
Sounds good. Done 👍

ghostdogpr (approved):
LGTM!

Would you mind moving the docs changes to another PR? Then I can merge this one first and the docs one after the next release (merging docs triggers an update immediately).

PR #1557, opened 2022-12-28
Caching of validation in the `ApolloPersistedQueries` wrapper

kyri-petrou:
Followup of [chat on Discord](https://discord.com/channels/629491597070827530/633200096393166868/1056792140983828652).

~~I gave it a good go with not breaking the `ValidationWrapper` API but I couldn't figure out a decent way to skip the validation altogether since the `Validator.prepare` and `VariablesCoercer.coerceVariables` methods also transform the data. However, I _think_ that unless we want to refactor these methods, it makes sense for the ValidationWrapper to be able to skip validation at least via the `skipValidation` argument - happy to hear your thoughts.~~

~~The other thing I'm not too sure about is the usage of `.hashCode()`. We could probably use `.toString` and hash the resulting string using `MD5` or `SHA1` instead, which would also allow distributed caching (e.g., via Redis), although I'm not sure whether any users do use APQ with a distributed cache.~~

~~Leaving this as draft for now just to discuss the breaking changes and the method used to cache the input variables~~


core/src/main/scala/caliban/wrappers/ApolloPersistedQueries.scala
```diff
@@ -1,58 +1,124 @@
...
+          case FloatValue.DoubleNumber(_)     => ZIO.succeed(FloatValue.DoubleNumber(0d))
+          case FloatValue.BigDecimalNumber(_) => ZIO.succeed(FloatValue.BigDecimalNumber(0))
+          case StringValue(_)                 => ZIO.succeed(StringValue("?"))
+          case BooleanValue(_)                => ZIO.succeed(BooleanValue(true))
+          case EnumValue(v)                   => ZIO.succeed(EnumValue(v))
+          case InputValue.ListValue(l)        => ZIO.foreach(l)(stripVariableValues).map(InputValue.ListValue(_))
+          case InputValue.ObjectValue(o)      =>
+            val sortedMap = TreeMap.empty[String, InputValue] ++ o
+            ZIO.foreach(sortedMap) { case (k, v) => stripVariableValues(v).map(k -> _) }.map(InputValue.ObjectValue(_))
+          case InputValue.VariableValue(_)    => ZIO.succeed(InputValue.VariableValue("?"))
```
kyri-petrou: I couldn't find anything to suggest that for `VariableValue` we perform validation against the actual value, but leaving this comment just in case I missed something

core/src/main/scala/caliban/GraphQL.scala
```diff
@@ -117,19 +117,20 @@ trait GraphQL[-R] { self =>
...
+                                          .flatMap { validatedRequest =>
+                                            Validator.prepare(
+                                              input.document,
+                                              typeToValidate,
+                                              schemaToExecute,
+                                              validatedRequest.operationName,
+                                              validatedRequest.variables.getOrElse(Map.empty),
+                                              input.skipValidation
+                                            )
+                                          }
```
ghostdogpr: Shall we just skip the validation wrappers if `skipValidation` is true? 🤔

kyri-petrou: I think that's a good idea, although what worries me is whether there are any users that have validation wrappers (either custom or provided) and specify `skipValidation=true`. I'm happy to make the change if we think that won't be an issue.

A potential safe(r) way to do it (albeit somewhat messy) would be to add another field to the [ValidationWrapperInput](https://github.com/ghostdogpr/caliban/pull/1557/files/3883d333ced66bb8bcbcf290f6e1b9051de0a536#diff-8ffb6cba0c54bbf92fc33727497416acdd30f5a0c20c20109ccf5a79908c3512R153) (e.g., `skipValidationWrappers: Boolean`).

kyri-petrou: Actually, now that I think about it a bit more, I don't think we can skip validation wrappers altogether. Ones like the `CostEstimation.queryCost` and `ApolloTracing.apolloTracing` might be a bit too messy to skip

paulpdaniels:
I'm on vacation now. I'll take a look at this next week if you don't mind 🙏

core/src/main/scala/caliban/GraphQL.scala
kyri-petrou: @ghostdogpr I have some local changes to skip validation wrappers (ones that make sense) when `skipValidation=true`. Do you prefer if I add them to this PR or to a followup one so that we can discuss them separately?

core/src/main/scala/caliban/wrappers/ApolloPersistedQueries.scala
```diff
@@ -1,58 +1,123 @@
...
+      ZIO.serviceWithZIO[ApolloPersistence](_.isValidated(hash, variablesHashCode))
+    def registerValidation(hash: String, variablesHashCode: Int): ZIO[ApolloPersistence, Nothing, Unit] =
+      ZIO.serviceWithZIO[ApolloPersistence](_.registerValidation(hash, variablesHashCode))
 
-    val live: UIO[ApolloPersistence] = Ref.make[Map[String, Document]](Map()).map { cache =>
-      new ApolloPersistence {
-        override def get(hash: String): UIO[Option[Document]]      = cache.get.map(_.get(hash))
-        override def add(hash: String, query: Document): UIO[Unit] = cache.update(_.updated(hash, query))
+    val live: UIO[ApolloPersistence] =
+      (ConcurrentMap.empty[String, Document] <*> ConcurrentSet.empty[(String, Int)]).map {
```
kyri-petrou: I think using `ConcurrentMap[String, Document]` is a better choice here since it avoids copying the underlying Map on each update.

Also, these data structures are ultra thin wrappers around `ConcurrentHashMap`, so we could avoid the `zio-concurrent` dependency and use those directly. Let me know what you think

ghostdogpr: In general I want to limit dependencies but it might be worth it for this one. I wonder if there are any other places this could be used.

core/src/main/scala/caliban/GraphQL.scala
ghostdogpr: Maybe do all here if it's a breaking change on the same thing

core/src/main/scala/caliban/wrappers/ApolloPersistedQueries.scala
kyri-petrou: > I wonder if there are any other places this could be used.

That was my initial thinking as well. I had a look around the codebase but didn't find any obvious ones. The only `Ref` that I found that wraps an immutable collection is [this one](https://github.com/ghostdogpr/caliban/blob/cd672ccb6cdb38a015a780f3e7a9da2186bbc196/core/src/main/scala/caliban/execution/Executor.scala#L166), but I think that's fine since it's unlikely to have enough errors to make a difference + prepending on a List doesn't create a whole new object and is very fast.

I took it a step further and had a look at places where we have `ListA :+ element` or `ListA ++ ListB` which are quite bad for performance. There are a couple of places in the codebase although I'm not sure what is the expected size of the lists in those places to justify changing them to a more appropriate collection for append ops like a Vector

ghostdogpr: Yeah I did a round of optimization a while ago and use mutable builders in places it mattered.

core/src/main/scala/caliban/GraphQL.scala
kyri-petrou: Done!

core/src/main/scala/caliban/wrappers/Wrapper.scala
```diff
@@ -21,8 +21,13 @@ import scala.annotation.tailrec
  * - `FieldWrapper` to wrap each field execution
  *
  * It is also possible to combine wrappers using `|+|` and to build a wrapper effectfully with `EffectfulWrapper`.
+ *
+ * Implementations can control the order at which this wrapper is executed by overriding the `priority` value.
+ * Setting a higher `priority` value will be executed first.
  */
 sealed trait Wrapper[-R] extends GraphQLAspect[Nothing, R] { self =>
+  val priority: Int = 0
```
kyri-petrou, line 29: Should this be `private[caliban] val priority = ...` instead?

core/src/main/scala/caliban/wrappers/ApolloPersistedQueries.scala
```diff
@@ -1,58 +1,125 @@
...
+            case Some((hash, _, None))   => f(query).tap(doc => ApolloPersistence.add(hash, doc))
+            case None                    => f(query)
           }
     }
 
+  private def validationWrapper(
+    docVar: Promise[Nothing, Option[(String, GraphQLRequest, Option[Document])]]
+  ): ValidationWrapper[ApolloPersistence] =
+    new ValidationWrapper[ApolloPersistence] {
+      override val priority: Int = Int.MaxValue // Run this wrapper at the outmost boundary
```
paulpdaniels: This feels like a tricky new semantic. Because if you had two wrappers that _needed_ to run last, the order of definition would still then be important right (assuming the underlying sort is stable). Do we actually need it? Or could we just tell the user that they need to put this one last?

ghostdogpr:
I like the concept of priority, it's simple and doesn't require any change for existing users while giving more power to advanced ones. 

Regarding passing skipValidation to wrappers, I think maybe we don't need that? The typical use case for it is to have validation in dev and no validation in prod (for performance) but then you can also control the wrappers you attach in each case. Curious to have others' opinion about it.

kyri-petrou:
> I like the concept of priority, it's simple and doesn't require any change for existing users while giving more power to advanced ones.
> 
> Regarding passing skipValidation to wrappers, I think maybe we don't need that? The typical use case for it is to have validation in dev and no validation in prod (for performance) but then you can also control the wrappers you attach in each case. Curious to have others' opinion about it.

Interesting, I wasn't aware that the common pattern is to disable validation in prod. What's the "worst-case scenario" if validation is disabled and the user input is "wrong" (for lack of a better word)?

ghostdogpr:
> Interesting, I wasn't aware that the common pattern is to disable validation in prod. What's the "worst-case scenario" if validation is disabled and the user input is "wrong" (for lack of a better word)?

Maybe a stack-overflow in case of cyclic fragments? In general I would say unexpected errors during the execution phase.

core/src/main/scala/caliban/wrappers/ApolloPersistedQueries.scala
frekw: I think we might need slightly more advanced data structures here that limit the number of entries in both caches. Right now an attacker would be able to DoS the server by sending a lot of unique docs to fill up the caches.

kyri-petrou: > I think we might need slightly more advanced data structures here that limit the number of entries in both caches. Right now an attacker would be able to DoS the server by sending a lot of unique docs to fill up the caches.

That's a good point. In fact at work I provide a custom implementation based on [Caffeine](https://github.com/ben-manes/caffeine) for this reason.

I doubt this wrapper alone justifies bringing in such a dependency, although I think we could detect the number of entries and clear some / all when adding a new key if its size exceeds a certain size. The runtime overhead should be minimal since in most cases new entries should be rare except in the case of an attack.

kyri-petrou:
@ghostdogpr Apologies, I've been very busy this past week and forgot to respond

> Maybe a stack-overflow in case of cyclic fragments

I'm guessing that validation should not be disabled for internet-facing APIs? This sounds like something that could be exploited by an attacker.

ghostdogpr:
> I'm guessing that validation should not be disabled for internet-facing APIs? 

It's up to the user 😄 But yeah in general it's recommended to keep it.

core/src/main/scala/caliban/GraphQL.scala
```diff
@@ -117,19 +117,20 @@ trait GraphQL[-R] { self =>
...
-                                          .prepare(
-                                            doc,
-                                            typeToValidate,
-                                            schemaToExecute,
-                                            validatedRequest.operationName,
-                                            validatedRequest.variables.getOrElse(Map.empty),
-                                            skipValidation
-                                          )
-                  executionRequest <- wrap(validate)(validationWrappers, doc)
+                  validate          = (input: ValidationWrapperInput) =>
```
paulpdaniels: Just a thought but what if we pushed the `skipValidation` into a `FiberRef` similar to how `ZQuery` handles [caching](https://github.com/zio/zio-query/blob/series/2.x/zio-query/shared/src/main/scala/zio/query/ZQuery.scala#L1474). I think that would require fewer API changes

core/src/main/scala/caliban/wrappers/ApolloPersistedQueries.scala
paulpdaniels: Do the other implementations do this as well? I'm just curious if we actually _need_ to cache the variables or if it's a quirk of our current flow.

kyri-petrou:
I recently realised an issue with this PR and I'm not sure how to proceed with it. It currently assumes that variables will be of the same type but different values across different invocations of the same query. This works for most datatypes but not for dynamic ones such as JSON, which would force the validation on each invocation if the JSON shape has changed.

What do we think it's the worst-case scenario if we validated variables only on the first invocation (i.e., before the query is cached) and skip validation altogether (including variables) if it's cached? Are there risks of SO errors or is that not possible?

core/src/main/scala/caliban/GraphQL.scala
kyri-petrou: Hmm that could work. I'll give that a go!

core/src/main/scala/caliban/wrappers/ApolloPersistedQueries.scala
kyri-petrou: What if we changed this from `Int.MaxValue` to some arbitrary number (e.g, 100)? As for the ordering, it might work but that requires existing users to be aware of this change. My thinking here was that existing users get the benefit without having to do any changes to their code.

One other benefit of using a number for the priority is for wrappers that have multiple underlying wrappers, but you'd want those to run in different order

core/src/main/scala/caliban/wrappers/ApolloPersistedQueries.scala
kyri-petrou: I think you might be right. I've identified another issue with this approach (see https://github.com/ghostdogpr/caliban/pull/1557#issuecomment-1445474759). Happy to hear your thoughts

kyri-petrou:
> I like the concept of priority, it's simple and doesn't require any change for existing users while giving more power to advanced ones.
> 
> Regarding passing skipValidation to wrappers, I think maybe we don't need that? The typical use case for it is to have validation in dev and no validation in prod (for performance) but then you can also control the wrappers you attach in each case. Curious to have others' opinion about it.

@ghostdogpr I followed @paulpdaniels suggestion and modelled the propagation of the `skipValidation` flag via a `FiberRef` to avoid breaking changes to the API. I also made the caching not be dependent on the variables to avoid issues with dynamic types (such as JSON). I haven't found something to justify that it'll cause issues, but if you have any tests in mind that it'd be good to add let me know.

core/src/main/scala/caliban/wrappers/ApolloPersistedQueries.scala
kyri-petrou: I disabled the skipping of `VariablesCoercer.coerceVariables` for APQs and it seems that's all we need to validate the inputs. This step is very quick so we still get most of the benefit from caching the query validation

core/src/main/scala/caliban/wrappers/ApolloPersistedQueries.scala
kyri-petrou: As mentioned on Discord, I changed the implementation to use Java's `ConcurrentHashMap` directly. It felt a bit too much to bring in the `zio.concurrent` dependency for such a thin wrapper

core/src/main/scala/caliban/wrappers/ApolloPersistedQueries.scala
kyri-petrou: Changed this to `100` instead, this way users can have a higher priority wrapper if they wish so

core/src/main/scala/caliban/wrappers/CostEstimation.scala
```diff
@@ -201,13 +203,15 @@ object CostEstimation {
...
         (doc: Document) =>
-          for {
-            req  <- process(doc)
-            cost <- computeCostZIO(req.field)(f)
-            _    <- ZIO.when(cost > maxCost)(
-                      ZIO.fail(ValidationError(s"Query costs too much: $cost. Max cost: $maxCost.", ""))
-                    )
-          } yield req
+          process(doc).tap { req =>
+            ZIO.unlessZIO(Validator.skipQueryValidationRef.get) {
```
paulpdaniels: I'm wondering if this needs another parameter to allow the user to bypass the validation skip here. Since it is possible that the `Double` is returning some aggregate cost value. So the user may want this to apply to all queries regardless of if they are persisted

kyri-petrou: That's a good point. I added the `skipForPersistedQueries: Boolean = false` argument to the `maxCostXX` methods. I used `false` as the default value so that we don't affect the behaviour for existing / unaware users

core/src/main/scala/caliban/wrappers/Wrapper.scala
```diff
@@ -49,7 +54,7 @@ object Wrapper {
 
   /**
    * Wrapper for the query validation stage.
-   * Wraps a function from a `Document` to a `ZIO[R, ValidationError, ExecutionRequest]`.
+   * Wraps a function from a `ValidationInput` to a `ZIO[R, ValidationError, ExecutionRequest]`.
```
ghostdogpr: It's still `Document` after all

ghostdogpr (commented):
Looks okay to me but I am not too familiar with these wrappers. @paulpdaniels you good?

core/src/main/scala/caliban/wrappers/Wrapper.scala
kyri-petrou: Oops 😅 Done

paulpdaniels (approved):
LGTM

ghostdogpr (approved):

PR #1562, opened 2023-01-03
refactor: add make case classes final

tusharmath:
fixes: Unable to generate ZIO Schema for the ADT.

### DIFF
--- a/core/src/main/scala/caliban/Value.scala
+++ b/core/src/main/scala/caliban/Value.scala
@@ -123,19 +123,19 @@ object Value {
         Try(LongNumber(s.toLong)) getOrElse
         BigIntNumber(BigInt(s))
 
-    case class IntNumber(value: Int)       extends IntValue {
+    final case class IntNumber(value: Int)       extends IntValue {
       override def toInt: Int       = value
       override def toLong: Long     = value.toLong
       override def toBigInt: BigInt = BigInt(value)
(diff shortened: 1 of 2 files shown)


ghostdogpr:
How about just doing it for the ADTs? We don't need it in tests and it make the code longer.

tusharmath:
Made the changes.

tusharmath:
Will this be automatically publish in the snapshot?

core/src/main/scala/caliban/Value.scala
```diff
@@ -149,19 +149,19 @@ object Value {
...
       override def toFloat: Float           = value.toFloat
       override def toDouble: Double         = value
       override def toBigDecimal: BigDecimal = BigDecimal(value)
       override def toString: String         = value.toString
     }
-    case class BigDecimalNumber(value: BigDecimal) extends FloatValue {
+    final case class BigDecimalNumber(value: BigDecimal) extends FloatValue {
       override def toFloat: Float           = value.toFloat
       override def toDouble: Double         = value.toDouble
       override def toBigDecimal: BigDecimal = value
```
tusharmath, line 167: This was required for ZIO Schema to encode `Document`

ghostdogpr (approved):

ghostdogpr:
Yep, a snapshot will be created in a few minutes.

PR #1591, opened 2023-01-29
Introduce semi-auto derivation

### LINKED ISSUE #1579: Experiment semi-auto derivation
Automatic derivation is great when you get started but for large schemas it's more efficient to explicit call `gen` for every type. The problem is that because auto-derivation is always on, it's hard to control which types are derived automatically or not. 

We could try to remove automatic derivation and add an import for it instead.

ghostdogpr:
Closes #1579

Scala 3 only ❤️:
```scala
// semi-auto derivation, needs a Schema for inner types
final case class Foo(value: String) derives Schema.SemiAuto

// auto derivation, generates a Schema for inner types
final case class Foo(value: String) derives Schema.Auto

// when R in Schema[R, A] is not Any
object CustomSchema extends SchemaDerivation[MyEnv]
final case class Foo(value: String) derives CustomSchema.SemiAuto
final case class Foo(value: String) derives CustomSchema.Auto

// semi-auto derivation without the derives keyword
implicit val fooSchema: Schema[MyEnv, Foo] = Schema.gen

// auto derivation without the derives keyword
import schema.auto._
implicit val fooSchema: Schema[MyEnv, Foo] = Schema.gen
```
Scala 2 (also cross-compile with Scala 3):
```scala
// semi-auto derivation, needs Schema for inner types
implicit val fooSchema: Schema[Any, Foo] = Schema.gen

// auto derivation, generates a Schema for inner types
import Schema.auto._
implicit val fooSchema: Schema[Any, Foo] = Schema.genAll

// semi-auto derivation when R in Schema[R, A] is not Any
object schema extends GenericSchema[MyEnv]
implicit val fooSchema: Schema[MyEnv, Foo] = schema.gen

// auto derivation when R in Schema[R, A] is not Any
object schema extends GenericSchema[MyEnv]
import schema.auto._
implicit val fooSchema: Schema[MyEnv, Foo] = schema.genAll
```

TODO: 
- docs


ghostdogpr:
`scripted3_jdk` has been flaky for some time, it is really annoying. There is no error, the test is just interrupted.
`test213_docs` fails because of the docs that I haven't updated yet.

ghostdogpr:
Using a large resource class in CircleCI seems to have fixed the issue.
Next up I'll apply the change to `ArgBuilder`!

core/src/main/scala-2/caliban/schema/SchemaDerivation.scala
```diff
@@ -231,20 +231,28 @@ trait SchemaDerivation[R] extends LowPriorityDerivedSchema {
...
+  def gen[R0, T]: Typeclass[T] = macro Magnolia.gen[T]
 
   /**
    * Returns an instance of `Schema` for the given type T.
-   * For a case class or sealed trait, you can call `genMacro[T].schema` instead to get more details if the
-   * schema can't be derived.
+   * This method will automatically generate missing `Schema` for all types nested inside T that are case classes or sealed traits.
    */
-  def gen[R0, T](implicit derived: Derived[Schema[R0, T]]): Schema[R0, T] = derived.schema
+  def genAll[R0, T](implicit derived: Derived[Schema[R0, T]]): Schema[R0, T] = derived.schema
```
guizmaii: Shouldn't we put this in the `AutoSchemaDerivation` as it's a kind of auto mode? 🤔

core/src/main/scala-2/caliban/schema/SchemaDerivation.scala
```diff
@@ -231,20 +231,28 @@ trait SchemaDerivation[R] extends LowPriorityDerivedSchema {
...
 
+  object auto extends AutoSchemaDerivation[R]
```
guizmaii, line 246: I think this encourages the user to use this mode, what about not adding it here?

guizmaii:
> scripted3_jdk has been flaky for some time, it is really annoying. There is no error, the test is just interrupted.

OOM? 🤔

core/src/main/scala-2/caliban/schema/SchemaDerivation.scala
ghostdogpr: ah so it would be available only as `Schema.auto.genAll`, why not 🤔

core/src/main/scala-2/caliban/schema/SchemaDerivation.scala
ghostdogpr: where do you suggest then?

core/src/main/scala-2/caliban/schema/SchemaDerivation.scala
ghostdogpr: Moved it

core/src/main/scala-2/caliban/schema/SchemaDerivation.scala
guizmaii: In the `AutoSchemaDerivation` trait maybe? But not in the non-automatic mode

guizmaii: Sorry to come back on this but why do you want to add this `auto` thing in this `SchemaDerivation` trait? 🤔

core/src/main/scala-3/caliban/schema/SchemaDerivation.scala
```diff
@@ -248,8 +248,21 @@ trait SchemaDerivation[A] {
...
     annotations.collectFirst { case GQLDefault(v) => v }
+}
 
-  inline given gen[R, A]: Schema[R, A] = derived[R, A]
+trait SchemaDerivation[R] extends CommonSchemaDerivation {
+  inline def gen[R, A]: Schema[R, A] = derived[R, A]
 
   inline def genDebug[R, A]: Schema[R, A] = PrintDerived(derived[R, A])
+
+  final lazy val auto = new AutoSchemaDerivation[Any] {}
```
guizmaii, line 315: Same question as here: https://github.com/ghostdogpr/caliban/pull/1591/files#r1104112132 🤔

core/src/main/scala-2/caliban/schema/SchemaDerivation.scala
ghostdogpr: I want to enabled auto-derivation using this import: `import caliban.schema.Schema.auto._`

guizmaii: Why not put this object into the `Schema` object then? 🤔

ghostdogpr: Because when `R` is not `Any` in Scala 2, you need to extend `GenericSchema[R]` and import `auto._` from that object. `Schema` is fixed to `Any`.

kyri-petrou:
Great work on this! I built this locally and tried it in a Scala 3 codebase and works great! Got just a couple of questions:

1. Is it expected that a schema is required for fields excluded via `@GQLExcluded`? e.g., the following fails unless we also derive a schema for `Foo`.

```scala 
case class Foo(foo: String)
case class Bar(bar: String, @GQLExcluded foo: Foo)
given Schema[Any, Bar] = Schema.gen
```

2. In Scala 3, is `Schema.auto.genAll` meant to work without importing `Schema.auto._` or is that just for Scala 2?

ghostdogpr:
> 1. Is it expected that a schema is required for fields excluded via `@GQLExcluded`? e.g., the following fails unless we also derive a schema for `Foo`.

Ah yeah, the exclusion happens much later currently.

> 2. In Scala 3, is `Schema.auto.genAll` meant to work without importing `Schema.auto._` or is that just for Scala 2?

You still need the import with both Scala 2 and 3. With Scala 3 `genAll` does exactly the same thing as `gen`, I just added it for the ability to cross-compile.

kyri-petrou:
> Ah yeah, the exclusion happens much later currently.

Outside the scope of this PR, but I _think_ it should be possible to exclude them with something along the lines of (for Scala 3 at least): 

```scala
// Macros.scala
  inline def isExcluded[T]: Boolean                         = ${ isExludedImpl[T] }
  def isExludedImpl[T: Type](using qctx: Quotes): Expr[Boolean] = {
    import qctx.reflect.*
    val tpe = TypeRepr.of[T]
    Expr(tpe.typeSymbol.annotations.exists { v =>
      v.tpe =:= TypeRepr.of[GQLExcluded]
    })
  }

// SchemaDerivation.scala
  inline def recurse[R, Label, A <: Tuple](index: Int = 0): List[(String, List[Any], Schema[R, Any], Int)] =
    inline erasedValue[(Label, A)] match {
      case (_: (name *: names), _: (t *: ts)) =>
        if Macros.isExluded[t] then recurse[R, names, ts](index + 1)
        else {
          val label = constValue[name].toString
          val annotations = Macros.annotations[t]
          val builder = summonInline[Schema[R, t]].asInstanceOf[Schema[R, Any]]
          (label, annotations, builder, index) :: recurse[R, names, ts](index + 1)
        }
      case (_: EmptyTuple, _)                 => Nil
    }
```

Probably best though to have my PR and this one merged in as is and add that change later if we think it's good to have it

kyri-petrou (approved):
Awesome work with this PR! By the way I'm wondering whether due to these + the JSON implicits changes the next release should be marked as 3.x?

ghostdogpr:
> Awesome work with this PR! By the way I'm wondering whether due to these + the JSON implicits changes the next release should be marked as 3.x?

I was thinking to do 2.1 to keep the alignment with zio versions and keep it simple (otherwise it might get confusing when zio 3 comes out). I didn't really follow semantic versioning so far because Caliban is not a library that is depended upon by other libraries. But I'll have to make it very clear in release notes that there are some breaking changes. I also need to think about how I'll handle the docs. If I only keep the new way, people with older versions might be confused.

ghostdogpr:
> > Ah yeah, the exclusion happens much later currently.
> 
> 
> 
> Outside the scope of this PR, but I _think_ it should be possible to exclude them with something along the lines of (for Scala 3 at least): 
> 
> 
> 
> ```scala
> 
> // Macros.scala
> 
>   inline def isExcluded[T]: Boolean                         = ${ isExludedImpl[T] }
> 
>   def isExludedImpl[T: Type](using qctx: Quotes): Expr[Boolean] = {
> 
>     import qctx.reflect.*
> 
>     val tpe = TypeRepr.of[T]
> 
>     Expr(tpe.typeSymbol.annotations.exists { v =>
> 
>       v.tpe =:= TypeRepr.of[GQLExcluded]
> 
>     })
> 
>   }
> 
> 
> 
> // SchemaDerivation.scala
> 
>   inline def recurse[R, Label, A <: Tuple](index: Int = 0): List[(String, List[Any], Schema[R, Any], Int)] =
> 
>     inline erasedValue[(Label, A)] match {
> 
>       case (_: (name *: names), _: (t *: ts)) =>
> 
>         if Macros.isExluded[t] then recurse[R, names, ts](index + 1)
> 
>         else {
> 
>           val label = constValue[name].toString
> 
>           val annotations = Macros.annotations[t]
> 
>           val builder = summonInline[Schema[R, t]].asInstanceOf[Schema[R, Any]]
> 
>           (label, annotations, builder, index) :: recurse[R, names, ts](index + 1)
> 
>         }
> 
>       case (_: EmptyTuple, _)                 => Nil
> 
>     }
> 
> ```
> 
> 
> 
> Probably best though to have my PR and this one merged in as is and add that change later if we think it's good to have it

Yeah why not do that but separately 👍

kyri-petrou:
> I was thinking to do 2.1 to keep the alignment with zio versions and keep it simple (otherwise it might get confusing when zio 3 comes out). I didn't really follow semantic versioning so far because Caliban is not a library that is depended upon by other libraries. But I'll have to make it very clear in release notes that there are some breaking changes. I also need to think about how I'll handle the docs. If I only keep the new way, people with older versions might be confused.

Sounds good. Might be worth to define the [versioning scheme](https://www.scala-sbt.org/1.x/docs/Publishing.html#Version+scheme) to `versionScheme := Some("pvp")` in the `build.sbt` file, which indicates that a bump in both `x` and `y` (in `x.y.z`) are major changes

ghostdogpr:
Very good point!

build.sbt
```diff
@@ -549,7 +549,8 @@ lazy val commonSettings = Def.settings(
     case Some((3, _)) =>
       Seq(
         "-explain-types",
-        "-Ykind-projector"
+        "-Ykind-projector",
+        "-no-indent"
```
frekw, line 553: ❤️

core/src/main/scala/caliban/relay/Base64Cursor.scala
```diff
@@ -35,7 +33,4 @@ object Base64Cursor {
 
     def value(cursor: Base64Cursor): Int = cursor.value
   }
-
-  implicit val schema: Schema[Any, Base64Cursor] =
```
frekw, line 39: Isn't this needed anymore? I think we should provide a schema for this 🤔

frekw: Nevermind, just saw that you  moved it.

frekw (approved):
Fantastic work, can't wait to give this a try in some of our larger APIs. I'm suspecting it'll have a noticeable effect on compilation times! 🎉

kolemannix:
Ditto! I am thrilled to see this; thank you!

.circleci/config.yml
```diff
@@ -162,25 +163,25 @@ jobs:
...
-            cd ..
-            rm -rf docs
-            cp -r vuepress/docs/.vuepress/dist docs
-            git add docs
-            git config user.name "Pierre Ricadat"
-            git config user.email "ghostdogpr@gmail.com"
-            git commit -m "[ci skip] update docs"
-            git push
-          fi
+#  microsite:
```
ghostdogpr, line 166: I am disabling the docs publishing in the CI in order to merge this PR, I will re-enable it when we release.
That way we can update docs for now and it won't be reflected on the site yet.

PR #1596, opened 2023-02-02
Add parsing of introspection response JSON string

lassebn:
Based on the explorations of this [proof-of-concept repo](https://github.com/undo-insurance/internal-gateway-poc).

The goal is to be able to stitch a Sangria graph at the root level within the same codebase, without doing network calls.

In order to do that we need to build a Caliban `__Schema` from a `sangria.schema.Schema`

**Possible adaptations?**

1. Change the `IntrospectionClient` to take a function which produces the introspection JSON data structure as a string.
```scala
def introspect(produceIntrospectionJsonString: _ => String): UIO[Document]
```
2. Expose the introspection JSON payload parsing from the introspection client, so consumers can produce a `Document` on their own from an introspection JSON representation.
3. Additional ideas we haven't thought of 😀

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -60,6 +60,7 @@ jobs:
   scripted3_jdk:
     docker:
       - image: cimg/openjdk:11.0-node
+    resource_class: medium+
     steps:
       - checkout
       - restore_cache:
--- a/tools/src/main/scala/caliban/tools/IntrospectionClient.scala
+++ b/tools/src/main/scala/caliban/tools/IntrospectionClient.scala
@@ -19,12 +19,6 @@ import zio.URIO
 
 object IntrospectionClient {
 
-  def introspect(introspectionJson: String): UIO[Document] = {
-    val result: Either[CalibanClientError, Document] =
-      introspection.decode(introspectionJson).map { case (result, errors, extensions) => result }
-    ZIO.succeed(result).absolve.orDie
-  }
-
   def introspect(uri: String, headers: Option[List[Options.Header]]): RIO[SttpClient, Document] =
     for {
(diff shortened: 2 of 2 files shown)


ghostdogpr:
Or how about we just make `private val introspection` public so that users can do whatever they want with it?

lassebn:
@ghostdogpr That is a very minimal change that would indeed enable our use case. So if there is no need to keep `introspection` private, I am all up for that 👍 🙌 

I've updated the PR to make that change instead.

.circleci/config.yml
```diff
@@ -60,6 +60,7 @@ jobs:
   scripted3_jdk:
     docker:
       - image: cimg/openjdk:11.0-node
+    resource_class: medium+
```
lassebn, line 63: The CI step `scripted3_jdk` kept being killed and it seems like it is because of an out of memory exception. Memory usage was reaching 100%.

I've checked a couple of runs of `scripted3_jdk` on other branches, and they went to 98% memory usage, so it seems the runner was utilized to the max 💪 

Maybe something about this change causes more memory to be used, I'm not sure why though. Making the val public might force the compiler to hold more stuff in scope? 🤷 

The `medium+` resource class is included on free plans, but it burns
15 credits a minute over 10/minute for the medium class, [Pricing info here](https://circleci.com/product/features/resource-classes/). It has 6GB of ram, an increase of 2GB.

I don't know what plan the project is running on, and I hope this doesn't incur unwarranted cost.

ghostdogpr: Ah yeah, I fixed this in another PR that I haven't merged yet (I even went on using `large` instances). This is definitely not caused by your change. Thanks!

ghostdogpr (approved):

PR #1607, opened 2023-02-16
Optimize generated code size (Scala 3)

kyri-petrou:
As discussed [on Discord](https://discord.com/channels/629491597070827530/633200096393166868/1075604452641361991), the Scala 3 generated code is currently quite large, mostly due to boilerplate code being inlined for every derived type.

~In this PR, we use macros along with `inline if` so that we don't generate redundant code for branches we already know during compile-time that won't be called during runtime.~

In this PR, we move the creation of the `Schema[R, A]` object into a private method so that the code is not generated for each derived type

I tested these changes on a medium size application, and the application code size went down from 2.7MB to 1.3MB. The number of classes generated via Caliban also decreased from ~900 to ~300. This makes the Scala 3 generated code roughly the same size as the Scala 2 counterpart

TODO:

* [x] Optimise ArgBuilder
* [x] ~Figure out how to refactor the `recurse` and `isEnum` into a macro~ Rewrite `recurse` to be tail recursive (generated code becomes a flat list)


kyri-petrou:
I can't seem to figure out how to rewrite [this](https://github.com/ghostdogpr/caliban/pull/1607/files#diff-88ee5242eb50f55a96a2ae7a676d4b116c93fe6f9653dbe2ed8afa6193dbd16cL34-L42) into a macro due to it being recursive. I haven't worked with macros before though so I'm probably missing something super easy / obvious: Any pointers are much appreciated 😄

Observing the generated scala code (using the `-Xprint:erasure` compiler flag) this is causing quite a fair bit of code to be generated. If we manage to convert that into a macro, `isEnum` I think it'll cut down the size of the generated code by another 20-30%, which will bring it fairly close to the levels of Magnolia. Unlike magnolia, for most parts we're not just creating a super generic template method but rather get the compiler to strip out the parts we don't need, which I think might be better for runtime performance

ghostdogpr:
Ugh, I wanted to avoid macros as much as possible because I can't maintain them 😂

So the private def functions that don't have the inline keyword are inlined? I didn't expect that 🤔

kyri-petrou:
> Ugh, I wanted to avoid macros as much as possible because I can't maintain them 😂

Hah I'm totally the same 🥲. This is actually the first time I touched macros... ever. [The ones that I added](https://github.com/ghostdogpr/caliban/pull/1607/files#diff-9f8fb5fe145689da082577535c038e431584b0873871c7bf91ba12e97b4d75cdR67-R108) seem quite trivial, they're just a typecheck really. I'm happy to remove them and capitalise more on private methods (see more below) if you think they're going to make it more difficult maintaining the library

> So the private def functions that don't have the inline keyword are inlined? I didn't expect that 🤔

The private functions are not inlined no, that's why I moved [these](https://github.com/ghostdogpr/caliban/pull/1607/files#diff-88ee5242eb50f55a96a2ae7a676d4b116c93fe6f9653dbe2ed8afa6193dbd16cR209-R261) into private methods to avoid the `.map`, `.filter` and all that being generated for each schema (TIL that these methods generate quite a bit of bytecode). The main issue is with [this method](https://github.com/ghostdogpr/caliban/blob/series/2.x/core/src/main/scala-3/caliban/schema/SchemaDerivation.scala#L34-L42), which inlines a lot of code for nested types. If we could refactor to a macro, that would reduce the size of the generated code quite a bit (and improve compile time as well). Having said that, I think we should only do that if there's a simple-ish way to do it, otherwise adding too much complexity for the sake of performance is not worth it IMO

kyri-petrou:
Benchmarks for old vs new implementation. There is a small-ish improvement, most of it which seems to come from GC (which I guess is expected). I'll make the changes discussed on discord and rerun the tests

### Old

```
[info] GraphQLBenchmarks.fragmentsCaliban                                         thrpt    5        59.247 ±    1.307   ops/s
[info] GraphQLBenchmarks.fragmentsCaliban:·gc.alloc.rate                          thrpt    5      3498.826 ±   77.144  MB/sec
[info] GraphQLBenchmarks.fragmentsCaliban:·gc.alloc.rate.norm                     thrpt    5  61926024.267 ± 1015.661    B/op
[info] GraphQLBenchmarks.fragmentsCaliban:·gc.count                               thrpt    5       210.000             counts
[info] GraphQLBenchmarks.fragmentsCaliban:·gc.time                                thrpt    5       213.000                 ms

[info] GraphQLBenchmarks.introspectCaliban                                        thrpt    5      3175.157 ±   83.557   ops/s
[info] GraphQLBenchmarks.introspectCaliban:·gc.alloc.rate                         thrpt    5      3229.301 ±   84.998  MB/sec
[info] GraphQLBenchmarks.introspectCaliban:·gc.alloc.rate.norm                    thrpt    5   1066508.796 ±   66.851    B/op
[info] GraphQLBenchmarks.introspectCaliban:·gc.count                              thrpt    5       324.000             counts
[info] GraphQLBenchmarks.introspectCaliban:·gc.time                               thrpt    5       144.000                 ms

[info] execution.NestedZQueryBenchmark.deepParallelQuery1000                      thrpt    5       107.663 ±    1.895   ops/s
[info] execution.NestedZQueryBenchmark.deepParallelQuery1000:·gc.alloc.rate       thrpt    5      7016.640 ±  123.402  MB/sec
[info] execution.NestedZQueryBenchmark.deepParallelQuery1000:·gc.alloc.rate.norm  thrpt    5  68341118.284 ± 1784.521    B/op
[info] execution.NestedZQueryBenchmark.deepParallelQuery1000:·gc.count            thrpt    5       115.000             counts
[info] execution.NestedZQueryBenchmark.deepParallelQuery1000:·gc.time             thrpt    5       683.000                 ms
```

### New (with macros)

```
[info] GraphQLBenchmarks.fragmentsCaliban                                         thrpt    5        62.621 ±    1.534   ops/s
[info] GraphQLBenchmarks.fragmentsCaliban:·gc.alloc.rate                          thrpt    5      3536.092 ±   86.562  MB/sec
[info] GraphQLBenchmarks.fragmentsCaliban:·gc.alloc.rate.norm                     thrpt    5  59213344.178 ±  817.945    B/op
[info] GraphQLBenchmarks.fragmentsCaliban:·gc.count                               thrpt    5       186.000             counts
[info] GraphQLBenchmarks.fragmentsCaliban:·gc.time                                thrpt    5       195.000                 ms

[info] GraphQLBenchmarks.introspectCaliban                                        thrpt    5      3219.513 ±    8.513   ops/s
[info] GraphQLBenchmarks.introspectCaliban:·gc.alloc.rate                         thrpt    5      3166.384 ±    8.356  MB/sec
[info] GraphQLBenchmarks.introspectCaliban:·gc.alloc.rate.norm                    thrpt    5   1031309.281 ±   63.995    B/op
[info] GraphQLBenchmarks.introspectCaliban:·gc.count                              thrpt    5       259.000             counts
[info] GraphQLBenchmarks.introspectCaliban:·gc.time                               thrpt    5       125.000                 ms

[info] execution.NestedZQueryBenchmark.deepParallelQuery1000                      thrpt    5       109.888 ±    2.408   ops/s
[info] execution.NestedZQueryBenchmark.deepParallelQuery1000:·gc.alloc.rate       thrpt    5      6819.714 ±  149.341  MB/sec
[info] execution.NestedZQueryBenchmark.deepParallelQuery1000:·gc.alloc.rate.norm  thrpt    5  65078023.859 ± 1232.956    B/op
[info] execution.NestedZQueryBenchmark.deepParallelQuery1000:·gc.count            thrpt    5       124.000             counts
[info] execution.NestedZQueryBenchmark.deepParallelQuery1000:·gc.time             thrpt    5       578.000                 ms
```

kyri-petrou:
@ghostdogpr  Okay so it seems the benefits the macros are providing is marginal. I removed them and introduced `makeXXSchema` methods instead.

With these changes, the Scala 3 codegen is now about the same as Scala 2 :)

core/src/main/scala-3/caliban/schema/ArgBuilderDerivation.scala
```diff
@@ -11,61 +11,79 @@ import scala.deriving.Mirror
...
-                case Right(value)                     => acc.map(value *: _)
-              }
-            }.map(m.fromProduct)
+            buildProduct(fields, annotations)(input).map(m.fromProduct)
         }
     }
+  end derived
+
+  private def buildSum[A](
+    subTypes: List[(String, List[Any], ArgBuilder[Any])],
```
ghostdogpr: Should parameters be by-name since the vals are lazy?

core/src/main/scala-3/caliban/schema/ArgBuilderDerivation.scala
```diff
@@ -11,61 +11,79 @@ import scala.deriving.Mirror
...
-            }.foldRight[Either[ExecutionError, Tuple]](Right(EmptyTuple)) { case (item, acc) =>
-              item match {
-                case error: Left[ExecutionError, Any] => error.asInstanceOf[Left[ExecutionError, Tuple]]
-                case Right(value)                     => acc.map(value *: _)
-              }
-            }.map(m.fromProduct)
+            buildProduct(fields, annotations)(input).map(m.fromProduct)
         }
     }
+  end derived
```
ghostdogpr: I'm planning to add the no-indent option, I think this won't be accepted 😅 (yeah, I hate the significant indentation 😆)

ghostdogpr (commented):
Looks good!

core/src/main/scala-3/caliban/schema/ArgBuilderDerivation.scala
kyri-petrou: Removed 😅. On a side note I used to really hate it as well but it's slowly growing on me.

core/src/main/scala-3/caliban/schema/ArgBuilderDerivation.scala
kyri-petrou: Good point. Done!

ghostdogpr (approved):

PR #1614, opened 2023-02-19
Scala 3 schema derivation via `derives` keyword

kyri-petrou:
Okay this turned out to be a bit more involved / difficult than anticipated. Turns out there's no other way than to make `toType` public because of inlining. It was also tricky getting the fully automatic derivation to work, since if we do `import auto.*` it causes some issues with circular references. I managed to resolve this by extending `Auto[A]` with `LowPriorityDerivedSchema`

Also, I figured out a small "hack" to make it safe to have `toType` as public without risking accidental usage. I've annotated `toType` with `@deprecated`, which way we get compiler warnings / errors if used. Let me know if you prefer we didn't do that and I can remove the annotation


ghostdogpr (approved):
Awesome!

PR #1615, opened 2023-02-19
Don't require a schema for excluded fields (Scala 3 only)

kyri-petrou:
I'm putting this here before I forget about it. We don't need to merge it in now, it can be done after the semi-derivation PR is finalised and merged in.

With these changes, we no longer need to define a Schema for types that are exluded via the  `@GQLExcluded` annotation.

Unfortunately I couldn't figure out how to make this work for Scala 2 as well


ghostdogpr:
LGTM, let's get it merged after the derivation PR is merged.

PR #1616, opened 2023-02-20
feat: Add observability wrappers

frekw:
Hey!

This adds some wrappers for improving the observability of Caliban APIs.

There's a metrics middleware that both measures the duration of fields as well as the total execution count for a field, handy for e.g specifying SLOs per field (such as successsful fields resolving faster than 500ms / all fields). In order to get any sort of sane names for these metrics I had to add the parent type to `FieldInfo` (which can probably be useful in other situations as well).

It also adds `caliban-tracing` which contains integration with zio-telemetry. It's a bit finicky at the moment because it only works if your resolvers actually uses batching via DataSources, but @adamgfraser would look into adding a more general `QueryAspect` which would make it possible to do for all requests.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -184,19 +184,13 @@ lazy val tracing = project
   .settings(name := "caliban-tracing")
   .settings(commonSettings)
   .settings(
-    scalaVersion := scala213,
-    crossScalaVersions -= scala212, // zio-telemetry is only published for 2.13 and 3.x
-    buildInfoKeys    := Seq[BuildInfoKey](
-      "scalaPartialVersion" -> CrossVersion.partialVersion(scalaVersion.value),
-      "scalafmtVersion"     -> scalafmtVersion
-    ),
     buildInfoPackage := "caliban.tracing",
     buildInfoObject  := "BuildInfo"
   )
   .settings(
     testFrameworks := Seq(new TestFramework("zio.test.sbt.ZTestFramework")),
     libraryDependencies ++= Seq(
-      "dev.zio"         %% "zio-opentelemetry"         % "3.0.0-RC2",
+      "dev.zio"         %% "zio-opentelemetry"         % "3.0.0-RC3",
       "dev.zio"         %% "zio-test"                  % zioVersion % Test,
       "dev.zio"         %% "zio-test-sbt"              % zioVersion % Test,
       "io.opentelemetry" % "opentelemetry-sdk-testing" % "1.23.0"   % Test
--- a/build.sbt
+++ b/build.sbt
@@ -27,13 +27,8 @@ val zioInteropReactiveVersion = "2.0.1"
 val zioConfigVersion          = "3.0.7"
 val zqueryVersion             = "0.4.0"
 val zioJsonVersion            = "0.4.2"
-<<<<<<< HEAD
 val zioHttpVersion            = "0.0.4"
-val zioOpenTelemetryVersion   = "3.0.0-RC4"
-=======
-val zioHttpVersion            = "2.0.0-RC10"
 val zioOpenTelemetryVersion   = "3.0.0-RC5"
->>>>>>> 9d1ee721 (bump deps)
 
 inThisBuild(
   List(
--- a/build.sbt
+++ b/build.sbt
@@ -27,8 +27,13 @@ val zioInteropReactiveVersion = "2.0.1"
 val zioConfigVersion          = "3.0.7"
 val zqueryVersion             = "0.4.0"
 val zioJsonVersion            = "0.4.2"
+<<<<<<< HEAD
 val zioHttpVersion            = "0.0.4"
 val zioOpenTelemetryVersion   = "3.0.0-RC4"
+=======
+val zioHttpVersion            = "2.0.0-RC10"
+val zioOpenTelemetryVersion   = "3.0.0-RC5"
+>>>>>>> 9d1ee721 (bump deps)
 
 inThisBuild(
   List(
--- a/build.sbt
+++ b/build.sbt
@@ -178,6 +178,32 @@ lazy val tools = project
   )
   .dependsOn(core, clientJVM)
 
+lazy val tracing = project
+  .in(file("tracing"))
+  .enablePlugins(BuildInfoPlugin)
+  .settings(name := "caliban-tracing")
+  .settings(commonSettings)
+  .settings(
+    scalaVersion := scala213,
+    crossScalaVersions -= scala212, // zio-telemetry is only published for 2.13 and 3.x
+    buildInfoKeys    := Seq[BuildInfoKey](
+      "scalaPartialVersion" -> CrossVersion.partialVersion(scalaVersion.value),
+      "scalafmtVersion"     -> scalafmtVersion
+    ),
+    buildInfoPackage := "caliban.tracing",
+    buildInfoObject  := "BuildInfo"
+  )
+  .settings(
+    testFrameworks := Seq(new TestFramework("zio.test.sbt.ZTestFramework")),
+    libraryDependencies ++= Seq(
+      "dev.zio"         %% "zio-opentelemetry"         % "3.0.0-RC2",
+      "dev.zio"         %% "zio-test"                  % zioVersion % Test,
+      "dev.zio"         %% "zio-test-sbt"              % zioVersion % Test,
+      "io.opentelemetry" % "opentelemetry-sdk-testing" % "1.23.0"   % Test
+    )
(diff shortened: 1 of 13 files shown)


frekw:
The JDK8 failures is coming from Sttp using [`HttpTimeoutException`](https://docs.oracle.com/en/java/javase/12/docs/api/java.net.http/java/net/http/HttpTimeoutException.html) introduced in Java 11. Not sure what to make of that.

core/src/main/scala/caliban/wrappers/FieldMetrics.scala
```diff
@@ -0,0 +1,90 @@
...
+    extraLabels: Set[MetricLabel]
+  ): Wrapper.FieldWrapper[Any] =
+    new Wrapper.FieldWrapper[Any] {
+      override def wrap[R](
+        query: ZQuery[R, CalibanError.ExecutionError, ResponseValue],
+        info: FieldInfo
+      ): ZQuery[R, CalibanError.ExecutionError, ResponseValue] =
+        for {
+          summarized            <-
+            query
```
ghostdogpr, line 52: Is that the part that `ZQuery` changes will make better?
Shall we wait for it? I see there is an open PR

ghostdogpr: A maybe it's more about the `@@` on the data source. But same question anyway 😄

vuepress/docs/docs/middleware.md
```diff
@@ -225,3 +227,23 @@ to fields. In this case the field resolver will override the type cost, however
 We may also specify a "multipliers" argument when using arguments. This will match the argument names and use numeric argument values to multiply the base value.
 In the above case this means that a query that specifies a `limit` of `10` for the `spokenLines` field will have the base field cost multipled by 10 resulting in a total cost
 of execution of `1000`
+
+## OpenTelemetry Tracing
+
+Caliban ships with support for OpenTelemtry tracing via integration with [zio-telemetry](https://github.com/zio/zio-telemetry) via the `caliban-tracing` package.
```
ghostdogpr:
```suggestion
Caliban ships with support for OpenTelemetry tracing via integration with [zio-telemetry](https://github.com/zio/zio-telemetry) in the `caliban-tracing` package.
```

vuepress/docs/docs/middleware.md
```diff
@@ -225,3 +227,23 @@ to fields. In this case the field resolver will override the type cost, however
...
+```
+
+Then add it to your schema as any other wrapper:
+
+```scala
+import caliban.tracing.TracingWrapper
+import zio.telemetry.opentelemetry.Tracing
+
+val api: GraphQL[Any] = ???
+val tracedApi = api @@ TracingWrapper.traced
```
ghostdogpr, line 250: Maybe add more details like, for what it's going to create spans? It would be good to know what spans will look like before trying it yourself.

core/src/main/scala/caliban/wrappers/FieldMetrics.scala
frekw: No, I don't think so actually (but worth a try). The problem is that we're getting the execution time of all subsequent queries at the top level. Given `{ a (1s) { b (2s) { c (3s) } } }`, `a` will time in at 6 seconds. So what we do is keep track of the current path at each point and subtract the max execution time for our children, so e.g `c` will correctly get measured as `3`, `b` will get measured as `5 - 3 = 2` and `a` will be `6 - 5 = 1`. A bit messy / annoying but seems to work well!

The Apollo Tracing wrapper contains similar complexity due to the same reason.

frekw: The PR was just merged so I dug into it a bit. It fixes the tracing (:tada:) but it doesn't really make a difference here. Since we need the `ResponseValue`, the [wrapper will effectively wrap all underlying queries needed to produce the `ResponseValue`](https://github.com/ghostdogpr/caliban/blob/1f2232c80d3df055986a324990bc9567be93ef04/core/src/main/scala/caliban/execution/Executor.scala#L139) and not just the immediate step needed to resolve the given field.

I guess what we could do is to introduce another kind of wrapper that only wraps the execution of a `QueryStep`, but I'm not sure if you could do much else with it rather than use it for metrics and tracing (since its input would essentially be a `ReducedStep[R]`). WDYT @ghostdogpr ?

ghostdogpr: That's fine as it is 😄

build.sbt
```diff
@@ -25,7 +25,7 @@ val zioInteropCats2Version    = "22.0.0.0"
 val zioInteropCats3Version    = "23.0.0.1"
 val zioInteropReactiveVersion = "2.0.0"
 val zioConfigVersion          = "3.0.7"
-val zqueryVersion             = "0.3.6"
+val zqueryVersion             = "0.3.6+1-45b3ae66-SNAPSHOT"
```
ghostdogpr: I think there was a release so you can use the official one now 🎉

build.sbt
```diff
@@ -178,6 +178,26 @@ lazy val tools = project
...
+  .settings(name := "caliban-tracing")
+  .settings(commonSettings)
+  .settings(
+    buildInfoPackage := "caliban.tracing",
+    buildInfoObject  := "BuildInfo"
+  )
+  .settings(
+    testFrameworks := Seq(new TestFramework("zio.test.sbt.ZTestFramework")),
+    libraryDependencies ++= Seq(
+      "dev.zio"         %% "zio-opentelemetry"         % "3.0.0-RC3",
```
ghostdogpr: There was a RC4 published a few days ago

ghostdogpr: Can you move the version to a variable?

tracing/src/main/scala/caliban/tracing/TracingWrapper.scala
```diff
@@ -0,0 +1,5 @@
+package caliban.tracing
+
+object TracingWrapper {
+  val traced = SchemaTracer.wrapper @@ FieldTracer.wrapper
```
ghostdogpr, line 4: 🔥

ghostdogpr (commented):
Looks good? Just some minor remarks

build.sbt
frekw: Done! 👍

build.sbt
frekw: Seems like it failed to get published, I'll ping Adam on discord!

build.sbt
frekw: RC4 seems to be missing from Sonatype as well :(

build.sbt
frekw: Bumped 👍

core/src/main/scala/caliban/wrappers/Wrappers.scala
```diff
@@ -44,6 +46,17 @@ object Wrappers {
...
+   * Returns a wrapper that logs slow queries
+   * @param duration threshold above which queries are considered slow
+   */
+  def logSlowQueries(duration: Duration): OverallWrapper[Any] =
+    onSlowQueries(duration) { case (time, query) =>
+      ZIO.logAnnotate("query", query) {
+        ZIO.logWarning(s"Slow query took ${time.render}:\n$query")
+      }
+    }
+
```
kyri-petrou: ❤️

core/src/main/scala/caliban/wrappers/FieldMetrics.scala
```diff
@@ -0,0 +1,90 @@
...
+                                     popped      <- ref.modify { state =>
+                                                      val (popped, rest) = state.partition { case (path, _) =>
+                                                        path.startsWith(currentPath)
+                                                      }
+                                                      (popped, rest)
+                                                    }
+                                     offset       = if (popped.isEmpty) ("", 0L)
+                                                    else popped.maxBy { case (_, duration) => duration }
+                                     duration     = end - start - offset._2
+                                     _           <- ref.update(current => (currentPath, duration + offset._2) :: current)
```
paulpdaniels, line 67: Why split this up from the `modify`? Is it to reduce the duration of exclusivity?

frekw: Yup, exactly. Probably too small a thing to matter but I usually try to scope my ref updates to be as small as possible 😅

ghostdogpr (approved):

ghostdogpr:
@frekw you can merge if it's ready 👍

PR #1617, opened 2023-02-20
initial repeatable directives support

jgulotta:
Part of #1159

This parses and renders repeatable directives, and skips them in uniqueness checks

I could not tell if any changes were necessary to actually use repeatable directives, but this was enough for our use case to generate a client from a schema that had repeatable directives in it. If such changes are needed I could use some guidance on what those are

When accounting for repeatability in uniqueness checks, I checked repeatability from the definition of the directive, but I made an assumption that directives were defined only once and that all referenced directives have a definition, further assuming that was validated elsewhere. If this is inaccurate I could also use some guidance on what I missed and/or how you'd like to address it

I wanted to add a test that showed "repeated repeatable directives do not error" but the `ValidationSpec` only has failure tests and I didn't notice any similar happy-path tests where I should place it. If you'd like to see the test I'm willing to add such a test wherever you prefer

There are a few places in the spec that mention non-repeatable directives in validation for extensions, but I didn't find any existing validation for extensions to alter. If I missed them then point me in their direction and I'll try to filter out repeatable directives there too

### DIFF
--- a/core/src/main/scala/caliban/validation/Validator.scala
+++ b/core/src/main/scala/caliban/validation/Validator.scala
@@ -9,6 +9,7 @@ import caliban.introspection.adt._
 import caliban.introspection.adt.__TypeKind._
 import caliban.parsing.SourceMapper
 import caliban.parsing.adt.Definition.ExecutableDefinition.{ FragmentDefinition, OperationDefinition }
+import caliban.parsing.adt.Definition.TypeSystemDefinition.DirectiveDefinition
 import caliban.parsing.adt.Definition.{ TypeSystemDefinition, TypeSystemExtension }
 import caliban.parsing.adt.OperationType._
 import caliban.parsing.adt.Selection.{ Field, FragmentSpread, InlineFragment }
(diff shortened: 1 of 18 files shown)


jgulotta:
Haven't had any luck figuring out why CI is failing - those exact commands succeed for me. It feels like there is an incremental or non-deterministic compilation issue but I don't have anything concrete to say what it might be

ghostdogpr:
Hmm yeah the CI error is super weird, I don't see what in your code might trigger it. But this never happens with the existing code so there must be something 🤔

jgulotta:
It seems to happen due to the extra lines in the `RendererSpec`, like the new lines with quotes in the raw string are somehow throwing off the parsing, sometimes.

There is some interaction with incremental compilation. If you have a successful compile, and then add those lines to the string, it will compile and run correctly. If you clean after the successful compile, it will fail again.

I think there might be some interaction with macros too. If I move the string to a field in the class, it seems to compile reliably, so this is what I've updated to.

I am sensing a scalac bug, though I couldn't find an existing one that seemed related.

core/src/main/scala/caliban/validation/Validator.scala
```diff
@@ -212,23 +213,25 @@ object Validator {
...
+                                checkDirectivesUniqueness(fragment.directives, directiveDefinitions)
+                                  .as(fragment.directives.map((_, __DirectiveLocation.FRAGMENT_DEFINITION)))
+                              )
+      selectionDirectives  <- collectDirectives(context.selectionSets, directiveDefinitions)
     } yield opDirectives.flatten ++ fragmentDirectives.flatten ++ selectionDirectives
 
   private def collectDirectives(
-    selectionSet: List[Selection]
+    selectionSet: List[Selection],
+    directiveDefintions: Map[String, List[DirectiveDefinition]]
```
ghostdogpr: typo: `directiveDefintions`

core/src/main/scala/caliban/validation/Validator.scala
```diff
@@ -248,16 +251,26 @@ object Validator {
...
-    directives.groupBy(_.name).find { case (_, v) => v.length > 1 } match {
+  private def checkDirectivesUniqueness(
+    directives: List[Directive],
+    directiveDefinitions: Map[String, List[DirectiveDefinition]]
+  ): IO[ValidationError, Unit] =
+    directives
+      .groupBy(_.name)
+      .find { case (n, v) =>
+        // only non-repeatable directives count against uniqueness
+        !directiveDefinitions.get(n).exists(_.exists(_.repeatable)) & (v.length > 1)
```
ghostdogpr: Shouldn't it be `&&`?

ghostdogpr (commented):
Looks nice!

@paulpdaniels do you want to give it a look? I know you make advanced usage of directives

core/src/main/scala/caliban/validation/Validator.scala
jgulotta: Fixed!

core/src/main/scala/caliban/validation/Validator.scala
jgulotta: I can make it `&&` if you like, There's no data dependency between the two nor much computation to be saved with a short circuit, and in such cases I have a habit of using `&`. I don't think it'll make a noticeable difference here either way but `&` can be friendlier to CPU instruction pipelines.

I suppose the main argument for `&&` is that `v` is a `List` and so `v.length` is not a constant time call, but my preferred solution for that is to not use `List`

federation/src/main/scala/caliban/federation/FederationV1.scala
```diff
@@ -10,17 +10,37 @@ class FederationV1
...
         ),
-        __Directive("requires", None, locations = Set(__DirectiveLocation.FIELD_DEFINITION), args = _FieldSet :: Nil),
-        __Directive("provides", None, locations = Set(__DirectiveLocation.FIELD_DEFINITION), args = _FieldSet :: Nil),
         __Directive(
           "key",
           None,
           locations = Set(__DirectiveLocation.OBJECT, __DirectiveLocation.INTERFACE),
-          args = _FieldSet :: Nil
+          args = _FieldSet :: Nil,
+          repeatable = false
```
paulpdaniels: This is actually repeatable: https://www.apollographql.com/docs/federation/subgraph-spec/#subgraph-schema-additions

core/src/main/scala/caliban/validation/Validator.scala
paulpdaniels: I forgot when it was added but wouldn't it be more performant to run the length check first using `lengthIs`?

```scala
(v.lengthIs > 1) & !directiveDefinitions.get(n).exists(_.exists(_.repeatable))
```

That should be effectively constant time since it only ever has to check if there are at least two items in the list. That also saves on having to do the `repeatability` check against all directives

paulpdaniels (approved):

core/src/main/scala/caliban/validation/Validator.scala
jgulotta: I would have changed it to `lengthIs` but that's not available in 2.12. I could use `v.lengthCompare(1) > 0`, which is a less clear version of `lengthIs > 1`

I would expect no performance difference when using `&` since it does not short circuit and both sides get evaluated

Using `lengthCompare` first coupled with `&&` would likely have the best behavior in pathological cases where many directives are repeated many times. In more usual cases I expect these lists would have very few elements and you'd probably need a microbenchmark to tell them apart

I'd probably choose `lengthCompare` with `&&` myself, but I'll go with whatever you want to read and maintain

federation/src/main/scala/caliban/federation/FederationV1.scala
jgulotta: Made `key` repeatable

core/src/main/scala/caliban/validation/Validator.scala
ghostdogpr: How about simply `(v.lengthIs > 1) && !directiveDefinitions.get(n).exists(_.exists(_.repeatable))` 😄 
(we're nitpicking, it's not gonna make much of a different anyway)

jgulotta: I can't do `lengthIs` because it's not in 2.12, which is part of the cross build. So I did the `lengthCompare` version with a note about why it's there

ghostdogpr (approved):
Thanks!

ghostdogpr:
Hmm the CI issue is back, weird 🤔

PR #1620, opened 2023-02-23
Let requestInterceptor eliminate part of the environment

lassebn:
This is a rough showcase for how `requestInterceptor` could be changed to eliminate the part of the environment `R` of GraphQLInterpreter[R, E] which should be resolved at runtime. E.g. for auth tokens.

This allows `HttpAdaptor.makeHttpService` (as well as the upload adaptor and websocket adaptor) to only require the environment needed to build the route. Currently it is suggested to provide dummy values in a `FiberRef` which the interceptors can then override at runtime.

As I was playing around with some APIs, I found it awkward that the types state that the request interceptors would provide some of the environment, but I still needed to provide the full environment when creating the route.

I've commented out a lot of files to keep this proof-of-concept slim. It is relevant to look at

* `adapters/akka-http/src/main/scala/caliban/AkkaHttpAdapter.scala`
* `examples/src/main/scala/example/akkahttp/AuthExampleApp.scala`
* `interop/tapir/src/main/scala/caliban/interop/tapir/RequestInterceptor.scala`
* `interop/tapir/src/main/scala/caliban/interop/tapir/TapirAdapter.scala`

Would you be interested in accepting a change like this? And do let me know if my motivation for this isn't clear enough 😊

### DIFF
--- a/examples/src/main/scala/example/akkahttp/AuthExampleApp.scala
+++ b/examples/src/main/scala/example/akkahttp/AuthExampleApp.scala
@@ -32,7 +32,7 @@ object AuthExampleApp extends App {
         case header if header.is("token") => header.value
       } match {
         case Some(token) =>
-          effect.provideSomeLayer[R](ZLayer.succeed[Any](AuthToken(token)))
+          effect.provideSomeLayer[R](ZLayer.succeed[Auth](AuthToken(token)))
         case _           => ZIO.fail(TapirResponse(StatusCode.Forbidden))
       }
   }
@@ -54,7 +54,7 @@ object AuthExampleApp extends App {
                         .orElse(request.connectionInfo.remote.map(_.getAddress.getHostAddress))
                         .map(IP)
                     }.orDie
-        clientIP <- effect.provideSomeLayer[R](ZLayer.succeed[Any](ip))
+        clientIP <- effect.provideSomeLayer[R](ZLayer.succeed[ClientIP](ip))
       } yield clientIP
   }
 
@@ -63,8 +63,8 @@ object AuthExampleApp extends App {
   case class Query(token: RIO[Auth, String], ip: RIO[ClientIP, String])
   private val resolver                          = RootResolver(
     Query(
-      token = ZIO.serviceWithZIO[Auth](_.get).map(_.value),
-      ip = ZIO.serviceWithZIO[ClientIP](_.get).map(_.map(_.value).getOrElse("no ip"))
+      token = ZIO.serviceWith[Auth](_.value),
+      ip = ZIO.serviceWith[ClientIP](_.map(_.value).getOrElse("no ip"))
     )
   )
   private val api                               = graphQL(resolver)
--- a/examples/src/main/scala/example/akkahttp/AuthExampleApp.scala
+++ b/examples/src/main/scala/example/akkahttp/AuthExampleApp.scala
@@ -22,7 +22,7 @@ object AuthExampleApp extends App {
 
   case class AuthToken(value: String)
 
-  type Auth = FiberRef[Option[AuthToken]]
+  type Auth = FiberRef[AuthToken]
 
   object AuthInterceptor extends RequestInterceptor[Any, Auth] {
     override def apply[R, A](
@@ -32,7 +32,7 @@ object AuthExampleApp extends App {
         case header if header.is("token") => header.value
       } match {
         case Some(token) =>
-          effect.provideSomeLayer[R](ZLayer.scoped[Any](FiberRef.make(Option.apply(AuthToken(token)))))
+          effect.provideSomeLayer[R](ZLayer.scoped[Any](FiberRef.make(AuthToken(token))))
         case _           => ZIO.fail(TapirResponse(StatusCode.Forbidden))
       }
   }
@@ -60,10 +60,10 @@ object AuthExampleApp extends App {
 
   val schema: GenericSchema[Auth with ClientIP] = new GenericSchema[Auth with ClientIP] {}
   import schema._
-  case class Query(token: RIO[Auth, Option[String]], ip: RIO[ClientIP, String])
+  case class Query(token: RIO[Auth, String], ip: RIO[ClientIP, String])
   private val resolver                          = RootResolver(
     Query(
-      token = ZIO.serviceWithZIO[Auth](_.get).map(_.map(_.value)),
+      token = ZIO.serviceWithZIO[Auth](_.get).map(_.value),
       ip = ZIO.serviceWithZIO[ClientIP](_.get).map(_.map(_.value).getOrElse("no ip"))
     )
   )
--- a/examples/src/main/scala/example/akkahttp/AuthExampleApp.scala
+++ b/examples/src/main/scala/example/akkahttp/AuthExampleApp.scala
@@ -15,6 +15,8 @@ import zio.{ FiberRef, RIO, Runtime, Unsafe, ZIO, ZLayer }
 
 import scala.concurrent.ExecutionContextExecutor
 import scala.io.StdIn
+import zio.Scope
+import _root_.cats.effect.Fiber
 
 object AuthExampleApp extends App {
 
@@ -22,14 +24,15 @@ object AuthExampleApp extends App {
 
   type Auth = FiberRef[Option[AuthToken]]
 
-  object AuthInterceptor extends RequestInterceptor[Auth] {
-    override def apply[R <: Auth, A](
+  object AuthInterceptor extends RequestInterceptor[Any, Auth] {
+    override def apply[R, A](
       request: ServerRequest
-    )(effect: ZIO[R, TapirResponse, A]): ZIO[R, TapirResponse, A] =
+    )(effect: ZIO[R with Auth, TapirResponse, A]): ZIO[R, TapirResponse, A] =
       request.headers.collectFirst {
         case header if header.is("token") => header.value
       } match {
-        case Some(token) => ZIO.serviceWithZIO[Auth](_.set(Some(AuthToken(token)))) *> effect
+        case Some(token) =>
+          effect.provideSomeLayer[R](ZLayer.scoped[Any](FiberRef.make(Option.apply(AuthToken(token)))))
         case _           => ZIO.fail(TapirResponse(StatusCode.Forbidden))
       }
   }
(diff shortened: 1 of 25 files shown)


interop/tapir/src/main/scala/caliban/interop/tapir/RequestInterceptor.scala
```diff
@@ -8,17 +8,20 @@ import zio.ZIO
...
+      ): ZIO[R2, TapirResponse, A] =
+        that.apply[R2, A](request)(self.apply[R2 with DynR2, A](request)(e))
+    }
 }
 
 object RequestInterceptor {
-  val empty: RequestInterceptor[Any] = new RequestInterceptor[Any] {
-    override def apply[R, A](request: ServerRequest)(e: ZIO[R, TapirResponse, A]): ZIO[R, TapirResponse, A] = e
+  val empty: RequestInterceptor[Any, Any] = new RequestInterceptor[Any, Any] {
+    override def apply[R, A](request: ServerRequest)(e: ZIO[R with Any, TapirResponse, A]): ZIO[R, TapirResponse, A] = e
```
ghostdogpr, line 25: Can remove `with Any`

examples/src/main/scala/example/akkahttp/AuthExampleApp.scala
```diff
@@ -11,53 +11,83 @@ import caliban.{ AkkaHttpAdapter, RootResolver }
 import sttp.model.StatusCode
 import sttp.tapir.json.circe._
 import sttp.tapir.model.ServerRequest
-import zio.{ FiberRef, RIO, Runtime, Unsafe, ZIO, ZLayer }
+import zio.{ RIO, Runtime, Unsafe, ZIO, ZLayer }
 
 import scala.concurrent.ExecutionContextExecutor
 import scala.io.StdIn
+import zio.Scope
```
ghostdogpr: (minor) Imports need to be sorted

examples/src/main/scala/example/akkahttp/AuthExampleApp.scala
```diff
@@ -11,53 +11,83 @@ import caliban.{ AkkaHttpAdapter, RootResolver }
...
+    override def apply[R, A](
       request: ServerRequest
-    )(effect: ZIO[R, TapirResponse, A]): ZIO[R, TapirResponse, A] =
+    )(effect: ZIO[R with Auth, TapirResponse, A]): ZIO[R, TapirResponse, A] =
       request.headers.collectFirst {
         case header if header.is("token") => header.value
       } match {
-        case Some(token) => ZIO.serviceWithZIO[Auth](_.set(Some(AuthToken(token)))) *> effect
+        case Some(token) =>
+          effect.provideSomeLayer[R](ZLayer.succeed[Auth](AuthToken(token)))
```
ghostdogpr, line 33: I'm wondering if performance is not going to get worse with that change. `provide` in general has some cost because it's merging environments together. I know recently zio-grpc completely removed the usage of environment to pass context for this reason (which requires many heavy changes for me at work 😭 ).

On the other hand it does look simpler to use since `FiberRef` is no longer necessary. I'm interested in other opinions @paulpdaniels @frekw @kyri-petrou ?

interop/tapir/src/main/scala/caliban/interop/tapir/RequestInterceptor.scala
lassebn: If I remove `with Any` I get this compile error 🤔 
```scala
object creation impossible, since method apply in trait RequestInterceptor of type [R1, A](request: sttp.tapir.model.ServerRequest)(e: zio.ZIO[R1 with Any,caliban.interop.tapir.TapirAdapter.TapirResponse,A])zio.ZIO[R1,caliban.interop.tapir.TapirAdapter.TapirResponse,A] is not defined
  val empty: RequestInterceptor[Any, Any] = new RequestInterceptor[Any, Any] {
```

examples/src/main/scala/example/akkahttp/AuthExampleApp.scala
lassebn: Is there a benchmark I can run to see what the impact is? I guess that should actually take the number of interceptors into account 🤔 If needed I can set up a benchmark to get some actual numbers on the impact.

interop/tapir/src/main/scala/caliban/interop/tapir/RequestInterceptor.scala
kyri-petrou: Out of curiosity, which Scala versions give this compilation error? I'm guessing 2.12 for sure but just curious if 2.13 does has the same issue.

Also, does changing the context bound to `[R <: Any, A]` solve this?

lassebn: 2.12 as well as 2.13, and changing the context bound like you suggest sadly doesn't solve it.

examples/src/main/scala/example/akkahttp/AuthExampleApp.scala
kyri-petrou: I _think_ that this change is okay because the number of interceptors is likely going to be very small (most cases 0 or 1), so it'll be called just once per request. Also, users looking to maximise performance are likely going to implement this without merging context anyways. Since this is just the example, we could add a comment to indicate that users wishing to maximise performance could implement this via a `FiberRef` instead. Or is it no longer possible to implement this using a `FiberRef` given the changes in this PR?

interop/tapir/src/main/scala/caliban/interop/tapir/RequestInterceptor.scala
kyri-petrou: Might be a bug in `scalac` (I'm guessing Scala 3 allows this right?) or that some magical import / implicit / compiler flag is required 😓. I haven't used Scala 2 in a while so I'm not sure what the solution is here; although if using `R with Any` gets it to work then I'm guessing that's fine?

examples/src/main/scala/example/akkahttp/AuthExampleApp.scala
lassebn: I just added commit https://github.com/ghostdogpr/caliban/pull/1620/commits/15227758479e714f62bde79e50aec52f1bf97c16 which brings back a "performance" example using the original `FiberRef` approach.

The type signature for the interceptor becomes `RequestInterceptor[Auth, Auth]` though, which is less nice, since it doesn't eliminate `Auth` so the second type parameter is "lying" now.

To keep things "clean" I think we're actually looking at adding 3 interceptors at work. But maybe the right solution is to only have one interceptor, if the runtime overhead is too big.

interop/tapir/src/main/scala/caliban/interop/tapir/RequestInterceptor.scala
lassebn: I'm actually unsure how I get to compile the interop module with Scala 3. Setting the whole build to Scala 3 fails because `sbt-scripted` isn't available.

examples/src/main/scala/example/akkahttp/AuthExampleApp.scala
frekw: I don't have any performance number, but I think using `ZLayer.apply` in ZIO2 is more efficient, and probably what we want in most cases (eliminating a single dep from `R`):

```scala
ZLayer.succeed(AuthToken(token))(effect)
```

kyri-petrou:
I'm a bit divided about these changes. On one hand, I do really see a lot of potential for advanced users. On the other hand, I'm worried about the added complexity / breaking changes this brings to existing and first-time users; especially `cats-effect` and `akka` users who are not familiar with ZIO environments. To a ZIO user it might seem trivial to set `R` and `IR` to the same type if you don't plan to eliminate anything from the environment, _but_ to a user that hasn't used ZIO before it's not as straightforward as one would think (talking from experience since my first interactions with ZIO were through Caliban). This becomes even more unclear if someone is not planning to use a `RequestInterceptor` at all but the `makeHttpService` method require 2 type arguments for the environment. 

This is more "food for thought" rather than a concrete recommendation; I'm wondering whether it's possible and/or desirable to keep the existing API as is via inheritance and overloaded methods where `R =:= IR`. e.g.,

```scala
trait RequestInterceptor2[-R, +DynR2] {
 ...
}

trait RequestInterceptor[R] extends RequestInterceptor2[R, R]
```

and then we overload methods to accept either `RequestInterceptor2` or `RequestInterceptor`. I haven't tested how any of this looks so it might not be possible, just thought to bring it up just in case

examples/src/main/scala/example/akkahttp/AuthExampleApp.scala
frekw: I am supportive of this change since it's also more inline with how `R` is described for ZIO2:

> The most idiomatic use of the ZIO environment is for:
> - Describing as values workflows that use capabilities that are "local" to a particular context
> - Where tracking the use of this context at the type level is helpful for reasoning about programs

(from https://zio.dev/reference/contextual/zio-environment-use-cases#1-local-contexts)

interop/tapir/src/main/scala/caliban/interop/tapir/RequestInterceptor.scala
kyri-petrou: I usually just run this https://github.com/ghostdogpr/caliban/blob/series/2.x/.circleci/config.yml#L97 in the terminal 😅

lassebn: Ah, smart 😄 It does compile on Scala 3 without `with Any` 🤷‍♂️ I guess this could be acceptable. I can add a comment on it not being needed for Scala 3.

examples/src/main/scala/example/akkahttp/AuthExampleApp.scala
ghostdogpr: Yeah I think it should be okay performance-wise, and it's quite simpler than with `FiberRef` actually. I think you can remove the `FiberRef` example, if we go for this solution it should be the recommended way.

The only concern I have is complexity as @kyri-petrou mentioned (and also things like having to write `with Any`).

interop/tapir/src/main/scala/caliban/interop/tapir/RequestInterceptor.scala
ghostdogpr: Hmm that's weird, I never had this issue for example with ZIO environment. It could be quite confusing for users if they have to do this to make it work 🤔

lassebn:
Todo for self:

- [ ] Add the overloaded methods with `R =:= IR`
- [ ] Check how usage looks in a Scala 3 project, and confirm we don't need `with Any`
- [ ] See if we can use `.apply` directly on the layer

ghostdogpr:
Closing in favor of https://github.com/ghostdogpr/caliban/pull/1707

Thanks for the work, it inspired the current solution!

lassebn:
I'm glad you could use it for inspiration. I like what you're doing with the refactors in #1707  🙌

PR #1629, opened 2023-02-25
Bypass validation when a field is null but a default value exists

ghostdogpr:
Solves this problem https://scastie.scala-lang.org/CecZcqy9RPyPdUtWAtHOTA

### DIFF
--- a/core/src/main/scala/caliban/validation/ValueValidator.scala
+++ b/core/src/main/scala/caliban/validation/ValueValidator.scala
@@ -73,10 +73,14 @@ object ValueValidator {
             argValue match {
               case ObjectValue(fields) =>
                 ZIO.foreachDiscard(inputType.inputFields.getOrElse(List.empty)) { f =>
-                  val value =
-                    fields.collectFirst { case (name, fieldValue) if name == f.name => fieldValue }
-                      .getOrElse(NullValue)
-                  validateType(f.`type`(), value, context, s"Field ${f.name} in $errorContext")
+                  fields.collectFirst { case (name, fieldValue) if name == f.name => fieldValue } match {
+                    case Some(value) =>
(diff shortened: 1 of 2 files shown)


paulpdaniels (approved):

kyri-petrou (approved):

frekw (approved):
LGTM!

valentin-panalyt:
Just for reference, this fixes the behaviour that you can see here: https://scastie.scala-lang.org/CecZcqy9RPyPdUtWAtHOTA

In case scastie doesn't work:
```
// import zio
import zio._
import zio.ZIOAppDefault
import caliban.schema.Annotations._

object MyApp {
  def run = {
    case class Foo(inner: Boolean)
    case class Outer(
        bar: String,
        // If not specified, {inner: true} should be used as default. 
        @GQLDefault("""{inner: true}""") foo: Foo
    )

    case class GetFooArgs(outer: Outer)

    case class Queries(example: GetFooArgs => UIO[String])

    def resolveExample(args: GetFooArgs) = ZIO.succeed("Done")

    val queries = Queries(example = args => resolveExample(args))

    import caliban.GraphQL.graphQL
    import caliban.RootResolver

    val api = graphQL(RootResolver(queries))

    val query = """
  {
    example(outer: {
       bar: "bar"
       # foo: {inner: false} # Uncomment this to make it work
       }
    )
  }"""

    for {
      _ <- ZIO.succeed(println("start"))
      result <- interpreter.execute(query)
      _ <- zio.ZIO.debug(result.data.toString)
      _ <- zio.ZIO.debug(result.errors.toString)
      _ <- ZIO.succeed(println("done"))
    } yield ()
  }
}

zio.Unsafe.unsafe(implicit us =>
  scala.concurrent.Await.result(
    zio.Runtime.default.unsafe.runToFuture(MyApp.run),
    scala.concurrent.duration.Duration.Inf
  )
)
```

PR #1630, opened 2023-02-26
Don't require a schema for excluded fields (Scala 3 only)

kyri-petrou:
Reopening #1615 targeting the main branch.

The changes in this PR allow us to skip the creation of a Schema for types that are annotated with `@GQLExcluded` in Scala 3.

```scala
final case class Foo(value: String)
final case class Bar(value: String, @GQLExcluded foo: Foo, num: Int) derives Schema.SemiAuto
```

Ideally, we'd also like this behaviour to be consistent between Scala 2 and 3, but I couldn't figure out how to implement this using Magnolia


ghostdogpr (approved):
👍

PR #1631, opened 2023-02-26
Add support for federation v2.3

### LINKED ISSUE #1611: [federation] support new `@composeDirective`
Apollo Federation v2.1 introduced new `@composeDirective` directive that allows users to specify directives that should be preserved in the supergraph composition (by default composition strips out most directives from supergraph).

```graphql
extend schema
  @link(
    url: "https://specs.apollo.dev/federation/v2.3",
    import: ["@composeDirective", "@extends", "@external", "@key", "@inaccessible", "@interfaceObject", "@override", "@provides", "@requires", "@shareable", "@tag"]
  )
  @link(url: "https://myspecs.dev/myCustomDirective/v1.0", import: ["@custom"])
  @composeDirective(name: "@custom")

// will be present in the supergraph definition
directive @custom on OBJECT

type Product @custom @key(fields: "id")
  id: ID!
  // other fields omitted for clarity
}
```

Additional resources:
* [`@composeDirective` directive definition](https://www.apollographql.com/docs/federation/federated-types/federated-directives#composedirective)

---
New directive functionality can be tested using [Apollo Federation Subgraph Compatibility](https://www.npmjs.com/package/@apollo/federation-subgraph-compatibility) NPX script (and [Github Action](https://github.com/marketplace/actions/apollo-federation-subgraph-compatibility)). Example integration project is already provided in the [subgraph compatibility testing repository](https://github.com/apollographql/apollo-federation-subgraph-compatibility/tree/main/implementations/caliban).

### LINKED ISSUE #1612: [federation] support new `@interfaceObject` directive
Apollo Federation v2.3 introduced new `@interfaceObject` directive that allows users to extend entity functionality through inheritance, i.e. given subgraph A

```graphql
interface Product {
 id: ID!
 title: String
}

# Book entity with key
type Book implements Product @key(fields: "id") {
 id: ID!
 title: String
 pages: Int
}

# Movie entity with key
type Movie implements Product @key(fields: "id") {
 id: ID!
 title: String
 duration: Int
}
```

We can generically extend the Product interface in other subgraphs by treating it as local object type with `@interfaceObject` directive. This new directive informs composition logic that it is actually an entity interface. This allows us to add new functionality in subgraph B without knowing any existing implementation details (i.e. without knowing anything about Book and Movie types).

```graphql
type Product @key(fields: "id") @interfaceObject {
 id: ID!
 reviews: [Review!]!
}

type Review {
 author: String
 text: String
 rating: Int
}
```

Additional resources:
* [@interfaceObject directive definition](https://www.apollographql.com/docs/federation/federated-types/federated-directives#interfaceobject)
* [Entity Interface documentation](https://www.apollographql.com/docs/federation/federated-types/interfaces)
* [Announcement blog post](https://www.apollographql.com/blog/announcement/platform/apollo-federation-and-graphos-add-support-for-entity-interfaces-to-streamline-collaboration/)

---
New directive functionality can be tested using [Apollo Federation Subgraph Compatibility](https://www.npmjs.com/package/@apollo/federation-subgraph-compatibility) NPX script (and [Github Action](https://github.com/marketplace/actions/apollo-federation-subgraph-compatibility)). Example integration project is already provided in the [subgraph compatibility testing repository](https://github.com/apollographql/apollo-federation-subgraph-compatibility/tree/main/implementations/caliban).

paulpdaniels:
Closes #1611 , Closes #1612 

Adds support for federation v2.1 and v2.3 (2.2 didn't seem to have any user facing changes). Since caliban 2.1 will likely be a "breaking" this includes some changes to make the package structure slightly more intuitive. Now users will import the `caliban.federation.v<major>[_<minor>]` package that reflects the support they need. If they need to make use of the new `composeDirective` mechanism they can also create a new instance of `FederationV2` that includes support for their new directives.

### DIFF
--- a/federation/src/main/scala/caliban/federation/package.scala
+++ b/federation/src/main/scala/caliban/federation/package.scala
@@ -1,11 +1,13 @@
 package caliban
 
-import caliban.federation.v2.{ FederationDirectivesV2_3, FederationV2, Versions }
+import caliban.federation.v2x.{ FederationDirectivesV2_3, FederationV2, Versions }
 
 package object federation {
 
   lazy val v1   = new FederationV1 with FederationDirectives
   lazy val v2_0 = new FederationV2(List(Versions.v2_0))
+  @deprecated("Use the new differentiated package structure instead (e.g. v2_3) instead", "2.1.0")
+  lazy val v2 = v2_0
   lazy val v2_1 = new FederationV2(List(Versions.v2_1))
   lazy val v2_3 = new FederationV2(List(Versions.v2_3)) with FederationDirectivesV2_3
 
--- a/federation/src/main/scala/caliban/federation/package.scala
+++ b/federation/src/main/scala/caliban/federation/package.scala
@@ -1,3 +1,12 @@
 package caliban
 
-package object federation extends FederationV1 with FederationDirectives
+import caliban.federation.v2.{ FederationDirectivesV2_3, FederationV2, Versions }
+
+package object federation {
+
+  lazy val v1   = new FederationV1 with FederationDirectives
+  lazy val v2_0 = new FederationV2(List(Versions.v2_0))
+  lazy val v2_1 = new FederationV2(List(Versions.v2_1))
+  lazy val v2_3 = new FederationV2(List(Versions.v2_3)) with FederationDirectivesV2_3
+
+}
--- a/vuepress/docs/docs/federation.md
+++ b/vuepress/docs/docs/federation.md
@@ -25,7 +25,7 @@ interaction with the gateway.
 If you already have a graph you can add federation simply by adding the `federated` annotation:
 
 ```scala
-import caliban.federation._
+import caliban.federation.v1._
 
 val schema: GraphQL[R] = graphQL(RootResolver(Queries(
   characters = List(Character("Amos"))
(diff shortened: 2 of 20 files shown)


vuepress/docs/docs/federation.md
```diff
@@ -25,7 +25,7 @@ interaction with the gateway.
 If you already have a graph you can add federation simply by adding the `federated` annotation:
 
 ```scala
-import caliban.federation._
+import caliban.federation.v1._
```
ghostdogpr, line 28: So we need to mention in the release notes that this is necessary, right? There is no more "default" version.

federation/src/main/scala/caliban/federation/package.scala
```diff
@@ -1,3 +1,12 @@
...
 
-package object federation extends FederationV1 with FederationDirectives
+import caliban.federation.v2.{ FederationDirectivesV2_3, FederationV2, Versions }
+
+package object federation {
+
+  lazy val v1   = new FederationV1 with FederationDirectives
+  lazy val v2_0 = new FederationV2(List(Versions.v2_0))
+  lazy val v2_1 = new FederationV2(List(Versions.v2_1))
+  lazy val v2_3 = new FederationV2(List(Versions.v2_3)) with FederationDirectivesV2_3
```
kyri-petrou: I think it might be good to add a comment here on which one to use to retain the same behaviour as the old `v2` package. It might be unclear to users which one is the drop-in replacement. We might want to add that to the release notes / docs as well.

kyri-petrou (commented):
I'm not greatly familiar with federated schemas but looks good to me overall. Just a very minor comment below

ghostdogpr (approved):

frekw (approved):
Not much to say other than that it looks good and that I'm digging the new package structure!

vuepress/docs/docs/federation.md
paulpdaniels: Yep, I suppose I could add another constant `v2` pointed at `v2_0` and then deprecate it as well? Though I'd need to move the structure around a bit since it would conflict with the current `v2` package

paulpdaniels: Oh actually misread what you were saying, correct there is no more default version since there is so much difference between the latest versions of federation I think it makes sense to make that explicit

kyri-petrou: > Yep, I suppose I could add another constant v2 pointed at v2_0 and then deprecate it as well

I was going to recommend that at first, but then I realised there's a `v2` package already, so I thought it wouldn't be possible to define a `v2` constant. But if it is possible, I think it's a good idea

paulpdaniels: I moved the `v2` package to `v2x` instead and then made a new deprecated `v2` constant that points at `v2_0`. It definitely isn't bincompat but it should at least still be source compatible for the most part.

federation/src/main/scala/caliban/federation/package.scala
paulpdaniels: See below, I was able to keep the `v2` constant but mark it deprecated to move users onto the new package structure

kyri-petrou (approved):

dariuszkuc:
just an fyi Federation v2.2 changed `@shareable` definition to `repeatable` (to allow specifying it on both type and its extension) -> [reference 2.3 schema](https://specs.apollo.dev/federation/v2.3/federation-v2.3.graphql)

PR #1633, opened 2023-02-27
Use ZPure validation

paulpdaniels:
Switches validation over to using ZPure from zio-prelude. The idea here being that since it is a mainly CPU bound computation we should be able to improve performance by avoiding the lifting into the ZIO environment until the last moment. Marking as draft for the moment because I want to run the benchmarks to compare against the old version.


core/src/main/scala/caliban/validation/ValueValidator.scala
```diff
@@ -111,14 +115,14 @@ object ValueValidator {
         }
     }
 
-  def validateEnum(value: String, inputType: __Type, errorContext: String): IO[ValidationError, Unit] = {
+  def validateEnum(value: String, inputType: __Type, errorContext: String): EReader[Any, ValidationError, Unit] = {
```
ghostdogpr, line 116: How about making a type alias for that so we only specify the `A`?

core/src/main/scala/zio/prelude/ops/package.scala
```diff
@@ -0,0 +1,40 @@
...
+      pure: => ZPure[W, S1, S1, R, E, A]
+    ): ZPure[W, S1, S1, R, E, Option[A]] =
+      when[W, S1, R, E, A](!condition)(pure)
+
+    def when[W, S1, R, E, A](condition: => Boolean)(
+      pure: => ZPure[W, S1, S1, R, E, A]
+    ): ZPure[W, S1, S1, R, E, Option[A]] =
+      ZPure.suspend(if (condition) pure.map(Some(_)) else ZPure.succeed(None))
+
+    def forEachDiscard[W, S1, R, E, A](as: Iterable[A])(
```
ghostdogpr: Oh they didn't exist?

paulpdaniels: There’s `forEach_` which works on things with an implicit `Foreach`, but the inference seemed somewhat troublesome (and for maps I had to convert them to lists first)

frekw:
Will be really interesting to see the benchmarks if it makes any difference!

core/src/main/scala/zio/prelude/ops/package.scala
```diff
@@ -0,0 +1,40 @@
...
+        if (iterator.hasNext) {
+          var result: ZPure[W, S1, S1, R, E, Any] = f(iterator.next())
+          while (iterator.hasNext)
+            result = result *> f(iterator.next())
+          result.unit
+        } else {
+          ZPure.unit
+        }
+      }
+  }
```
kyri-petrou: My gut feeling is that the usage of by-name arguments (except `pure: => ZPure[W, S1, S1, R, E, A]` since we probably want that to be by-name) and `ZPure.suspend(...)` here will have a rather noticeable performance impact. Would something like this work or do we strictly need the methods to return lazily constructed `ZPure`s?

```scala
    def whenCase[W, S1, R, E, A, B](a: A)(
      pf: PartialFunction[A, ZPure[W, S1, S1, R, E, B]]
    ): ZPure[W, S1, S1, R, E, Option[B]] =
      pf.lift(a).map(_.map(Some(_))).getOrElse(ZPure.succeed(None))

    def unless[W, S1, R, E, A](condition: Boolean)(
      pure: => ZPure[W, S1, S1, R, E, A]
    ): ZPure[W, S1, S1, R, E, Option[A]] =
      when[W, S1, R, E, A](!condition)(pure)

    def when[W, S1, R, E, A](condition: Boolean)(
      pure: => ZPure[W, S1, S1, R, E, A]
    ): ZPure[W, S1, S1, R, E, Option[A]] =
      if (condition) pure.map(Some(_)) else ZPure.succeed(None)

    def forEachDiscard[W, S1, R, E, A](as: Iterable[A])(
      f: A => ZPure[W, S1, S1, R, E, Any]
    ): ZPure[W, S1, S1, R, E, Unit] = {
        val iterator = as.iterator
        if (iterator.hasNext) {
          var result: ZPure[W, S1, S1, R, E, Any] = f(iterator.next())
          while (iterator.hasNext)
            result = result *> f(iterator.next())
          result.unit
        } else {
          ZPure.unit
        }
      }
  }
```

core/src/main/scala/caliban/validation/ValueValidator.scala
```diff
@@ -7,40 +7,44 @@ import caliban.introspection.adt._
...
 import zio.{ IO, ZIO }
+import zio.prelude._
+import zio.prelude.ops._
 
 object ValueValidator {
-  def validateDefaultValue(field: __InputValue, errorContext: String): IO[ValidationError, Unit] =
-    ZIO
-      .whenCase(field.defaultValue) { case Some(v) =>
+  def validateDefaultValue(field: __InputValue, errorContext: String): EReader[Any, ValidationError, Unit] =
+    ZPure.suspend(field.defaultValue match {
```
kyri-petrou: Do we need the `ZPure.suspend` here?

core/src/main/scala/caliban/validation/Validator.scala
```diff
@@ -115,26 +130,19 @@ object Validator {
...
     }
-  }
+  }.toZIO
 
   private def check(
     document: Document,
     rootType: RootType,
     variables: Map[String, InputValue]
-  ): IO[ValidationError, Map[String, FragmentDefinition]] = {
+  ): EReader[Any, ValidationError, Map[String, FragmentDefinition]] = ZPure.suspend {
```
kyri-petrou: Same as previously, I think the `ZPure.suspend` might not be needed

kyri-petrou (commented):
Great job with this PR, looks awesome! My main questionmark is whether we need all the `ZPure.suspend` wrapping. 

Assuming they're not strictly necessary, would it be okay to run the performance tests with and without them just to see how big of a difference they make?

core/src/main/scala/zio/prelude/ops/package.scala
ghostdogpr: if we make our own `when`/`whenCase`, maybe we could return `Unit` as it used to be, so that we don't need to do `.unit` everywhere 😄

frekw: Screenshot 2023-02-27 at 16 29 27

core/src/main/scala/caliban/validation/ValueValidator.scala
paulpdaniels: So it should have `Context` fixed as the environment?

paulpdaniels:
```
Before:

[info] Benchmark                             Mode  Cnt      Score      Error  Units
[info] ValidationBenchmark.deep100          thrpt    5  36587.287 ±  597.830  ops/s
[info] ValidationBenchmark.deep1000         thrpt    5  38774.520 ±  962.654  ops/s
[info] ValidationBenchmark.deep10000        thrpt    5  37434.806 ± 1955.669  ops/s
[info] ValidationBenchmark.multifield100    thrpt    5  40530.176 ± 1331.577  ops/s
[info] ValidationBenchmark.multifield1000   thrpt    5  40654.093 ±  811.066  ops/s
[info] ValidationBenchmark.multifield10000  thrpt    5  40736.894 ±  391.088  ops/s
[info] ValidationBenchmark.simple100        thrpt    5  57284.387 ± 1203.689  ops/s
[info] ValidationBenchmark.simple1000       thrpt    5  59926.856 ±  689.663  ops/s
[info] ValidationBenchmark.simple10000      thrpt    5  60312.916 ± 2149.860  ops/s

After:

[info] Benchmark                             Mode  Cnt      Score      Error  Units
[info] ValidationBenchmark.deep100          thrpt    5  43956.629 ± 1114.713  ops/s
[info] ValidationBenchmark.deep1000         thrpt    5  41390.666 ± 1131.941  ops/s
[info] ValidationBenchmark.deep10000        thrpt    5  44897.465 ± 1168.360  ops/s
[info] ValidationBenchmark.multifield100    thrpt    5  43935.164 ± 1409.333  ops/s
[info] ValidationBenchmark.multifield1000   thrpt    5  46596.305 ± 1716.184  ops/s
[info] ValidationBenchmark.multifield10000  thrpt    5  44627.265 ± 1096.116  ops/s
[info] ValidationBenchmark.simple100        thrpt    5  62381.322 ±  957.098  ops/s
[info] ValidationBenchmark.simple1000       thrpt    5  66442.423 ± 3072.536  ops/s
[info] ValidationBenchmark.simple10000      thrpt    5  66518.309 ± 3723.042  ops/s
```

core/src/main/scala/zio/prelude/ops/package.scala
adamgfraser: These operators are built in now.

kyri-petrou:
I was curious as to what difference (if any) [these changes](https://github.com/ghostdogpr/caliban/pull/1633#discussion_r1118378243) would make, and to my surprise, it's quite a lot more significant than I'd imagine. They seem to add another ~15-20% improvement on top of the ~10% we already got from this PR

```
Before (as currently in this PR)
[info] GraphQLBenchmarks.fragmentsCaliban       thrpt    5     80.099 ±   0.957  ops/s
[info] execution.ValidationBenchmark.deep10000  thrpt    5  78225.327 ± 557.705  ops/s

After
[info] GraphQLBenchmarks.fragmentsCaliban       thrpt    5     98.080 ±   1.572  ops/s
[info] execution.ValidationBenchmark.deep10000  thrpt    5  88084.419 ± 484.028  ops/s
```

I only run the test suite against the fragmentsCaliban benchmark (since that query is quite beefy) and just one of the validation ones, but I guess the performance increase will be more or less the same across the other benchmarks

core/src/main/scala/zio/prelude/ops/package.scala
paulpdaniels: Removed

core/src/main/scala/caliban/validation/Validator.scala
paulpdaniels: Removed

kyri-petrou:
I rerun the 2 benchmarks above and it seems that while performance for the `ValidationXX` benchmark is roughly the same, the fragmentsBenchmark has really taken a hit (lower than current master) after switching to zio-prelude 1.0.0-RC17:

```
[info] GraphQLBenchmarks.fragmentsCaliban       thrpt    5     52.417 ±   0.273  ops/s
[info] execution.ValidationBenchmark.deep10000  thrpt    5  77426.451 ± 290.797  ops/s

series/2.x branch:
[info] GraphQLBenchmarks.fragmentsCaliban  thrpt    5  82.412 ± 0.635  ops/s
```

I'm thinking it's probably due to couple of reasons:

1. ~~The `ZPure.foreachXX` [methods](https://github.com/zio/zio-prelude/blob/series/2.x/core/shared/src/main/scala/zio/prelude/fx/ZPure.scala#L1080-L1115) use recursion instead of a while loop as in @paulpdaniels custom implementation, which is likely less performant.~~
2. `ZPure.when`: [the pure argument](https://github.com/zio/zio-prelude/blob/series/2.x/core/shared/src/main/scala/zio/prelude/fx/ZPure.scala#L1209) is by-value instead of by-name, which means we end up creating all the `ValidationError`s unnecessarily.

Since optimizations seem to be needed on the zio-prelude side, I'll take this discussion there. Otherwise, I'm happy with the changes in this PR - great job @paulpdaniels!

**Edit**: the culprit is definitely `ZPure.when`. Changing the `pure` argument to be by-name yields the following results:

```
[info] Benchmark                                 Mode  Cnt      Score      Error  Units
[info] GraphQLBenchmarks.fragmentsCaliban       thrpt    5     96.725 ±    0.420  ops/s
[info] execution.ValidationBenchmark.deep10000  thrpt    5  93019.457 ± 2991.515  ops/s
```

kyri-petrou (approved):
Great job with this PR 🎉 

We might want to temporarily use a custom `when` method or hold off merging this in until https://github.com/zio/zio-prelude/pull/1099 makes it into a new release.

core/src/main/scala/caliban/validation/ValueValidator.scala
ghostdogpr: Ah I see, there are 2 different types... I guess what you did is simple enough 👍

ghostdogpr:
> Great job with this PR 🎉
> 
> We might want to temporarily use a custom `when` method or hold off merging this in until [zio/zio-prelude#1099](https://github.com/zio/zio-prelude/pull/1099) makes it into a new release.

+1 on this, let's have the best perf we can 😄

ghostdogpr (approved):

paulpdaniels:
Ok seems like a new version of prelude should be imminent so I'll wait for that

kyri-petrou:
That was quick - seems RC18 is out already 🎉

core/src/main/scala/caliban/validation/Validator.scala
```diff
@@ -447,18 +475,18 @@ object Validator {
...
           val possibleTypes         = getPossibleTypes(currentType).flatMap(_.name)
           val possibleFragmentTypes = getPossibleTypes(fragmentType).flatMap(_.name)
           val applicableTypes       = possibleTypes intersect possibleFragmentTypes
-          ZIO.when(applicableTypes.isEmpty)(
+          ZPure.when(applicableTypes.isEmpty)(
             failValidation(
               s"${name.fold("Inline fragment spread")(n => s"Fragment spread '$n'")} is not possible: possible types are '${possibleTypes
                 .mkString(", ")}' and possible fragment types are '${possibleFragmentTypes.mkString(", ")}'.",
-              "Fragments are declared on a type and will only apply when the runtime object type matches the type condition. They also are spread within the context of a parent type. A fragment spread is only valid if its type condition could ever apply within the parent type."
+              "Fragments are declared on a type and will only apply ZPure.when the runtime object type matches the type condition. They also are spread within the context of a parent type. A fragment spread is only valid if its type condition could ever apply within the parent type."
```
ghostdogpr, line 489: This one looks wrong

ghostdogpr: There are a few occurrences 😄

ghostdogpr: Fixed!

PR #1635, opened 2023-03-02
Micro-optimization for resolving custom object schema fields

kyri-petrou:
I was looking at some application profiling results at work, and I've noticed that there was a rather surprising amount of CPU time (~2 - 3% of the total application CPU load) that was spent on this method (note that the API has many recursive types which require hand-rolled object schemas).

I _think_ the reason for that is that we are calling the `FieldAttributes => List[(__Field, V => Step[R1])]` lambda on each invocation of the `resolve` method when it's not dependent on the `value: T` input, whereas we probably don't need to


paulpdaniels (approved):
Nice!

ghostdogpr (approved):
Thanks!

Issue #1637, opened 2023-03-05
Fix RenderingSpec


ghostdogpr:
The large string literals in `RenderingSpec` break the scala 2 compiler 😱 
The whole test has been commented out for now.

An alternative implementation could be to parse the result of `render` and do a schema comparison (available in `tools`) instead of a string comparison.

PR #1639, opened 2023-03-05
Switch to jsoniter client

paulpdaniels:
Migrates the client utilities to jsoniter for better performance. As a plus, because there is no built in AST we now use the `__ObjectValue` in most places which means we don't need to expose the underlying json lib anymore.

### DIFF
--- a/client-laminext/src/main/scala/io/laminext/websocket/jsoniter/WebSocketReceiveBuilderJsoniterOps.scala
+++ b/client-laminext/src/main/scala/io/laminext/websocket/jsoniter/WebSocketReceiveBuilderJsoniterOps.scala
@@ -1,0 +1,26 @@
+package io.laminext.websocket.jsoniter
+
+import com.github.plokhotnyuk.jsoniter_scala.core.{ readFromString, writeToString, JsonValueCodec }
+import io.laminext.websocket.{ initialize, receive, send, WebSocketBuilder, WebSocketReceiveBuilder }
+
+import scala.util.control.NonFatal
+
+class WebSocketReceiveBuilderJsoniterOps(b: WebSocketReceiveBuilder) {
+
+  @inline def json[Receive, Send](implicit
+    receiveCodec: JsonValueCodec[Receive],
+    sendCodec: JsonValueCodec[Send]
+  ): WebSocketBuilder[Receive, Send] =
+    new WebSocketBuilder[Receive, Send](
+      url = b.url,
+      protocol = b.protocol,
+      initializer = initialize.text,
+      sender = send.text[Send](writeToString(_)),
+      receiver = receive.text[Receive] { text =>
+        try Right(readFromString[Receive](text))
+        catch {
+          case NonFatal(e) => Left(e)
+        }
+      }
+    )
+}
--- a/client/src/main/scala/caliban/client/GraphQLRequest.scala
+++ b/client/src/main/scala/caliban/client/GraphQLRequest.scala
@@ -1,7 +1,7 @@
 package caliban.client
 
-import io.circe.syntax._
-import io.circe.{ Encoder, Json }
+import com.github.plokhotnyuk.jsoniter_scala.core.JsonValueCodec
+import com.github.plokhotnyuk.jsoniter_scala.macros.JsonCodecMaker
 
 /**
  * Represents a GraphQL request, containing a query and a map of variables.
@@ -10,12 +10,6 @@ case class GraphQLRequest(query: String, variables: Map[String, __Value])
 
 object GraphQLRequest {
 
-  implicit val encoder: Encoder[GraphQLRequest] = (req: GraphQLRequest) =>
-    Json.obj(
-      "query"     -> Json.fromString(req.query),
-      "variables" -> Json.obj(req.variables.map { case (k, v) =>
-        k -> v.asJson
-      }.toList: _*)
-    )
+  implicit val jsonEncoder: JsonValueCodec[GraphQLRequest] = JsonCodecMaker.makeCirceLike[GraphQLRequest]
 
 }
--- a/client/src/main/scala/caliban/client/__Value.scala
+++ b/client/src/main/scala/caliban/client/__Value.scala
@@ -1,6 +1,7 @@
 package caliban.client
 
-import com.github.plokhotnyuk.jsoniter_scala.core.{ JsonReader, JsonValueCodec, JsonWriter }
+import com.github.plokhotnyuk.jsoniter_scala.core.{ writeToString, JsonReader, JsonValueCodec, JsonWriter }
+import com.github.plokhotnyuk.jsoniter_scala.macros.JsonCodecMaker
 
 import scala.annotation.switch
 import scala.collection.immutable.TreeMap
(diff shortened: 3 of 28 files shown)


ghostdogpr:
If you get the latest master, the weird error should be gone (see https://github.com/ghostdogpr/caliban/issues/1637)

client-laminext/src/main/scala/io/laminext/websocket/jsoniter/WebSocketReceiveBuilderJsoniterOps.scala
```diff
@@ -0,0 +1,26 @@
+package io.laminext.websocket.jsoniter
+
+import com.github.plokhotnyuk.jsoniter_scala.core.{ readFromString, writeToString, JsonValueCodec }
+import io.laminext.websocket.{ initialize, receive, send, WebSocketBuilder, WebSocketReceiveBuilder }
+
+import scala.util.control.NonFatal
+
+class WebSocketReceiveBuilderJsoniterOps(b: WebSocketReceiveBuilder) {
+
+  @inline def json[Receive, Send](implicit
```
kyri-petrou, line 10: Does this also work for Scala 3 or does it ignore this annotation?

client/src/main/scala/caliban/client/__Value.scala
```diff
@@ -20,6 +24,9 @@ sealed trait __Value { self =>
 }
 
 object __Value {
+  private final val MaxDepth                                    = 512
```
kyri-petrou: Do we think there might be cases that the users want to grab more than 512 fields deep? In the server we were comfortable in having this since it's up to the user to choose a JSON implementation, but here we're "forcing" one on them - so I'm not sure whether we need to think of any potential edge cases

kyri-petrou (commented):
PR looks great! Just a couple of minor(ish) comments below.

Also, I've seen the `JsonCodecMaker.makeCirceLike` being used to derive most codecs. I get the feeling that default codec (via`JsonCodecMaker.make`) which has a more efficient encoder would also work for most if not all the cases. Did you come across any issues using the default codec maker or was it just so that we could mimic circe's behaviour as much as possible?

ghostdogpr:
FYI we finally decided to drop JDK8 support 😄 It's updated in the main branch

guizmaii:
> FYI we finally decided to drop JDK8 support 😄 It's updated in the main branch

🎉 

@ghostdogpr Can we then remove the dependency on `async-http-client` and replace it with the default JDK `HttpClient`, maybe?

ghostdogpr:
> > FYI we finally decided to drop JDK8 support 😄 It's updated in the main branch
> 
> 
> 
> 🎉 
> 
> 
> 
> @ghostdogpr Can we then remove the dependency on `async-http-client` and replace it with the default JDK `HttpClient`, maybe?

Yeah, sounds good!

ghostdogpr:
Since we dropped jdk8 support, I think this one can be finalized? Can you update to the latest master?

I think it's the last one we can merge before doing the 2.1 😄

client/src/main/scala/caliban/client/__Value.scala
paulpdaniels: I'm not sure, that seems like it would be _really_ deep as a structure. Maybe we could provide it as an `implicit` would that be better? 

```scala
case class MaxDepth(depth: Int)

object MaxDepth {
  implicit val default: MaxDepth = MaxDepth(512)
}
```

paulpdaniels:
Hmmm The subscription tests have started failing, but the only major difference there is that I migrated from the `AsyncHttpBackendClient` to the `HttpBackendClient`. The error seems like it is a decoding error but it is very unclear so I need to dig a bit more into it.

client/src/main/scala/caliban/client/__Value.scala
kyri-petrou: To be honest I agree with you, these are _really_ nested objects. I doubt there are many, if any, real use-cases where one would need to query this deep. 

I think we could probably get away with increasing this value to 1023 and see if anyone ever requests for this limit to be increased or made configurable?

kyri-petrou:
@paulpdaniels I remember having the same issue some time ago when `AsyncHttpClientZioBackend` was deprecated and attempted switching over to `HttpClientZioBackend`. If I recall correctly the client fails to decode the websocket responses. I couldn't for the life of me figure out how to get it to work, and then `AsyncHttpClientZioBackend` got de-deprecated so we just continued using that.

frekw:
Yeah I’ve been down that rabbit hole as well without having figured it out. As @kyri-petrou  it looks like a decoding error

ghostdogpr:
How about just keep using the old backend then? If someone figures it out later, they can make another PR but it seems not directly related to this PR.

plokhotnyuk:
TIL that an active support of JDK 11 ends in 6 months: https://endoflife.date/java

So we are rising the bar just in time ;)

client-laminext/src/main/scala/io/laminext/websocket/jsoniter/WebSocketReceiveBuilderJsoniterOps.scala
paulpdaniels: I'm not actually sure, the compiler didn't complain so I'm guessing it might ignore it?

core/src/main/scala/caliban/interop/jsoniter/jsoniter.scala
```diff
@@ -1,7 +1,7 @@
 package caliban.interop.jsoniter
 
 import caliban.Value._
-import caliban._
+import caliban.{ GraphQLResponse, _ }
```
ghostdogpr: Super minor: this change is not needed

ghostdogpr (approved):
Just a tiny remark, feel free to merge after that 👍

client-laminext/src/main/scala/io/laminext/websocket/jsoniter/WebSocketReceiveBuilderJsoniterOps.scala
kyri-petrou: Yeah I printed the compiler output and it seems that it's ignored. I'm guessing it's okay if it doesn't get inlined right?

client/src/main/scala/caliban/client/GraphQLRequest.scala
```diff
@@ -10,12 +10,6 @@ case class GraphQLRequest(query: String, variables: Map[String, __Value])
...
 object GraphQLRequest {
 
-  implicit val encoder: Encoder[GraphQLRequest] = (req: GraphQLRequest) =>
-    Json.obj(
-      "query"     -> Json.fromString(req.query),
-      "variables" -> Json.obj(req.variables.map { case (k, v) =>
-        k -> v.asJson
-      }.toList: _*)
-    )
+  implicit val jsonEncoder: JsonValueCodec[GraphQLRequest] = JsonCodecMaker.makeCirceLike[GraphQLRequest]
```
kyri-petrou, line 13: Just out of curiosity, do we need this to be `CirceLike` (and the other Codecs) or is it only to keep the encoding / decoding as similar as what we previously had?

paulpdaniels: This is just to keep the codecs as close to the original behavior. I'm not sure if there would be a noticeable difference but I feel it would be hard to decode if these changed in subtle ways under the hood.

PR #1646, opened 2023-03-12
Unified fastparse 3.x parser

kyri-petrou:
## Benchmarks

TLDR:

* Scala 2 parser's performance is roughly the same between fastparse 2 & 3
* Scala 3 parser is now as performant as Scala 2 🚀
* Caliban parser ~x2 faster than Sangria (based on the introspection query)

### Caliban: Fastparse 2.x / cats-parse:
```
Scala 2.13:
[info] GraphQLBenchmarks.parserCaliban  thrpt    5  39264.507 ± 1454.536  ops/s

Scala 3:
[info] GraphQLBenchmarks.parserCaliban  thrpt    5  14765.473 ±  57.193  ops/s
```

### Caliban: Fastparse 3.x:

```
Scala 2:
[info] GraphQLBenchmarks.parserCaliban  thrpt    5  39054.533 ± 440.924  ops/s

Scala 3:
[info] GraphQLBenchmarks.parserCaliban  thrpt    5  40977.712 ± 965.025  ops/s
```

### Sangria:
```
Scala 2 & 3 are roughly the same:
GraphQLBenchmarks.parserSangria  thrpt    5  20894.900 ± 172.430  ops/s
```


ghostdogpr (approved):
Wow this is so good

core/src/main/scala/caliban/parsing/parsers/StringParsers.scala
```diff
@@ -32,9 +32,9 @@ private[caliban] trait StringParsers {
         (state: @switch) match {
           case Normal                  =>
             (currentChar: @switch) match {
+              case CR                                    => loop(index + 1, state = DetermineLineBreakStart)
```
kyri-petrou, line 35: This is a micro-optimization since new lines are very common in graphQL queries, and it seems to perform slightly better if we have this case on the top since it contains only 1 character (as opposed to 5 in the case below)

PR #1650, opened 2023-03-18
Memoize and optimize selection of fragments during `Field` creation

kyri-petrou:
As discussed on Discord, memoizing the creation of fragments provides a significant performance boost. For the `fragmentsQuery` benchmark, this translates to:

```shell
[info] GraphQLBenchmarks.fragmentsCaliban  thrpt    5  104.550 ± 0.505  ops/s # Before
[info] GraphQLBenchmarks.fragmentsCaliban  thrpt    5  135.401 ± 0.987  ops/s # After
```

My big concern however is that we perhaps can't out-right memoize this due to the computation being dependent on `innerType` (more info in the comments below). However, all the tests seem to pass and I couldn't manage to create a case where the same `FragmentSpread(...)` would result in different `Field` being generated based on different `innerType`. Happy to hear your thoughts on what I could further try to test this

I've also included some other minor optimizations and cleanups in this PR - happy to remove them if we want to have a cleaner view of the memoization changes

### DIFF
--- a/core/src/main/scala/caliban/execution/Field.scala
+++ b/core/src/main/scala/caliban/execution/Field.scala
@@ -118,10 +118,7 @@ object Field {
 
               val _fields = if (checkDirectives(resolvedDirectives)) {
                 fragments.get(name).map { f =>
-                  val t = innerType.possibleTypes
-                    .flatMap(_.find(_.name.contains(f.typeCondition.name)))
-                    .orElse(rootType.types.get(f.typeCondition.name))
-                    .getOrElse(fieldType)
+                  val t = rootType.types.getOrElse(f.typeCondition.name, fieldType)
                   loop(f.selectionSet, t).map { field =>
                     if (field._condition.isDefined) field
                     else
--- a/core/src/main/scala/caliban/execution/Field.scala
+++ b/core/src/main/scala/caliban/execution/Field.scala
@@ -1,6 +1,5 @@
 package caliban.execution
 
-import scala.collection.mutable.ArrayBuffer
 import caliban.Value.BooleanValue
 import caliban.introspection.adt.{ __DeprecatedArgs, __Type }
 import caliban.parsing.SourceMapper
@@ -9,6 +8,8 @@ import caliban.parsing.adt.Selection.{ Field => F, FragmentSpread, InlineFragmen
 import caliban.parsing.adt.{ Directive, LocationInfo, Selection, VariableDefinition }
 import caliban.schema.{ RootType, Types }
 import caliban.{ InputValue, Value }
+
+import scala.jdk.CollectionConverters._
 
 /**
  * Represents a field used during the exeuction of a query
@@ -67,25 +68,15 @@ object Field {
     directives: List[Directive],
     rootType: RootType
   ): Field = {
-    def loop(selectionSet: List[Selection], fieldType: __Type): Field = {
-      val fieldList  = ArrayBuffer.empty[Field]
-      val map        = collection.mutable.Map.empty[(String, String), Int]
-      var fieldIndex = 0
+    val memoizedFragments = collection.mutable.Map.empty[FragmentSpread, Option[List[(Field, String)]]]
+
+    def loop(selectionSet: List[Selection], fieldType: __Type): List[Field] = {
+      val map = new java.util.LinkedHashMap[(String, String), Field](selectionSet.length)
 
       def addField(f: Field, condition: Option[String]): Unit = {
         val name = f.alias.getOrElse(f.name)
         val key  = (name, condition.getOrElse(""))
-        map.get(key) match {
-          case None        =>
-            // first time we see this field, add it to the array
-            fieldList += f
-            map.update(key, fieldIndex)
-            fieldIndex = fieldIndex + 1
-          case Some(index) =>
-            // field already existed, merge it
-            val existing = fieldList(index)
-            fieldList(index) = existing.combine(f)
-        }
+        map.compute(key, (_, existing) => if (existing == null) f else existing.combine(f))
       }
 
       val innerType            = Types.innerType(fieldType)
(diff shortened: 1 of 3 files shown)


core/src/main/scala/caliban/execution/Field.scala
```diff
@@ -121,68 +111,72 @@ object Field {
...
           if (checkDirectives(resolvedDirectives)) {
-            val t = innerType.possibleTypes
+            val t      = innerType.possibleTypes
               .flatMap(_.find(_.name.exists(typeCondition.map(_.name).contains)))
               .orElse(typeCondition.flatMap(typeName => rootType.types.get(typeName.name)))
               .getOrElse(fieldType)
-            if (selectionSet.nonEmpty) {
-              val fields = loop(selectionSet, t).fields
-              typeCondition match {
-                case None           => fieldList ++= fields
```
kyri-petrou, line 164: It seems to me that this was a bug? If we appended `fields` onto `fieldsList` without incrementing `fieldIndex`, wouldn't that potentially cause the wrong field to be merged?

core/src/main/scala/caliban/execution/Field.scala
```diff
@@ -121,68 +111,72 @@ object Field {
...
+          val fields = memoizedFragments.getOrElseUpdate(
+            fr, {
+              val resolvedDirectives = directives.map(resolveDirectiveVariables(variableValues, variableDefinitions))
+
+              if (checkDirectives(resolvedDirectives)) {
+                fragments
+                  .get(name)
+                  .map { f =>
+                    val t = innerType.possibleTypes
+                      .flatMap(_.find(_.name.contains(f.typeCondition.name)))
```
kyri-petrou, line 124: This is the part that concerns me the most, since `innerType` will be different each time we are here. If we memoize this, it seems that it won't take the new `innerType` into consideration. But in all cases that I tested this against, `innerType.possibleTypes` was `None`, so I'm not sure what needs to be done for it to be `Some(...)`

core/src/main/scala/caliban/execution/Field.scala
```diff
@@ -121,68 +111,70 @@ object Field {
...
+          val (fields, condition) = memoizedFragments.getOrElseUpdate(
+            fr, {
+              val resolvedDirectives = directives.map(resolveDirectiveVariables(variableValues, variableDefinitions))
+
+              val _fields = if (checkDirectives(resolvedDirectives)) {
+                fragments.get(name).map { f =>
+                  val t = innerType.possibleTypes
+                    .flatMap(_.find(_.name.contains(f.typeCondition.name)))
+                    .orElse(rootType.types.get(f.typeCondition.name))
+                    .getOrElse(fieldType)
```
kyri-petrou: This is the part that concerns me the most, since innerType will be different each time we are here. If we memoize this, it seems that it won't take the new innerType into consideration. But in all cases that I tested this against, innerType.possibleTypes was None, so I'm not sure what needs to be done for it to be Some(...)

core/src/main/scala/caliban/execution/Field.scala
ghostdogpr: I agree, it looks like a bug!

ghostdogpr: If it's not too difficult, a test case that reproduced the bug would be awesome

core/src/main/scala/caliban/execution/Field.scala
ghostdogpr: In test `fragment on union` in `FragmentSpec`, possibleTypes is not None (it contains the subtypes of a union), but `find` doesn't find anything because the fragment is on the parent type, not one of the subtypes. Is it even possible? It sounds like even if we removed the find, it would find the type in `rootType.types` anyway, no?

core/src/main/scala/caliban/execution/Field.scala
kyri-petrou: I'll look into how I can reproduce. It seems that this case would only be reachable with inlined fragments that don't have a type condition. From a quick search, that means inlined fragments that use the `@skip` or `@include` directives

core/src/main/scala/caliban/execution/Field.scala
kyri-petrou: That's an interesting point - it seems we should always be able to get the type from `rootType.types`, but I'm not sure if it's possible to get a different type from `innerType.possibleTypes`? Or is that definitely not possible?

Another approach I just thought of if we wanted to be _extremely_ cautious is to move the memoization after L121-122 and only fetch from the cache if those 2 lines return `Some(...)` different value than L123. It does add a bit of overhead but at least this way we can be more certain we're not introducing a regression?

ghostdogpr: I think it only makes sense on inline fragments where you have one fragment for each subtype, but for fragment spreads I think you have to use the parent type anyway, so I think we can just go straight to the `orElse`.

core/src/main/scala/caliban/execution/Field.scala
kyri-petrou: I've managed to create a test case where the wrong fields get combined. Given the query below, with the current code instead of merging `name` + `name`, we end up combining `origin` + `name`. However, since the fields are combined, there's no overriding. I'm not sure how to translate that into an actual issue or get a test to fail. Any thoughts?

```gql
query {
  amos: character(name: "Amos Burton") {
    ... @include(if: true) {
      origin
    }
    name
    name
  }
}
```

ghostdogpr: If you make the resolver of type `Field => ...`, then you can return the list of requested fields, and it would be wrong I think?

kyri-petrou: Okay took me a while but got there eventually 😅 As I mentioned on Discord, the test would fail without the change in the `Field.apply` method

https://github.com/ghostdogpr/caliban/compare/series/2.x...reproduce-inline-fragment-bug

kyri-petrou: Test added in this PR as well

PR #1666, opened 2023-03-29
Fix `derives Schema.Auto` derivation

kyri-petrou:
Fixes https://discord.com/channels/629491597070827530/633200096393166868/1090541477395959920

### DIFF
--- a/core/src/test/scala-3/caliban/schema/Scala3DerivesSpec.scala
+++ b/core/src/test/scala-3/caliban/schema/Scala3DerivesSpec.scala
@@ -38,25 +38,13 @@ object Scala3DerivesSpec extends ZIOSpecDefault {
 
         assertTrue(gql.render == expected)
       },
-      test("Auto derivation reuses implicit derivations in GenericSchema") {
-        case class A(a: String)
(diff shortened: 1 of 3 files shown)


core/src/test/scala-3/caliban/schema/Scala3DerivesSpec.scala
```diff
@@ -6,81 +6,80 @@ import zio.test.{ assertTrue, ZIOSpecDefault }
...
+
+        |type Bar {
+        |  foo: Foo!
+        |}
+
+        |type Foo {
+        |  value: String!
+        |}""".stripMargin
+
+    List(
```
kyri-petrou, line 24: Seems that I messed this one up when I was writing the tests at first. Good thing is that all of the "old" tests were passing

kyri-petrou:
I copied the `SchemaSpec` tests into the Scala 3 folder and slightly rewrote them to use `derives Schema.Auto`, so that we test against a much wider range of cases to make sure there are no other issues with the usage of it.

Note that quite a few of these tests were failing on the main branch

ghostdogpr (approved):

PR #1676, opened 2023-04-06
Added ZIO Caliban real technical interview project to Resources

infradev1:
GitHub link to real technical interview project using ZIO, Caliban, Doobie, Cats core, and domain error modeling.


ghostdogpr:
Ah the html files are auto-generated, instead you need to modify the markdown files that are in the vuepress folder. Thanks, it will be a great addition!

infradev1:
@ghostdogpr PR updated!

kyri-petrou:
Slightly off-topic, but I realised recently that [methods inside `Http.collectHttp`](https://github.com/CarlosLaraFP/Cases/blob/de4d0bd5cc015cf5c2d816acfba2444dd05e5d28/src/main/scala/com/company/cases/CaseApp.scala#L33-L38) are invoked on each request. So in the link above we end up recreating the service for each request. I'm not sure if this is the desired behaviour or a bug in zio-http

@ghostdogpr I wonder whether we should change [the example here](https://github.com/ghostdogpr/caliban/blob/series/2.x/examples/src/main/scala/example/ziohttp/ExampleApp.scala#L23-L24) so that the creation of the http / ws service is outside the partial function

ghostdogpr:
> @ghostdogpr I wonder whether we should change [the example here](https://github.com/ghostdogpr/caliban/blob/series/2.x/examples/src/main/scala/example/ziohttp/ExampleApp.scala#L23-L24) so that the creation of the http / ws service is outside the partial function

I wonder if this is a bug as well, we definitely don't want that!

infradev1:
@ghostdogpr @kyri-petrou The official ZIO guide also follows this approach: [How To Build a GraphQL Web Service](https://zio.dev/guides/tutorials/build-a-graphql-webservice)

I would expect the zio-http implementation guarantees a single instance of the service. This is certainly the case for REST APIs. I wonder why it would not be the case for GraphQL APIs.

Is this something we need to resolve before merging the PR?

kyri-petrou:
> Is this something we need to resolve before merging the PR?

Not at all. I'm happy to have this merged in. My comment was more oriented about changing the example

kyri-petrou (approved):

PR #1682, opened 2023-04-11
Build caliban-client for Scala Native

kyri-petrou:
As discussed on Discord, this PR builds the Caliban client for Scala Native. While it's a bit of a niche usecase, it potentially allows for users to build native apps similar to `cURL` for GraphQL using Scala & Caliban.

You might notice that we're not running the tests for Scala Native. I couldn't get `zio-test` to execute the tests using Scala Native. I believe this is related to these 2 PRs:

- https://github.com/zio/zio/pull/7883
- https://github.com/zio/zio/pull/7911

I picked a few tests and rewrote them using JUnit locally (which is officially supported by Scala Native) and they were passing. For now, the only thing tested is the compilation to native executable.I'm open to ideas on other ways we could test that this work.

Also, I added the native tests alongside the JS CI. It felt like a bit of an overkill to create 2 new CI workflows just for building the client for native, but I can do that if we think it's better.

Leaving the PR as draft until CI passes as there are likely some packages required for the build that I missed.


kyri-petrou:
Hmmm weird.. The `test212_jdk` test failed twice at the exact same spot due to a `Received "killed" signal`. I'm thinking it might not like the `-Xmx3G` in `.sbtops`

ghostdogpr:
About the memory issue, I added `resource_class: large` in the CI config for the scripted steps, but you can add it for the others if necessary.

kyri-petrou:
By the way it seems that the addition of the flags to `.sbtopts` has made a rather noticeable improvement to the `scriptedX` steps (~6 mins down from 8)

ghostdogpr (approved):

ghostdogpr:
Anything else we're waiting?

kyri-petrou:
> Anything else we're waiting?

Nope all good 👍 - just wanted to get CI to pass before marking it as ready

PR #1688, opened 2023-04-13
Allow passing of directives to making scalars

yarian:
We ran into an issue when upgrading federation 2. We need to be able to add `@tag` directives to  scalar type definitions. This currently doesn't seem supported but it seems like an oversight given that it's there for enums, etc...


yarian:
CI is failing only on here. My fork has it running fine. I found an issue here:
https://discuss.circleci.com/t/could-not-find-a-usable-config-yml-you-may-have-revoked-the-circleci-oauth-app/46274/5

Which indicated that closing and re-opening pull request fixed it for some people. So going to try that.

PR #1694, opened 2023-04-18
Allow passing of directives to making scalars

yarian:
We ran into an issue when upgrading federation 2. We need to be able to add @tag directives to scalar type definitions. This currently doesn't seem supported but it seems like an oversight given that it's there for enums, etc...


core/src/main/scala/caliban/schema/Schema.scala
```diff
@@ -137,11 +137,12 @@ trait GenericSchema[R] extends SchemaDerivation[R] with TemporalSchema {
     name: String,
     description: Option[String],
     specifiedBy: Option[String],
+    directives: Option[List[Directive]],
```
ghostdogpr, line 141: Can you update the scaladoc?

core/src/main/scala/caliban/schema/Schema.scala
```diff
@@ -131,17 +131,19 @@ trait GenericSchema[R] extends SchemaDerivation[R] with TemporalSchema {
    * @param name name of the scalar type
    * @param description description of the scalar type
    * @param specifiedBy URL of the scalar specification
+   * @param directives The directives to add to the type
```
ghostdogpr:
```suggestion
   * @param directives the directives to add to the type
```
nitpicking just to be consistent with the other parameters 😄

core/src/main/scala/caliban/Rendering.scala
```diff
@@ -20,7 +20,9 @@ object Rendering {
               if (isBuiltinScalar(name)) None
               else
                 Some(
-                  s"""${renderDescription(t.description)}scalar $name${renderSpecifiedBy(t.specifiedBy)}""".stripMargin
+                  s"""${renderDescription(t.description)}\n${renderDirectives(
```
ghostdogpr: According to the spec, the directives come after the name: https://spec.graphql.org/October2021/#sec-Scalars

yarian: Ah, TIL. Thanks will fix.

ghostdogpr: Also I think you don't need any space or line break, if you look at other usage of `renderDirectives`

yarian:
Looks like CircleCI got its act together 🥳

ghostdogpr (approved):
Thanks!

PR #1698, opened 2023-04-20
Stop filtering directives on inputs and input fields

yarian:
This is another difference between federation 1 and federation 2.

https://www.apollographql.com/docs/graphos/delivery/contracts/#valid-tag-locations

In federation 2 input types and their fields can and should be tagged when using contract graphs. We were previously filtering out their directives. I suspect this is so that a type that is both a normal type and an input type would not need to be defined by library users twice, but it is problematic in Federation 2.

I am not sure how Caliban normally handles these sort of splits or if it does or if we just say that the latest version doesn't work for Fed 1 but this seems like a breaking change without some mechanism to mitigate it.


ghostdogpr (approved):
LGTM

ghostdogpr:
cc @paulpdaniels

PR #1700, opened 2023-04-23
Add MiMa checks and fix binary incompatible changes

### LINKED ISSUE #1702: Setup MiMa to prevent unintended breaking changes
Now that Caliban has the versioning scheme defined as `pvp`, it would be good to setup mima checks during CI so that we can catch any unintended breaking changes.

kyri-petrou:
This PR adds MiMa checks during CI. See comments below for some more info. Just some points for discussion:

1. Should we be strict when the code is source compatible but not binary compatible? (e.g., the `makeScalar` method wasn't bincompat)
2. What's the best way to ensure that we clear the MiMa filters after each release?

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -63,6 +63,7 @@ jobs:
   test213_jdk:
     docker:
       - image: cimg/openjdk:11.0-node
+    resource_class: large
     steps:
       - checkout
       - restore_cache:
@@ -77,6 +78,7 @@ jobs:
   test3_jdk:
     docker:
       - image: cimg/openjdk:11.0-node
+    resource_class: large
     steps:
       - checkout
       - restore_cache:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -66,7 +66,7 @@ jobs:
       - checkout
       - restore_cache:
           key: sbtcache
-      - run: sbt ++2.13.10 core/test http4s/test akkaHttp/test play/test zioHttp/test examples/compile catsInterop/compile benchmarks/compile monixInterop/compile tapirInterop/test clientJVM/test federation/test reporting/test tracing/test
+      - run: sbt ++2.13.10 core/test http4s/test akkaHttp/test play/test zioHttp/test examples/compile catsInterop/compile benchmarks/compile monixInterop/compile tapirInterop/test clientJVM/test tools/test federation/test reporting/test tracing/test
       - save_cache:
           key: sbtcache
           paths:
@@ -80,7 +80,7 @@ jobs:
       - checkout
       - restore_cache:
           key: sbtcache
-      - run: sbt ++3.2.2 core/test catsInterop/compile benchmarks/compile monixInterop/compile clientJVM/test clientJS/compile zioHttp/test tapirInterop/test http4s/test federation/test reporting/test tracing/test
+      - run: sbt ++3.2.2 core/test catsInterop/compile benchmarks/compile monixInterop/compile clientJVM/test clientJS/compile zioHttp/test tapirInterop/test http4s/test federation/test tools/test reporting/test tracing/test
       - save_cache:
           key: sbtcache
           paths:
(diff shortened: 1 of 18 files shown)


.circleci/config.yml
```diff
@@ -80,7 +80,7 @@ jobs:
       - checkout
       - restore_cache:
           key: sbtcache
-      - run: sbt ++3.2.2 core/test catsInterop/compile benchmarks/compile monixInterop/compile clientJVM/test clientJS/compile zioHttp/test tapirInterop/test http4s/test federation/test reporting/test tracing/test
+      - run: sbt ++3.2.2 core/test catsInterop/compile benchmarks/compile monixInterop/compile clientJVM/test clientJS/compile zioHttp/test tapirInterop/test http4s/test federation/test tools/test reporting/test tracing/test
```
kyri-petrou, line 86: We were only testing `tools` against Scala 2.12. Was that intentional?

core/src/main/scala/caliban/schema/Types.scala
```diff
@@ -7,14 +7,19 @@ import scala.annotation.tailrec
...
-    directives: Option[List[Directive]] = None
+    specifiedBy: Option[String] = None
+  ): __Type =
+    makeScalar(name, description, specifiedBy, None)
+
+  def makeScalar(
+    name: String,
+    description: Option[String],
+    specifiedBy: Option[String],
+    directives: Option[List[Directive]]
```
kyri-petrou: So this change was source compatible but not binary compatible. I'm 50-50 whether I should revert this change and add this method as a filter instead

build.sbt
```diff
@@ -579,6 +595,22 @@ lazy val commonSettings = Def.settings(
...
 )
 
+lazy val enableMimaSettingsJVM =
+  Def.settings(
+    mimaPreviousArtifacts := previousStableVersion.value.map(organization.value %% moduleName.value % _).toSet,
+    mimaBinaryIssueFilters ++= Seq( // Filters should be cleared after a new release
+      ProblemFilters.exclude[MissingTypesProblem]("caliban.schema.SchemaDerivation$Auto"),
+      ProblemFilters.exclude[ReversedMissingMethodProblem](
+        "caliban.schema.SchemaDerivation#Auto.caliban$schema$SchemaDerivation$Auto$$$outer"
+      )
```
kyri-petrou: The changes to `Schema.Auto` are not bincompat but that's okay since it wasn't working as intended previously

core/src/main/scala/caliban/schema/Types.scala
ghostdogpr: Considering the changes I'm working on with the adapters, I think we may go 2.2 directly so maybe it's fine to keep those breaking changes.

.circleci/config.yml
ghostdogpr: Probably an oversight 👍

ghostdogpr:
> 1. Should we be strict when the code is source compatible but not binary compatible? (e.g., the `makeScalar` method wasn't bincompat)

I think we should be strict for patch versions.

> 2. What's the best way to ensure that we clear the MiMa filters after each release?

I have a step where I update the version in docs, I can do it at the same time (it's manual though).

How do we make Mima not fail when we know the next release will be a minor version release (in which case, we allow breaking changes)?

core/src/main/scala/caliban/schema/Types.scala
kyri-petrou: > How do we make Mima not fail when we know the next release will be a minor version release (in which case, we allow breaking changes)?

@ghostdogpr I  added a flag in `build.sbt` to disable CI breaking on binary incompatible changes. The good thing about this flag is that the plugin will still report the bincompat issues in the console, so we can always look at it when we're close to releasing to know what bin / source incompatible changes we introduced

Also, Since the next release will likely be 2.2.0, I reverted the changes to fix the bincompat issue in `makeScalar`.

kyri-petrou:
> How do we make Mima not fail when we know the next release will be a minor version release (in which case, we allow breaking changes)?

@ghostdogpr I  added a flag in `build.sbt` to disable CI breaking on binary incompatible changes. The good thing about this flag is that the plugin will still report the bincompat issues in the console, so we can always look at it when we're close to releasing to know what bin / source incompatible changes we introduced

Also, Since the next release will likely be 2.2.0, I reverted the changes to fix the bincompat issue in `makeScalar`.

core/src/main/scala/caliban/introspection/adt/__Directive.scala
```diff
@@ -17,4 +17,7 @@ case class __Directive(
       isRepeatable,
       locations.map(_.toDirectiveLocation)
     )
+
+  @deprecated("Use isRepeatable")
+  def repeatable: Boolean = isRepeatable
```
ghostdogpr: Can we delete these then?

core/src/main/scala/caliban/schema/Schema.scala
```diff
@@ -147,6 +147,15 @@ trait GenericSchema[R] extends SchemaDerivation[R] with TemporalSchema {
       override def resolve(value: A): Step[Any]                              = PureStep(makeResponse(value))
     }
 
+  @deprecated("use the overloaded method with directives instead")
```
ghostdogpr: Delete?

core/src/main/scala/caliban/schema/Types.scala
```diff
@@ -7,9 +7,6 @@ import scala.annotation.tailrec
 
 object Types {
 
-  /**
```
ghostdogpr: Any reason to remove this?

core/src/main/scala/caliban/introspection/adt/__Directive.scala
kyri-petrou: Done 👍 My bad, had to head out in a hurry earlier and didn't cleanup properly

core/src/main/scala/caliban/schema/Types.scala
kyri-petrou: I deleted the extra method but didn't realise I also deleted the scaladoc 😅

ghostdogpr (approved):

PR #1701, opened 2023-04-23
Add type alias for `ArgBuilder.Auto` and cleanup `Schema.Auto`

kyri-petrou:
Leaving this as draft until we hear back whether this is indeed a bug with the Scala 3 compiler.

I've also removed the extension of `GenericSchema` from `Schema.Auto`. This doesn't seem to be needed now that we're using `using NotGiven[...]`.


kyri-petrou:
@ghostdogpr still haven't heard back on the issue on dotty's page. What do you think regarding this approach? Or should we avoid the alias altogether by renaming `ArgBuilder.Auto` to `ArgBuilder.GenAuto`(or otherwise) instead?

ghostdogpr:
Let's go on with this, no choice

ghostdogpr (approved):

PR #1705, opened 2023-04-25
update tapir=1.2.13 zio-http=3.0.0-RC1

jgulotta:


jgulotta:
I don't really understand the error here. The changes are pretty trivial but maybe something needs to be added to get websockets working again?

kyri-petrou:
Hmm the elusive decoding error is back. We had this issue in the past but it was when we tried to use a different sttp backend in the tests I believe.

We still don't know what the root cause is, but in the past we'd get it with all server implementations. We believe it's related to the sttp client not being able to decode the websocket payload correctly. In this case it seems only the zio-http server implementation is failing

jgulotta:
I figured out this case at least. The issue was in the protocol selection

```scala
        val protocol = req.headers.get(Header.SecWebSocketProtocol) match {
          case Some(value) => Protocol.fromName(value.toString())
          case None        => Protocol.Legacy
        }
```

The `toString()` is not correct for the new header API and does not give the actual header value.

So `Protocol.fromName` was not receiving the actual header value and defaulting to the legacy case instead of the proper case

```scala
  def fromName(name: String): Protocol = name match {
    case GraphQLWS.name => GraphQLWS
    case _              => Legacy
  }
```

 To get the actual value one needs to use `renderedValue` instead of `toString()`, which then resolves the correct protocol.

examples/src/main/scala/example/ziohttp/AuthExampleApp.scala
```diff
@@ -61,7 +61,7 @@ object Auth {
   }
 
   def middleware[R] = HttpAppMiddleware.customAuthZIO { headers =>
-    val user = headers.authorization.map(_.toString())
+    val user = headers.get(Header.Authorization).map(_.toString())
```
kyri-petrou: Should this be `.renderedValue` as well?

jgulotta: Yes! Fixed

ghostdogpr (approved):
Thanks!

PR #1707, opened 2023-04-28
Refactor request interceptors and execution configuration

### LINKED ISSUE #1478: RequestInterceptor is limiting changes on resource axis
Current type declares idempotence for ZIO resource:
```scala
trait RequestInterceptor[-R] { self =>
  def apply[R1 <: R, A](request: ServerRequest)(e: ZIO[R1, TapirResponse, A]): ZIO[R1, TapirResponse, A]
  ...
}
```
while it can be beneficial to remove that limit to:
- be able to `provideSomeLayer`, or
- declare more resource dependency

Sort of:
```scala
trait RequestInterceptor[-R] { self =>
  def apply[R2, R1 <: R, A](request: ServerRequest)(e: ZIO[R1, TapirResponse, A]): ZIO[R2, TapirResponse, A]
  ...
```

As for now the `provideSomeLayer` requires a dirty lie:
```scala
  override def apply[R1 <: R, A](
        request: ServerRequest
      )(e: ZIO[R1, TapirAdapter.TapirResponse, A]
      ): ZIO[R1, TapirAdapter.TapirResponse, A] =
      for {
        userInfo <- authenticate(request)
        ne <- e.provideSomeLayer[R1](ZLayer.succeed(userInfo)) // TODO: WTF!!!
      } yield ne
  }
```
where in fact `ne` does not need `UserInfo` in it's `R`.

ghostdogpr:
Closes #1478


adapters/akka-http/src/main/scala/caliban/AkkaHttpAdapter.scala
```diff
@@ -24,56 +24,89 @@ import scala.concurrent.{ ExecutionContext, Future }
...
+              StatusCode,
+              (String, CalibanPipe),
+              ZioWebSockets,
+              RIO[R, *]
+            ]
+          ]
+      )
+    )
+
+  @deprecated
```
ghostdogpr: I kept the old functions but deprecated and without the `requestInterceptor`. Is it worth it?

core/src/main/scala/caliban/Configurator.scala
```diff
@@ -0,0 +1,43 @@
...
+  private[caliban] def skipValidation: UIO[Boolean] =
+    skipQueryValidationRef.get
+
+  private[caliban] def enableIntrospection: UIO[Boolean] =
+    enableIntrospectionRef.get
+
+  private[caliban] def queryExecution: UIO[QueryExecution] =
+    queryExecutionRef.get
+
+  def setSkipValidation(skip: Boolean): UIO[Unit] =
```
ghostdogpr: I was not sure to make these `ZIO` or `ZLayer`. If you write your own interceptor logic, `ZIO` is more convenient, but if you wanna use it as it is, `ZLayer` is better 🤔

core/src/main/scala/caliban/Configurator.scala
```diff
@@ -0,0 +1,43 @@
...
+    skipQueryValidationRef.set(skip).unit
```
paulpdaniels: I was thinking we have two maybe:

```scala
def setSkipValidation(skip: Boolean): ULayer[Unit] = ZLayer.scoped(setSkipValidationScoped(skip))

def setSkipValidationScoped(skip: Boolean): URIO[Scope, Unit]
```

Then you could configure it as:

```scala
.configure(
  Configurator.setSkipValidation(true) ++ Configurator.setEnableIntrospection(true)
)
```

Without necessarily needing to wrap everything

examples/src/main/scala/example/akkahttp/AuthExampleApp.scala
```diff
@@ -19,44 +19,37 @@ object AuthExampleApp extends App {
...
-  implicit val runtime: Runtime[Auth] = Unsafe.unsafe(implicit u => Runtime.unsafe.fromLayer(initLayer))
+  implicit val runtime: Runtime[Any]                      = Runtime.default
 
   val interpreter = Unsafe.unsafe(implicit u => runtime.unsafe.run(api.interpreter).getOrThrow())
   val adapter     = AkkaHttpAdapter.default
 
   val route =
     path("api" / "graphql") {
-      adapter.makeHttpService(interpreter, requestInterceptor = AuthInterceptor)
+      adapter.makeHttpService(HttpAdapter(interpreter).configure(conf))
```
paulpdaniels: 💯 I like this new syntax

interop/tapir/src/main/scala/caliban/interop/tapir/HttpUploadAdapter.scala
```diff
@@ -0,0 +1,120 @@
...
+
+      def executeRequest(
+        graphQLRequest: GraphQLRequest,
+        serverRequest: ServerRequest
+      ): ZIO[R1, TapirResponse, GraphQLResponse[E]] =
+        self
+          .executeRequest(graphQLRequest, serverRequest)
+          .provideSome[R1](ZLayer.succeed(serverRequest), configurator)
+    }
+}
```
paulpdaniels: Just a thought, but if we made this trait sealed could we then optimize the `provideSome` by doing something like:

```scala
case class Configured[R1, R](layer: ZLayer[R1 & ServerRequest, TapirResponse, R]) extends HttpUploadAdapter[R1, E] {
  def configure[R2](configurator: ZLayer[R2 & ServerRequest, TapirResponse, R1]) = 
    Configured(ZLayer.make[R1](configurator, layer))
}
```

I haven't tried to compile that so not sure if that would work.

core/src/main/scala/caliban/Configurator.scala
```diff
@@ -0,0 +1,43 @@
...
+    Unsafe.unsafe(implicit u => FiberRef.unsafe.make(DefaultValidations))
+
+  private val skipQueryValidationRef: FiberRef[Boolean] =
+    Unsafe.unsafe(implicit u => FiberRef.unsafe.make(false))
+
+  private val enableIntrospectionRef: FiberRef[Boolean] =
+    Unsafe.unsafe(implicit u => FiberRef.unsafe.make(true))
+
+  private val queryExecutionRef: FiberRef[QueryExecution] =
+    Unsafe.unsafe(implicit u => FiberRef.unsafe.make(QueryExecution.Parallel))
```
paulpdaniels: I'm wondering if it would optimize this to make them all part of a single configuration case class? Then during execution phase we only need to examine a single `FiberRef`.

e.g. `case class ExecutionConfiguration(skipValidation: Boolean, queryExecution: QueryExecution)`

core/src/main/scala/caliban/Configurator.scala
ghostdogpr: Yeah, good idea 👍

core/src/main/scala/caliban/Configurator.scala
ghostdogpr: Sounds good, will do.

ghostdogpr: Done

core/src/main/scala/caliban/Configurator.scala
ghostdogpr: Done

interop/tapir/src/main/scala/caliban/interop/tapir/HttpUploadAdapter.scala
ghostdogpr: Applied on `HttpAdapter`. Let me know what you think, then I'll apply elsewhere.

adapters/akka-http/src/main/scala/caliban/AkkaHttpAdapter.scala
ghostdogpr: I got rid of it

adapters/zio-http/src/main/scala/caliban/ZHttpAdapter.scala
```diff
@@ -1,47 +1,27 @@
...
-      requestInterceptor
-    )
-    ZioHttpInterpreter(serverOptions).toHttp(endpoints)
-  }
+  ): HttpApp[R, Throwable] =
+    ZioHttpInterpreter(serverOptions).toHttp(adapter.serverEndpoints[R])
 
   def makeWebSocketService[R, E](
     interpreter: GraphQLInterpreter[R, E],
-    skipValidation: Boolean = false,
```
ghostdogpr, line 41: Argh, the zio-http websocket support still doesn't use Tapir!

ghostdogpr:
TODO & questions:
- zio-http websocket doesn't use Tapir, not sure if I keep the old way or if I try to make it work with `WebSocketAdapter`
- http4s and zio-http auth examples use a middleware at the http level, not a request interceptor, should I change them?
- refactor all examples to use the module pattern 2.0

interop/tapir/src/main/scala/caliban/interop/tapir/HttpAdapter.scala
```diff
@@ -0,0 +1,135 @@
...
+
+  def serverEndpoints[R1 <: R]: List[ServerEndpoint[Any, RIO[R1, *]]] = {
+    def logic(request: (GraphQLRequest, ServerRequest)): RIO[R1, Either[TapirResponse, GraphQLResponse[E]]] = {
+      val (graphQLRequest, serverRequest) = request
+      executeRequest(graphQLRequest, serverRequest).either
+    }
+    endpoints.map(_.serverLogic(logic))
+  }
+
+  def configure[R1](configurator: ZLayer[R1 & ServerRequest, TapirResponse, R]): HttpAdapter[R1, E] =
```
ghostdogpr: Should I name this one `configureLayer`?

core/src/main/scala/caliban/Configurator.scala
```diff
@@ -0,0 +1,47 @@
...
+import caliban.validation.Validator.DefaultValidations
+import zio._
+import zio.prelude.EReader
+
+object Configurator {
+  case class ExecutionConfiguration(
+    skipValidation: Boolean = false,
+    enableIntrospection: Boolean = true,
+    queryExecution: QueryExecution = QueryExecution.Parallel,
+    validations: List[EReader[Context, ValidationError, Unit]] = DefaultValidations
```
kyri-petrou: Really love this! Gives a lot of flexibility to advanced users to customize which validations to perform.

However, I feel like `EReader[Context, ValidationError, Unit]` might be bit confusing to users at a first glance, especially for those that haven't used `zio-prelude` before. Should we use a type alias `Validation` (or similar)?

kyri-petrou: On a similar note, should we add an object `Validations` in `Validator` that contains all the validations that Caliban provides? Or perhaps rename `DefaultValidations` to `AllValidations` so that it's clear to the user that that this is a list of all the validations available?

core/src/main/scala/caliban/Configurator.scala
```diff
@@ -0,0 +1,47 @@
...
+import caliban.CalibanError.ValidationError
+import caliban.execution.QueryExecution
+import caliban.validation.Context
+import caliban.validation.Validator.DefaultValidations
+import zio._
+import zio.prelude.EReader
+
+object Configurator {
+  case class ExecutionConfiguration(
+    skipValidation: Boolean = false,
```
kyri-petrou, line 17: I feel like this might cause some confusion on how `skipValidation = true` differs from `validations = Nil`. I doubt we could get rid of `skipValidation` altogether, so I'm not sure what's the best way forward. Just a thought (probably not the best) - we could change `validations` into a `NonEmptyList` from prelude to perhaps prevent some confusions?

core/src/main/scala/caliban/Configurator.scala
```diff
@@ -0,0 +1,47 @@
...
+
+import caliban.CalibanError.ValidationError
+import caliban.execution.QueryExecution
+import caliban.validation.Context
+import caliban.validation.Validator.DefaultValidations
+import zio._
+import zio.prelude.EReader
+
+object Configurator {
+  case class ExecutionConfiguration(
```
kyri-petrou, line 16: I think we need to bring the deleted scaladoc from `GraphQL.execute` here to explain the different arguments

interop/tapir/src/main/scala/caliban/interop/tapir/HttpAdapter.scala
kyri-petrou: Is this layer meant for request interception alone? Perhaps we could name it `withRequestInterception`? If we use the term "configure" might create some confusion that it should be used with the `Configurator` case class somehow

interop/tapir/src/main/scala/caliban/interop/tapir/HttpAdapter.scala
```diff
@@ -0,0 +1,135 @@
...
+    HttpAdapter.Configured(self, configurator)
+
+  def configure(configurator: URIO[Scope, Unit]): HttpAdapter[R, E] =
```
kyri-petrou: Just for my own understanding, how is this meant to be used?

core/src/main/scala/caliban/GraphQL.scala
```diff
@@ -100,36 +100,34 @@ trait GraphQL[-R] { self =>
...
+          trace: Trace
+        ): URIO[R, GraphQLResponse[CalibanError]] =
           decompose(wrappers).flatMap {
             case (overallWrappers, parsingWrappers, validationWrappers, executionWrappers, fieldWrappers, _) =>
               wrap((request: GraphQLRequest) =>
                 (for {
                   doc              <- wrap(Parser.parseQuery)(parsingWrappers, request.query.getOrElse(""))
                   intro             = doc.isIntrospection
-                  _                <- ZIO.when(intro && !enableIntrospection) {
+                  config           <- Configurator.configuration
```
kyri-petrou: This feels a bit dangerous as we could accidentally use it further down in the pipeline after the underlying FiberRef has been modified. I'm not sure whether there's anything we could do about it though - I guess as long as we're aware of it should be fine

kyri-petrou (commented):
Looks great! Really excited for these changes!

Some very minor comments below, mostly around API semantics

interop/tapir/src/main/scala/caliban/interop/tapir/HttpAdapter.scala
paulpdaniels: @kyri-petrou it should now all be combined for efficiency. So I'd prefer to keep it as `configure` and people can use it with the configurator

ghostdogpr: How about the naming of the 2 variants though? Should I keep the overload or name it `configure`/`configureLayer`?

core/src/main/scala/caliban/Configurator.scala
ghostdogpr: It might be even more confusing if you provide `skipValidation = true` but you see you have to provide a non-empty list too. I will add a scaladoc with an explicit mention that `validations` are ignored when `skipValidation` is true.

core/src/main/scala/caliban/Configurator.scala
ghostdogpr: Added an alias `QueryValidation` with a comment explaining what it is.

ghostdogpr: Renamed to `AllValidations`

core/src/main/scala/caliban/Configurator.scala
ghostdogpr: Good point, done.

interop/tapir/src/main/scala/caliban/interop/tapir/HttpAdapter.scala
ghostdogpr: `adapter.configure(Configurator.setSkipValidationScoped(true))`

Using a `ZIO` is a little easier to write than a `ZLayer` when you don't use the incoming part of the environment.

core/src/main/scala/caliban/GraphQL.scala
ghostdogpr: Yeah I was careful to call it again in the validation and execution code since it can be modified by a wrapper.
I agree with you so I did this trick:
```scala
config                               <- Configurator.configuration.map { config =>
                                          (config.skipValidation, config.enableIntrospection)
                                        }
(skipValidation, enableIntrospection) = config
```

interop/tapir/src/main/scala/caliban/interop/tapir/HttpAdapter.scala
ghostdogpr: One idea: we could have this:
```scala
type RequestInterceptor[R1, R]: ZLayer[R1 & ServerRequest, TapirResponse, R]
type Configurator[R1]: URIO[R1 & Scope, Unit]

def intercept[R1](interceptor: RequestInterceptor[R1, R]): HttpAdapter[R1, E]
def configure[R1](configurator: Configurator[R1]): HttpAdapter[R & R1, E]
```

Even though under the hood we use the same thing. Since it's now optimized, you can even call them and the layers will be combined.

What do you think?

kyri-petrou: Personally, I really like it. I think it's a good middleground hiding implementation details from the users while retaining efficiency

core/src/main/scala/caliban/Configurator.scala
kyri-petrou: Sounds good!

interop/tapir/src/main/scala/caliban/interop/tapir/HttpAdapter.scala
paulpdaniels: Do we need the `ZIO` variant at all if we have `ZLayer`? My thought for how people would use the `ZIO` configurator variants would only be to roll their own logic there:

```scala
val myConfig = ZLayer.scoped(for {
  request  <- ZIO.service[ServerRequest]
  session  <- Auth.run(request)
  _        <- Configurator.setAllowIntrospectionScoped(true).when(session.user.isSuperAdmin)
  _.       <- Configuration.setSkipValidationScoped(true).when(session.isInternal)  
} yield session)
```

But I would think having one way to do things is easier (meaning the `Layer` since it has the most power

adapters/zio-http/src/main/scala/caliban/ZHttpAdapter.scala
```diff
@@ -1,47 +1,27 @@
...
-    enableIntrospection: Boolean = true,
     keepAliveTime: Option[Duration] = None,
-    queryExecution: QueryExecution = QueryExecution.Parallel,
     webSocketHooks: WebSocketHooks[R, E] = WebSocketHooks.empty
```
paulpdaniels: Could we add `WebSocketHooks` and `keepAliveTime` to the configurator too? Or would that be confusing you think?

examples/src/main/scala/example/akkahttp/AuthExampleApp.scala
```diff
@@ -4,13 +4,13 @@ import akka.actor.ActorSystem
 import akka.http.scaladsl.Http
 import akka.http.scaladsl.server.Directives.{ getFromResource, path, _ }
 import caliban._
-import caliban.interop.tapir.RequestInterceptor
+import caliban.interop.tapir.HttpAdapter
```
paulpdaniels: Question: We have some overloading of `*Adapter` now, we have `TapirAdapter` + `<webstack>Adapter` and now we have `HttpAdapter`, it seems like it might be slightly confusing. Should we look at renaming this or the various adapters?

interop/tapir/src/main/scala/caliban/interop/tapir/HttpUploadAdapter.scala
```diff
@@ -0,0 +1,132 @@
...
+import sttp.tapir.Codec.JsonCodec
+import sttp.tapir._
+import sttp.tapir.model.ServerRequest
+import sttp.tapir.server.ServerEndpoint
+import zio._
+
+import java.nio.charset.StandardCharsets
+import scala.util.Try
+
+sealed trait HttpUploadAdapter[-R, E] { self =>
```
paulpdaniels: Isn't `E` covariant? Not sure if it hurts anything for it not to be here.

interop/tapir/src/main/scala/caliban/interop/tapir/HttpAdapter.scala
ghostdogpr: The problem is that configure with a ZLayer doesn't work well when all you want is just skip validations. `Configurator.setSkipValidation` returns a layer of `Unit`, which doesn't match the signature of `configure` which expects a layer of `R` as a result. So you have to do some annoying conversion which is what the second `configure` allows you to avoid.

interop/tapir/src/main/scala/caliban/interop/tapir/HttpUploadAdapter.scala
ghostdogpr: I couldn't make this work because it's used in both covariant and contravariant inside the implementation. But I think it doesn't matter since it's used only in the constructor.

examples/src/main/scala/example/akkahttp/AuthExampleApp.scala
ghostdogpr: I agree. I'd rather keep `Adapter` for what it is currently to avoid more changes, how about `HttpInterpreter`, etc? It is effectively a wrapper around the initial `interpreter` with the protocol added on top.

ghostdogpr: Renamed

adapters/zio-http/src/main/scala/caliban/ZHttpAdapter.scala
ghostdogpr: I thought about it, but I kinda like to have them only for WebSockets, I don't really want to put them in `ExecutionConfiguration`.
It could be a separate case class and fiber ref, but there's also the problem that `WebSocketHooks` is parameterized in R and E.

ghostdogpr:
Resolved the zio-http websocket issue!

interop/tapir/src/main/scala/caliban/interop/tapir/HttpAdapter.scala
ghostdogpr: If I do my suggestion above (intercept/configure), I can delete the methods in Configurator that return a ZLayer and keep only the ones that return ZIO.

The flows would be:
- for simple configuration: use configure + Configurator.set (ZIO)
- for complex workflows: use intercept + ZLayer (you're free to use Configurator in there, or you can call configure separately)

I think it's pretty good like this!

kyri-petrou:
> http4s and zio-http auth examples use a middleware at the http level, not a request interceptor, should I change them

Hmm good question. To be honest I see value in keeping them; especially for http4s since users might want to use cats-effect to implement the authentication. As for zio-http, I think the options are relatively similar, but having auth as a middleware allows that to be reused in REST endpoints as well for cases that the server exposes both. Perhaps we can leave them as is and leave a comment for users to check out one of the other examples if they wish to implement auth via caliban?

ghostdogpr:
> > http4s and zio-http auth examples use a middleware at the http level, not a request interceptor, should I change them
> 
> 
> 
> Hmm good question. To be honest I see value in keeping them; especially for http4s since users might want to use cats-effect to implement the authentication. As for zio-http, I think the options are relatively similar, but having auth as a middleware allows that to be reused in REST endpoints as well for cases that the server exposes both. Perhaps we can leave them as is and leave a comment for users to check out one of the other examples if they wish to implement auth via caliban?

Yeah I'll leave as it is, we can point at the Akka/Play examples if needed. Good to have all cases in different examples.

interop/tapir/src/main/scala/caliban/interop/tapir/WebSocketInterpreter.scala
```diff
@@ -0,0 +1,123 @@
...
+      protocol: String
+    ): URIO[R, Either[TapirResponse, (String, CalibanPipe)]] =
+      Protocol
+        .fromName(protocol)
+        .make(interpreter, keepAliveTime, webSocketHooks)
+        .map(res => Right((protocol, res)))
+  }
+
+  private case class Configured[R1, R, E](
+    adapter: WebSocketInterpreter[R, E],
```
ghostdogpr: I need to rename these to interpreter

ghostdogpr: Done

adapters/zio-http/src/main/scala/caliban/ZHttpAdapter.scala
ghostdogpr: Solved

interop/tapir/src/main/scala/caliban/interop/tapir/HttpAdapter.scala
ghostdogpr: Done!

interop/tapir/src/main/scala/caliban/interop/tapir/WebSocketInterpreter.scala
```diff
@@ -0,0 +1,121 @@
...
+import sttp.capabilities.zio.ZioStreams
+import sttp.model.{ headers => _ }
+import sttp.tapir.Codec.JsonCodec
+import sttp.tapir._
+import sttp.tapir.model.{ ServerRequest, UnsupportedWebSocketFrameException }
+import sttp.tapir.server.ServerEndpoint
+import sttp.ws.WebSocketFrame
+import zio._
+
+trait WebSocketInterpreter[-R, E] { self =>
```
paulpdaniels: Should we seal it? Should users be able to make their own version of this?

ghostdogpr: Ah, I forgot. Fixed.

paulpdaniels (approved):

kyri-petrou (approved):
Looks great! Really looking forward to the next release!

PR #1712, opened 2023-05-01
fix defer streaming

paulpdaniels:
@kyri-petrou found a bug when using compression on zio-http. Initial investigation seems to indicate that using `peel` doesn't behave correctly with streaming under compression (not sure why). However, we can avoid the partial unpacking of the stream by using a `ZChannel` directly, this should also result in a more performant pipeline.

*Edit* 06/05/23

Found a subsequent issue related to how the streams are processed. In a nutshell, in order to properly declare the stream format we need to declare a `oneOf` variant that defines the stream io. However, this requires that we _know_ at the time of construction what type of stream we'll be using (which is more or less hidden from the HttpInterpreter), before we were casting somewhat blindly and this caused a `ClassCastException` deep within the tapir runtime. My solution was to introduce a type parameter when calling these methods which allows us to pass through the stream type at construction time. In order to do so we also required a new `StreamConstructor` type which is fed to the `Interpreter` to transform the `ZStream` into the desired stream type.

The advantage of this approach is that our streams are now self contained within a separate block, and we get to avoid the additional typecasting we did before because now all of the types correctly line up at compile time.

### DIFF
--- a/adapters/akka-http/src/main/scala/caliban/AkkaHttpAdapter.scala
+++ b/adapters/akka-http/src/main/scala/caliban/AkkaHttpAdapter.scala
@@ -6,11 +6,10 @@ import akka.stream.{ Materializer, OverflowStrategy }
 import akka.util.ByteString
 import caliban.AkkaHttpAdapter._
 import caliban.interop.tapir.TapirAdapter._
-import caliban.interop.tapir.{ HttpInterpreter, HttpUploadInterpreter, WebSocketInterpreter }
+import caliban.interop.tapir.{ HttpInterpreter, HttpUploadInterpreter, StreamConstructor, WebSocketInterpreter }
 import sttp.capabilities.WebSockets
 import sttp.capabilities.akka.AkkaStreams
 import sttp.capabilities.akka.AkkaStreams.Pipe
-import sttp.capabilities.zio.ZioStreams
 import sttp.model.{ MediaType, StatusCode }
 import sttp.tapir.Codec.JsonCodec
 import sttp.tapir.PublicEndpoint
@@ -29,7 +28,9 @@ class AkkaHttpAdapter private (private val options: AkkaHttpServerOptions)(impli
     interpreter: HttpInterpreter[R, E]
   )(implicit runtime: Runtime[R], materializer: Materializer): Route =
     akkaInterpreter.toRoute(
-      interpreter.serverEndpoints[R].map(convertHttpStreamingEndpoint[R, (GraphQLRequest, ServerRequest)])
+      interpreter
+        .serverEndpoints[R, AkkaStreams](AkkaStreams)
+        .map(convertHttpStreamingEndpoint[R, (GraphQLRequest, ServerRequest)])
     )
 
   def makeHttpUploadService[R, E](interpreter: HttpUploadInterpreter[R, E])(implicit
@@ -38,7 +39,7 @@ class AkkaHttpAdapter private (private val options: AkkaHttpServerOptions)(impli
     requestCodec: JsonCodec[GraphQLRequest],
     mapCodec: JsonCodec[Map[String, Seq[String]]]
   ): Route =
-    akkaInterpreter.toRoute(convertHttpStreamingEndpoint(interpreter.serverEndpoint[R]))
+    akkaInterpreter.toRoute(convertHttpStreamingEndpoint(interpreter.serverEndpoint[R, AkkaStreams](AkkaStreams)))
 
   def makeWebSocketService[R, E](
     interpreter: WebSocketInterpreter[R, E]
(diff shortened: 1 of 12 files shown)


interop/tapir/src/test/scala/caliban/interop/tapir/TestData.scala
```diff
@@ -22,13 +23,25 @@ object TestData {
...
   case class CharacterArgs(name: String)
 
   val sampleCharacters = List(
-    Character("James Holden", List("Jim", "Hoss"), EARTH, Some(Captain("Rocinante"))),
+    Character(
+      "James Holden",
+      List("Jim", "Hoss"),
+      EARTH,
+      Some(Captain("Rocinante")),
+      labels = ZIO.succeed(List("Captain"))
```
kyri-petrou, line 43: Is it possible to delay this effect and test that the initial payload arrives immediately when this is deferred or would that be too difficult?

interop/tapir/src/main/scala/caliban/interop/tapir/TapirAdapter.scala
```diff
@@ -309,22 +309,30 @@ object TapirAdapter {
...
-            .intersperse(InnerBoundary, InnerBoundary, EndBoundary)
+    val pipeline = ZPipeline.fromChannel {
+      lazy val reader: ZChannel[Any, Throwable, Chunk[ResponseValue], Any, Throwable, Chunk[ResponseValue], Any] =
+        ZChannel.readWithCause(
+          (in: Chunk[ResponseValue]) =>
+            in.headOption match {
+              case Some(value) =>
+                ZChannel.write(in.updated(0, resp.copy(data = value).toResponseValue)) *>
+                  ZChannel.identity[Throwable, Chunk[ResponseValue], Any]
+              case None        => reader
```
kyri-petrou, line 135: Question for my own understanding: under what conditions would we reach this case? Or is it technically unreachable?

kyri-petrou (commented):
LGTM, just a stretch comment below regarding testing delayed effects. Feel free to ignore it if it's too hard to test for though

interop/tapir/src/main/scala/caliban/interop/tapir/TapirAdapter.scala
paulpdaniels: If you received an empty chunk from upstream, you would just want to keep reading reading until you received the first value.

interop/tapir/src/test/scala/caliban/interop/tapir/TestData.scala
paulpdaniels: That might be slightly tricky so I'd say we could do that as a follow up. My thought is that we would need to record when we receive each event and verify that at least a certain amount of time has passed between events

adapters/akka-http/src/main/scala/caliban/AkkaHttpAdapter.scala
```diff
@@ -29,7 +28,9 @@ class AkkaHttpAdapter private (private val options: AkkaHttpServerOptions)(impli
     interpreter: HttpInterpreter[R, E]
   )(implicit runtime: Runtime[R], materializer: Materializer): Route =
     akkaInterpreter.toRoute(
-      interpreter.serverEndpoints[R].map(convertHttpStreamingEndpoint[R, (GraphQLRequest, ServerRequest)])
+      interpreter
+        .serverEndpoints[R, AkkaStreams](AkkaStreams)
```
paulpdaniels, line 32: I could probably go further an introduce a partially applied method here, but I don't think it is necessary since this really should only be used by people who know what they are doing

adapters/zio-http/src/main/scala/caliban/ZHttpAdapter.scala
```diff
@@ -17,7 +18,7 @@ object ZHttpAdapter {
   def makeHttpService[R, E](interpreter: HttpInterpreter[R, E])(implicit
     serverOptions: ZioHttpServerOptions[R] = ZioHttpServerOptions.default[R]
   ): HttpApp[R, Throwable] =
-    ZioHttpInterpreter(serverOptions).toHttp(interpreter.serverEndpoints[R])
+    ZioHttpInterpreter(serverOptions).toHttp(interpreter.serverEndpoints[R, ZioStreams](ZioStreams))
```
paulpdaniels, line 21: As usual `zio-http` gets the most succinct option 😸

adapters/zio-http/src/test/scala/caliban/ZHttpAdapterSpec.scala
```diff
@@ -54,7 +54,7 @@ object ZHttpAdapterSpec extends ZIOSpecDefault {
     suite.provideShared(
       apiLayer,
       Scope.default,
-      Server.defaultWithPort(8089)
+      Server.defaultWith(_.port(8089).responseCompression())
```
paulpdaniels, line 57: Turned this on because we saw different behavior vs with it off.

adapters/akka-http/src/test/scala/caliban/AkkaHttpAdapterSpec.scala
```diff
@@ -20,7 +20,9 @@ import sttp.tapir.json.play._
 import zio._
 import zio.test._
 
+import scala.concurrent.ExecutionContext.Implicits.global
 import scala.language.postfixOps
+import scala.util.Success
```
ghostdogpr: Clean imports?

ghostdogpr (approved):

adapters/akka-http/src/test/scala/caliban/AkkaHttpAdapterSpec.scala
paulpdaniels: Fixed!

kyri-petrou (approved):
Nice trick with the `StreamConstructor`! Looks good to me

PR #1719, opened 2023-05-07
Jsoniter codec micro-optimizations

kyri-petrou:
I realised I had these sitting in a branch for a while now. It shouldn't produce any notable performance improvement but I thought it'd be good to clean these up


plokhotnyuk:
I suggest moving the case of string decoding to be in the first place of the decoding match.

The number decoding could be before array and object decoding if the case expression will be simplified to `case x if x == '-' || (x >= '0' && x <= '9') =>`.

kyri-petrou:
Hi @plokhotnyuk, thanks for your feedback!

It's not shown in the PR change diff, but the decoders use a `@switch` annotation at the start of the matching, which means the pattern patching will be compiled to a `lookupswitch`. My understanding is that when a `lookupswitch` is used, which has an O(log(n)) complexity vs O(n) when using standard pattern matching, the ordering of the cases is irrelevant. Unfortunately we can't use the `@switch` annotation with expressions such as `case x if x == '-' || (x >= '0' && x <= '9')`.

Do you recon it's better to forgo the `@switch` annotation for having a reduced number of cases and also being able to control the order that the character is checked against?

plokhotnyuk:
The idea behind my proposition is to minimise number of branches for most frequently used values like strings and numbers. Having `lookupswitch` for more than 16 values leads to more than 4 checks for each JSON value type. Also, contemporary JVMs are able to simplify `x >= '0' && x <= '9'` to 1 add an 1 compare machine operations.

kyri-petrou:
I think that makes sense, thanks again for the feedback! I re-arranged some of the cases based on your suggestions

ghostdogpr (approved):

PR #1723, opened 2023-05-12
Fix return type of ZHttpAdapter.makeHttpService

ghostdogpr:
`makeHttpService` was returning `HttpApp[R, Throwable]` while `makeWebSocketService` returned `HttpApp[R, Response]` aka `App[R]`. Considering `serve` expects an `App[R]`, it would be better to fix `makeHttpService`.

### DIFF
--- a/adapters/zio-http/src/main/scala/caliban/ZHttpAdapter.scala
+++ b/adapters/zio-http/src/main/scala/caliban/ZHttpAdapter.scala
@@ -17,8 +17,10 @@ object ZHttpAdapter {
 
   def makeHttpService[R, E](interpreter: HttpInterpreter[R, E])(implicit
     serverOptions: ZioHttpServerOptions[R] = ZioHttpServerOptions.default[R]
-  ): HttpApp[R, Throwable] =
-    ZioHttpInterpreter(serverOptions).toHttp(interpreter.serverEndpoints[R, ZioStreams](ZioStreams))
+  ): App[R] =
+    ZioHttpInterpreter(serverOptions)
+      .toHttp(interpreter.serverEndpoints[R, ZioStreams](ZioStreams))
+      .withDefaultErrorResponse
 
   def makeWebSocketService[R, E](
     interpreter: WebSocketInterpreter[R, E]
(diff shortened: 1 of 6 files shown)


kyri-petrou:
Isn't this technically a breaking change? I'm surprised the MiMa checks didn't pick it up.

ghostdogpr:
> Isn't this technically a breaking change? I'm surprised the MiMa checks didn't pick it up.

Yeah i was expecting MiMa to break. But I think it's reasonable to release this as 2.2.1 (it's kind of a bug actually).

kyri-petrou (approved):

adapters/zio-http/src/main/scala/caliban/ZHttpAdapter.scala
```diff
@@ -17,8 +17,10 @@ object ZHttpAdapter {
 
   def makeHttpService[R, E](interpreter: HttpInterpreter[R, E])(implicit
     serverOptions: ZioHttpServerOptions[R] = ZioHttpServerOptions.default[R]
-  ): HttpApp[R, Throwable] =
-    ZioHttpInterpreter(serverOptions).toHttp(interpreter.serverEndpoints[R, ZioStreams](ZioStreams))
+  ): App[R] =
+    ZioHttpInterpreter(serverOptions)
+      .toHttp(interpreter.serverEndpoints[R, ZioStreams](ZioStreams))
+      .withDefaultErrorResponse
```
paulpdaniels, line 23: The only thing I don't like about this is that it throws away all information regarding the error so the user can't choose to take a different action on it

ghostdogpr: True, but this is a one-liner they can easily copy. Otherwise it forces them to add this handler themselves which is more annoying.

ghostdogpr: Also it seems the tapir code already convert errors like Bad Request into a Response, so it will throw away only the unhandled errors which should be very rare (since Caliban also catches errors on its own).

paulpdaniels (approved):

PR #1725, opened 2023-05-22
Cache creation of interpreter

kyri-petrou:
With this PR, the validation of the root schema returns an `Either` instead of an `IO`. This in turn allows us to cache the creation of the interpreter, to avoid accidentally recreating it per-request which would be very detrimental to performance.

For example, the code below would previously recreate the interpreter on each request, whereas with the changes in this PR the interpreter would be created only on the first request, and then be memoized.

```scala
    val apiRoutes = (??? : GraphQL[Any]).interpreter.map { interpreter =>
      ZHttpAdapter.makeHttpService(HttpInterpreter(interpreter))
    }
    
    Http.fromHttpZIO[Request] {
      case (GET | POST) -> !! / "api" / "graphql" => apiRoutes
      case _ => Http.empty
    }
```
At the moment, `cachedInterpreter` is private, but we could potentially make it public as well. Note that some of the methods (see comments) are no longer used, but kept for binary and source compatibility

### DIFF
--- a/core/src/main/scala/caliban/GraphQL.scala
+++ b/core/src/main/scala/caliban/GraphQL.scala
@@ -28,7 +28,10 @@ trait GraphQL[-R] { self =>
   protected val features: Set[Feature]
 
   private[caliban] def validateRootSchema(implicit trace: Trace): IO[ValidationError, RootSchema[R]] =
-    Validator.validateSchema(schemaBuilder)
+    ZIO.fromEither(validatedRootSchemaCached)
+
+  private lazy val validatedRootSchemaCached: Either[ValidationError, RootSchema[R]] =
+    Validator.validateSchemaEither(schemaBuilder)
 
   /**
    * Returns a string that renders the API types into the GraphQL format.
@@ -77,8 +80,8 @@ trait GraphQL[-R] { self =>
    * adding some middleware around the query execution.
    * Fails with a [[caliban.CalibanError.ValidationError]] if the schema is invalid.
    */
-  final def interpreter(implicit trace: Trace): IO[ValidationError, GraphQLInterpreter[R, CalibanError]] =
-    validateRootSchema.map { schema =>
+  final def interpreter(implicit trace: Trace): IO[ValidationError, GraphQLInterpreter[R, CalibanError]] = {
+    ZIO.fromEither(validatedRootSchemaCached).map { schema =>
       lazy val rootType =
         RootType(
           schema.query.opType,
@@ -151,6 +154,7 @@ trait GraphQL[-R] { self =>
           }
       }
     }
+  }
 
   /**
    * Attaches a function that will wrap one of the stages of query processing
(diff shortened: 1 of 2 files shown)


core/src/main/scala/caliban/validation/Validator.scala
```diff
@@ -52,12 +52,17 @@ object Validator {
    * Verifies that the given schema is valid. Fails with a [[caliban.CalibanError.ValidationError]] otherwise.
    */
   def validateSchema[R](schema: RootSchemaBuilder[R]): IO[ValidationError, RootSchema[R]] = {
```
kyri-petrou, line 54: this seems to be unused now, so we could mark it as `@deprecated`

core/src/main/scala/caliban/GraphQL.scala
```diff
@@ -78,7 +78,10 @@ trait GraphQL[-R] { self =>
    * Fails with a [[caliban.CalibanError.ValidationError]] if the schema is invalid.
    */
   final def interpreter(implicit trace: Trace): IO[ValidationError, GraphQLInterpreter[R, CalibanError]] =
-    validateRootSchema.map { schema =>
+    ZIO.fromEither(cachedInterpreter)
```
kyri-petrou: I'm thinking whether this should initialise the interpreter upfront or lazily, e.g.,

```scala
val i = cachedInterpreter
ZIO.fromEither(i)
```

Thoughts?

ghostdogpr: Makes sense yeah

core/src/main/scala/caliban/validation/Validator.scala
ghostdogpr: I would keep that, just in case it's useful for some

ghostdogpr (approved):

ghostdogpr (approved):
Feel free to merge 👍

core/src/main/scala/caliban/GraphQL.scala
```diff
@@ -77,8 +77,13 @@ trait GraphQL[-R] { self =>
    * adding some middleware around the query execution.
    * Fails with a [[caliban.CalibanError.ValidationError]] if the schema is invalid.
    */
-  final def interpreter(implicit trace: Trace): IO[ValidationError, GraphQLInterpreter[R, CalibanError]] =
-    validateRootSchema.map { schema =>
+  final def interpreter(implicit trace: Trace): IO[ValidationError, GraphQLInterpreter[R, CalibanError]] = {
+    val i = cachedInterpreter
+    ZIO.fromEither(i)
```
paulpdaniels, line 82: Was not in lining this deliberate ?

kyri-petrou: Initially in the PR it wasn't inlined; but then I was on the fence whether we should lazily initialise the interpreter when the method is invoked or not (see [here](https://github.com/ghostdogpr/caliban/pull/1725/files/f3df7371f233a54494a6453721cfc58a5d49f7af#r1201307120))

PR #1731, opened 2023-05-24
Validate names in validateSchema

### LINKED ISSUE #1729: GraphQL.render does not validate type and field names
For example, for a type defined like
```scala
@GQLName("Foo { str: String }\ntype Bar")
case class Foo(
    @GQLName("bool: Boolean\n  num")
    num: Int
)
```
the renderer simply emits
```graphql
type Foo { str: String }
type Bar {
  bool: Boolean
  num: Int!
}
```
without any concern.

Pretty much same effect can be achieved by naming classes and fields directly via backticks:
```scala
case class `Foo { str: String }\ntype Bar`(
  `bool: Boolean\n  num`: Int
)
```

ghostdogpr:
Closes #1729

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -123,6 +123,12 @@ lazy val core = project
   .settings(commonSettings)
   .settings(enableMimaSettingsJVM)
   .settings(
+    mimaBinaryIssueFilters ++= Seq(
+      ProblemFilters.exclude[IncompatibleMethTypeProblem]("caliban.validation.Validator.doesNotStartWithUnderscore"),
+      ProblemFilters.exclude[DirectMissingMethodProblem]("caliban.validation.Validator.doesNotStartWithUnderscore")
+    )
+  )
+  .settings(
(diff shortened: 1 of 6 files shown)


ghostdogpr:
@kyri-petrou any idea why Mima is complaining about private methods?

frekw (approved):
Nice!

kyri-petrou:
> @kyri-petrou any idea why Mima is complaining about private methods?

My guess is that it's because they're `package` private, which means that users could have namespaced the file under `package caliban` to use them.

Another reason I guess might be in cases that a user is using `caliban-core` 2.2.2 and `caliban-<something>` at 2.2.1 pulled from different dependencies

Given that there are almost no other libraries that depend on caliban directly to create such mismatch, and that end-users namespacing code under `package caliban` are likely not publishing artifacts, I think we're safe to add these as ignored in mima.

ghostdogpr:
Adding an exclusion for mima

PR #1732, opened 2023-05-24
Make support for isRepeatable optional

### LINKED ISSUE #1726: Code generation plugin fails when generating schema for older Caliban servers
Code generation plugin gets a `Cannot query field 'isRepeatable' on type '__Directive'` response when trying to generate schema for an older server. Tested with Caliban client 2.2.1 and server 1.2.1. Issue does not seem to be present on clients earlier than 2.1.0.

ghostdogpr:
Closes #1726

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -145,8 +145,8 @@ jobs:
       - checkout
       - restore_cache:
           key: sbtcache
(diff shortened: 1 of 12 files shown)


ghostdogpr:
Hmm I think we may exclude `tools` from Mima @kyri-petrou ?

kyri-petrou:
> Hmm I think we may exclude `tools` from Mima @kyri-petrou ?

I think it's a bit tricky. My gut says it should be fine, but we might need to do it on a case-by-case basis given that the `tracing` module lightly depends on it.

Also, the other part that worries me is that the codegen plugin uses it. In some rare cases where code is generated by a different service / pipeline than the service using the generated code, this (might?) cause issues?

ghostdogpr:
That code is used only during the code generation, not by the generated code so I think we should be fine.

kyri-petrou (approved):
LGTM!

PR #1734, opened 2023-05-26
Modify rendering spec to parse the rendered schema and compare definitions

### LINKED ISSUE #1637: Fix RenderingSpec

nox213:
Closes #1637
Here's a rough idea
 - First render API
 - Parse rendered API
 - Compare parsed API with original API

To do this, we need to fix `index` of `Directive` because values are set differently when we parse rendered API.
`fixDefinitions` is a helper function for that work.


ghostdogpr (approved):
Awesome!

Issue #138, opened 2019-12-26
It seems that some data types in Cats(OptionT, fs2.Stream) is not support?


UkonnRa:
Say having such the Queries:

```scala
final case class Queries[F[_]: Effect]( // Or any other traits?
    users: F[List[User]],
    userById: UserId => F[Option[User]],
    userByEmail: Email => F[Option[User]]
)

final case class PrettierQueries[F[_]: Effect]( // Or any other traits?
    users: fs2.Stream[F, User],
    userById: UserId => OptionT[F, User],
    userByEmail: Email => OptionT[F, User]
)
```

Neither of them can pass the compile. Just throw the errors:
```
Cannot find a Schema for type com.ukonnra.wonderland.user.endpoint.graphql.Queries[F].
[error]      Caliban derives a Schema automatically for basic Scala types, case classes and sealed traits, but
[error]      you need to manually provide an implicit Schema for other types that could be nested in endpoint.graphql.Queries[F].
[error]      If you use a custom type as an argument, you also need to provide an implicit ArgBuilder for that type.
[error]      See https://ghostdogpr.github.io/caliban/docs/schema.html for more information.
[error]   val interpreter = GraphQL.graphQL(RootResolver(queries)).interpreter
```

ghostdogpr:
Did you add the caliban-cats module and the import mentioned in the [interop docs](https://ghostdogpr.github.io/caliban/docs/interop.html)? I think the first example should work as long as it’s able to derive a schema for `User`. If not, you can try to call `gen` directly and Magnolia might tell you what’s missing. 

For the second one, you need a `Schema` instance for `OptionT` and fs2 `Stream`. This would be a nice addition to the caliban-cats module if you manage to do it. 

I’m away from computer for the holiday until January 6. When I come back I’ll be able to test this code and help you more if needed 😀

UkonnRa:
Thanks for quick response, I just wrote the fs2.Stream and OptionT Schemas:
```scala
  implicit def optionTSchema[F[_]: Effect, R, A](implicit ev: Schema[R, Option[A]]): Schema[R, OptionT[F, A]] =
    new Schema[R, OptionT[F, A]] {
      override def toType(isInput: Boolean): __Type =
        ev.toType(isInput)

      override def optional: Boolean =
        ev.optional

      override def resolve(value: OptionT[F, A]): Step[R] =
        QueryStep(ZQuery.fromEffect(value.value.toIO.to[Task].map(ev.resolve)))
    }

[...]

PR #1737, opened 2023-05-28
FS2 Stream interop

satorg:
Enables `Schema` derivation for the `fs2.Stream` type.
See also #138.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -11,6 +11,7 @@ val akkaVersion               = "2.6.20"
 val catsEffect3Version        = "3.5.0"
 val catsMtlVersion            = "1.3.0"
 val circeVersion              = "0.14.5"
+val fs2Version                = "3.7.0"
 val http4sVersion             = "0.23.19"
 val javaTimeVersion           = "2.5.0"
 val jsoniterVersion           = "2.23.1"
@@ -263,6 +264,7 @@ lazy val catsInterop = project
       else Seq(compilerPlugin(("org.typelevel" %% "kind-projector" % "0.13.2").cross(CrossVersion.full)))
     } ++ Seq(
       "org.typelevel" %% "cats-effect"      % catsEffect3Version,
+      "co.fs2"        %% "fs2-core"         % fs2Version,
       "dev.zio"       %% "zio-interop-cats" % zioInteropCats3Version,
       "dev.zio"       %% "zio-test"         % zioVersion % Test,
       "dev.zio"       %% "zio-test-sbt"     % zioVersion % Test
--- a/examples/src/main/scala/example/interop/cats/ExampleCatsInterop.scala
+++ b/examples/src/main/scala/example/interop/cats/ExampleCatsInterop.scala
@@ -1,47 +1,93 @@
 package example.interop.cats
 
+import caliban.ResponseValue.{ ObjectValue, StreamValue }
 import caliban._
+import caliban.interop.cats.ToEffect
+import caliban.interop.cats.implicits._
+import caliban.interop.fs2.implicits._
 import caliban.schema.Schema.auto._
+import cats.effect.std.{ Console, Dispatcher, Random }
+import cats.effect.syntax.all._
+import cats.effect.{ ExitCode, IO, IOApp, LiftIO }
+import fs2.Stream
+import zio.ZIO
+import zio.interop.catz._
+import zio.stream.ZStream
+import zio.stream.interop.fs2z._
 
-import cats.effect.{ ExitCode, IO, IOApp }
-import cats.effect.std.Dispatcher
-import zio.Runtime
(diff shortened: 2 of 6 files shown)


build.sbt
```diff
@@ -263,6 +264,7 @@ lazy val catsInterop = project
       else Seq(compilerPlugin(("org.typelevel" %% "kind-projector" % "0.13.2").cross(CrossVersion.full)))
     } ++ Seq(
       "org.typelevel" %% "cats-effect"      % catsEffect3Version,
+      "co.fs2"        %% "fs2-core"         % fs2Version,
```
satorg, line 267: This dependency is made `Optional` in the `zio-interop-cats` module. Should we make it optional here too?

ghostdogpr: Yeha we could do that.

interop/cats/src/main/scala/caliban/interop/fs2/Fs2Interop.scala
```diff
@@ -0,0 +1,38 @@
+package caliban.interop.fs2
+
+import caliban.interop.cats.FromEffect
```
ghostdogpr, line 3: Can you organize imports everywhere?

ghostdogpr (commented):
Excellent! Can you also add an example to `ExampleCatsInterop`? You can take a look at the code in `ExampleMonixInterop` (which is commented out because Monix is not compatible with CE3), there is an example of Subscription with `Observable` which you can adapt to fs2. Thanks!

build.sbt
kyri-petrou: To be honest I think we're okay with making this a required dependency. Http4s depends on fs2 as well, and the likelihood of a cats-effect app not having dependency on fs2 is very very low

interop/cats/src/main/scala/caliban/interop/fs2/Fs2Interop.scala
satorg: Actually, I did organize the imports. I use VSCode+Metals and this is how their function "Organize imports" works.

Would you prefer to have them organized in another style? Like folded one or something? I could either try to find out if it is possible to configure VSCode that way or even re-organize them by hand, if it'd make sense.

ghostdogpr: Ah yeah I use the IntelliJ style which is group items together and use `_` if there are more than 5 of them. You can see how it is in other files. Now sure if Metals lets you do that easily 😄

satorg: @ghostdogpr, seems I figured out how to configure the "organize imports" action in Metals. And I managed to set it to an IntelliJ-like style, check it out please.

Also I added a `.scalafix.conf` to the root of the repo which does the trick. So that if someone else also would be using Metals, they could get the IntelliJ style in there without any hustle. But if you don't like that change – let me know please, I am up to roll it back.

ghostdogpr: I think that's ok 👍

satorg:
@ghostdogpr ,
> Can you also add an example to `ExampleCatsInterop`? 

Done, check it out please.

examples/src/main/scala/example/interop/cats/ExampleCatsInterop.scala
```diff
@@ -1,47 +1,93 @@
...
+  case class Queries(numbers: List[Number], randomNumber: IO[Number], moreRandomNumbers: Stream[IO, Number])
 
-  val numbers      = List(1, 2, 3, 4).map(Number)
-  val randomNumber = IO(scala.util.Random.nextInt()).map(Number)
+  case class Subscriptions(numbers: Stream[IO, Number])
 
-  val queries = Queries(numbers, randomNumber)
+  val numbers             = List(1, 2, 3, 4).map(Number)
+  val randomNumber        = IO(scala.util.Random.nextInt()).map(Number)
+  val moreRandomNumbers   = Stream.repeatEval(random.nextInt).take(5).map(Number)
```
ghostdogpr: How about using the same random function in both cases?

ghostdogpr (commented):
Looks great, just a small suggestion for simplification!

examples/src/main/scala/example/interop/cats/ExampleCatsInterop.scala
satorg: Oh.. Sure, will change. I overlooked it. Thank you.

satorg: @ghostdogpr fixed, take another look please. thx.

ghostdogpr (approved):
Thanks!

PR #1742, opened 2023-05-30
Render schema directives even if no queries/mutations/subscriptions

yarian:
Per discord chat. The current behavior is problematic for Apollo federation subgraphs that only have entity resolvers. This changes rendering to include an extend schema statement.

Rendering tests seem commented out so didn't mess with that but tested with a local example I was running into and it works as expected.

### DIFF
--- a/core/src/main/scala/caliban/GraphQL.scala
+++ b/core/src/main/scala/caliban/GraphQL.scala
@@ -43,7 +43,7 @@ trait GraphQL[-R] { self =>
     val flattenedParts   = parts.flatten.mkString("\n")
     val schema           = (flattenedParts, schemaDirectives) match {
       case ("", "")                      => ""
-      case ("", schemaDirectives)        => s"extend schema $schemaDirectives\n"
+      case ("", schemaDirectives)        => s"extend schema $schemaDirectives"
       case (something, schemaDirectives) => s"""schema $schemaDirectives{
                                                |$something
                                                |}""".stripMargin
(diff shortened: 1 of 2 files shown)


core/src/main/scala/caliban/GraphQL.scala
```diff
@@ -42,7 +42,7 @@ trait GraphQL[-R] { self =>
     val schemaDirectives = renderSchemaDirectives(schemaBuilder.schemaDirectives)
 
     val schema = parts.flatten.mkString("\n") match {
-      case ""        => ""
+      case ""        => s"extend schema $schemaDirectives\n"
```
yarian: Let me know if you rather double `\n`.

ghostdogpr: Maybe only do this when schema directives are not empty?

ghostdogpr:
Actually rendering tests were fixed in https://github.com/ghostdogpr/caliban/pull/1734 last Friday so you could add a test 😄 I think `fixDefinitions` will need to be fixed in the test because it's ignoring extensions at the moment.

core/src/main/scala/caliban/GraphQL.scala
```diff
@@ -40,12 +40,13 @@ trait GraphQL[-R] { self =>
...
-
-    val schema = parts.flatten.mkString("\n") match {
-      case ""        => ""
-      case something => s"""schema $schemaDirectives{
-                           |$something
-                           |}""".stripMargin
+    val flattenedParts   = parts.flatten.mkString("\n")
+    val schema           = (flattenedParts, schemaDirectives) match {
+      case ("", "")                      => ""
+      case ("", schemaDirectives)        => s"extend schema $schemaDirectives\n"
```
ghostdogpr: I think the \n is not needed

ghostdogpr (commented):
Just a minor comment, LGTM!

PR #1744, opened 2023-05-31
Update zio-http to 3.0.0-RC2

kyri-petrou:
zio-http 3.0.0-RC2 brought websocket refactoring, which now requires subprotocols to be defined in the server config. Since we don't have any control over that, my guess is that the best approach would be to highlight this required change in the next release notes.


adapters/zio-http/src/main/scala/caliban/ZHttpAdapter.scala
```diff
@@ -50,11 +48,14 @@ object ZHttpAdapter {
...
-                       case ChannelEvent(ch, UserEventTriggered(HandshakeComplete)) =>
-                         out.runForeach(ch.writeAndFlush(_)).race(ch.awaitClose)
-                       case ChannelEvent(_, ChannelRead(WebSocketFrame.Text(text))) =>
+          socket = Handler.webSocket { ch =>
+                     ch.receiveAll {
+                       case ChannelEvent.UserEventTriggered(UserEvent.HandshakeComplete) =>
+                         out
+                           .interruptWhen(ch.awaitShutdown)
+                           .runForeach(msg => ch.send(ChannelEvent.Read(msg)))
+                           .forkDaemon
```
kyri-petrou, line 62: I could only get this to work by forking this effect. Not sure if this is the right approach, any thoughts @frekw?

adapters/zio-http/src/test/scala/caliban/ZHttpAdapterSpec.scala
```diff
@@ -53,7 +54,14 @@ object ZHttpAdapterSpec extends ZIOSpecDefault {
     suite.provideShared(
       apiLayer,
       Scope.default,
-      Server.defaultWith(_.port(8089).responseCompression())
+      Server.defaultWith(
+        _.port(8089)
+          .withWebSocketConfig(
```
ghostdogpr: Well that sucks. It doesn't work without it, right?

kyri-petrou: Nope 😞 I wonder whether we can create a default websocket config in the `adapters/zio-http` module and advice users that they need to use it.

By the way I think this might be a good thing for users to be able to customise. In the past I wanted to disable legacy WS protocol support in a service, and I had to write a custom middleware to do so. Being able to do that via the server config can be handy. Although it's a bit annoying that we can't provide a default without users making changes to their code

ghostdogpr: Yeah it would have been nice if it worked by default and then we could override it by config.

adapters/zio-http/src/main/scala/caliban/ZHttpAdapter.scala
frekw: I haven't had the time to dive into the new WebSocket APIs yet unfortunately :/

But I agree that this `forkDaemon` feels off. In general I'm not super fond of our setup where we create a separate queue for responses. I wonder if that contributes to the weirdness in this case, that we are adding more abstraction on top of zio-http's handling? 🤔

adapters/zio-http/src/test/scala/caliban/ZHttpAdapterSpec.scala
kyri-petrou: Yeah that would have been nice indeed. As a middleground, I added a default websocket config to the ZHttpAdapter which I hope it'll improve UX a bit

ghostdogpr: What's your opinion on this btw https://discord.com/channels/629491597070827530/633200096393166868/1113331097120555029

kyri-petrou:
@ghostdogpr @frekw @paulpdaniels After our discussions on Discord, I'm going to mark this PR as draft and try and get some clarifications on the following:

- [ ] What's the (new) recommended approach for using streams
- [ ] Whether we can work out something so that we can dynamically set the WS subprotocol to avoid having the user define them as per our case
- [ ] What other major API changes are planned prior to the stable release

Once we have more clarity on these we can decide what to do with this PR. What do you think?

ghostdogpr:
Sounds good! Regarding the protocol, I'm also thinking that ultimately we want WebSockets to use tapir and I think with tapir we are able to define the protocol when we define the endpoint. So I wonder if they'll be able to support zio-http with the way it's configured.

frekw:
@ghostdogpr that's an excellent point!

I was hoping that we'd be able to start working on adding zio-http WebSocket support to Tapir to be able to decrease the number of WebSocket implementations, so that's definitely in our interest as well.

kyri-petrou:
I'm closing this since we'll definitely be skipping RC2, and RC3 which also contains a lot of breaking changes by the look of it so nothing in this PR is really relevant / salvageable

PR #1758, opened 2023-06-09
Add derives Schema.SemiAuto in schema code generation

### LINKED ISSUE #1667: Add derives Schema.SemiAuto in schema code generation
The only tricky part is that it should only be there for Scala 3. If we can't detect it automatically, we could make this behavior enabled by a parameter.

nox213:
closes #1667

-Add `addDerives` flag to toggle this feature
-Change`scalafmt - use runner.dialect to use scala3` because `derive` keyword is only supported from scala3. And `schemaWriterSpec` run scalafmt before comparing results.

### DIFF
--- a/.scalafmt-for-test.conf
+++ b/.scalafmt-for-test.conf
@@ -1,6 +1,6 @@
 version = "3.1.2"
 
-runner.dialect = scala213
+runner.dialect = scala3
 maxColumn = 120
 align.preset = most
 continuationIndent.defnSite = 2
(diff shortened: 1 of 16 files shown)


ghostdogpr:
Hmm there are fmt errors because of the scala3 dialect 🤔

ghostdogpr:
Instead of changing the project .scalafmt.conf, in `SchemaWriterSpec` you can pass a path to `Formatter.format` that points to another .scalafmt.conf file, that can be used for the test only.

nox213:
I fixed the problem. Thank you for your review. @ghostdogpr

ghostdogpr (approved):
Great, thanks!

PR #1803, opened 2023-07-20
Support interfaces implementing interfaces

doohochang:
* make `Parsers.interfaceTypeDefinition` to support interfaces implementing interfaces.
* add tests into `ParserSpec`, `ClientWriterSpec`.


ghostdogpr:
Looks great!
Mima is failing but I think it's fine to ignore those since those classes should not be used much by library users. Can you add exclusions like this: https://github.com/ghostdogpr/caliban/blob/660962e53d2121f9306dc1ace6aaeb67b166cf3c/build.sbt#L127

ghostdogpr (approved):
Thanks!

PR #1813, opened 2023-07-26
Improved request interception semantics and `HttpAdapter` configuration performance

kyri-petrou:
This PR:

1. Adds an `interceptWithZIO` method on `HttpAdapter` which can be used to intercept requests and apply some logic without erasing part of the environment.
2. Improves performance for configured `HttpAdapter`s by avoiding modification of the ZIO environment (see [Discord post](https://discord.com/channels/629491597070827530/633200096393166868/1133309512284114965) for more info)

⚠️ Bincompat Breaking change ⚠️

To provide a unified `configure` implementation, the method now requires an implicit `Tag[R1]`. This is not necessary if we wish to have different `configure` implementations across `Base` and `Intercepted`. Let me know whether you think it's better to avoid the breaking change.

On this note, I think we might want to also add an implicit `Tag[R1]` to `intercept` since it seems that different layer composition methods require one, which we might need to use in the future

## Todo

- [ ] Add tests for `interceptWithZIO`

### DIFF
--- a/interop/tapir/src/main/scala/caliban/interop/tapir/HttpInterpreter.scala
+++ b/interop/tapir/src/main/scala/caliban/interop/tapir/HttpInterpreter.scala
@@ -10,6 +10,7 @@ import sttp.tapir._
 import zio._
 
 sealed trait HttpInterpreter[-R, E] { self =>
+
   protected def endpoints[S](streams: Streams[S]): List[
     PublicEndpoint[(GraphQLRequest, ServerRequest), TapirResponse, CalibanResponse[streams.BinaryStream], S]
   ]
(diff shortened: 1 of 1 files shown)


interop/tapir/src/main/scala/caliban/interop/tapir/HttpInterpreter.scala
```diff
@@ -52,15 +64,43 @@ object HttpInterpreter {
...
+      interceptor: ServerRequest => ZIO[R2, TapirResponse, Unit]
+    )(implicit tag: Tag[R2]): HttpInterpreter[R1 & R2, E] =
+      copy[R1 & R2, R & R2, E](layer =
+        layer ++ ZLayer.environment[R2] ++ ZLayer(ZIO.serviceWithZIO[ServerRequest](interceptor))
+      )
+
+    def configure[R2](configurator: Configurator[R2])(implicit tag: Tag[R2]): HttpInterpreter[R1 & R2, E] =
+      Intercepted[R1 & R2, R & R2, E](
+        interpreter.configure[R2](configurator),
+        layer ++ ZLayer.environment[R2]
```
ghostdogpr, line 110: Using `ZLayer.makeSome` doesn't require a `Tag` and is faster, I think?

interop/tapir/src/main/scala/caliban/interop/tapir/HttpInterpreter.scala
```diff
@@ -52,15 +64,43 @@ object HttpInterpreter {
       graphQLRequest: GraphQLRequest,
       serverRequest: ServerRequest
     )(implicit streamConstructor: StreamConstructor[BS]): ZIO[R, TapirResponse, CalibanResponse[BS]] =
-      interpreter.executeRequest(graphQLRequest).map(buildHttpResponse[E, BS])
+      ZIO.scoped[R](config *> interpreter.executeRequest(graphQLRequest).map(buildHttpResponse[E, BS]))
```
ghostdogpr: Seems pity to add overhead on the simple case.

interop/tapir/src/main/scala/caliban/interop/tapir/HttpInterpreter.scala
kyri-petrou: I’ve tried everything I could think of to get rid of the `Tag` / `++` and I can’t seem to get it to work (same for `configure`). 

When using `ZLayer.makeSome` alongside `ZLayer.environment`, a circular dependency compilation error is raised if the layer output contains a union type.

ghostdogpr:
I'm not so keen on making the API more complex 🤔 How about something like this instead? https://github.com/ghostdogpr/caliban/pull/1816

PR #1815, opened 2023-07-27
Exclude fields with different arguments from interfaces

kyri-petrou:
There seems to be a bug with the schema generation for interfaces; When fields have the same underlying type, the arguments are not considered. For example, with the code below we would include the `different` field in the definition of the `MyInterface` interface.

With the changes in this PR we ensure that the arguments of `__Field` are identical when merging the types

```scala
  @GQLInterface
  sealed trait MyInterface

  object MyInterface   {
    case class A(common: String, different: String)           extends MyInterface
    case class B(common: String, different: String => String) extends MyInterface
  }
```

@ghostdogpr the `Types.unify` method is not used anywhere anymore. Is it possible that users are using this method? I haven't removed it just to maintain bin compatibility


ghostdogpr (approved):
I guess we could delete `unify`. There are already a bunch of breaking changes in the upcoming release. And this one is actually buggy.

ghostdogpr (approved):

PR #1816, opened 2023-07-28
Allow access ServerRequest in configure

ghostdogpr:


kyri-petrou (approved):
Given the simplicity of this approach, I also think we should go with it even if it doesn’t solve the performance overhead issue of modifying the zio environment for cases we want to configure the interpreter but don’t need to access the `ServerRequest`

paulpdaniels (approved):

PR #1830, opened 2023-08-10
Optimize Scala3's codegen

kyri-petrou:
Seems that we were still creating a bit too much redundant code during Schema derivation in Scala 3.

With these changes, we avoid creating any methods / lambdas during while we're inlining to reduce the size of the generated code. Tested the changes by compiling an API at work which has a medium-sized schema and the generated code size was down by ~10%

I also tried to cleanup a bit of the derived code in ArgBuilder's derivation (see last commit). Let me know if you prefer I rollback those changes

### DIFF
--- a/core/src/main/scala-3/caliban/schema/ArgBuilderDerivation.scala
+++ b/core/src/main/scala-3/caliban/schema/ArgBuilderDerivation.scala
@@ -30,60 +30,66 @@ trait CommonArgBuilderDerivation {
   inline def derived[A]: ArgBuilder[A] =
     inline summonInline[Mirror.Of[A]] match {
       case m: Mirror.SumOf[A] =>
-        lazy val subTypes   = recurse[m.MirroredElemLabels, m.MirroredElemTypes]()
-        lazy val traitLabel = constValue[m.MirroredLabel]
-        new ArgBuilder[A] {
-          def build(input: InputValue): Either[ExecutionError, A] =
-            buildSum[A](subTypes, traitLabel)(input)
-        }
+        makeSumArgBuilder[A](
+          recurse[m.MirroredElemLabels, m.MirroredElemTypes](),
+          constValue[m.MirroredLabel]
+        )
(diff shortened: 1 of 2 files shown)


core/src/main/scala-3/caliban/schema/ArgBuilderDerivation.scala
```diff
@@ -30,60 +30,66 @@ trait CommonArgBuilderDerivation {
   inline def derived[A]: ArgBuilder[A] =
     inline summonInline[Mirror.Of[A]] match {
       case m: Mirror.SumOf[A] =>
-        lazy val subTypes   = recurse[m.MirroredElemLabels, m.MirroredElemTypes]()
-        lazy val traitLabel = constValue[m.MirroredLabel]
-        new ArgBuilder[A] {
```
kyri-petrou, line 35: Note to self: Never call `new ...` on traits / abstract classes inside an `inline` method 😄

core/src/main/scala-3/caliban/schema/SchemaDerivation.scala
```diff
@@ -58,24 +58,28 @@ trait CommonSchemaDerivation {
...
 
   private def makeSumSchema[R, A](
-    members: => List[(String, List[Any], Schema[R, Any], Int)],
+    _members: => List[(String, List[Any], Schema[R, Any], Int)],
     info: TypeInfo,
     annotations: List[Any]
-  )(m: Mirror.SumOf[A]): Schema[R, A] = new Schema[R, A] {
+  )(ordinal: A => Int): Schema[R, A] = new Schema[R, A] {
+
+    private lazy val members = _members
```
kyri-petrou, line 81: It's better to move the `lazy val` inside the class. `lazy val`s produce a fair bit of code, and creating one per schema was adding some overhead when compiling

ghostdogpr (approved):
LGTM!

PR #1834, opened 2023-08-13
Auto enum field derivation [Scala 3]

kyri-petrou:
Not sure if this is a good or bad idea, but I thought it might be worth to discuss it at least.

With semi-auto derivation, we need to create a custom schema for each of the enum fields. e.g.,

Example 1:
```scala
enum Foo {
  case A, B
}

given Schema[Any, Foo]          = Schema.derived
given Schema[Any, Foo.A.type]   = Schema.derived
given Schema[Any, Foo.B.type]   = Schema.derived
```

In a perfect world, this would be allowed, but unfortunately Scala 3 doesn't allow the use of the `derives` keyword on enum fields:

```scala
enum Foo derives Schema.SemiAuto {
  case A derives Schema.SemiAuto
  case B derives Schema.SemiAuto
}
```
One alternative is to use auto-derivation for enums, but this already becomes a bit annoying when there are many enums in the schema:

```scala
given Schema[Any, Foo] = {
  import Schema.auto.*
  Schema.derived
}
```

Where this becomes problematic is when enums are used as unions instead. In the case below, auto-derivation will generate a schema for `FooInner` too, which is something we don't want if we're using semi-auto derivation.

Example 2:
```scala
case class FooInner(value: String)

@GQLUnion
enum Foo {
  case A()
  case B(value: FooInner)
}
```

The changes in this PR automatically generate a schema for enum fields, but not for any of their child fields (if applicable).

Schema definition then becomes:

```scala
// Example 1:
given Schema[Any, Foo] = Schema.derived

// Example 2:
given Schema[Any, Foo]    = Schema.derived
given Schema[Any, FooInner] = Schema.derived
```

What are your thoughts on this? Do you think that this is simplifying creation of schemas for enums or it might be making things unnecessarily complex?

### DIFF
--- a/core/src/main/scala-3/caliban/schema/ArgBuilderDerivation.scala
+++ b/core/src/main/scala-3/caliban/schema/ArgBuilderDerivation.scala
@@ -22,7 +22,7 @@ trait CommonArgBuilderDerivation {
           (
             constValue[name].toString,
             Macros.annotations[t], {
-            if (Macros.isEnumField[P, t])
+              if (Macros.isEnumField[P, t])
                 if (!Macros.implicitExists[ArgBuilder[t]]) derived[t]
                 else summonInline[ArgBuilder[t]]
               else summonInline[ArgBuilder[t]]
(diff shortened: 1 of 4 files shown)


ghostdogpr:
Looks okay to me, I like it.

ghostdogpr (approved):

Issue #1794, opened 2023-07-14
Stitching does not escape backslashes consistently


pellekrogholt:
Problem: 

Stitching graphs with Json as inputs that contain `\n`, `\r`, `\”` and so forth is not correctly escaped in `RemoteQuery.QueryRenderer.render(_)` at least some invalid json is produced and breaks stitching.

This stitching code is fragile:

```scala
  case class StringValue(value: String)   extends Value {
    override def toString: String = s""""${value.replace("\"", "\\\"").replace("\n", "\\n")}""""
  }
```

Reproduced here: https://scastie.scala-lang.org/wJwNYBRVR1aYJPsaDU0mvQ

ghostdogpr:
I believe it was fixed in https://github.com/ghostdogpr/caliban/pull/1835

emilgodsk:
Hi @ghostdogpr - I do not believe it was fixed in #1835 

See the updated Scastie using the new render https://scastie.scala-lang.org/C1nQo987RU2Sm8NLrisWbA as well as my fork, adding some tests: https://github.com/ghostdogpr/caliban/compare/series/2.x...emilgodsk:caliban:series/2.x

Here, the following tests are failing:
```
test("correctly renders a query for a field of json escaped characters")
test("correctly renders a query for a field of json escaped escaped quote")
test("correctly renders a query for a field of json escaped newline")
test("correctly renders a query for a field of json escaped carriage return")
test("correctly renders a query for a field of json escaped tab")
```

The test `test("correctly renders a query for a field of json escaped characters")` basically tests all the escaped characters at once. Here it is worth noting, that I added circe as a test dependency, to validate that the JSON we expect is valid, however, the actual output is invalid.

ghostdogpr:
@paulpdaniels if you have a chance to take a look

PR #1835, opened 2023-08-13
Add fast rendering

paulpdaniels:
Introduces a `Renderer` type and a `DocumentRenderer` implementation which generalizes the the rendering of a `Document` so that it can work with both query and schema documents (the inverted parser).

Running the benchmarks:

```
// Without document caching
[info] Benchmark                             Mode  Cnt       Score      Error  Units
[info] GraphQLBenchmarks.renderCaliban      thrpt    5  103325.148 ± 5125.692  ops/s
[info] GraphQLBenchmarks.renderCalibanFast  thrpt    5  235363.260 ± 1367.168  ops/s

// With document caching
[info] Benchmark                             Mode  Cnt       Score      Error  Units
[info] GraphQLBenchmarks.renderCaliban      thrpt    5  106606.129 ±  511.746  ops/s
[info] GraphQLBenchmarks.renderCalibanFast  thrpt    5  842012.538 ± 1521.568  ops/s
```

The above benchmarks show a 2x improvement over the base rendering (calling `GraphQL#render`) when using `.toDocument`, however, if we cache the document (something I added on this PR) we see an 8x improvement on rendering.

This should help with use cases that require the document to be re-serialized (e.g. federation, stitching, etc.)

TODO:

- [x] Finish Compact parsing
- [x] Migrate remaining "rendering" components
- [x] Add more tests


ghostdogpr:
Can this replace the implementation of `.render` entirely?

core/src/main/scala/caliban/rendering/Renderer.scala
```diff
@@ -0,0 +1,559 @@
...
+      var first = true
+      value.foreach { v =>
+        if (first) first = false
+        else write.append(separator)
+        self.unsafeRender(v, indent, write)
+      }
+    }
+  }
+
+  def unsafeRender(value: A, indent: Option[Int], write: StringBuilder): Unit
```
kyri-petrou: Should this be made a `protected def` instead?

core/src/main/scala/caliban/rendering/Renderer.scala
```diff
@@ -0,0 +1,559 @@
...
+    def nlOrSp  = if (newline) writer += '\n' else writer += ' '
+    def nlOrNot = if (newline) writer += '\n' else writer
+
+    description.foreach {
+      case value if value.contains('\n') =>
+        val valueEscaped = value.replace(tripleQuote, s"\\$tripleQuote")
+        // check if it ends in quote but it is already escaped
+        if (value.endsWith("\\\"")) {
+          writer ++= tripleQuote
+          nlOrNot
```
kyri-petrou: I had to stare at this for the better part of a minute before realising it's a side-effecting method 😅. Maybe rename them to  `nlOrSp()` and `nlOrNot()` to make it more readable?

core/src/main/scala/caliban/rendering/Renderer.scala
```diff
@@ -0,0 +1,559 @@
...
+          writer ++= valueEscaped
+          nlOrNot
+          writer ++= tripleQuote
+        } else if (value.last == '"') {
```
kyri-petrou: Is there any chance that `description` can be an empty string? If yes, `.last` will throw a `NoSuchElementException`

core/src/main/scala/caliban/rendering/Renderer.scala
```diff
@@ -0,0 +1,559 @@
...
+  private[caliban] lazy val fieldDefinitionRenderer: Renderer[FieldDefinition] = new Renderer[FieldDefinition] {
+    override def unsafeRender(definition: FieldDefinition, indent: Option[Int], builder: StringBuilder): Unit =
+      definition match {
+        case FieldDefinition(description, name, arguments, tpe, directives) =>
+          builder += '\n'
+          unsafeRenderDescription(description, indent, builder)
+          pad(indent, builder)
+          builder ++= name
+          unsafeRenderArguments(arguments, builder)
+          builder ++= s": $tpe"
```
kyri-petrou: Nit: I think this will be slightly more performant
```suggestion
          builder ++= s": "
          builder ++= tpe.toString
```

core/src/main/scala/caliban/rendering/Renderer.scala
```diff
@@ -0,0 +1,559 @@
...
+    }
+  }
+
+  private def increment(indentation: Option[Int]): Option[Int] = indentation.map(_ + 1)
+
+  private def unsafeFastString(value: String, builder: StringBuilder) = {
+    builder.append('"')
+    var i = 0
+    while (i < value.length) {
+      value.charAt(i) match {
```
kyri-petrou: I believe that in most cases the char will be the default `c`, is that right? If that's the case, consider using a lookup switch (binary search) so that we check only 3 (maybe 4) of the conditions instead of all 7

```suggestion
      (value.charAt(i): @annotation.switch) match {
```

core/src/main/scala/caliban/rendering/Renderer.scala
```diff
@@ -0,0 +1,559 @@
...
+    case TypeDefinition.ScalarTypeDefinition(_, name, _)          => (0, name)
+    case TypeDefinition.UnionTypeDefinition(_, name, _, _)        => (1, name)
+    case TypeDefinition.EnumTypeDefinition(_, name, _, _)         => (2, name)
+    case TypeDefinition.InputObjectTypeDefinition(_, name, _, _)  => (3, name)
+    case TypeDefinition.InterfaceTypeDefinition(_, name, _, _, _) => (4, name)
+    case TypeDefinition.ObjectTypeDefinition(_, name, _, _, _)    => (5, name)
+  }
+
+  def render(document: Document): String = {
+    val builder = new StringBuilder
```
kyri-petrou: The default StringBuilder has a size of 16, which causes it to keep resizing once we reach that size. Given that these rendered strings are likely to be more than 16 chars, is there a more sensible default we could use? Perhaps 32 or 64?

kyri-petrou (commented):
Great work! Had a look through the code, couldn't find anything really standing out. Just some minor recommendations for performance

frekw:
Super clean! I wonder if this also solves https://github.com/ghostdogpr/caliban/issues/1794?

core/src/main/scala/caliban/rendering/Renderer.scala
paulpdaniels: I'm not sure if it's worth it but perhaps it could be estimated by the size of the document? Actually we have a `SourceMapper` in the `Document` which for "parsed" documents is backed by the original document string. Maybe I could preset it to that size? For schema docs it's a little trickier, but you're right we could probably fix a constant or a constant * # of definitions.

core/src/main/scala/caliban/rendering/Renderer.scala
paulpdaniels: Ok will fix

core/src/main/scala/caliban/rendering/Renderer.scala
paulpdaniels: You're right I missed this bit

core/src/main/scala/caliban/rendering/Renderer.scala
paulpdaniels: Ah nice!

core/src/main/scala/caliban/rendering/Renderer.scala
paulpdaniels: It should be covered by the case match `.contains('\n')` so we know that there is at least one character

core/src/main/scala/caliban/rendering/Renderer.scala
kyri-petrou: Yeah I think presetting it to the size of `SourceMapper` (or rounding up to the next power of 2?) makes sense since the rendered document is likely going to be smaller than that? Unless the query is minimized of course.

core/src/main/scala/caliban/rendering/Renderer.scala
paulpdaniels: Done

core/src/main/scala/caliban/rendering/Renderer.scala
paulpdaniels: Done (I think)

core/src/main/scala/caliban/Rendering.scala
```diff
@@ -1,259 +0,0 @@
-package caliban
-
-import caliban.Value._
-import caliban.introspection.adt._
-import caliban.introspection.adt.__TypeKind._
-import caliban.parsing.adt.Directive
-
-object Rendering {
```
kyri-petrou, line 8: I'm thinking whether we want to keep this object for binary compatibility, annotate it as deprecated and have its public methods and proxy to `rendering.Renderer` instead? Or is is not possible to implement all the methods?

core/src/main/scala/caliban/execution/Field.scala
```diff
@@ -58,6 +59,50 @@ case class Field(
         case (None, None)         => None
       }
     )
+
+  def toSelection: Selection = {
```
ghostdogpr: lazy val?

federation/src/test/scala/caliban/federation/FederationV1Spec.scala
```diff
@@ -9,6 +9,8 @@ import caliban.Value.{ BooleanValue, StringValue }
 import caliban.schema.Schema.auto._
 import caliban.schema.ArgBuilder.auto._
 import caliban.federation.v1._
+import caliban.parsing.adt.Directive
```
ghostdogpr: Are these needed?

core/src/main/scala/caliban/Rendering.scala
ghostdogpr: Only a few of them were public so yeah would be nice to keep them as deprecated, if possible

core/src/main/scala/caliban/execution/Field.scala
paulpdaniels: Done

federation/src/test/scala/caliban/federation/FederationV1Spec.scala
paulpdaniels: Removed

core/src/main/scala/caliban/Rendering.scala
paulpdaniels: Fixed!

kyri-petrou (approved):
LGTM 👍

ghostdogpr (approved):

PR #1837, opened 2023-08-15
Allow skipping `ValidationWrapper` for introspection queries

kyri-petrou:
~~Introspection queries are relatively large + deep, so in order for introspection to be allowed we need for these wrappers to allow 250+ fields and 12 levels depth. With this PR, we can disable these checks for introspection queries.~~

~~One other way that this could be implemented is to be able to set a separate limit for introspections, e.g.,:~~

```scala
def maxFields(maxFields: Int, maxIntrospectionFields: Option[Int] = None): ValidationWrapper[Any] = ???
```

~~Thoughts?~~

EDIT: This PR now adds the `skipForIntrospection` method which returns a new ValidationWrapper that is skipped for introspection queries.

### DIFF
--- a/core/src/main/scala/caliban/wrappers/Wrapper.scala
+++ b/core/src/main/scala/caliban/wrappers/Wrapper.scala
@@ -56,7 +56,23 @@ object Wrapper {
    * Wrapper for the query validation stage.
    * Wraps a function from a `Document` to a `ZIO[R, ValidationError, ExecutionRequest]`.
    */
-  trait ValidationWrapper[-R] extends SimpleWrapper[R, ValidationError, ExecutionRequest, Document]
+  trait ValidationWrapper[-R] extends SimpleWrapper[R, ValidationError, ExecutionRequest, Document] { self =>
+
+    final def skipForIntrospection: ValidationWrapper[R] = new ValidationWrapper[R] {
+      override val priority: Int = self.priority
+
+      override def apply[R1 <: R](that: GraphQL[R1]): GraphQL[R1] = self.apply(that)
+
+      override def |+|[R1 <: R](that: Wrapper[R1]): Wrapper[R1] = self |+| that
+
+      override def wrap[R1 <: R](
+        f: Document => ZIO[R1, ValidationError, ExecutionRequest]
+      ): Document => ZIO[R1, ValidationError, ExecutionRequest] =
+        (doc: Document) =>
+          if (doc.isIntrospection) f(doc)
+          else self.wrap(f)(doc)
+    }
+  }
 
   /**
    * Wrapper for the query execution stage.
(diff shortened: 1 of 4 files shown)


frekw:
In my opinion we should always skip it for introspection. Not much point of having it in that case, right?

kyri-petrou:
> In my opinion we should always skip it for introspection. Not much point of having it in that case, right?

Ideally I think we should do this, I agree. My main concern though is that our [`isIntrospection` check](https://github.com/ghostdogpr/caliban/blob/98aa20bad287c072bc28c9bcb347e73e29a67bb0/core/src/main/scala/caliban/introspection/Introspector.scala#L92-L100) is not particularly strict, so we probably want to be able to give the option to limit introspection queries as well.

This is the main reason I'm also partially inclined for the method to accept a `maxIntrospectionFields: Option[Int]` argument instead.

ghostdogpr:
How about something more generic, like `def skipForIntrospection(validationWrapper: ValidationWrapper): ValidationWrapper`, that would check if it's introspection and run or not the inner wrapper. It feels more re-usable and prevent adding a boolean everywhere.
Maybe we can add some syntax to make it nicer.

kyri-petrou:
@ghostdogpr what do you think about making it a final method on `ValidationWrapper`? Or do you prefer having it as a method / syntax in `Wrappers` which we can import instead?

ghostdogpr:
Method sounds good!

ghostdogpr (approved):

kyri-petrou:
@ghostdogpr I'll add a unit test just in case before merging in

core/src/main/scala/caliban/wrappers/Wrapper.scala
```diff
@@ -56,7 +56,22 @@ object Wrapper {
...
+    final def skipForIntrospection: ValidationWrapper[R] = new ValidationWrapper[R] {
+      override val priority: Int = self.priority
+
+      override def wrap[R1 <: R](
+        f: Document => ZIO[R1, ValidationError, ExecutionRequest]
+      ): Document => ZIO[R1, ValidationError, ExecutionRequest] =
+        (doc: Document) =>
+          if (doc.isIntrospection) f(doc)
+          else self.wrap(f)(doc)
+    }
```
kyri-petrou, line 73: @ghostdogpr turns out we can't simply override `apply` and `|+|`. Do we think there might be cases where someone implemented custom logic in those methods?

For `apply`, I could get this to work, but looks rather weird so I left it out - let me know if you think we should add it:

```scala
override def apply[R1 <: R](that: GraphQL[R1]): GraphQL[R1] = super.apply(self.apply(that))
```

ghostdogpr: It's probably okay

ghostdogpr (approved):

PR #1845, opened 2023-08-21
GraphQLTransportWS: Always return an id upon errors

### LINKED ISSUE #1844: ConnectionInit in GraphQLTransportWS returns faulty message upon error
https://github.com/ghostdogpr/caliban/blob/52540e3d8dc9200443054dc57e23ce76ca02fe44/interop/tapir/src/main/scala/caliban/interop/tapir/ws/Protocol.scala#L76C52-L80

If the beforeInit Webhook returns an error it seems that the protocol is broken since the connection_init message doesn't contain an id. In order for it to work correctly it would need to set the id field as well as per the [spec](https://github.com/enisdenjo/graphql-ws/blob/master/PROTOCOL.md#error).
```
interface ErrorMessage {
  id: '<unique-operation-id>';
  type: 'error';
  payload: GraphQLError[];
}
```

SvenW:
Fixes #1844

### DIFF
--- a/interop/tapir/src/main/scala/caliban/interop/tapir/ws/Protocol.scala
+++ b/interop/tapir/src/main/scala/caliban/interop/tapir/ws/Protocol.scala
@@ -9,6 +9,7 @@ import caliban.interop.tapir.WebSocketHooks
 import zio.stm.TMap
 import zio.stream.{ UStream, ZStream }
 import zio.{ Duration, Promise, Queue, Ref, Schedule, UIO, URIO, ZIO }
+import java.util.UUID
 
 sealed trait Protocol {
   def name: String
@@ -49,15 +50,20 @@ object Protocol {
       override def complete(id: String): GraphQLWSOutput =
         GraphQLWSOutput(Ops.Complete, Some(id), None)
 
-      override def error[E](id: Option[String], e: E): GraphQLWSOutput =
+      override def error[E](id: Option[String], e: E): GraphQLWSOutput = {
+        val outputId = id match {
+          case None    => Some(UUID.randomUUID().toString)
+          case Some(_) => id
+        }
         GraphQLWSOutput(
           Ops.Error,
-          id,
+          outputId,
           Some(ResponseValue.ListValue(List(e match {
             case e: CalibanError => e.toResponseValue
             case e               => StringValue(e.toString)
           })))
         )
+      }
     }
 
     override def make[R, E](
--- a/interop/tapir/src/main/scala/caliban/interop/tapir/ws/Protocol.scala
+++ b/interop/tapir/src/main/scala/caliban/interop/tapir/ws/Protocol.scala
@@ -8,8 +8,7 @@ import caliban.interop.tapir.TapirAdapter.CalibanPipe
 import caliban.interop.tapir.WebSocketHooks
 import zio.stm.TMap
 import zio.stream.{ UStream, ZStream }
-import zio.{ Duration, Promise, Queue, Ref, Schedule, UIO, URIO, ZIO }
-import java.util.UUID
+import zio.{ Duration, Promise, Queue, Random, Ref, Schedule, UIO, URIO, ZIO }
 
 sealed trait Protocol {
   def name: String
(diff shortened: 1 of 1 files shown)


ghostdogpr (approved):
Seems ok to me.

@paulpdaniels wanna take a look?

kyri-petrou:
~Seems ok to me too. Do we also need to fix this for the legacy protocol too, or is the spec different for that one?~

SvenW:
> Seems ok to me too. Do we also need to fix this for the legacy protocol too, or is the spec different for that one?

From what I can tell connection errors do not include the id in the old [spec](https://github.com/apollographql/subscriptions-transport-ws/blob/master/PROTOCOL.md#gql_connection_error)

kyri-petrou:
@SvenW the error you referred to in the new spec is a response to the `Subscribe` message, not to the connection message. 

image

This is the equivalent of `GQL_ERROR` in the legacy spec: https://github.com/apollographql/subscriptions-transport-ws/blob/master/PROTOCOL.md#gql_error

I can't seem to find anywhere how an error due to connection is meant to look like though in the new spec

kyri-petrou:
According to the spec, it seems that we need to use HTTP codes to reject `connection_init` messages

>If the server wishes to reject the connection, for example during authentication, it is recommended to close the socket with 4403: Forbidden.

https://github.com/enisdenjo/graphql-ws/blob/master/PROTOCOL.md#connectioninit

SvenW:
> According to the spec, it seems that we need to use HTTP codes to reject `connection_init` messages
> 
> > If the server wishes to reject the connection, for example during authentication, it is recommended to close the socket with 4403: Forbidden.
> 
> https://github.com/enisdenjo/graphql-ws/blob/master/PROTOCOL.md#connectioninit

Yeah, you're right. We need to handle the authentication errors for the connection_init events as well with the correct code when closing the socket. However from what I can tell the `error` message should always contain the id though.

kyri-petrou:
@SvenW I agree, although I don't think this will solve the linked issue. Or have you tested and it does solve it?

SvenW:
> @SvenW I agree, although I don't think this will solve the linked issue. Or have you tested and it does solve it?

Well it solves it but not completely according to the spec. Let me see if I can find some time during the day and solve the unauthenticated part as well

SvenW:
> @SvenW I agree, although I don't think this will solve the linked issue. Or have you tested and it does solve it?

From what I can tell graphql-transport-ws always closes the socket on `connection_init` with a 4403, Forbidden as can be seen [here](https://github.com/enisdenjo/graphql-ws/blob/master/src/server.ts#L608-L614). What do you think @kyri-petrou ?

interop/tapir/src/main/scala/caliban/interop/tapir/ws/Protocol.scala
```diff
@@ -49,15 +50,20 @@ object Protocol {
       override def complete(id: String): GraphQLWSOutput =
         GraphQLWSOutput(Ops.Complete, Some(id), None)
 
-      override def error[E](id: Option[String], e: E): GraphQLWSOutput =
+      override def error[E](id: Option[String], e: E): GraphQLWSOutput = {
+        val outputId = id match {
+          case None    => Some(UUID.randomUUID().toString)
```
paulpdaniels: Nit: It would be better to use the `zio.Random` construct, though you'll then need to push the id generation up where you will still be in ZIO land

kyri-petrou:
> From what I can tell graphql-transport-ws always closes the socket on connection_init with a 4403, Forbidden as can be seen [here](https://github.com/enisdenjo/graphql-ws/blob/master/src/server.ts#L608-L614). What do you think @kyri-petrou ?

I think as long as we're within the spec and it works that's fine. Have you tested whether this solves the original issue?

SvenW:
> > From what I can tell graphql-transport-ws always closes the socket on connection_init with a 4403, Forbidden as can be seen [here](https://github.com/enisdenjo/graphql-ws/blob/master/src/server.ts#L608-L614). What do you think @kyri-petrou ?
> 
> I think as long as we're within the spec and it works that's fine. Have you tested whether this solves the original issue?

I've tested it with the AuthExampleApp and it seems to behave as expected, both with error messages and with the connection_init event with authorization errors.

paulpdaniels (approved):

kyri-petrou (approved):
Nice!

interop/tapir/src/main/scala/caliban/interop/tapir/ws/Protocol.scala
ghostdogpr: @SvenW can you just update that one? Then it's good to go 🙏

ghostdogpr (approved):

PR #1846, opened 2023-08-21
Add support for `@oneOf` inputs

kyri-petrou:
VERY early draft for adding support for the `@oneOf` inputs, just to discuss whether we're on the right track. See current version of the RFC [here](https://github.com/graphql/graphql-spec/pull/825)

Some of the assumptions I made:

1. We probably want to require an annotation to mark an input sealed trait as a `oneOf`. I _think_ that makes sense since the spec on `@oneOf` is very strict, so it's very unlikely that these sealed traits will be used for any return types
2. Since in Scala 3 we control derivation, I enforced the checks that the sealed trait is a valid `@oneOf` via macros. I think it's nicer UX if we give compile-time errors instead of runtime ones for invalid types. However, in Scala 2 we don't control the derivation so we have to enforce these checks during runtime. I understand if we want to unify the validation, so I'm happy to remove the macros
3. I couldn't find anything in the spec (might have missed it though) but I assumed we don't want to mark inputs as `@oneOf` if they have only 1 field (kind of defeats the purpose?)

## TODO:

- [x] Add validations for `@oneOf` inputs in `Validator` (both against the Schema & user inputs)
- [x] Add tests where `@oneOf` inputs are a mix of input objects and Scalars
- [ ] Add documentation
- [ ] Discuss on what we should do with the client


core/src/main/scala-2/caliban/schema/ArgBuilderDerivation.scala
```diff
@@ -54,6 +57,24 @@ trait CommonArgBuilderDerivation {
...
+  private def makeOneOffBuilder[A](ctx: SealedTrait[ArgBuilder, A]): ArgBuilder[A] =
+    new ArgBuilder[A] {
+      private lazy val builders = ctx.subtypes.map(_.typeclass)
+
+      def build(input: InputValue): Either[ExecutionError, A] = input match {
+        case InputValue.ObjectValue(fields) if fields.size == 1 =>
+          builders.view
+            .map(_.build(input))
+            .find(_.isRight)
+            .getOrElse(Left(ExecutionError(s"Invalid oneOf input $fields for trait ${ctx.typeName.short}}")))
```
kyri-petrou: Open to suggestions on how we can make this more efficient. I can't figure out a way to extract the field argument here unfortunately :(

core/src/main/scala-2/caliban/schema/ArgBuilderDerivation.scala
```diff
@@ -54,6 +57,24 @@ trait CommonArgBuilderDerivation {
...
+        case InputValue.ObjectValue(_)                          =>
+          Left(ExecutionError(s"Exactly one key must be specified for oneOf inputs"))
```
kyri-petrou: TODO: Move this check into `Validator`

core/src/main/scala/caliban/introspection/Introspector.scala
```diff
@@ -67,14 +78,17 @@ object Introspector extends IntrospectionDerivation {
...
+
     val resolver = __Introspection(
       __Schema(
         rootType.description,
         rootType.queryType,
         rootType.mutationType,
         rootType.subscriptionType,
         types,
-        directives ++ rootType.additionalDirectives
+        directives ++ (if (hasOneOf) List(oneOfDirective) else Nil) ++ rootType.additionalDirectives
```
kyri-petrou, line 94: We can enable this by default. I thought we might want to add it only when defined in the schema at first

core/src/main/scala-2/caliban/schema/ArgBuilderDerivation.scala
```diff
@@ -54,6 +57,24 @@ trait CommonArgBuilderDerivation {
         }
       case None        => Left(ExecutionError(s"Can't build a trait from input $input"))
     }
+
+  private def makeOneOffBuilder[A](ctx: SealedTrait[ArgBuilder, A]): ArgBuilder[A] =
```
ghostdogpr: Minor: typo, should be OneOf

core/src/main/scala-2/caliban/schema/ArgBuilderDerivation.scala
ghostdogpr: Isn't there a possibility that some types can successfully be parsed as another type of the oneof?

core/src/main/scala-2/caliban/schema/ArgBuilderDerivation.scala
ghostdogpr: Seen other occurrences in the rest of the code, might have to do a search and replace 😄

core/src/main/scala-3/caliban/schema/macros/Macros.scala
```diff
@@ -88,4 +89,26 @@ private[caliban] object Macros {
     Expr(TypeRepr.of[P].typeSymbol.flags.is(Flags.Enum) && TypeRepr.of[T].typeSymbol.flags.is(Flags.Enum))
   }
 
+  def hasOneOfInputAnnotationImpl[T: Type](using q: Quotes): Expr[Boolean] = {
+    import q.reflect.*
+    Expr(TypeRepr.of[T].typeSymbol.annotations.exists(_.tpe.typeSymbol.name == "GQLOneOfInput"))
+  }
+
+  def isValidOneOffInputImpl[T: Type](using q: Quotes): Expr[Boolean] = {
```
ghostdogpr: I think I would rather have it in the schema validation. Your point is right about UX, I am just wary of macros because they might break (like of them broke with Java 17) and aren't easy to maintain.

Schema validation happens when you start your server to you find quickly anyway no?

core/src/main/scala/caliban/introspection/Introspector.scala
ghostdogpr: I agree

ghostdogpr (commented):
Looks very nice!

core/src/main/scala-2/caliban/schema/ArgBuilderDerivation.scala
paulpdaniels: We get the subclass name though right? I think we could recapitalize the type name `Dog => dog` as the default strategy, and allow people to override that explicitly with an annotation

kyri-petrou: > Isn't there a possibility that some types can successfully be parsed as another type of the oneof?

@ghostdogpr No because we ensure that the argument names of the leaf cases are unique

>  I think we could recapitalize the type name Dog => dog as the default strategy, and allow people to override that explicitly with an annotation

@paulpdaniels  I think this could work, but just to be on the same page. Let's say we want to implement this:

```gql
input BarInput {
  a: String!
  b: Int!
}

input FooInput @oneOf {
  stringArg: String
  barArg: BarInput
}
```

Currently, we can do it as:

```scala
case class Bar(a: String, b: Int)

sealed trait Foo
object Foo:
  case class ArgA(stringArg: String) extends Foo
  case class ArgB(barArg: Bar) extends Foo
```
With your recommendation @paulpdaniels, we need to do the following:

```scala
sealed trait Foo
object Foo:
  @GQLValueType
  @GQLOneOfFieldName("stringArg")  // Otherwise it'll be named `argA`
  case class ArgA(stringArg: String) extends Foo
  
  @GQLOneOfFieldName("barArg")
  case class Bar(a: String, b: Int) extends Foo

  // Alternatively:
  @GQLInputName("BarInput")
  case class BarArg(a: String, b: Int) extends Foo
```

If this is what you meant, I like it. ~It reuses the existing annotations~, and prevents having to create an extra case class as in the case of `case class ArgB(barArg: Bar) extends Foo` in my first example

EDIT: At first I used the `GQLName` annotation, but on second thought, we might want to introduce another annotation as I think if we use `@GQLName` to rename the field it'll cause confusion since we normally use it to rename the type. Open to suggestion for a better annotation name :)

core/src/main/scala-2/caliban/schema/ArgBuilderDerivation.scala
kyri-petrou: Oops 😅 Fixed

core/src/main/scala-3/caliban/schema/macros/Macros.scala
kyri-petrou: I think you're right - it's nice to have these checks during compiling but they do add unnecessary maintenance (and to some degree, compiling time) overhead. Removed

core/src/main/scala-2/caliban/schema/ArgBuilderDerivation.scala
ghostdogpr: `GQLName`is already used for renaming fields though

kyri-petrou: @ghostdogpr are you suggesting we should use `GQLName` to rename the fields here as well? My only concern is that in this case `GQLName` annotates the case class, so if someone wants to use this `sealed trait` as a union, then the type will not be named properly

ghostdogpr: Actually I think I prefer the initial way:
```scala
case class Bar(a: String, b: Int)

sealed trait Foo
object Foo {
  case class ArgA(stringArg: String) extends Foo
  case class ArgB(barArg: Bar) extends Foo
}
```

That's how arguments work currently and I think it's good to be consistent.

kyri-petrou: From a UX perspective, I prefer this as well. However, this approach makes 2 things a bit more complicated:

1. We have to iterate through the leaf `ArgBuilder`s in order to find which one matches the input as per my comment above. Personally, I don't think this is much of an issue. We could even do something clever and memoize this after the iteration succeeds for the first time.
2. We have to ensure that `ArgA`, `ArbB` etc have only 1 argument. Unfortunately, the only time that we can do that is during derivation as we need to flatten the list of arguments when we create the schema. This probably means we need to decide on what's best UX wise for this case. We can either:
    1. Throw an error during derivation (perhaps use macros in Scala 3?). This is not very pure but it's safe as the error will be thrown during the application startup
    2. Drop any extra arguments from the schema, but since we can't do any validation this will likely result in unexpected runtime errors
    3. Drop any leaf cases that are invalid. This might cause some confusion but if we document it well then it shouldn't be too big of an issue
    4. ???

kyri-petrou: Annoyingly, this would be the easiest way to define a schema / arg builder for a `@oneOf` input, but then it's really annoying to extract the argument that was passed:

```scala
@GQLOneOfInput
case class Foo(
  stringArg: Option[String],
  barArg: Option[Bar]
)
```

ghostdogpr: Couldn't we use `__Type#origin` to find during validation that 2 fields come from the same object and fail in that case?

kyri-petrou: Unfortunately no. For Scalars, we don't set a value for the origin field. I actually tried setting it manually only for oneOf input arguments and it didn't work either, it caused issues with the `validateClashingTypes` validation.

The only way I could figure out how to propagate the type info was to add a field on `__InputValue` and exclude that field from any schema generation: https://github.com/ghostdogpr/caliban/pull/1846/files#diff-7fd5a1e4c9e85d511a507e4a9d0ddb63b131a5be6decb8feb3f95d0c00e5f31cR16

It's a bit of a hack unfortunately, but it works. Other than binary compatibility (which we're breaking with this PR anyways), do we see any other issues with doing this?

ghostdogpr: Seems okay to me, I don't mind that hack so much if things stay simple on the user side.

kyri-petrou: I also rewrote the `ArgBuilder`s for OneOf inputs to use fallback `orElse`. Under the hood it's still the same as iterating through them I suppose but at least this feels more Caliban-y 😅

core/src/main/scala/caliban/validation/Validator.scala
```diff
@@ -823,27 +860,49 @@ object Validator {
...
     def validateFields(fields: List[__InputValue]): EReader[Any, ValidationError, Unit] =
       ZPure.foreachDiscard(fields)(validateInputValue(_, inputObjectContext)) *>
-        noDuplicateInputValueName(fields, inputObjectContext)
+        noDuplicateInputValueName(fields)
+
+    def validateOneOfFields(fields: List[__InputValue]): EReader[Any, ValidationError, Unit] =
+      failValidationWhen(fields.size <= 1)(
+        s"OneOf $inputObjectContext has less than 2 fields",
+        "OneOf input objects must have at least 2 possible fields"
+      ) *>
```
kyri-petrou: ~This is not mentioned anywhere in the spec (i.e., it's not explicitly allowed or disallowed), but it doesn't seem to me like something that should be allowed normally.~

~The only case I can think of that might require this is if you currently have only 1 type of input but you plan to add more cases in the future and don't want to introduce a breaking change..~

~Thoughts?~

[Just checked](https://github.com/graphql/graphql-spec/pull/825#issuecomment-1696975451), they're allowed. I'll remove the validation

core/src/main/scala-2/caliban/schema/ArgBuilderDerivation.scala
```diff
@@ -53,6 +56,27 @@ trait CommonArgBuilderDerivation {
...
+
+  private def makeOneOfBuilder[A](ctx: SealedTrait[ArgBuilder, A]): ArgBuilder[A] =
+    new ArgBuilder[A] {
+
+      private def inputError(input: InputValue) =
+        ExecutionError(s"Invalid oneOf input $input for trait ${ctx.typeName.short}")
+
+      private val combined = ctx.subtypes.map(_.typeclass).toList.asInstanceOf[List[ArgBuilder[A]]] match {
+        case head :: tail =>
+          tail.foldLeft(head)(_ orElse _).orElse(input => Left(inputError(input)))
```
paulpdaniels: This isn’t super nice, maybe we could have ArgBuilder have a partial function that can select if it can handle a particular input function. Then by default it would just be `def partial: PartialFunction[InputValue, Either[ExecutionError, A]] = PartialFunction(build)`. That way we could delegate to the ArgBuilder Wdyt?

kyri-petrou: I gave it a go implementing your recommendation. Let me know what you think

core/src/main/scala-3/caliban/schema/ArgBuilderDerivation.scala
```diff
@@ -91,28 +97,69 @@ trait CommonArgBuilderDerivation {
...
+    private def fromFields(fields: Map[String, InputValue]): Either[ExecutionError, A] = {
+      var i          = 0
+      val l          = params.length
+      var acc: Tuple = EmptyTuple
+      while (i < l) {
+        val (label, default, builder) = params(i)
+        val field                     = fields.getOrElse(label, null)
+        val value                     = if (field ne null) builder.build(field) else default
+        value match {
+          case Right(v)    => acc :*= v
```
paulpdaniels, line 155: What is this operator?

paulpdaniels (approved):

core/src/main/scala-3/caliban/schema/ArgBuilderDerivation.scala
kyri-petrou: It's shorthand for `acc = acc :* v`, where `:*` is the Scala 3 `Tuple` equivalent of a `Seq`'s `:+` (append)

core/src/main/scala/caliban/introspection/adt/__Type.scala
```diff
@@ -21,7 +21,8 @@ case class __Type(
   ofType: Option[__Type] = None,
   specifiedBy: Option[String] = None,
   @GQLExcluded directives: Option[List[Directive]] = None,
-  @GQLExcluded origin: Option[String] = None
+  @GQLExcluded origin: Option[String] = None,
+  isOneOf: Option[Boolean] = None
```
ghostdogpr, line 25: Shall we exclude this field when `oneof` is not used?

ghostdogpr: Nope

ghostdogpr (approved):

tjarvstrand:
🚀 ❤️

PR #1856, opened 2023-08-26
Optimize `VariableCoercer`

kyri-petrou:
Seems that when we refactored the `Validator` to use `ZPure` we forgot(?) to do it with the `VariableCoercer`. Reality is that in most cases the variables will be a small / shallow object so this doesn't _really_ provide much performance benefits for most users, except perhaps in cases where the inputs are deeply nested JSON files.

In this PR there are 2 different optimizations:

1. Use ZPure in favour of ZIO effects when recursing
2. Make `context` a by-name argument. This prevents us building the error context except when we need to fail the validation

### Benchmarking results

Before:
```
[info] Benchmark                              Mode  Cnt      Score     Error  Units
[info] ValidationBenchmark.variableCoercer100   thrpt    5  14006.365 ± 180.586  ops/s
[info] ValidationBenchmark.variableCoercer1000  thrpt    5    518.925 ±  18.125  ops/s
```

After:
```
[info] Benchmark                                 Mode  Cnt      Score     Error  Units
[info] ValidationBenchmark.variableCoercer100   thrpt    5  87682.811 ± 661.300  ops/s
[info] ValidationBenchmark.variableCoercer1000  thrpt    5   8429.327 ±  28.180  ops/s
```

### DIFF
--- a/core/src/main/scala/caliban/GraphQL.scala
+++ b/core/src/main/scala/caliban/GraphQL.scala
@@ -105,7 +105,9 @@ trait GraphQL[-R] { self =>
                                                           }
                   typeToValidate                        = if (intro) introspectionRootType else rootType
                   schemaToExecute                       = if (intro) introspectionRootSchema else schema
-                  validatedReq                         <- VariablesCoercer.coerceVariables(request, doc, typeToValidate, skipValidation)
+                  validatedVariables                   <-
+                    VariablesCoercer
+                      .coerceVariables(request.variables.getOrElse(Map.empty), doc, typeToValidate, skipValidation)
                   validate                              = (doc: Document) =>
                                                             for {
                                                               config       <- Configurator.configuration
@@ -113,8 +115,8 @@ trait GraphQL[-R] { self =>
                                                                                 doc,
                                                                                 typeToValidate,
                                                                                 schemaToExecute,
-                                                                                validatedReq.operationName,
-                                                                                validatedReq.variables.getOrElse(Map.empty),
+                                                                                request.operationName,
+                                                                                validatedVariables,
                                                                                 config.skipValidation,
                                                                                 config.validations
                                                                               )
--- a/core/src/main/scala/caliban/GraphQL.scala
+++ b/core/src/main/scala/caliban/GraphQL.scala
@@ -105,9 +105,8 @@ trait GraphQL[-R] { self =>
                                                           }
                   typeToValidate                        = if (intro) introspectionRootType else rootType
                   schemaToExecute                       = if (intro) introspectionRootSchema else schema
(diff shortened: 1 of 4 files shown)


core/src/main/scala/caliban/parsing/VariablesCoercer.scala
```diff
@@ -23,64 +25,77 @@ object VariablesCoercer {
...
+  )(implicit trace: Trace): IO[ValidationError, GraphQLRequest] =
+    coerceVariables(req.variables.getOrElse(Map.empty), doc, rootType, skipValidation)
+      .map(m => req.copy(variables = Some(m)))
+
+  def coerceVariables(
+    variables: Map[String, InputValue],
+    doc: Document,
+    rootType: RootType,
+    skipValidation: Boolean
+  )(implicit trace: Trace): IO[ValidationError, Map[String, InputValue]] = {
```
kyri-petrou, line 37: Not necessary as an optimization, but I found it a bit weird that we need to pass the `GraphQLRequest` to this method only to extract the variables and copy them back in.

Do we see a case in the future where we might need more information from the `GraphQLRequest` case class? Happy to rollback this change if we don't want to add this overloaded method

core/src/main/scala/caliban/GraphQL.scala
```diff
@@ -105,16 +105,18 @@ trait GraphQL[-R] { self =>
                                                           }
                   typeToValidate                        = if (intro) introspectionRootType else rootType
                   schemaToExecute                       = if (intro) introspectionRootSchema else schema
-                  validatedReq                         <- VariablesCoercer.coerceVariables(request, doc, typeToValidate, skipValidation)
+                  validatedVariables                   <-
+                    VariablesCoercer
```
ghostdogpr: This is super minor but how about extracting `request.variables.getOrElse(Map.empty)` into a variable so that hopefully this fits in one line and stays easier to read?

core/src/main/scala/caliban/parsing/VariablesCoercer.scala
ghostdogpr: I think what you did is okay

ghostdogpr (approved):
Nice!

ghostdogpr (approved):

core/src/main/scala/caliban/GraphQL.scala
kyri-petrou: I was a bit wary of extracting them into a variable so we don't accidentally use them elsewhere but yeah the formatting was annoying me as well. Hopefully naming them `unsafeVars` (`unsafeVariables` didn't fit in a single line) will be clear enough that they're not to be used 😅

PR #1858, opened 2023-08-29
Update magnolia to v1

kyri-petrou:
### ⚠️ _Very_ binary incompatible change ⚠️

While working on the `@oneOf` inputs feature, I've noticed that we use a very outdated version of magnolia (3+ years old by now). It seems that magnolia (now v1) is published under the `com.softwaremill` organisation, and includes some very small changes to their API. However, I'm fairly certain that the code generated by magnolia v0.x will not be binary compatible with magnolia's v1.x, so we definitely need to bump the version to `2.4.0` if we include this in the next release.

Since next version will likely be `2.4.0` if we include the oneOf inputs feature, perhaps we should also update magnolia?


build.sbt
```diff
@@ -608,12 +605,14 @@ lazy val commonSettings = Def.settings(
...
         "-opt:l:inline",
+        "-opt:l:method",
         "-explaintypes"
       )
     case Some((2, 13)) =>
       Seq(
         "-Xlint:-byname-implicit",
-        "-explaintypes"
+        "-explaintypes",
+        "-opt:l:method"
```
kyri-petrou, line 638: I noticed that we don't have method-local optimizations enabled. These are pretty safe and binary compatible so I enabled them. I can remove them if needed

ghostdogpr:
I'm okay with this change but not sure if 2.4.0 will be the next one.
There are several binary compatible changes we could release as 2.3.1 first, so maybe let's not merge this one just yet.

kyri-petrou:
> I'm okay with this change but not sure if 2.4.0 will be the next one. There are several binary compatible changes we could release as 2.3.1 first, so maybe let's not merge this one just yet.

Sounds good to me 👍

core/src/main/scala-2/caliban/schema/ArgBuilderDerivation.scala
```diff
@@ -5,8 +5,7 @@ import caliban.InputValue
 import caliban.Value._
 import caliban.schema.Annotations.GQLDefault
 import caliban.schema.Annotations.GQLName
-import magnolia._
-import mercator.Monadic
+import magnolia1._
```
paulpdaniels, line 8: Really like that softwaremill versions their packages like this, it's so helpful when you have lagging dependencies

frekw: 💯 agreed. We cross publish some internal deps for both Scala 2&3 and this has been a godsend.

frekw (approved):
LGTM! But perhaps hold off merging until 2.4?

kyri-petrou:
> LGTM! But perhaps hold off merging until 2.4?

Yeap that's the plan

ghostdogpr:
Can you rebase? Then let's merge this one

Issue #1823, opened 2023-07-31
Pekko variant?


MichelEdkrantz:
Hi!
Thanks for this great lib :) Now that Pekko 1.0 has been released, is there any interest from your end to provide a pekko variant of this?

https://github.com/apache/incubator-pekko

I believe it would be most a copy of this, but with imports changed to pekko. 
https://github.com/ghostdogpr/caliban/blob/series/2.x/adapters/akka-http/src/main/scala/caliban/AkkaHttpAdapter.scala

Tapir has no pekko support yet, but looks like it's in the making
https://github.com/softwaremill/tapir/issues/2884

ghostdogpr:
Yeah, I'm definitely okay with having a pekko-http adapter.

As you said, we need tapir support first but as soon as it's available we can add that adapter (it'll be quite easy).

MichelEdkrantz:
Created a fork and a PR in #1862. Some issues remain but hoping to move this forward somehow 😅

PR #1862, opened 2023-08-30
Add support for pekko

MichelEdkrantz:
Hi, 
Trying to add pekko support after the discussion in #1823. I've barely used sbt before, so any pointers to what I missed is greatly appreciated. Tried to copy the parts from akka, and changed to a unique port for the tests. 

pekko-http also support scala 3, so the skip-part there should probably be changed? 

Tapir needs to be updated, since pekko support was added in 1.7.0. Not sure if this should be a separate PR? 

This is causing the build to fail in the ZHttpAdapter. Without touching anything else, the build starts failing with tapir version 1.5.0 that introduces zio-http version 3.0.0-RC2
https://github.com/softwaremill/tapir/releases/tag/v1.5.0

One compile time issue seems to be breaking interface changes from zio-http 3.0.0.RC2 regarding zio.http.socket
https://github.com/zio/zio-http/pull/2152

Any ideas on how to proceed? Hoping this investigation will help moving this forward, but it seems to be some general blockers that need to be untangled before introducing anything with pekko. 
Should we create a separate issue for this?


build.sbt
```diff
@@ -18,11 +18,12 @@ val jsoniterVersion           = "2.23.3"
 val laminextVersion           = "0.16.2"
 val magnoliaVersion           = "0.17.0"
 val mercatorVersion           = "0.2.1"
+val pekkoVersion              = "1.0.1"
 val playVersion               = "2.8.20"
 val playJsonVersion           = "2.9.4"
 val scalafmtVersion           = "3.7.13"
 val sttpVersion               = "3.9.0"
-val tapirVersion              = "1.4.0"
+val tapirVersion              = "1.7.3"
```
kyri-petrou, line 26: We can't do this currently because the version of zio-http we depend on in one of the modules works only with `tapir-zio-http-server=1.4.0`.

On that note, @ghostdogpr if this becomes blocking for new features, I believe we should be able to update `tapir-core` et.al to the latest version and pin `tapir-zio-http-server` to `1.4.0` until a new release of `zio-http` is out. What do you think?

ghostdogpr: Will it work? I don't think tapir core 1.7.3 will be compatible with tapir-zio-http-server 1.4.0?

build.sbt
```diff
@@ -84,6 +85,7 @@ lazy val root = project
     core,
     http4s,
     akkaHttp,
+    pekkoHttp,
```
ghostdogpr, line 89: Can you also add it to `.circleci/config.yml` so that the CI will run it?

build.sbt
```diff
@@ -369,6 +371,26 @@ lazy val akkaHttp = project
   )
   .dependsOn(core, tapirInterop % "compile->compile;test->test")
 
+lazy val pekkoHttp = project
+  .in(file("adapters/pekko-http"))
+  .settings(name := "caliban-pekko-http")
+  .settings(commonSettings)
+  .settings(enableMimaSettingsJVM)
+  .settings(
+    skip           := (scalaVersion.value == scala3),
```
ghostdogpr: If it works on Scala 3, these can be removed.

build.sbt
ghostdogpr: An alternative is going back to not using tapir for zio-http at all...
Or we just move to RC2... 🤔

kyri-petrou: I haven't tested it but I don't see why it wouldn't work. Assuming tapir-core doesn't have any package-private breaking changes, all the core APIs needed by tapir-zio-http-server 1.4.0 from core should still be there in 1.7.3

ghostdogpr: Dunno, I thought that between 3 major versions the chance of breaking changes would be pretty high 😄

kyri-petrou: Tapir uses `semver` versioning though, right? So that that would be 3 minor versions. Unless they also use `pvp` then yes I wouldn't do that 😄

ghostdogpr: Looks like they do! Hopefully it works 🤞

MichelEdkrantz:
Hi again, thanks for your quick feedback 
After some more fiddling. Here are some of my thoughts

**Scala 3**
Scala3 does not work for pekko, since tapir-json-play does not publish any _3 artifact. 
https://app.circleci.com/pipelines/github/ghostdogpr/caliban/5302/workflows/80f80262-8a02-458c-a74d-f5334b1b63bb/jobs/32666

It seems it's only used for tests? Maybe moving this to Circe or so could help?
import sttp.tapir.json.play._

**Tapir Versions**
Tried to put different versions of tapir for zioHttp and http4s. But the tapir-interop part is still set as `tapirVersion` (1.7.3) so that leaves the build in a tapir version mixed mode, which seems ugly and potentially harmful. I don't know if there is an sbt way to solve this, or if that's undesired for any reason? 

Updating the socket parts to zio-http to 3.0.0-RC2 could probably solve some these issues. They introduced breaking changes in the refactor here https://github.com/zio/zio-http/pull/2152. I still think it would be good to update the tapir version in a separate PR and not here. 
https://zio.dev/zio-http/examples/advanced/websocket-server/

ghostdogpr:
Yeah you can change to another Json like circe if that enables Scala 3.

For Tapir let's see if it works first 😅

kyri-petrou:
@MichelEdkrantz there are 2 main issue with updating to zio-http 3.0.0-RC2:

1. The current main of `zio-http` (soon(?) to be RC3) contains many breaking changes compared to RC2. We would have to re-rewrite parts of the code once RC3 is out so we would be doubling down on the work necessary
2. In RC2 we can't set the WS subprotocol dynamically when processing the request. For subscriptions to work, it requires users to add extra config params to their server config, so unless someone reads the release notes very carefully they might end up with some unpleasant surprises after deployment. This was fixed / added back in the current main, so if we wait for RC3 we won't have to "break" things

I think let's give it a go with changing the `tapirVersion` variable to `1.7.3` and hard-coding `1.4.0` [here](https://github.com/ghostdogpr/caliban/blob/series/2.x/build.sbt#L345) and see if the test suite passes

MichelEdkrantz:
Ok, cool :) Thanks for the clarification. 
I got pekko with scala3 to build now, changing from play to circe in the test suite so it seems doable. 

If hard coding the tapirVersion for zioHttp to 1.40, how will it work with the dependsOn tapir-interop. 
That line uses a different tapir version. Maybe that's ok?

kyri-petrou:
@MichelEdkrantz The way that SBT works is that if you have conflicting version dependencies (in this case, `tapir-interop` depends on `tapir-core` 1.7.3, and `interop-zio-http` depends on `1.4.0`, the latest version will be selected. Since the only thing that has a dependency on `tapir-zio-http-server` is the zio-interop module, the final resolution will be something along these lines:

```
tapir-core -> 1.7.3
tapir-json-X -> 1.7.3
tapir-zio -> 1.7.3
tapir-zio-http-server -> 1.4.0
```
This _should_ be okay, because we only have forward version dependencies since `tapir-zio-http-server` sits at the top of this dependency tree, and all packages it depends on are `>= 1.4.0` but still binary compatible according to `semver` versioning (`major.minor.patch`).

In short, it means that all the APIs that `tapir-zio-http-server` needs to work should all be there in tapir-core & tapir-zio 1.7.3.

This would have been a problem if we wanted to update `tapir-zio-http-server` to `1.7.3` but keep `tapir-core` to `1.4.0`. Does that make sense?

If you want to get a better understanding of the dependency tree, you can run this, which will open a browser window with the dependency tree

```shell
sbt zioHttp/dependencyBrowseTree
```

MichelEdkrantz:
Hi, 
yes. Thanks. That's what I was getting at also with the comment about "tapir version mixed mode". Just wanted to highlight it so it's ok with you. Any pointers for how to best proceed with this now? Not sure if there is anything more than I can do at this stage?

.circleci/config.yml
```diff
@@ -83,7 +83,7 @@ jobs:
       - checkout
       - restore_cache:
           key: sbtcache
-      - run: sbt ++3.3.0 core/test catsInterop/compile benchmarks/compile monixInterop/compile clientJVM/test clientJS/compile zioHttp/test tapirInterop/test http4s/test federation/test tools/test reporting/test tracing/test
+      - run: sbt ++3.3.0 core/test catsInterop/compile benchmarks/compile monixInterop/compile clientJVM/test clientJS/compile zioHttp/test tapirInterop/test pekkoHttp/test http4s/test federation/test tools/test reporting/test tracing/test
```
kyri-petrou, line 86: We need to add the `pekkoHttp` to the `test_mima` job(s) as well

adapters/pekko-http/src/main/scala/caliban/PekkoHttpAdapter.scala
```diff
@@ -0,0 +1,198 @@
...
+import sttp.tapir.PublicEndpoint
+import sttp.tapir.model.ServerRequest
+import sttp.tapir.server.ServerEndpoint
+import sttp.tapir.server.pekkohttp.{ PekkoHttpServerInterpreter, PekkoHttpServerOptions }
+import zio._
+import zio.stream.ZStream
+
+import scala.concurrent.{ ExecutionContext, Future }
+
+class PekkoHttpAdapter private (private val options: PekkoHttpServerOptions)(implicit ec: ExecutionContext) {
```
kyri-petrou: Bit of a nit: I think the `private val` part can be removed

kyri-petrou (commented):
I'm not too familiar with the Akka ecosystem but overall LGTM. Before merging in though we need to:

- [ ] Enable mima tests for both Scala 2.13 & 3
- [ ] Add a section under `./examples`. I believe this should be more or less similar to the Akka one (don't forget to add the new module in `./examples/README.md` too

(@ghostdogpr correct me if I'm wrong) but I believe we'll also need to add the relevant doco in a separate PR under `vuepress/docs/docs/README.md` and `vuepress/docs/docs/adapters.md`. Reason for having it in a separate PR is so that it can be merged when the next release is ready, as the microsite is updated automatically via CI

kyri-petrou:
Regarding this:
> Any pointers for how to best proceed with this now? Not sure if there is anything more than I can do at this stage?

I think that as long as the test suite passes and we're not seeing any weird warnings from SBT or the compiler we should be OK (famous last words 😅 ). One thing I'm not sure is whether we'll be including this new adapter in the 2.3.1 or a subsequent release. If it's in a subsequent release, ZIO http might have an RC3 version out which will likely align the tapir versions again.

ghostdogpr:
I think we can include this in 2.3.1

ghostdogpr:
@MichelEdkrantz as @kyri-petrou mentioned, can you do the documentation changes in a separate PR? The reason is that docs are automatically published after merge, so I prefer to do it only after I release a new version.

.circleci/config.yml
MichelEdkrantz: added a line here, but it fails. I guess since this is since there is nothing to compare to. should this be a separate PR later once 1 published version exists?

MichelEdkrantz:
> @MichelEdkrantz as @kyri-petrou mentioned, can you do the documentation changes in a separate PR? The reason is that docs are automatically published after merge, so I prefer to do it only after I release a new version.

Sorry, will fix this 👍

.circleci/config.yml
ghostdogpr: Ah probably yeah

ghostdogpr: Maybe just comment it out so it will be easy to add later

kyri-petrou:
@ghostdogpr as an extra step of precaution, I'd like to get [this PR](https://github.com/ghostdogpr/caliban/pull/1867) merged in first just to cover all angles with the tapir version changes in this PR

kyri-petrou (dismissed):
Okay sorry for going rounds on this one. It seems we can't do this. Tapir doesn't seem to be strictly binary compatible against minor versions.

Actually I'm not even sure how did the test suite even pass, all I did was create an interceptor:

```shell
    Exception in thread "zio-fiber-22989,22940" java.lang.NoSuchMethodError: 'sttp.tapir.server.interceptor.log.DefaultServerLog sttp.tapir.server.interceptor.log.DefaultServerLog$.apply(scala.Function1, scala.Function2, scala.Function2, scala.Function2, java.lang.Object, boolean, boolean, boolean, boolean, scala.Function1, scala.Function1, scala.Function1, boolean, java.time.Clock)'
    	at sttp.tapir.server.ziohttp.ZioHttpServerOptions$.defaultServerLog(ZioHttpServerOptions.scala:42)
    	at sttp.tapir.server.ziohttp.ZioHttpServerOptions$.customiseInterceptors(ZioHttpServerOptions.scala:36)
    	at sttp.tapir.server.ziohttp.ZioHttpServerOptions$.default(ZioHttpServerOptions.scala:56)
    	at caliban.ZHttpAdapter$.makeHttpService$default$2(ZHttpAdapter.scala:19)
```

kyri-petrou:
Ok I see what happened. I tried to update tapir to 1.7.3 in a different codebase but used a version of Caliban that was compiled against 1.4.0. It seems that a user can only use the tapir version that Caliban is compiled against due to default method / case class arguments. So if Caliban is compiled against tapir-core 1.7.3 then it works.

ghostdogpr (approved):

ghostdogpr:
@MichelEdkrantz can you rebase to the latest master? That way we'll have the additional tests from https://github.com/ghostdogpr/caliban/pull/1867. Then we can merge this if CI passes.

ghostdogpr:
Thanks!

MichelEdkrantz:
Hi,
What's the typical process for publishing a release for this? 
I also rebased the docs PR #1866 here.

ghostdogpr:
Just having enough stuff to release (which is the case now) and me having the time to take care of it 😄 I'll try to do it this weekend or the next at last. There's a snapshot published so you can already play with it.

PR #1867, opened 2023-09-01
Add test for HTTP `GET` method

kyri-petrou:
Saw some time ago that we don't have a test against HTTP `GET` methods. Adding this as an extra precaution now that we're going to use a different tapir version for core / zio-http


ghostdogpr (approved):

ghostdogpr:
Ah just saw that CI was red after approving 😄

kyri-petrou:
~Huh.. So the `GET` endpoint for the `play` adapter seems to be broken? I'm surprised that no one reported this yet, I wonder whether it's recent (or that very few users use the play adapter? 😅)~

🤦 Nevermind me, just needed to add the route to the test

PR #1868, opened 2023-09-01
Fix DocumentRenderer

paulpdaniels:
Added a new test harness for verifying the output of the `DocumentRenderer` and found some bugs as a result.


core/src/main/scala/caliban/rendering/DocumentRenderer.scala
```diff
@@ -91,9 +114,13 @@ object DocumentRenderer extends Renderer[Document] {
     }
 
   private lazy val documentRenderer: Renderer[Document] = Renderer.combine(
-    directiveDefinitionsRenderer.contramap(_.directiveDefinitions),
+    (directiveDefinitionsRenderer ++
```
kyri-petrou, line 117: Bit of a nitpick: I wonder if we should change `Renderer.Combined` to use a `Vector` instead, which has has `~O(1)` complexity over `List`'s `O(N)` for appending new elements. Or do we never have enough renderers combined to make a difference?

kyri-petrou (approved):
Nice, thanks for the fixes! Also, love the new tests :) Just a tiny nitpick comment

core/src/main/scala/caliban/rendering/DocumentRenderer.scala
paulpdaniels: I think that might be overkill, since for the most part renderers won't be constructed dynamically so the construction cost will be amortized.

PR #1875, opened 2023-09-07
Adjust tracing span name to adhere to semantic guidelines

Fluxx:
Adjusts the tracing span name to adhere to [the published OpenTelemetry semantic guidelines for GraphQL](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/database/graphql.md). Mostly, this was just two things:

1. Renaming the `"query"` attribute to `"document"`, and
2. Renaming the span from the generic `"query"` to `"<operationType>[ <operationName>]"`

To achieve 2, I had to modify the `ExecutionRequest` type to include the `operationName` as an argument. This wasn't too hard, as `ExecutionRequest` is only constructed in one spot.

### DIFF
--- a/tracing/src/main/scala/caliban/tracing/SchemaTracer.scala
+++ b/tracing/src/main/scala/caliban/tracing/SchemaTracer.scala
@@ -44,8 +44,8 @@ object SchemaTracer {
 
   private def spanName(request: ExecutionRequest): String = {
     val operationTypeString = request.operationType match {
-      case OperationType.Query => "query"
-      case OperationType.Mutation => "mutation"
+      case OperationType.Query        => "query"
+      case OperationType.Mutation     => "mutation"
       case OperationType.Subscription => "subscription"
     }
 
--- a/tracing/src/main/scala/caliban/tracing/SchemaTracer.scala
+++ b/tracing/src/main/scala/caliban/tracing/SchemaTracer.scala
@@ -43,7 +43,7 @@ object SchemaTracer {
 
   private def attributes[T, R](
     field: Field
-  ) = List("query" -> graphQLQuery(field))
+  ) = List("document" -> graphQLQuery(field))
 
   private def graphQLQuery(field: Field): String =
     RemoteQuery.apply(maskField(field)).toGraphQLRequest.query.getOrElse("")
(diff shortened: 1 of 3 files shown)


Fluxx:
I see there is some binary compatibility issues with my change, since it is a breaking change. Lemme know how I should handle this. Probably easiest is to make the new `operationName` argument have a default?

tracing/src/main/scala/caliban/tracing/SchemaTracer.scala
```diff
@@ -30,7 +31,7 @@ object SchemaTracer {
         else
           ZIO.serviceWithZIO[Tracing](tracer =>
             tracer.span[R, Nothing, GraphQLResponse[CalibanError]](
-              "query",
+              spanName(request),
```
kyri-petrou, line 34: I believe we can achieve the same using combination of an `OverallWrapper` and `EffectfulWrapper` to extract the operation name and populate a Promise or Ref with it. This way we can avoid the breaking change.

You can take a look at how we do something similar in other wrappers such as `ApolloPersistedQueries` and `FieldMetrics`

kyri-petrou:
> Probably easiest is to make the new operationName argument have a default?

Making `operationName` have a default argument is still a breaking change unfortunately. See my recommendation above, I believe we can avoid adding the field altogether

tracing/src/main/scala/caliban/tracing/SchemaTracer.scala
Fluxx: Ah that is a good idea. I will take a look at doing that. My ZIO skills are only moderate, so I may come back here for help but will let you know. The `ApolloPersistedQueries` and `FieldMetrics` are good reference though, thanks for mentioning those.

kyri-petrou:
@Fluxx are you still planning on working on this? 

By the way given that we're introducing breaking changes in the next release, we might as well keep this as is @ghostdogpr or is it better not to add more fields to the `ExecutionRequest`?

ghostdogpr:
Yeah I am good with merging as it is. @Fluxx can you just rebase once to make sure there's no issue with the latest code?

Fluxx:
I was still planning on working on it, but hadn't had time in the past couple weeks. I will be able to rebase this pr this week though.

Fluxx:
@ghostdogpr rebased from upstream `series/2.x`.

ghostdogpr (approved):
Thanks!

PR #1876, opened 2023-09-08
Optimizations round 14

kyri-petrou:
Stuck on a plane for 14 hours with nothing better to do 😄

Mima seems to complain about the change of type of `String` to `=> String`. The change is source compatible but not binary compatible. Shall we merge this in after 3.3.1?


ghostdogpr:
Yep let's keep 2.3.1 fully compatible and ship all breaking changes in 2.4.0

ghostdogpr:
You can add the Mima exclusion now that we're heading to 2.4.0

kyri-petrou:
@ghostdogpr done

PR #1878, opened 2023-09-09
Caching v2

### LINKED ISSUE #1840: Apollo Caching implementation is obsolete
The current implementation apollo caching uses an outdated method that relies on schema extensions. However, this has been superceded and should now actually be returned via the `Cache-Control` header in http. We'll still need to compute the cache control via wrapper but will then need to do an additional step in the TapirAdapter to convert that extension into a single cache-control header.

https://www.apollographql.com/docs/apollo-server/performance/caching/#calculating-cache-behavior

paulpdaniels:
Apollo has defined a new [caching protocol](https://www.apollographql.com/docs/apollo-server/performance/caching/) (not sure when they switched over but I think my federation caching has been broken for a while now). Anyway it isn't specific to their router anymore and can instead be used to generate http cache-control directives that are useful if your graph supports `GET` and sits behind a CDN.

This PR adds support for it with a new aspect.

Closes https://github.com/ghostdogpr/caliban/issues/1840


core/src/main/scala/caliban/GraphQLAspect.scala
```diff
@@ -16,3 +19,19 @@ trait GraphQLAspect[+LowerR, -UpperR] { self =>
...
+  def withSchemaDirectives(directives: List[Directive]): GraphQLAspect[Nothing, Any] = new GraphQLAspect[Nothing, Any] {
+    def apply[R](gql: GraphQL[R]): GraphQL[R] = gql.withSchemaDirectives(directives)
+  }
+
+  def withDirectives(directives: List[__Directive]): GraphQLAspect[Nothing, Any] = new GraphQLAspect[Nothing, Any] {
+    def apply[R](gql: GraphQL[R]): GraphQL[R] = gql.withAdditionalDirectives(directives)
+  }
+
+  def withTypes(types: List[__Type]): GraphQLAspect[Nothing, Any] = new GraphQLAspect[Nothing, Any] {
+    def apply[R](gql: GraphQL[R]): GraphQL[R] = gql.withAdditionalTypes(types)
```
paulpdaniels, line 34: I didn't end up needing these, but I think they are useful so I kept them in place. It makes it easy if you are defining a new aspect that also requires new directives/types.

core/src/main/scala/caliban/wrappers/Caching.scala
```diff
@@ -0,0 +1,288 @@
...
+
+  object CacheSettings {
+    val default: CacheSettings = CacheSettings(0.seconds, CacheScope.Public)
+  }
+
+  /**
+   * Overrides the cache hint for a particular field. This can be used to dynamically set the cache hint within
+   * a resolver
+   */
+  def setCacheHint(hint: CacheHint): UIO[Unit] = cacheOverride.set(Some(hint))
```
kyri-petrou: Nice ❤️

interop/tapir/src/main/scala/caliban/interop/tapir/TapirAdapter.scala
```diff
@@ -190,4 +196,22 @@ object TapirAdapter {
...
+          val scope  = fields.collectFirst {
+            case (Caching.ScopeName, Value.EnumValue("PRIVATE")) => CacheDirective.Private
+            case (Caching.ScopeName, Value.EnumValue("PUBLIC"))  => CacheDirective.Public
+          }.getOrElse(CacheDirective.Public)
+
+          maxAge.toList :+ scope
+        }
+      })
+      .flatten
+      .map(_.mkString(", "))
```
kyri-petrou: Might be asking for too much here, but would it be possible to chose if we want to set `s-maxage` instead of `max-age`? This would be useful for cases that we want to cache on CDN but not in the browser cache

core/src/main/scala/caliban/wrappers/Caching.scala
```diff
@@ -0,0 +1,288 @@
...
+    }
+
+    private def mostRestrictive(
+      parent: CacheHint,
+      current: CacheHint
+    ): CacheHint = {
+      val scope         = List(parent.scope, current.scope).flatten.minOption
+      val shouldInherit = current.inheritMaxAge
+      val maxAge        =
+        if (shouldInherit) parent.maxAge
```
kyri-petrou: Should the condition be `shouldInherit && parent.maxAge.isDefined` or do we want to inherit regardless whether it's set?

core/src/main/scala/caliban/wrappers/Caching.scala
```diff
@@ -0,0 +1,288 @@
...
+    private[caliban] def cacheHintFromType(typ: __Type): Option[CacheHint] = {
+      val key = typ.name.getOrElse(typ.toString)
+      Option(_typeCache.get(key)) match {
+        case Some(Some(hint)) => Some(hint)
+        case Some(None)       => None
+        case None             =>
+          typ.directives match {
+            case Some(dirs) =>
+              val hint = extractCacheDirective(dirs)
+              hint.foreach(h => _typeCache.put(key, Some(h)))
```
kyri-petrou: Should this be `_typeCache.put(key, hint)` instead?

core/src/main/scala/caliban/wrappers/Caching.scala
```diff
@@ -0,0 +1,288 @@
...
+              hint
+            case None       =>
+              _typeCache.put(key, None)
+              None
+          }
+      }
+    }
```
kyri-petrou: Assuming that we don't need the `.foreach` part as per my comment above, perhaps consider using `computeIfAbsent` instead:

```suggestion
    private[caliban] def cacheHintFromType(typ: __Type): Option[CacheHint] = {
      val key = typ.name.getOrElse(typ.toString)
      _typeCache.computeIfAbsent(key, _ => typ.directives.flatMap(extractCacheDirective))
    }
```

core/src/main/scala/caliban/wrappers/Caching.scala
```diff
@@ -0,0 +1,288 @@
...
+
+            field.fields.foldLeft(newPolicy)(loop)
+          }
+
+          val updated = loop(CachePolicy(CacheHint.default), request.field)
+
+          state.update(updated.merge).as(request)
+        }
+      }
+    }
```
kyri-petrou: I'll be honest, it took me a few minutes to figure out why you pulled this logic out of the `FieldWrapper` until I saw that it doesn't need to wrap pure values anymore. Nice!

core/src/main/scala/caliban/wrappers/Caching.scala
kyri-petrou: As a suggestion/question, should we check if the value was already set and use the most restrictive one instead? Something along:

```suggestion
  def setCacheHint(hint: CacheHint): UIO[Unit] =
    cacheOverride.update {
      case None           => Some(hint)
      case Some(existing) => Some(mostRestrictive(existing, hint))
    }
```

interop/tapir/src/main/scala/caliban/interop/tapir/HttpInterpreter.scala
```diff
@@ -1,14 +1,19 @@
...
+import sttp.model.headers.CacheDirective
 import sttp.model.{ headers => _, _ }
 import sttp.tapir.Codec.JsonCodec
 import sttp.tapir.model.ServerRequest
 import sttp.tapir._
 import zio._
 
+import java.util.concurrent.TimeUnit
+import scala.concurrent.duration.FiniteDuration
+
```
kyri-petrou: Nit: Added imports are unused

interop/tapir/src/main/scala/caliban/interop/tapir/TapirAdapter.scala
```diff
@@ -190,4 +196,22 @@ object TapirAdapter {
...
+    extensions
+      .flatMap(_.fields.collectFirst { case (Caching.DirectiveName, ResponseValue.ObjectValue(fields)) =>
+        fields.collectFirst { case ("policy", ResponseValue.ObjectValue(fields)) =>
+          val maxAge = fields.collectFirst { case (Caching.MaxAgeName, v: IntValue) =>
+            CacheDirective.MaxAge(FiniteDuration(v.toInt, TimeUnit.SECONDS))
+          }
+          val scope  = fields.collectFirst {
+            case (Caching.ScopeName, Value.EnumValue("PRIVATE")) => CacheDirective.Private
+            case (Caching.ScopeName, Value.EnumValue("PUBLIC"))  => CacheDirective.Public
+          }.getOrElse(CacheDirective.Public)
```
kyri-petrou: This feels like a dangerous default value. Perhaps something like this is a bit safer?

```scala
maxAge match {
  case None    => CacheDirective.NoCache
  case Some(_) => CacheDirective.Private
}
```

core/src/main/scala/caliban/wrappers/Caching.scala
```diff
@@ -0,0 +1,288 @@
...
+                          List(
+                            MaxAgeName -> IntValue(maxAge.toSeconds),
+                            ScopeName  -> StringValue(scope.toString)
+                          )
+                        )
+                      )
+                    )
+                  )
+                )((s, extensions) => extensions.fields ++ s)
+              )
```
kyri-petrou: Very nitpick (sorry 😅): I think this is a bit more performant / readable

```suggestion
              ResponseValue.ObjectValue(
                (DirectiveName -> ResponseValue.ObjectValue(
                  List(
                    "version" -> IntValue(2),
                    "policy" -> ResponseValue.ObjectValue(
                      List(
                        MaxAgeName -> IntValue(maxAge.toSeconds),
                        ScopeName -> StringValue(scope.toString)
                      )
                    )
                  )
                )) :: response.extensions.map(_.fields).getOrElse(Nil)
              )
```

kyri-petrou (commented):
Love this, great work! I think my only main comment is whether we need a more sensible default value for the scope in the Caching header

interop/tapir/src/main/scala/caliban/interop/tapir/TapirAdapter.scala
paulpdaniels: I could probably make that an option for the `CacheSettings`

core/src/main/scala/caliban/wrappers/Caching.scala
paulpdaniels: Yeah, technically I think we could even go further, and cache this per document that would save us some steps here.

core/src/main/scala/caliban/wrappers/Caching.scala
paulpdaniels: Done! But with `fold` so we don't need an intermediate `Option` construction

interop/tapir/src/main/scala/caliban/interop/tapir/TapirAdapter.scala
paulpdaniels: I think I captured this behavior

interop/tapir/src/main/scala/caliban/interop/tapir/TapirAdapter.scala
paulpdaniels: Added an option to `CacheSettings` that allows the caller to define how they want to convert the `CachePolicy` into a header.

core/src/main/scala/caliban/wrappers/Caching.scala
paulpdaniels: Changed

core/src/main/scala/caliban/wrappers/Caching.scala
paulpdaniels: Fixed!

core/src/main/scala/caliban/wrappers/Caching.scala
```diff
@@ -0,0 +1,329 @@
...
+          `type` = () => Types.boolean,
+          defaultValue = None
+        )
+      ),
+      isRepeatable = false
+    )
+  )
+
+  def aspect[R](output: Ref[CachePolicy] => Wrapper[R]): GraphQLAspect[Nothing, R] =
+    new Default(output)
```
paulpdaniels, line 56: Abstracted this out, this is to support a fast follow where I'd like to support full response caching within caliban.

core/src/main/scala/caliban/wrappers/Caching.scala
kyri-petrou: I'm guessing you're thinking of using the `d.sourceMapper` (or even `d.sourceMapper.hashCode`) as the cache key? I think that would work, although I'm a bit worried about using an unbounded cache here since that can theoretically lead to a memory leak (especially if exploited by an attacker).

Are there any bounded cache implementations available either in Java or ZIO that we could use? If not, I guess we could require a cache as a ZIO dependency of the wrapper (similar to what we do with the `ApolloPersistedQueries` wrapper, but that makes UX a bit worse.

core/src/main/scala/caliban/wrappers/Caching.scala
```diff
@@ -0,0 +1,329 @@
...
+  }
+
+  private class Default[-R1](inner: Ref[CachePolicy] => Wrapper[R1]) extends GraphQLAspect[Nothing, R1] {
+    private val _typeCache = new ConcurrentHashMap[String, Option[CacheHint]]()
+
+    def apply[R <: R1](gql: GraphQL[R]): GraphQL[R] = {
+      val wrapper = EffectfulWrapper(
+        cacheOverride.set(None) *> Ref
+          .make(CachePolicy(CacheHint.default))
+          .map(state => staticWrapper(state) |+| fieldWrapper(state) |+| inner(state))
```
kyri-petrou: Very minor: Perhaps skip the validation wrapper for introspection queries

```suggestion
          .map(state => staticWrapper(state).skipForIntrospection |+| fieldWrapper(state) |+| inner(state))
```

core/src/main/scala/caliban/wrappers/Caching.scala
```diff
@@ -0,0 +1,329 @@
...
+
+    private def staticWrapper(state: Ref[CachePolicy]): ValidationWrapper[Any] = new ValidationWrapper[Any] {
+      override def wrap[R](
+        f: Document => ZIO[R, CalibanError.ValidationError, ExecutionRequest]
+      ): Document => ZIO[R, CalibanError.ValidationError, ExecutionRequest] = { (d: Document) =>
+        f(d).flatMap { request =>
+          def loop(policy: CachePolicy, field: Field, parentHint: Option[CacheHint]): CachePolicy = {
+            val fieldHint   = extractCacheDirective(field.directives)
+            val isInherited = fieldHint.exists(_.inheritMaxAge)
+            val typeHint    = field.parentType.flatMap(cacheHintFromType)
```
kyri-petrou: Consider making this into a `def` instead so that it's not executed if not needed

core/src/main/scala/caliban/wrappers/Caching.scala
```diff
@@ -0,0 +1,329 @@
...
+      val cacheable      = computedMaxAge > 0L
+      if (!cacheable) "no-store"
+      else {
+        val b = new StringBuilder()
+        b.append("max-age=")
+        b.append(computedMaxAge)
+        scope.foreach { s =>
+          b.append(", ")
+          b.append(s.toString.toLowerCase)
+        }
```
kyri-petrou, line 220: By default, the `public` scope is used. Should we default to `private` instead?

> If you forget to add private to a response with personalized content, then that response can be stored in a shared cache and end up being reused for multiple users, which can cause personal information to leak.

https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Cache-Control

core/src/main/scala/caliban/wrappers/Caching.scala
```diff
@@ -0,0 +1,329 @@
...
+            ScopeName         -> scope.fold[InputValue](NullValue)(s => EnumValue(s.toString)),
+            InheritMaxAgeName -> BooleanValue(inheritMaxAge)
+          )
+        )
+      )
+
+  case class CacheSettings(
+    defaultMaxAge: Duration,
+    defaultScope: CacheScope,
+    toHeader: CachePolicy => String = _.hint.toHeaderString
```
kyri-petrou, line 112: ❤️

core/src/main/scala/caliban/wrappers/Caching.scala
kyri-petrou: Oh that sounds interesting!

core/src/main/scala/caliban/wrappers/Caching.scala
paulpdaniels: Not sure, I think the default from Apollo is public, and likewise it seems that's the case for the http header as well. Nonetheless, this already deviates somewhat so we could define our own semantic if we want.

core/src/main/scala/caliban/wrappers/Caching.scala
paulpdaniels: Probably would need to normalize the document first since there could be cosmetic changes to otherwise identical documents. But it requires a further definition of "equality" between two documents.

ghostdogpr:
If you rebase it should solve the mima issue

interop/tapir/src/main/scala/caliban/interop/tapir/HttpInterpreter.scala
```diff
@@ -1,8 +1,10 @@
 package caliban.interop.tapir
 
+import caliban.Value.IntValue
```
ghostdogpr: The imports don't seem to be used

interop/tapir/src/main/scala/caliban/interop/tapir/TapirAdapter.scala
```diff
@@ -2,9 +2,12 @@ package caliban.interop.tapir
 
 import caliban._
 import caliban.ResponseValue.StreamValue
+import caliban.Value.IntValue
```
ghostdogpr: Clean imports

interop/tapir/src/main/scala/caliban/interop/tapir/TapirAdapter.scala
```diff
@@ -190,4 +196,13 @@ object TapirAdapter {
 
   def isFtv1Header(r: Header): Boolean =
     r.name == GraphQLRequest.`apollo-federation-include-trace` && r.value == GraphQLRequest.ftv1
+
+  private def computeCacheDirective(extensions: Option[ResponseValue.ObjectValue]): Option[String] =
```
ghostdogpr, line 256: Shall we keep it in extensions after we used it?

core/src/main/scala/caliban/wrappers/Caching.scala
```diff
@@ -0,0 +1,329 @@
+package caliban.wrappers
```
ghostdogpr, line 1: What do we do with `ApolloCaching`? Might be confusing to have both along each other? Also this one doesn't mention Apollo, is it a standard used by others too?

ghostdogpr:
Can you add a separate PR with docs? 🙏

core/src/main/scala/caliban/wrappers/Caching.scala
kyri-petrou: Might be wrong, but I think that the main benefit of this one is that since it uses HTTP headers for caching it can be used by any (browser) client. I wonder though whether the `ApolloCaching` would work in cases where this one wouldn't

core/src/main/scala/caliban/wrappers/Caching.scala
kyri-petrou: Yeah after thinking it through a bit more, I think it might be better to keep behaviour as consistent with the standard semantics.

Perhaps we should add it in the scaladoc somewhere & release notes that if no scope is defined, `PUBLIC` is used by default. The `ApolloCaching` wrapper didn't use headers and it might cause some issues for users that migrate to this wrapper who forgot to set a scope

core/src/main/scala/caliban/wrappers/Caching.scala
paulpdaniels: Sorry didn't see the notifications for these comments. Yeah I removed the Apollo prefix because it doesn't actually anything apollo specific, it's using the standard http headers. I'm not sure if it is worth keeping the former I don't think apollo has supported it for a while without customization to the gateway so it may give people the false sense that caching is working (or as we saw recently convince them that caliban is doing some sort of caching).

interop/tapir/src/main/scala/caliban/interop/tapir/TapirAdapter.scala
paulpdaniels: Yeah probably not necessary, I'll strip it from the extensions.

core/src/main/scala/caliban/wrappers/Caching.scala
kyri-petrou: Yeah I'm in favour of deprecating it as well. When I first started using it, I was also under the impression it was caching the responses but found out later that it wasn't 😓

kyri-petrou (approved):
Very happy with these changes. I think the only very minor thing would be to deprecate the `ApolloCaching` wrapper, otherwise it looks great

core/src/main/scala/caliban/wrappers/Caching.scala
ghostdogpr: Yeah let's deprecate it with a message pointing to the new one 👍

kyri-petrou:
@paulpdaniels the PR I merged in caused a bunch of merge conflicts. I'm happy to fix them - just let me know if you're happy if I push to your branch and I'll do it

paulpdaniels:
> Can you add a separate PR with docs?
yep I can do that @ghostdogpr

paulpdaniels:
@kyri-petrou yeah you'd probably know better what you changed so go ahead!

kyri-petrou:
@paulpdaniels done 👍

kyri-petrou:
Oh one last thing; Doesn't have to be done in this PR but just leaving this here before I forget about it again. 

Should we respect the `cache-control` or `Pragma: no-cache` header if it exists in the request?

PR #1880, opened 2023-09-10
Fix Scala 3's `Warning: mocking up superclass for module class macros` warning

kyri-petrou:
When I first started using caliban with Scala 3, I was really perplexed about the `Warning: mocking up superclass for module class macros` warning the compiler was emitting, believing that there was something wrong with my code. It turns out that the issue was because the object containing the macros is defined as `private[caliban]`.

With this PR, we move the `Macros` object within the `caliban.schema` package and make it `private`. This way we get rid of the (very) annoying compiler warning.

Note that another option is to leave the Macros object within the `caliban.schema.macros` package and make it public


kyri-petrou:
On second thought, I decided to move it back to `caliban.schema.macros` and make it public to avoid the breaking change, made the `*Impl` methods private.

I don't really see an issue with making it public. What are your thoughts @ghostdogpr?

ghostdogpr (approved):
Looks good!

PR #1888, opened 2023-09-15
Fix duplicate types being output when calling render schema

### LINKED ISSUE #1887: caliban.renderSchemaWith() outputs duplicate types
I noticed that caliban.renderSchemaWith() has been added and immediately tried it on the project at hand, but it seems that some types, enums and unions are duplicated in the output.
I looked at the difference with GraphQL.render, and it seems to be due to the lack of processing done by groupBy, which is done by RootSchemaBuilder.

https://github.com/ghostdogpr/caliban/blob/c27deabd426f53f053136246e3e51ea9fc4d9a49/core/src/main/scala/caliban/schema/RootSchemaBuilder.scala#L31-L33

nox213:
closes #1887

### DIFF
--- a/core/src/main/scala/caliban/package.scala
+++ b/core/src/main/scala/caliban/package.scala
@@ -84,10 +84,13 @@ package object caliban {
           if (hasSubscription) subscriptionSchema.toType_().name else None,
           schemaDescription
         ) ::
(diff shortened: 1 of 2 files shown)


ghostdogpr (approved):
Thanks!

PR #1900, opened 2023-09-21
Make Introspection introspectable

### LINKED ISSUE #1897: Introspection schema should be introspectable
The spec is not super clear on this but it looks like most other libraries support it.

Example query:
```
query IntrospectionQuery {
  __type(name: "__Schema") {
    name
    fields {
      name
    }
  }
}
```

ghostdogpr:
Closes https://github.com/ghostdogpr/caliban/issues/1897


kyri-petrou (approved):
Damn those introspection tests are not pretty to add things to 😢 Had the same issue in the `@oneOf` PR

PR #1901, opened 2023-09-21
Update tapir to 1.7.5

kyri-petrou:


adapters/play/src/test/scala/caliban/PlayAdapterSpec.scala
```diff
@@ -24,7 +24,7 @@ import zio.test.{ Live, ZIOSpecDefault }
 import scala.language.postfixOps
 
 object PlayAdapterSpec extends ZIOSpecDefault {
-  import sttp.tapir.json.play._
```
ghostdogpr, line 27: Play can't use play? 😭

ghostdogpr, line 27: Play can't use play? 😭

kyri-petrou: It was already an issue - it just became obvious now that `tapir-json-play` updated `play-json` to `2.10.1`. We didn't realise earlier cause we don't have `play-json` as a dependency in the tests

I think we're fine since our compile dependencies are okay

kyri-petrou: Play 2.9.0 seems to be nearing (with Scala 3 support!) so just need to wait a bit more i guess. It's currently at RC2, although if zio-http is any indication it might take a fair bit till it makes it to a stable version

kyri-petrou:
Seems that tapir-play-json is not available yet for Scala 3, so I'm using it only for the AkkaHttp tests

kyri-petrou:
Eghh this is annoying. So both Akka + Play don't work with the latest `play-json` due to versioning mismatches in Jackson Databind. And we can't use `tapir-json-play` in modules that build for Scala 3 since it's not published for it yet

ghostdogpr:
It's okay, let's just use it later when it's available 😄

kyri-petrou:
@ghostdogpr do you mean waiting on updating tapir or just use a different json library for the tests?

ghostdogpr:
> @ghostdogpr do you mean waiting on updating tapir or just use a different json library for the tests?

Just use another one in tests and switch back to it later when it works

ghostdogpr (approved):

PR #1902, opened 2023-09-21
Remove deprecated code

ghostdogpr:
Since the next version will be a major version, let's get rid of deprecated code.

I disabled Mima because it's too annoying to add all the rules and it's already decided we will break.


kyri-petrou (approved):

paulpdaniels:
The `Rendering` code too?

ghostdogpr:
Too soon?

kyri-petrou:
I'm wondering whether we should aim to keep deprecated methods which are completely deleted for a few more major version just to give users enough time to migrate.

For legacy services I usually update non-CVE dependencies every 6 months or so, and it can be very annoying when methods completely disappear and I have to go looking where its replacement is - it makes a chore even more of a chore (not talking about Caliban ofc)

ghostdogpr:
The reason I removed it is because I think no one uses those directly, they should be using `.render` instead. Apart from that I agree with you.

ghostdogpr:
Restored Rendering

PR #1903, opened 2023-09-22
Schema transformations

### LINKED ISSUE #130: Support for schema transformations
Something similar to this Apollo-server feature: https://www.apollographql.com/docs/apollo-server/features/schema-transforms/

The current problem is that Caliban currently stores runtime schema into this object:
```scala
case class Operation[-R](opType: __Type, plan: Step[R])
```

`__Type` matches the graphql type used by introspection, while `Step` is a description of a computation, that might require running some effects to get the underlying steps. If you want to remove or rename a field, it needs to be applied in both structures, which isn't really possible in the case of `Step`.

We need to think of a different way of encoding the resolver to support this feature. Maybe `Step` is generated too early and there should be something in between.

ghostdogpr:
Closes https://github.com/ghostdogpr/caliban/issues/130

This is the first step towards a much bigger change that will introduce remote sources to Caliban and the ability to combine them into a single GraphQL API (giving Caliban the ability to act as a gateway).

Adds the ability to transform types of an existing schema:
- rename types, fields and arguments
- filter interfaces, fields and arguments
- custom transformations

I used `PartialFunction` rather than `String` to give the ability to handle multiple items at once and to write custom logic such as capitalization, prefixing, etc.

Example usage:
```scala
api
  .transform(Transformer.RenameType { case "authors_v1_Author" => "Author" })
  .transform(Transformer.FilterField {
    case ("Repository", "name") => true
    case ("Repository", _)      => false
  })
```

TODO:
- [x] tests
- [ ] docs


kyri-petrou:
Love these transformers! I can take a look at the code over the weekend when I have a bit more free time

core/src/main/scala/caliban/transformers/Transformer.scala
```diff
@@ -0,0 +1,178 @@
...
+import caliban.schema.Step
+import caliban.schema.Step.{ FunctionStep, MetadataFunctionStep, ObjectStep }
+
+/**
+ * A transformer is able to modify a type, modifying its schema and the way it is resolved.
+ */
+trait Transformer[-R] { self =>
+  val typeVisitor: TypeVisitor
+
+  def transformStep[R1 <: R]: PartialFunction[Step[R1], Step[R1]]
```
kyri-petrou: I wonder whether we should use a `Function1` as the return type here instead to make function composition a bit easier

core/src/main/scala/caliban/transformers/Transformer.scala
```diff
@@ -0,0 +1,178 @@
...
+
+  def |+|[R0 <: R](that: Transformer[R0]): Transformer[R0] = new Transformer[R0] {
+    val typeVisitor: TypeVisitor = self.typeVisitor |+| that.typeVisitor
+
+    def transformStep[R1 <: R0]: PartialFunction[Step[R1], Step[R1]] =
+      Function.unlift { step =>
+        val modifiedStep = self.transformStep.lift(step).getOrElse(step)
+        that.transformStep.lift(modifiedStep)
+      }
```
kyri-petrou: In case we want to keep the PartialFunction, I struggled a bit with following the composition here. Perhaps this might be a bit more readable although I'm not 100% sure whether they do the same thing:

```suggestion
    def transformStep[R1 <: R0]: PartialFunction[Step[R1], Step[R1]] = {
      def withDefault(tr: Transformer[R1]): PartialFunction[Step[R1], Step[R1]] =
        tr.transformStep[R1].orElse { case x => x }
      withDefault(self) andThen withDefault(that)
    }
```

core/src/main/scala/caliban/transformers/Transformer.scala
```diff
@@ -0,0 +1,178 @@
...
+        field.copy(name = f.lift((t.name.getOrElse(""), field.name)).getOrElse(field.name))
+      ) |+|
+        TypeVisitor.inputFields.modifyWith((t, field) =>
+          field.copy(name = f.lift((t.name.getOrElse(""), field.name)).getOrElse(field.name))
+        )
+
+    def transformStep[R]: PartialFunction[Step[R], Step[R]] = { case ObjectStep(typeName, fields) =>
+      ObjectStep(
+        typeName,
+        fields.map { case (fieldName, step) => f.lift((typeName, fieldName)).getOrElse(fieldName) -> step }
```
kyri-petrou, line 75: I doubt this has any meaningful performance difference, but using `.applyOrElse` _might_ help with readability (here and in other places). Haven't tried it though to see how it looks so it might actually make things worse

core/src/main/scala/caliban/execution/Executor.scala
```diff
@@ -52,7 +57,7 @@ object Executor {
       arguments: Map[String, InputValue],
       path: List[Either[String, Int]]
     ): ReducedStep[R] =
-      step match {
+      transformer.transformStep.lift(step).getOrElse(step) match {
```
kyri-petrou: If I'm understanding this correctly, we will be running the transformers on the field of each object in an `ObjectStep` prior to determining whether it'll be used, is that correct? I'm kind of worried about the runtime performance this might have on large schemas.

I'm still trying to wrap my head around how all this fits together, but I don't suppose it would be possible to apply the transformations during `caliban.execution.Field.apply`? Might be missing something here, but the benefit I see by doing that (assuming it's possible) is that is that we would only be transforming fields that are used as part of the query.

kyri-petrou (commented):
Still trying to wrap my head around how a few parts of this work, but looks great overall. The only thing I'm really wondering is whether it might be possible to apply the transformations prior to execution

core/src/main/scala/caliban/transformers/Transformer.scala
ghostdogpr: Yeah my code here was wrong, fixed it in my other branch, will propagate to this one.

ghostdogpr: Fixed

core/src/main/scala/caliban/transformers/Transformer.scala
ghostdogpr: I found `applyOrElse` having a super weird type signature and failing to infer types.
For example `self.transformStep.applyOrElse((step, field), _ => step)` gives me:
```
[error] /Users/pierre/GIT/caliban/core/src/main/scala/caliban/transformers/Transformer.scala:21:72: missing parameter type
[error]       val modifiedStep = self.transformStep.applyOrElse((step, field), _ => step)
[error]                                                                        ^
[error] /Users/pierre/GIT/caliban/core/src/main/scala/caliban/transformers/Transformer.scala:22:31: type mismatch;
[error]  found   : (caliban.schema.Step[Nothing], caliban.execution.Field)
[error]  required: (caliban.schema.Step[R1], caliban.execution.Field)
[error]       that.transformStep.lift((modifiedStep, field)).getOrElse(modifiedStep)
```

core/src/main/scala/caliban/transformers/Transformer.scala
ghostdogpr: I use `{ case (ObjectStep(name, fields), _)` a lot and having function would force me to have `case other => other`. On the other hand the usage is simpler.

core/src/main/scala/caliban/execution/Executor.scala
ghostdogpr: Transforming `Field.apply` would only modify the client request, but we need to change the step itself. A very simple example is if the client request `__typename` and you renamed that type. The name will come from the `ObjectStep`.

But your point about performance is very valid.

ghostdogpr: I could transform all the steps tree in advance and save it but the problem is in the `Extend` transformer I made, I also need access to `Field` when I transform the step 🤔 Need to think about this.

ghostdogpr: Maybe I can add `transformQuery` for my specific use case that uses `Field` and remove it from `transformStep`, then I can transform the steps in advance and keep the transformed value.

ghostdogpr: Ah even if I do it in advance it will run every time as soon as it's inside a `QueryStep` or `FunctionStep`...

core/src/main/scala/caliban/transformers/Transformer.scala
kyri-petrou: Agh yeah that'd be annoying indeed. I'm happy either way, just thought to bring it up in case composition was being a bit too cumbersome

core/src/main/scala/caliban/transformers/Transformer.scala
kyri-petrou: Aghh I see - I was trying things out using Scala 3 which seems to be okay with inference but yeah switching to Scala 2.12 / 2.13 seems to break things

core/src/main/scala/caliban/execution/Executor.scala
kyri-petrou: >Transforming Field.apply would only modify the client request, but we need to change the step itself. A very simple example is if the client request __typename and you renamed that type. The name will come from the ObjectStep.

Agh fair point, didn't think of that issue.

Just thinking out loud: I know you're going to work on a refactoring of how we build the graph, but in the meantime perhaps one way that we could avoid the performance overhead would be by having the transformers have some internal caching on whether they are known to transform / not transform a Step. This would require us however to have a way to assign a consistent hash to a step that doesn't need to be recalculated each time we're executing a request.

Can you think of a way that we could achieve that or would that be too difficult / not worth it given you're planning to refactor the graphql resolution?

kyri-petrou: Actually it doesn't need to be unique per se. It can be the hash code of the schema that created the Step perhaps?

ghostdogpr: It would still not work when it's depending on the field. Actually as you can see in the group chat I'm thinking of a completely different approach for handling the gateway feature because it is starting to feel wrong to reuse the current executor. If I do I guess transformers would move there too.

ghostdogpr:
Will adopt another approach

PR #1905, opened 2023-09-23
Add a simple wrapper to check directives

ghostdogpr:
It can be useful if you want to use an annotation to control permissions at the field level. Even if not super useful, it is a good example of what you can do with wrappers, and users can use that as a base to make their own.

### DIFF
--- a/core/src/main/scala/caliban/wrappers/Wrappers.scala
+++ b/core/src/main/scala/caliban/wrappers/Wrappers.scala
@@ -2,14 +2,15 @@ package caliban.wrappers
 
 import caliban.CalibanError.{ ExecutionError, ValidationError }
 import caliban.Value.NullValue
-import caliban.execution.{ ExecutionRequest, Field }
-import caliban.parsing.adt.Document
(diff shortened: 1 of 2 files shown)


paulpdaniels:
One limitation is that it doesn't work for pure fields right? I think a more general solution might be to use either a validation wrapper or an OverallWrapper if you want to allow partial failures

ghostdogpr:
It works if you use FieldWrapper with the Boolean turned to true (with a performance cost). With ValidationWrapper you would need to browse the whole tree yourself. My idea wasn't to make something super useful but more to show what's possible using wrappers.

kyri-petrou:
@ghostdogpr should we expose the `wrapPureValues: Boolean = false` (or even with `true` as the default) as an argument to the `checkDirectives` method? That way users can enable/disable that if they wish to take the performance hit without having to create another wrapper

ghostdogpr:
> @ghostdogpr should we expose the `wrapPureValues: Boolean = false` (or even with `true` as the default) as an argument to the `checkDirectives` method? That way users can enable/disable that if they wish to take the performance hit without having to create another wrapper

Done (named it the same way you just did in your PR)

PR #1906, opened 2023-09-23
Add federation 2.5 support

paulpdaniels:
https://www.apollographql.com/docs/federation/federated-types/federated-directives#controlling-access


ghostdogpr (approved):
We just add the directives, nothing special to do?

paulpdaniels:
Yeah, nothing special to do the gateway handles the behavior. Im thinking we could also add a utility version of this that lets people do add this kind of thing outside federation. But it should be kept separate so we can evolve the behavior of it

PR #1907, opened 2023-09-24
Optionally disable tracing of pure fields in `ApolloTracing` wrapper

kyri-petrou:
We previously discussed on Discord that it might be good to expose this as an option where users might prefer performance over correctness

### DIFF
--- a/benchmarks/src/main/scala/caliban/execution/NestedZQueryBenchmark.scala
+++ b/benchmarks/src/main/scala/caliban/execution/NestedZQueryBenchmark.scala
@@ -77,7 +77,7 @@ class NestedZQueryBenchmark {
     run(
       graphQL[Any, MultifieldRoot, Unit, Unit](
         RootResolver(NestedZQueryBenchmarkSchema.multifield1000Elements)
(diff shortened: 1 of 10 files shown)


ghostdogpr (approved):

ghostdogpr:
How about `ApolloFederatedTracing`?

kyri-petrou:
> How about `ApolloFederatedTracing`?

I didn't want to touch that since I wasn't sure whether it would cause any weird behaviour when queried fields are not included in the traces. Having said that, the same can be said about `ApolloTracing`. I'll add it and add a warning about it in both places

kyri-petrou:
@ghostdogpr done

ghostdogpr:
@paulpdaniels by any chance do you know if missing fields cause problems on Apollo side?

kyri-petrou:
@ghostdogpr what are your thoughts? Shall we merge this in?

I'm thinking that given that it's optional to disable the the wrapping of pure values, users can try it out and check whether it works with their setup so there isn't much harm done in adding this

ghostdogpr:
Sounds good 👍

ghostdogpr (approved):

PR #1909, opened 2023-09-24
Fix compliance warnings & suggestions

### LINKED ISSUE #1899: Full compliance with GraphQL over HTTP spec
See https://graphql-http.com/ and https://graphql.github.io/graphql-over-http/

```
GraphQL over HTTP audit report
60 audits in total
✅ 43 pass
💡 4 notices (suggestions)
⚠️ 13 warnings (optional)
Passing
4655 MUST accept application/json and match the content-type
47DE SHOULD accept */* and use application/json for the content-type
80D8 SHOULD assume application/json content-type when accept is missing
82A3 MUST use utf-8 encoding when responding
BF61 MUST accept utf-8 encoded request
78D5 MUST assume utf-8 in request if encoding is unspecified
2C94 MUST accept POST requests
5A70 MAY accept application/x-www-form-urlencoded formatted GET requests
9C48 MAY NOT allow executing mutations on GET requests
9ABE MAY respond with 4xx status code if content-type is not supplied on POST requests
03D4 MUST accept application/json POST requests
A5BF MAY use 400 status code when request body is missing on POST
LKJ0 MAY use 400 status code on object {query} parameter
LKJ1 MAY use 400 status code on number {query} parameter
LKJ2 MAY use 400 status code on boolean {query} parameter
LKJ3 MAY use 400 status code on array {query} parameter
13EE MUST allow string {query} parameter when accepting application/json
6C00 MAY use 400 status code on object {operationName} parameter
6C01 MAY use 400 status code on number {operationName} parameter
6C02 MAY use 400 status code on boolean {operationName} parameter
6C03 MAY use 400 status code on array {operationName} parameter
B8B3 MUST allow string {operationName} parameter when accepting application/json
0220 MUST allow null {variables} parameter when accepting application/json
0221 MUST allow null {operationName} parameter when accepting application/json
0222 MUST allow null {extensions} parameter when accepting application/json
4760 MAY use 400 status code on string {variables} parameter
4761 MAY use 400 status code on number {variables} parameter
4762 MAY use 400 status code on boolean {variables} parameter
4763 MAY use 400 status code on array {variables} parameter
28B9 MUST allow map {variables} parameter when accepting application/json
6A70 MAY allow URL-encoded JSON string {variables} parameter in GETs when accepting application/json
58B0 MAY use 400 status code on string {extensions} parameter
58B1 MAY use 400 status code on number {extensions} parameter
58B2 MAY use 400 status code on boolean {extensions} parameter
58B3 MAY use 400 status code on array {extensions} parameter
1B7A MUST allow map {extensions} parameter when accepting application/json
B6DC MAY use 4xx or 5xx status codes on JSON parsing failure
BCF8 MAY use 400 status code on JSON parsing failure
572B SHOULD use 200 status code on document parsing failure when accepting application/json
FDE2 SHOULD use 200 status code on document validation failure when accepting application/json
7B9B SHOULD use a status code of 200 on variable coercion failure when accepting application/json
865D SHOULD use 4xx or 5xx status codes on document parsing failure when accepting application/graphql-response+json
51FE SHOULD use 4xx or 5xx status codes on document validation failure when accepting application/graphql-response+json

Notices
The server MAY support these, but are truly optional. These are suggestions following recommended conventions.
423L MAY use 400 status code on missing {query} parameter
Response status code is not 400
D6D5 MAY allow URL-encoded JSON string {variables} parameter in GETs when accepting application/graphql-response+json
Response status code is not 200
8764 MAY use 4xx or 5xx status codes if parameters are invalid
Response status is not between 400 and 599
3E3A MAY use 400 status code if parameters are invalid
Response status code is not 400

Warnings
The server SHOULD support these, but is not required.
22EB SHOULD accept application/graphql-response+json and match the content-type
Response status code is not 200
34A2 SHOULD allow string {query} parameter when accepting application/graphql-response+json
Response status code is not 200
8161 SHOULD allow string {operationName} parameter when accepting application/graphql-response+json
Response status code is not 200
94B0 SHOULD allow null {variables} parameter when accepting application/graphql-response+json
Response status code is not 200
94B1 SHOULD allow null {operationName} parameter when accepting application/graphql-response+json
Response status code is not 200
94B2 SHOULD allow null {extensions} parameter when accepting application/graphql-response+json
Response status code is not 200
2EA1 SHOULD allow map {variables} parameter when accepting application/graphql-response+json
Response status code is not 200
428F SHOULD allow map {extensions} parameter when accepting application/graphql-response+json
Response status code is not 200
556A SHOULD use 400 status code on document parsing failure when accepting application/graphql-response+json
Response status code is not 400
D586 SHOULD not contain the data entry on document parsing failure when accepting application/graphql-response+json
Response body is not valid JSON
74FF SHOULD use 400 status code on document validation failure when accepting application/graphql-response+json
Response status code is not 400
5E5B SHOULD not contain the data entry on document validation failure when accepting application/graphql-response+json
Response body is not valid JSON
86EE SHOULD use a status code of 400 on variable coercion failure when accepting application/graphql-response+json
Response status code is not 400
```

kyri-petrou:
Followup of https://discord.com/channels/629491597070827530/633200096393166868/1154353360128987137

Closes https://github.com/ghostdogpr/caliban/issues/1899

This PR fixes the spec compliance warnings and applies (most) recommendations by:

1. Adding support when a client accepts the `application/graphql-response+json` content type
2. Responding with 400 status code on invalid graphql requests

The only recommendation that hasn't been applied is the `9C48 MAY NOT allow executing mutations on GET requests`. I didn't add support for this one for 2 reasons mainly:

1. Being able to determine whether a request is a mutation at the HTTP layer is tricky. If we want to perform this check, we'd somehow need to propagate the info to raise an error if the request is a mutation down to validation, and then convert this error to a 400 status code in the tapir adapter
2. Even if it's recommended not to accept mutations over GET requests, there is a chance that this might break _some_ existing clients.

### Spec compliance report

```
GraphQL over HTTP audit report
60 audits in total
✅ 59 pass
💡 1 notices (suggestions)
Passing
22EB SHOULD accept application/graphql-response+json and match the content-type
4655 MUST accept application/json and match the content-type
47DE SHOULD accept */* and use application/json for the content-type
80D8 SHOULD assume application/json content-type when accept is missing
82A3 MUST use utf-8 encoding when responding
BF61 MUST accept utf-8 encoded request
78D5 MUST assume utf-8 in request if encoding is unspecified
2C94 MUST accept POST requests
5A70 MAY accept application/x-www-form-urlencoded formatted GET requests
9ABE MAY respond with 4xx status code if content-type is not supplied on POST requests
03D4 MUST accept application/json POST requests
A5BF MAY use 400 status code when request body is missing on POST
423L MAY use 400 status code on missing {query} parameter
LKJ0 MAY use 400 status code on object {query} parameter
LKJ1 MAY use 400 status code on number {query} parameter
LKJ2 MAY use 400 status code on boolean {query} parameter
LKJ3 MAY use 400 status code on array {query} parameter
34A2 SHOULD allow string {query} parameter when accepting application/graphql-response+json
13EE MUST allow string {query} parameter when accepting application/json
6C00 MAY use 400 status code on object {operationName} parameter
6C01 MAY use 400 status code on number {operationName} parameter
6C02 MAY use 400 status code on boolean {operationName} parameter
6C03 MAY use 400 status code on array {operationName} parameter
8161 SHOULD allow string {operationName} parameter when accepting application/graphql-response+json
B8B3 MUST allow string {operationName} parameter when accepting application/json
94B0 SHOULD allow null {variables} parameter when accepting application/graphql-response+json
0220 MUST allow null {variables} parameter when accepting application/json
94B1 SHOULD allow null {operationName} parameter when accepting application/graphql-response+json
0221 MUST allow null {operationName} parameter when accepting application/json
94B2 SHOULD allow null {extensions} parameter when accepting application/graphql-response+json
0222 MUST allow null {extensions} parameter when accepting application/json
4760 MAY use 400 status code on string {variables} parameter
4761 MAY use 400 status code on number {variables} parameter
4762 MAY use 400 status code on boolean {variables} parameter
4763 MAY use 400 status code on array {variables} parameter
2EA1 SHOULD allow map {variables} parameter when accepting application/graphql-response+json
28B9 MUST allow map {variables} parameter when accepting application/json
D6D5 MAY allow URL-encoded JSON string {variables} parameter in GETs when accepting application/graphql-response+json
6A70 MAY allow URL-encoded JSON string {variables} parameter in GETs when accepting application/json
58B0 MAY use 400 status code on string {extensions} parameter
58B1 MAY use 400 status code on number {extensions} parameter
58B2 MAY use 400 status code on boolean {extensions} parameter
58B3 MAY use 400 status code on array {extensions} parameter
428F SHOULD allow map {extensions} parameter when accepting application/graphql-response+json
1B7A MUST allow map {extensions} parameter when accepting application/json
B6DC MAY use 4xx or 5xx status codes on JSON parsing failure
BCF8 MAY use 400 status code on JSON parsing failure
8764 MAY use 4xx or 5xx status codes if parameters are invalid
3E3A MAY use 400 status code if parameters are invalid
572B SHOULD use 200 status code on document parsing failure when accepting application/json
FDE2 SHOULD use 200 status code on document validation failure when accepting application/json
7B9B SHOULD use a status code of 200 on variable coercion failure when accepting application/json
865D SHOULD use 4xx or 5xx status codes on document parsing failure when accepting application/graphql-response+json
556A SHOULD use 400 status code on document parsing failure when accepting application/graphql-response+json
D586 SHOULD not contain the data entry on document parsing failure when accepting application/graphql-response+json
51FE SHOULD use 4xx or 5xx status codes on document validation failure when accepting application/graphql-response+json
74FF SHOULD use 400 status code on document validation failure when accepting application/graphql-response+json
5E5B SHOULD not contain the data entry on document validation failure when accepting application/graphql-response+json
86EE SHOULD use a status code of 400 on variable coercion failure when accepting application/graphql-response+json
Notices
The server MAY support these, but are truly optional. These are suggestions following recommended conventions.
9C48 MAY NOT allow executing mutations on GET requests
Response status is not between 400 and 499
```


interop/tapir/src/test/scala/caliban/interop/tapir/TestApi.scala
```diff
@@ -64,7 +64,6 @@ object TestApi extends GenericSchema[TestService with Uploads] {
       maxDepth(30) @@                 // query analyzer that limit query depth
       timeout(3 seconds) @@           // wrapper that fails slow queries
       printSlowQueries(500 millis) @@ // wrapper that logs slow queries
-      printErrors @@                  // wrapper that logs errors
```
kyri-petrou, line 67: I removed this since the tests were too noisy printing out all the (expected) errors from the added tests

paulpdaniels:
For the GET requests, we might be able to do something similar to Apollo federated tracing and map it into an extension. What if we made it part of the ExecutionContext that gets passed around? That's set as part of global fiberRef so it should be transparent to the user

kyri-petrou:
> For the GET requests, we might be able to do something similar to Apollo federated tracing and map it into an extension. What if we made it part of the ExecutionContext that gets passed around? That's set as part of global fiberRef so it should be transparent to the user

I'm a bit braindead after a long day of work, but I added this in the last commit. I think it can be potentially improved further, and happy to hear your feedback.

Note that I made the "validation" optional just in case there are clients that do want to enable support for this regardless of recommendations for best practises. With the latest commit:

image

interop/tapir/src/main/scala/caliban/interop/tapir/TapirAdapter.scala
```diff
@@ -82,27 +82,50 @@ object TapirAdapter {
...
+    request: ServerRequest
+  )(
     response: GraphQLResponse[E]
   )(implicit
     streamConstructor: StreamConstructor[BS],
     responseCodec: JsonCodec[ResponseValue]
-  ): (MediaType, CalibanBody[BS]) =
+  ): (MediaType, StatusCode, CalibanBody[BS]) = {
+    def acceptsGqlJson = request.acceptsContentTypes.fold(
+      _ => false, // NOTE: 1st January 2025 this should be changed to `true`
```
ghostdogpr: Maybe link to the reason for that? 😄

core/src/main/scala/caliban/GraphQL.scala
```diff
@@ -118,20 +126,22 @@ trait GraphQL[-R] { self =>
                                                                                 config.skipValidation,
                                                                                 config.validations
                                                                               )
+                                                              _            <- ZIO.unless(config.allowMutationsOverGetRequests)(checkHttpMethod(executionReq))
```
ghostdogpr: Can you check if it changes anything to benchmarks?

interop/tapir/src/main/scala/caliban/interop/tapir/TapirAdapter.scala
kyri-petrou: Done. Also changed the description a bit because it wasn't very accurate

core/src/main/scala/caliban/GraphQL.scala
kyri-petrou: I couldn't notice any difference in performance, but just to be absolutely sure, I ended up refactoring it a bit so that we only get the data from the `FiberRef` when the operation is a mutation. For queries, the method should be a no-op essentially

core/src/main/scala/caliban/HttpRequestMethod.scala
```diff
@@ -1,14 +1,23 @@
...
+
+  private val fiberRef: FiberRef[HttpRequestMethod] = Unsafe.unsafe(implicit u => FiberRef.unsafe.make(POST))
+
+  val get: UIO[HttpRequestMethod] = fiberRef.get
+
+  def setWith[R, E, B](method: HttpRequestMethod)(zio: ZIO[R, E, B]): ZIO[R, E, B] =
+    method match {
+      case POST => zio
+      case GET  => fiberRef.locally(method)(zio)
+    }
```
kyri-petrou, line 22: I changed this a little bit to hide the implementation detail of using a `FiberRef` and no-op for `POST` requests.

For `setWith`, it could also be implemented as below, but _I think_ `locally` would be more performant since it doesn't require adding `Scope` to the environment, and we can no-op it for `POST` requests

```scala
def set(method: HttpRequestMethod): URIO[Scope, Unit] = fiberRef.locallyScoped(method)
```

core/src/main/scala/caliban/HttpRequestMethod.scala
```diff
@@ -0,0 +1,23 @@
...
+import caliban.CalibanError.ValidationError
+import zio.{ FiberRef, Scope, UIO, URIO, Unsafe, ZIO }
+
+private[caliban] sealed trait HttpRequestMethod
+
+private[caliban] object HttpRequestMethod {
+  case object GET  extends HttpRequestMethod
+  case object POST extends HttpRequestMethod
+
+  val MutationOverGetError: ValidationError = ValidationError("Mutations are not allowed for GET requests", "")
```
ghostdogpr, line 12: Btw seeing this make me think we should probably extend `NoStackTrace` for `ValidationError` at least. What do you think?

ghostdogpr (approved):

core/src/main/scala/caliban/HttpRequestMethod.scala
kyri-petrou: Quite possibly yes. Actually I think we should probably do that with all the errors since we override `getCause` in `ParsingError` and `ExecutionError`, and the stacktrace of the cause is included in the reported error.

The only downside to that is that `getCause` will now be `null` for `ValidationError`, which might cause some some NPEs to users if they've been reckless with it. I think that should be fine though as long as we point it out in the release notes. 

Also, if one really needs the stack traces, they can set the `scala.control.noTraceSuppression` system property to `true`, which effectively makes `NoStackTrace` to behave like a normal `Throwable`.

I can make the changes in a followup PR

ghostdogpr: Sounds good!

kyri-petrou:
@ghostdogpr on second thought, I'm going try placing the Http method validation within `Validator`, and adding the validation as part of the list of validations we pass into `Validator.prepare`.

I'll let you know once I have that ready

kyri-petrou:
@ghostdogpr Okay after giving it a go, I don't think it's worth it, but thought to put that code in a separate branch just to see if we all agree: #1923

ghostdogpr (approved):
🎉

PR #1910, opened 2023-09-26
Update zio-http to RC2

kyri-petrou:
~Reopening #1744~

Huh, seems that the tapir WS support works - or at least that the tests are passing


ghostdogpr (approved):
Very happy about the websocket part

frekw:
Nice!

Hopefully we get 1.0 soon. Quite a lot has changed between RC2 and SNAPSHOT as well so we'll likely have to do something similar again soon. 😓

frekw (approved):

PR #1922, opened 2023-10-04
Another round of optimizations

kyri-petrou:
## Main optimizations:

- [Scala 3] Use a `Vector` instead of a `List` in derived `Sum` schemas. The benefit of this is that the `.apply` method on Vectors is O(1) whereas on Lists is O(N). The benefit of this becomes obvious only in presence of enums with many values
- Avoid checking whether to wrap pure values each time we reduce an object
- Add `allFieldsMap` lazy val on `__Type` to replace `.find` calls. This provides benefits mostly to objects with large number of fields as it avoids iterating through them all
- Use `MurmurHash3.unorderedHash` to calculate the hashCode of `ResponseValue.ObjectValue` the conversion of to Set. I'm not sure about the total impact of this since it's hard to identify how much it's been used.

## Minor / ultra-minor:

- Avoid conversions in `Executor.mergeFields` when no modifications have been made
- [Scala 3] Avoid recalculating all the field names each time `resolve` is called
- Split `resolve` method of derived Schemas into individual methods to aid JDK code caching
- Convert methods in anonymous classes to `private`
- Add `aliasedName` val on `Field` to avoid repeated calls to `alias.getOrElse(name)`
- Probably a bunch more that I missed

### DIFF
--- a/core/src/main/scala/caliban/execution/Executor.scala
+++ b/core/src/main/scala/caliban/execution/Executor.scala
@@ -58,10 +58,10 @@ object Executor {
           value match {
             case EnumValue(v) =>
               // special case of an hybrid union containing case objects, those should return an object instead of a string
-              val obj = currentField.fields.collectFirst {
-                case f: Field if f.name == "__typename" =>
+              val obj = currentField.fields.view.filter(_._condition.forall(_.contains(v))).collectFirst {
+                case f if f.name == "__typename" =>
                   ObjectValue(List(f.aliasedName -> StringValue(v)))
-                case f: Field if f.name == "_"          =>
+                case f if f.name == "_"          =>
                   NullValue
               }
               obj.fold(s)(PureStep(_))
--- a/core/src/main/scala/caliban/execution/Executor.scala
+++ b/core/src/main/scala/caliban/execution/Executor.scala
@@ -290,7 +290,9 @@ object Executor {
             val f = array(index)
             array(index) = f.copy(fields = f.fields ::: field.fields)
         }
-      } else modified = true
+      } else {
+        modified = true
+      }
     }
 
     // Avoid materializing the buffer if no modification took place
@@ -301,19 +303,19 @@ object Executor {
     FieldInfo(field.aliasedName, field, path, fieldDirectives, field.parentType)
 
   private def reduceList[R](list: List[ReducedStep[R]], areItemsNullable: Boolean): ReducedStep[R] = {
-    var isPure = true
+    var allPure = true
     val pures  = List.newBuilder[ResponseValue]
     var tail   = list
-    while (isPure && tail.nonEmpty)
+    while (allPure && tail.nonEmpty)
       tail.head match {
         case step: PureStep =>
           pures += step.value
           tail = tail.tail
         case _              =>
-          isPure = false
+          allPure = false
       }
 
-    if (isPure) PureStep(ListValue(pures.result()))
+    if (allPure) PureStep(ListValue(pures.result()))
     else ReducedStep.ListStep(list, areItemsNullable)
   }
 
@@ -321,7 +323,7 @@ object Executor {
     items: List[(String, ReducedStep[R], FieldInfo)],
     fieldWrappers: List[FieldWrapper[R]]
   ): ReducedStep[R] =
-    if (!fieldWrappers.exists(_.wrapPureValues) && items.forall(_._2.isInstanceOf[PureStep]))
+    if (!fieldWrappers.exists(_.wrapPureValues) && items.forall(_._2.isPure))
       PureStep(ObjectValue(items.asInstanceOf[List[(String, PureStep, FieldInfo)]].map { case (k, v, _) =>
         (k, v.value)
       }))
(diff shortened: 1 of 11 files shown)


core/src/main/scala/caliban/execution/Executor.scala
```diff
@@ -57,9 +58,9 @@ object Executor {
           value match {
             case EnumValue(v) =>
               // special case of an hybrid union containing case objects, those should return an object instead of a string
-              val obj = mergeFields(currentField, v).collectFirst {
+              val obj = currentField.fields.collectFirst {
```
kyri-petrou: @ghostdogpr do you see a reason why we'd need to merge the fields of an enum value here? Tests seem to pass without it and I'm not sure I can think of a case where we'd need to merge them.

If you can think of a case where this is needed let me know and I'll rollback the change and add a test instead

core/src/main/scala/caliban/execution/Field.scala
```diff
@@ -217,31 +214,34 @@ object Field {
 
   private def resolveDirectiveVariables(
     variableValues: Map[String, InputValue],
-    variableDefinitions: List[VariableDefinition]
+    variableDefinitions: Map[String, VariableDefinition]
   )(directive: Directive): Directive =
     if (directive.arguments.isEmpty) directive
     else directive.copy(arguments = resolveVariables(directive.arguments, variableDefinitions, variableValues))
 
+  private val emptyVariables = Map.empty[String, InputValue]
```
ghostdogpr: `Map.empty` already points to an object so you don't need to save it into a val.

core/src/main/scala/caliban/execution/Executor.scala
```diff
@@ -302,9 +302,9 @@ object Executor {
 
   private def reduceObject[R](
     items: List[(String, ReducedStep[R], FieldInfo)],
-    fieldWrappers: List[FieldWrapper[R]]
+    wrapPureValues: Boolean
   ): ReducedStep[R] =
-    if (!fieldWrappers.exists(_.wrapPureValues) && items.map(_._2).forall(_.isInstanceOf[PureStep]))
+    if (!wrapPureValues && items.forall(_._2.isPure))
```
ghostdogpr, line 307: Much nicer!

core/src/main/scala/caliban/execution/Field.scala
```diff
@@ -43,6 +43,8 @@ case class Field(
 ) { self =>
   lazy val locationInfo: LocationInfo = _locationInfo()
 
+  private[caliban] val aliasedName: String = alias.getOrElse(name)
```
ghostdogpr, line 46: Nice! I do it all the time.

core/src/main/scala/caliban/execution/Executor.scala
ghostdogpr: Yeah I don't think it's needed considering the `collectFirst` right after.

core/src/main/scala/caliban/execution/Executor.scala
```diff
@@ -268,32 +269,31 @@ object Executor {
 
   private[caliban] def mergeFields(field: Field, typeName: String): List[Field] = {
     // ugly mutable code but it's worth it for the speed ;)
-    val array = ArrayBuffer.empty[Field]
-    val map   = collection.mutable.Map.empty[String, Int]
-    var index = 0
+    val map      = new java.util.LinkedHashMap[String, Field]()
```
ghostdogpr, line 272: Did you measure this change?

ghostdogpr (commented):
Very nice!

core/src/main/scala/caliban/execution/Executor.scala
kyri-petrou: I didn't measure it this time around, but I did when I changed the same logic in `caliban.execution.Field`. The performance is the same with both implementations, with the difference that using a `LinkedHashMap` is cleaner. Resources online seem to suggest that the performance of `LinkedHashMap` and `HashMap` should be roughly identical as well

core/src/main/scala/caliban/execution/Executor.scala
kyri-petrou: Yeah that's what I thought as well. The only thing that might be needed is the filtering we do within `mergeFields`. Do you think that might be needed here?

```scala
if (field._condition.forall(_.contains(typeName))) { ... }
```

ghostdogpr: Ah... maybe if there is a fragment that applies only to a subtype of the enum?

ghostdogpr (approved):
Except the `mergeFields` that I'm not 100% sure about, everything looks great

core/src/main/scala/caliban/execution/Executor.scala
kyri-petrou: Just to be safe, I added back the filtering using a `.view` without  calling `mergeFields`. This way we avoiding filtering the entire collection in cases that it's not needed. I think the performance impact should be extremely minimal either way

ghostdogpr: Sounds good! And feel free to merge when CI is green 👍

PR #1923, opened 2023-10-05
Alt http method validation implementation

kyri-petrou:
After giving this a go, I don't think I like it. It requires too many breaking changes, as well as making `HttpRequestMethod` public, and the only _real_ benefit is that the validation now is now part of `AllValidations`. 

Leaving this here just to check whether we think this is worth it or not

### DIFF
--- a/core/src/main/scala/caliban/GraphQL.scala
+++ b/core/src/main/scala/caliban/GraphQL.scala
@@ -1,7 +1,6 @@
 package caliban
 
 import caliban.CalibanError.ValidationError
-import caliban.Configurator.ExecutionConfiguration
 import caliban.execution.{ ExecutionRequest, Executor, Feature }
 import caliban.introspection.Introspector
 import caliban.introspection.adt._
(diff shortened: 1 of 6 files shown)


core/src/main/scala/caliban/GraphQL.scala
```diff
@@ -120,16 +109,17 @@ trait GraphQL[-R] { self =>
                   validate                              = (doc: Document) =>
                                                             for {
                                                               config       <- Configurator.configuration
+                                                              reqMethod    <- HttpRequestMethod.get
```
ghostdogpr, line 112: This call can be moved inside of `prepare`, no?

What are the breaking changes?

kyri-petrou: Yes - I initially had it in there but somehow something feels weird for the behaviour of the `.prepare` method to be dependent on an argument that is provided internally via a `FiberRef`.

Only other breaking change is for `Caliban.validation.Context`, but I guess that's a tolerable one because it's very unlikely that users will be using it

kyri-petrou: There's also the downside that the `HttpRequestMethod` needs to be made public because `Context` is public (unless we use a boolean flag `isGetRequest` instead).

ghostdogpr: Okay let's keep the other PR then

Issue #1668, opened 2023-03-30
Allow mixing interfaces and unions in schema code generation


ghostdogpr:
Make this schema supported in server codegen:
```graphql
schema {
  query: Query
}

union AllErrors = Bar | Foo | FooBar

interface Error {
  message: String!
}

type Bar implements Error {
  message: String!
}

type Foo implements Error {
  message: String!
}

type FooBar implements Error {
  message: String!
}

type Query {
  errorInterface: Error!
  errorUnion: AllErrors!
}
```

Code should be something like this (not sure which companion object to use, the interface or the union):
```scala
case class Query(errorInterface: Error, errorUnion: AllErrors)

@GQLInterface
sealed trait Error {
  def message: String
}

sealed trait AllErrors

case class Foo(message: String) extends Error with AllErrors
case class Bar(message: String) extends Error with AllErrors
case class FooBar(message: String) extends Error with AllErrors
```

jmbeach:
This would be very nice for my use case. My schema uses quite a bit of implements etc

ghostdogpr:
Solved by https://github.com/ghostdogpr/caliban/pull/1925

PR #1925, opened 2023-10-05
Improve codegen

oyvindberg:
Hi there

I'm trying out caliban, and found that the codegen had some bugs, particularly related to interfaces. Here are some tests and rather quick and dirty patches for the problems I saw.

## inheriting fields with arguments from interface
I guess this was never tried. now it doesn't infer new `Args` types for the subtypes but reuses the one from the interface. Or well, there were none for the interface, but let's add those instead

## Missing prefix in scala code for object type namespaced by one type union

I don't like the logic for putting object types in namespaces and would prefer it all at the top-level with multiple extends, honestly. But the logic being what it is, this replicates some computation of prefixes from interfaces for type unions

## Missing `derives` for interface type
Scala 3 complained about this, so here is a fix.

--- 

I can break this up into consecutive PR's if needed, but since the first attempt gave conflicts I bundled them together

### DIFF
--- a/tools/src/main/scala/caliban/tools/SchemaWriter.scala
+++ b/tools/src/main/scala/caliban/tools/SchemaWriter.scala
@@ -164,7 +164,7 @@ object SchemaWriter {
 
     def writeInterface(interface: InterfaceTypeDefinition, impls: List[ObjectTypeDefinition]): String =
       s"""@GQLInterface
-        ${writeDescription(interface.description)}sealed trait ${interface.name} extends scala.Product with scala.Serializable {
+        ${writeDescription(interface.description)}sealed trait ${interface.name} extends scala.Product with scala.Serializable $derivesSchema {
          ${interface.fields.map(field => "def " + writeField(field, interface)).mkString("\n")}
         }
 
--- a/tools/src/main/scala/caliban/tools/SchemaWriter.scala
+++ b/tools/src/main/scala/caliban/tools/SchemaWriter.scala
@@ -1,8 +1,11 @@
 package caliban.tools
 
+import caliban.parsing.adt.Definition.TypeSystemDefinition.TypeDefinition
 import caliban.parsing.adt.Definition.TypeSystemDefinition.TypeDefinition._
 import caliban.parsing.adt.Type.{ ListType, NamedType }
 import caliban.parsing.adt.{ Document, Type }
+
+import scala.collection.compat._
 
 object SchemaWriter {
 
@@ -22,13 +25,23 @@ object SchemaWriter {
     val derivesSchemaAndArgBuilder: String =
       if (addDerives) " derives caliban.schema.Schema.SemiAuto, caliban.schema.ArgBuilder" else ""
 
-    val interfaceImplementationsMap = (for {
+    val interfaceImplementationsMap: Map[InterfaceTypeDefinition, List[ObjectTypeDefinition]] = (for {
       objectDef    <- schema.objectTypeDefinitions
       interfaceDef <- schema.interfaceTypeDefinitions
       if objectDef.implements.exists(_.name == interfaceDef.name)
     } yield interfaceDef -> objectDef).groupBy(_._1).map { case (definition, tuples) =>
       definition -> tuples.map(_._2)
     }
+
+    val interfacesExtendedForObject: Map[ObjectTypeDefinition, List[InterfaceTypeDefinition]] =
+      interfaceImplementationsMap.iterator.flatMap { case (i, os) => os.map(o => (o, i)) }.toList.groupMap {
+        case (o, _) => o
+      } { case (_, i) => i }
+
+    def inheritedFromInterface(obj: ObjectTypeDefinition, field: FieldDefinition): Option[InterfaceTypeDefinition] =
+      interfacesExtendedForObject.get(obj) flatMap { interfaces =>
+        interfaces.find(_.fields.exists(_.name == field.name))
+      }
 
     def reservedType(typeDefinition: ObjectTypeDefinition): Boolean =
       typeDefinition.name == "Query" || typeDefinition.name == "Mutation" || typeDefinition.name == "Subscription"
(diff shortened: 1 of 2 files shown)


tools/src/main/scala/caliban/tools/SchemaWriter.scala
```diff
@@ -1,9 +1,12 @@
 package caliban.tools
 
+import caliban.parsing.adt.Definition.TypeSystemDefinition.TypeDefinition
 import caliban.parsing.adt.Definition.TypeSystemDefinition.TypeDefinition._
 import caliban.parsing.adt.Type.{ ListType, NamedType }
 import caliban.parsing.adt.{ Document, Type }
 
+import scala.collection.compat.toTraversableLikeExtensionMethods
```
ghostdogpr: It looks like this import only works on 2.12. Maybe import `._` if it's not needed for 2.13 and 3?

ghostdogpr (commented):
Thanks for the contribution! 

By the way, if you're interested in changing the nested objects to be top-level, I think that makes sense. Having them nested causes some issues like when you have both a union and an interface https://github.com/ghostdogpr/caliban/issues/1668

tools/src/main/scala/caliban/tools/SchemaWriter.scala
oyvindberg: 👍

oyvindberg:
> By the way, if you're interested in changing the nested objects to be top-level, I think that makes sense. Having them nested causes some issues like when you have both a union and an interface https://github.com/ghostdogpr/caliban/issues/1668

I think I'll give this a shot today. 

There is another reason why I'd like flat namespace beyond complexity in the codegen. At that point we're very close to outputting graphql union types as scala union types for scala 3. Have you looked into it? I can open a separate issue for that that where we can discuss it.

My goal for all this is to generate server-side stubs by the way. I want to start with a graphql schema and implement the server. Have you done any work in that direction? I can open an issue for that too unless there is one.

ghostdogpr:
> There is another reason why I'd like flat namespace beyond complexity in the codegen. At that point we're very close to outputting graphql union types as scala union types for scala 3. Have you looked into it? I can open a separate issue for that that where we can discuss it.

The problem is that Scala 3 doesn't generate `Mirror` for union types, so we can't use typeclass derivation for it: https://github.com/lampepfl/dotty/pull/15279

> My goal for all this is to generate server-side stubs by the way. I want to start with a graphql schema and implement the server. Have you done any work in that direction? I can open an issue for that too unless there is one.

You mean like generate a fake implementation automatically?

oyvindberg:
> The problem is that Scala 3 doesn't generate Mirror for union types, so we can't use typeclass derivation for it: https://github.com/lampepfl/dotty/pull/15279

That's true. But, and I'm on very thin ice here, I *think* we can write a macro which pattern matches on the union type, picks out all the members and `summonAll`s `Schema`s for them. 

Then we need to generate some code like this:
```scala
  type SearchResult = Human | Droid | Starship

  given Schema[Any, SearchResult] = Schema.typeUnion[SearchResult]

  // which would expand to something like this
  given Schema[Any, SearchResult] with {
    val _1: Schema[Any, Human] = summon[Schema[Any, Human]]
    val _2: Schema[Any, Droid] = summon[Schema[Any, Droid]]
    val _3: Schema[Any, Starship] = summon[Schema[Any, Starship]]
    val subTypes = List(_1, _2, _3)

    def resolve(value: SearchResult): caliban.schema.Step[Any] =
      value match {
        case x: Human => _1.resolve(x)
        case x: Droid => _2.resolve(x)
        case x: Starship => _3.resolve(x)
      }

    def toType(isInput: Boolean, isSubscription: Boolean): caliban.introspection.adt.__Type =
      caliban.schema.Types.makeUnion(Some("SearchResult"), None, subTypes.map(_.toType_(isInput, isSubscription)))
  }

```


> You mean like generate a fake implementation automatically?

No, I mean generate a type-safe shim I can use to provide an implementation. Exactly what you would get if somebody gave you a graphql file and you implement it manually in caliban. 

The code I want is 99% what comes out of the codegen today, but the complication is that AFAIU we need to evaluate parts of the response lazily based on the query we're processing, that is some of the members of the case classes should be `UIO[Foo]` instead of `Foo` if they for instance correspond to a `join` in the database which is only used to compute some subset of fields. And this cannot be inferred from the schema itself, so I want to draft some kind of strategy you can plug in to the codegen to affect this.

Effectively a function `(objectName,fieldName) => Effectful | Pure`

Do you get what I'm aiming for? Thoughts?

oyvindberg:
The strategy I'd like to implement by the way would be "make all fields tagged with a directive called `X` effectful", where you can configure `X`. I imagine it could be called `@lazy`, `@slow`, or something like this. This would have the splendid side-effect of indicating in the API itself that fetching a given field may incur additional runtime-costs. A somewhat leaky abstraction, but a good tool if you're worried about performance for the graphql part of the system.

ghostdogpr:
> That's true. But, and I'm on very thin ice here, I _think_ we can write a macro which pattern matches on the union type, picks out all the members and `summonAll`s `Schema`s for them.

I see. I've been wary of macros so far because they can be a pain to maintain (I'm not proficient at all with it) but if it's something simple enough, why not. @kyri-petrou that might be of interest to you 😄 

> No, I mean generate a type-safe shim I can use to provide an implementation. Exactly what you would get if somebody gave you a graphql file and you implement it manually in caliban.

Ah okay, I was confused because it's what it's supposed to be already 😄 It was initially designed as a "bootstrap" feature to generate code for getting started and then you would modify that code. But I think more and more people are using it like you want (and it makes sense since the resolvers are defined separately). What you suggest totally makes sense.

oyvindberg:
I pushed a new version of the branch now which removes the namespacing of object types inside companions for type aliases and interfaces. Likely solves #1668

oyvindberg:
I created a separate issue for union types as union types, and won't touch it further in this PR.


If you're happy with what you're seeing here, I'll continue towards what I outlined above for generating server stubs with a strategy for choosing effectful fields in a follow-up PR

ghostdogpr:
Do you mind adding a test case for #1668? Just to make sure it's resolved.

oyvindberg:
Added 👍

ghostdogpr:
Thanks!

PR #1927, opened 2023-10-06
codegen: Recognize @lazy intention to generate side-effecting fields

oyvindberg:
The scenario here is code generation of server code from graphql file. Context and proposal in #1925 

If you generate code for server usage before this change, you'll need to evaluate and fetch all data to build a given response case class for a query.

However, Caliban supports suspending values of some fields in an effect in order to only evaluate those values if the client asked for the given field.

`EffectStrategy` allows you to plug in a mechanism to decide which fields should be wrapped in such a side-effect.

Let me know what you think of this direction @ghostdogpr ! :)


tools/src/test/scala/caliban/tools/SchemaWriterSpec.scala
```diff
@@ -630,6 +627,75 @@ object SchemaWriterSpec extends ZIOSpecDefault {
...
+      gen(
+        """
+        |directive @lazy on FIELD_DEFINITION
+        |
+        |type Foo {
+        |  bar: String!
+        |  baz: String! @lazy
+        |}""",
+        effectStrategy = EffectStrategy.EffectfulIfTaggedWith("lazy")
+      ),
```
kyri-petrou: I'm thinking whether we should consider all fields marked with `@lazy` as effectful, regardless of the effectStrategy. I can't really imagine any other usage of that directive that would mean something different

tools/src/main/scala/caliban/tools/EffectStrategy.scala
```diff
@@ -0,0 +1,47 @@
...
+      if (field.directives.exists(d => d.name == directiveName)) Pure else Effectful
+  }
+
+  object AlwaysPure extends EffectStrategy {
+    override def decide(owner: TypeDefinition, field: FieldDefinition): Maybe = EffectStrategy.Pure
+  }
+
+  object AlwaysEffectful extends EffectStrategy {
+    override def decide(owner: TypeDefinition, field: FieldDefinition): Maybe = EffectStrategy.Effectful
+  }
```
kyri-petrou: I think we should add some scaladoc here warning the user about the potential performance impact that this will have. (similar to FieldWrapper) https://github.com/ghostdogpr/caliban/blob/80ad988a401ce2903f834fcadd21d928937bfbe3/core/src/main/scala/caliban/wrappers/Wrapper.scala#L90

tools/src/test/scala/caliban/tools/SchemaWriterSpec.scala
oyvindberg: I didn't want to standardize on a name in this first version.

So to be clear: In order for `EffectStrategy.EffectfulIfTaggedWith` have an effect currently:
- you need to set it up with a name, `lazy` in this case
- define the `lazy` annotation yourself
- tag a field with it

An alternative approach would be to standardize on one name, and not make it configurable

tools/src/main/scala/caliban/tools/EffectStrategy.scala
oyvindberg: I can do that. We can also delete it, I'm not sure if it's useful in practise. it was just very work to add it

tools/src/test/scala/caliban/tools/SchemaWriterSpec.scala
ghostdogpr: Maybe just choosing a name is simpler? And it has the advantage of working out of the box with the sbt plugin (exposing options to the plug-in is rather painful).

kyri-petrou: I think let's use either `lazy` or `lazyField`, and _if_ users want to be able to customise it for some reason we can add the customisation back.

On that note, what do you think about adding an input argument to that directive with a default value `(effectful: Boolean = true)`, and generate a `() => A` instead of an effect when that is set to `false`?

ghostdogpr: Why not use `effectType: Option[String]` which lets you specify the effect type. That way you can mix different ones like `Task`/`UIO` or `ZIO`/`ZQuery`, which is more powerful than the current way. And `None` could be for `() =>`

kyri-petrou: Yeah that sounds like a better idea

oyvindberg: Some followups before I get back to work on this:

> Maybe just choosing a name is simpler? And it has the advantage of working out of the box with the sbt plugin (exposing options to the plug-in is rather painful). 

First complication: we only want to generate effect types for *server* code, it does not make sense on the client. So some parameterization is needed for the codegen either way. I can inline the `EffectStrategy.EffectfulIfTaggedWith` strategy, but how would you want the client/server distinction to be made?

> I think let's use either lazy or lazyField, and if users want to be able to customise it for some reason we can add the customisation back.

I think `lazy` is a good name for scala developers, but I don't think it communicates well to other people who consume the schema with this directive thrown in. I also don't have a better alternative at the top of my head, which is why I wanted to delay that decision and let it be user-configurable. Other suggestions?

> Why not use effectType: Option[String] which lets you specify the effect type. That way you can mix different ones like Task/UIO or ZIO/ZQuery, which is more powerful than the current way. And None could be for () =>

Do you mean to expose `effectType` in the _directive_? It seems to me like oversharing a lot in the public API.

I would want this to bake a bit honestly before we commit to it.
- we'll need the directive to apply to top-level queries as well. Those are unconditionally effectful currently, so the mechanism may be a bit different.
- need a fallback if effect type is not specified
- I think there will be some complexity around mixing effect types. `F` needs to be surrounded by a `class`, `IO` needs non-global type-class evidence in order to output `derives Schema`, `ZQuery` has multiple type arguments.

ghostdogpr: I'm fine with just adding a simple directive for now and see if people express the need to customize the effect type later. Also agree it makes sense to make it configurable.

kyri-petrou: > Do you mean to expose effectType in the directive? It seems to me like oversharing a lot in the public API.

Since these directives are used only for codegen customization, do you think we could (should?) exclude them from introspection?

Having said that, I agree with @ghostdogpr, let's start with something simple and we can add more customization in the future

oyvindberg: ok, I simplified the PR now. deleted `EffectStrategy` and made the codegen look for a `lazy` annotation. I realized the client sourcegen was a completely separate thing now :) 

WDYT?

tools/src/main/scala/caliban/tools/SchemaWriter.scala
```diff
@@ -63,7 +63,7 @@ object SchemaWriter {
       "%s:%s ZStream[Any, Nothing, %s]".format(
         safeName(field.name),
         if (field.args.nonEmpty) s" ${argsName(field, od)} =>" else "",
-        writeType(field.ofType)
+        writeMaybeEffectType(od, field)
```
ghostdogpr: I think this one should not be wrapped since it's already inside `ZStream`

oyvindberg: Good point, will fix it

oyvindberg: done

ghostdogpr (approved):
LGTM!

PR #1929, opened 2023-10-07
Executor optimizations

kyri-petrou:
I was having a look through the code the other day and we had some unnecessary flatmaps in the `makeQuery` method. I've also did some minor optimizations regarding the sequence of cases pattern matching, but those don't as much benefit as the changes within `makeQuery`. 

Benchmarks before and after (note I that I only run against the `Batched` mode, but can do it for parallel as well if needed).

TL;DR: We're seeing around 15% improvement in execution time 🎉 

Before:
```
[info] Benchmark                                          Mode  Cnt    Score   Error  Units
[info] NestedZQueryBenchmark.deepBatchedQuery1000        thrpt    5  361.232 ± 5.242  ops/s
[info] NestedZQueryBenchmark.multifieldBatchedQuery1000  thrpt    5  529.950 ± 6.783  ops/s
```

After:
```
[info] Benchmark                                          Mode  Cnt    Score    Error  Units
[info] NestedZQueryBenchmark.deepBatchedQuery1000        thrpt    5  438.154 ±  4.434  ops/s
[info] NestedZQueryBenchmark.multifieldBatchedQuery1000  thrpt    5  632.729 ± 13.082  ops/s
```

### DIFF
--- a/core/src/main/scala/caliban/execution/Executor.scala
+++ b/core/src/main/scala/caliban/execution/Executor.scala
@@ -54,28 +54,18 @@ object Executor {
       path: List[Either[String, Int]]
     ): ReducedStep[R] =
       step match {
-        case s @ PureStep(value)            =>
-          value match {
-            case EnumValue(v) =>
-              // special case of an hybrid union containing case objects, those should return an object instead of a string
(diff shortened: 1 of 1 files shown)


ghostdogpr (approved):
Very nice!

PR #1932, opened 2023-10-09
Support deprecation of input values

### LINKED ISSUE #1848: Support deprecation of InputValues
The support for deprecation of input values was added to the spec about a year a ago (https://github.com/graphql/graphql-spec/pull/805) would it be possible to modify [InputValue.scala#L8-L12 ](https://github.com/ghostdogpr/caliban/blob/series/2.x/core/src/main/scala/caliban/introspection/adt/__InputValue.scala#L8-L12) in a way so that it also honours the deprecation decorator (`GQLDeprecated`)?

ghostdogpr:
Closes https://github.com/ghostdogpr/caliban/issues/1848

I intentionally didn't change `IntrospectionClient` because it would break on servers that don't support it (similarly to repeated directives). Let's consider it once the spec is released and adopted.


paulpdaniels (approved):
LGTM

kyri-petrou (approved):
Nice!

PR #1933, opened 2023-10-10
Add option to dynamically enable / disable `ApolloTracing`

kyri-petrou:
The motivation behind this change is that tracing adds a significant overhead to requests, both in terms of execution time and response payload size.

With this change, the user can control dynamically whether tracing will be enabled for the current execution. This opens up possibilities to dynamically enable / disable tracing via the `HttpInterpreter` or HTTP middleware which optionally enable tracing based on HTTP headers.

~~PS: If we like this idea, should we also make it available for the federation tracing wrapper?~~

Actually I just realised that `ApolloFederatedTracing` is already controllable via the graphql extensions.

### DIFF
--- a/core/src/main/scala/caliban/wrappers/ApolloTracing.scala
+++ b/core/src/main/scala/caliban/wrappers/ApolloTracing.scala
@@ -15,6 +15,8 @@ import zio.query.ZQuery
 
 object ApolloTracing {
 
+  private val isEnabledRef = Unsafe.unsafe(implicit u => FiberRef.unsafe.make(true))
+
   /**
    * Returns a wrapper that adds tracing information to every response
    * following Apollo Tracing format: https://github.com/apollographql/apollo-tracing.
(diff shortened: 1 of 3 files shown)


core/src/main/scala/caliban/wrappers/Wrapper.scala
```diff
@@ -41,6 +41,16 @@ object Wrapper {
     def wrap[R1 <: R](f: Info => ZIO[R1, E, A]): Info => ZIO[R1, E, A]
   }
 
+  /**
+   * A wrapper that doesn't do anything.
+   * Useful for cases where we want to programmatically decide whether we'll use a wrapper or not
+   */
+  val empty: Wrapper[Any] = new OverallWrapper[Any] {
```
ghostdogpr: Override `|+|` to return only the right part?

core/src/main/scala/caliban/wrappers/ApolloTracing.scala
```diff
@@ -15,26 +15,45 @@ import zio.query.ZQuery
 
 object ApolloTracing {
 
+  private val isEnabledRef = Unsafe.unsafe(implicit u => FiberRef.unsafe.make(true))
```
ghostdogpr, line 18: Should it be lazy?

core/src/main/scala/caliban/wrappers/Wrapper.scala
kyri-petrou: I ended up creating an object for it which overrides both `|+|` and `apply`. Let me know what you think.

On that note, a bit more philosophical question: should it be called `Empty` or something like `Id` / `Identity`?

core/src/main/scala/caliban/wrappers/ApolloTracing.scala
kyri-petrou: In all other places where we use `FiberRef`s we define them as `val`. Since objects are loaded lazily on the JVM, I think it's fine if we have it as a `val` since it'd be initialised when we first access the object to create the wrapper rather than during execution?

Do you think they should be lazy instead?

core/src/main/scala/caliban/wrappers/Wrapper.scala
ghostdogpr: I like `Empty`, I used the same pattern here: https://github.com/ghostdogpr/caliban/blob/58c0d76fcc8ec894e5ffc727d59511dc343a425e/core/src/main/scala/caliban/introspection/adt/__Type.scala#L164

core/src/main/scala/caliban/wrappers/ApolloTracing.scala
ghostdogpr: Ah right, I thought it was going to be loaded even though we don't use it. Looks fine then.

core/src/main/scala/caliban/wrappers/Wrapper.scala
kyri-petrou: Ah nice :) Should I be making it into a `case object` instead? I'm not too sure what is the practical difference.

ghostdogpr: it makes `toString` nicer 😄 I don't think there's any other benefit

kyri-petrou: Looking at the implementations of `Nil`, `None` as well as your `TypeVisitor.Empty`, they all seem to be `case object`s so I'll use that instead

ghostdogpr: https://scastie.scala-lang.org/S8QMz72CRlWJQiBof5b0fw

ghostdogpr: Not super critical but sometimes when debugging it helps a bit.

kyri-petrou: Another thing I noticed is that it makes the `.hashCode()` method consistent when executing the same program multiple times. I'm not sure whether this has any real value though

ghostdogpr (approved):

paulpdaniels (approved):

PR #1934, opened 2023-10-12
docs: Add section on server codegen

oyvindberg:
Hey there. I think we need to showcase the new functionality for server stubs. Thoughts?


ghostdogpr (approved):
Looks good! I'll merge it only after the next release as it gets published automatically

oyvindberg:
nice! can we have a release soon? 🙏

ghostdogpr:
Yeah I think once we merge the Caching PR we could do a release. I'm traveling in Europe this week and the next so I may be less responsive than usual but will work on it when I can.

PR #1948, opened 2023-10-19
Update tapir to 1.8.2

kyri-petrou:


ghostdogpr (approved):

ghostdogpr (changes requested):
Ah it's still red 😅

kyri-petrou:
There seems to be an issue with tapir. The test passes if I remove the `boundary` and `deferSpec` parameters from the response's content-type. I'll open an issue there

kyri-petrou:
@ghostdogpr CI is now passing. Turns out there is a small bug in sttp:model which affected the test suite. The production code is not affected by it

interop/tapir/src/test/scala/caliban/interop/tapir/TapirAdapterSpec.scala
```diff
@@ -483,9 +483,7 @@ object TapirAdapterSpec {
       fromMetadata(
         asStringAlways.map(error => Left(HttpError(error, StatusCode.UnprocessableEntity))),
         ConditionalResponseAs(
-          _.contentType.exists(
-            MediaType.unsafeParse(_).matches(ContentTypeRange("multipart", "mixed", ContentTypeRange.Wildcard))
-          ),
+          _.contentType.exists(_.startsWith("multipart/mixed")),
```
ghostdogpr, line 486: Is it always first?

ghostdogpr (approved):

interop/tapir/src/test/scala/caliban/interop/tapir/TapirAdapterSpec.scala
kyri-petrou: Do you mean first in the list or at the start of the string?

ghostdogpr: Start of the string. Just wanted to double check `startsWith` is good enough

kyri-petrou: Ah yeah it will be AFAICT. We should also be able to revert this change once the underlying issue has been resolved

ghostdogpr: Good to go then 👍

PR #1952, opened 2023-10-21
Fix #1352 using magnolia macros

### LINKED ISSUE #1352: Runtime exception "cannot access class jdk.internal.ValueBased" on JDK 17
Hey,

we are using caliban 2.0 with JDK 17 and Scala 3. When I include a `java.time` type with automatic schema derivation everything compiles but throws a runtime error. I was able to reproduce this in Scastie

https://scastie.scala-lang.org/fUOJ5OqzQoqbY1UmkEVysg

kyri-petrou:
The main issue seems to be when using `Macro.anns` to inspect the annotations of parameters for Product types within the `recurse` method. After taking a closer look at the code, it seems that we are unnecessarily double-extracting the parameter annotations for Product due to reusing the same `recurse` method for both Sum and Product types.

With this PR, we use different recurse methods for Product and Sum types. This solves this issue since the parameter annotations are now extracted via `Macro.paramAnns` which works fine on JDK17+. To demonstrate, I added 2 unit tests for it and added test workflows for JDK17 for Scala 3 (also for Scala 2.13 for completeness).

While the underlying issue with Magnolia still exists (see [here](https://github.com/softwaremill/magnolia/issues/493)), I think we should still do the separation as now macros that are specific to product/sum types are only executed when needed, which results in reduced compilation time + codegen size

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -76,9 +76,9 @@ jobs:
             - "~/.ivy2/cache"
             - "~/.sbt"
             - "~/.m2"
-  test213_jdk21:
+  test213_jdk17:
     docker:
-      - image: cimg/openjdk:21.0-node
+      - image: cimg/openjdk:17.0-node
     resource_class: large
     steps:
       - checkout
@@ -106,9 +106,9 @@ jobs:
             - "~/.ivy2/cache"
             - "~/.sbt"
             - "~/.m2"
-  test3_jdk21:
+  test3_jdk17:
     docker:
-      - image: cimg/openjdk:21.0-node
+      - image: cimg/openjdk:17.0-node
     resource_class: large
     steps:
       - checkout
@@ -269,7 +269,7 @@ workflows:
           filters:
             tags:
               only: /^v(0|[1-9]\d*)\.(0|[1-9]\d*)\.(0|[1-9]\d*)(?:-((?:0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*)(?:\.(?:0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*))*))?(?:\+([0-9a-zA-Z-]+(?:\.[0-9a-zA-Z-]+)*))?$/
-      - test213_jdk21:
+      - test213_jdk17:
           requires:
             - lint
           filters:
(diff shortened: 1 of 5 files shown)


.circleci/config.yml
```diff
@@ -76,6 +76,21 @@ jobs:
             - "~/.ivy2/cache"
             - "~/.sbt"
             - "~/.m2"
+  test213_jdk17:
```
ghostdogpr, line 79: A bit concerned about the CI time and cost if we test all combinations 🤔 Shall we test JDK 11? OR maybe 2.13 with 11 and 3 with 17?

core/src/main/scala-3/caliban/schema/macros/Macros.scala
```diff
@@ -7,42 +7,18 @@ import scala.quoted.*
...
 
   /**
    * Tests whether type argument [[FieldT]] in [[Parent]] is annotated with [[GQLExcluded]]
    */
   private def isFieldExcludedImpl[Parent: Type, FieldT: Type](using qctx: Quotes): Expr[Boolean] = {
     import qctx.reflect.*
+    val fieldName = Type.valueOfConstant[FieldT]
     Expr(TypeRepr.of[Parent].typeSymbol.primaryConstructor.paramSymss.flatten.exists { v =>
-      Type.valueOfConstant[FieldT].map(_ == v.name).getOrElse(false)
+      fieldName.map(_ == v.name).getOrElse(false)
```
ghostdogpr, line 21: Nice

.circleci/config.yml
kyri-petrou: I think CI time will be mostly unaffected since the workflows are run in parallel. But I agree about cost.

The issue I see with testing Scala 2.13 with JDK11 and Scala 3 with JDK17 is that if there's an issue with the Scala 3 test suite we won't know whether it's cause of JDK17 or Scala 3.

Perhaps we can test Scala 3 against JDK 11 & 17 and leave Scala 2.13 tested just against 11?

ghostdogpr: I checked our CI concurrency limit (30) and credits (we use about 25%) so there's some margin. Maybe it's okay for now but it's going to be worse when we add JDK 21 so I wonder if there is a better way. Maybe we could do it on `series/2.x` only which means we find it after merging, but at least before doing a release.

kyri-petrou: Yeah I think that's a good tradeoff. Also _in case_ something breaks, it doesn't block other PRs from getting merged. Updated the PR for the JDK 17 workflows to run only when merging into series/2.x

ghostdogpr (approved):

PR #1953, opened 2023-10-22
Add helper method for creating custom enum schemas

kyri-petrou:
Currently, the only way for a user to create a custom enum schema is a bit verbose, and requires the user to look up the enum schema derivation in order to implement one:

```scala
trait Foo
object Foo {
  case object A extends Foo
  case object B extends Foo
}

implicit val fooSchema: Schema[Any, Foo] = new Schema[Any, Foo] {
  def toType(isInput: Boolean = false, isSubscription: Boolean = false): __Type =
    Types.makeEnum(
      Some("Foo"),
      None,
      List(__EnumValue("FooB", None, false, None, None),  __EnumValue("FooA", None, false, None, None)),
      None,
    )

  override def resolve(value: Foo): Step[R] = PureStep(StringValue("Foo" + value.toString))
}
```

With this PR, we add a helper method in the `Schema` companion object to make it easier to create one:

```scala
import Schema._

implicit val fooSchema: Schema[Any, Foo] = enumSchema[Foo](
  "Foo",
  values = List(enumValue("FooA"), enumValue("FooB")),
  repr = v => "Foo" + v.toString
)
```

### DIFF
--- a/core/src/main/scala/caliban/schema/Schema.scala
+++ b/core/src/main/scala/caliban/schema/Schema.scala
@@ -200,6 +200,7 @@ trait GenericSchema[R] extends SchemaDerivation[R] with TemporalSchema {
    * @param values       list of possible enum values
    * @param directives   the directives to add to the type
    * @param repr         function that defines how to convert A into a string. WARNING: The resulting string must be contained in the values list
+   * @see [[enumValue]]  convenience method for creating enum values required by this method
    */
   def enumSchema[A](
     name: String,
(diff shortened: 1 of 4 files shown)


core/src/main/scala/caliban/schema/Schema.scala
```diff
@@ -193,6 +193,48 @@ trait GenericSchema[R] extends SchemaDerivation[R] with TemporalSchema {
...
+  ): Schema[Any, A] = new Schema[Any, A] {
+    private val validEnumValues = values.map(_.name).toSet
+
+    override def toType(isInput: Boolean, isSubscription: Boolean): __Type =
+      makeEnum(Some(name), description, values, None, if (directives.nonEmpty) Some(directives) else None)
+
+    override def resolve(value: A): Step[Any] = {
+      val asString = repr(value)
+      if (validEnumValues.contains(asString)) PureStep(EnumValue(asString))
+      else QueryStep(ZQuery.fail(ExecutionError(s"Invalid enum value '$asString'")))
```
kyri-petrou, line 220: I'm not sure whether we should be failing the query here or just allow this (even if the resulting value is invalid according to the schema).

Happy to hear your thoughts

ghostdogpr: I think failing makes sense.

ghostdogpr (approved):

PR #1954, opened 2023-10-22
Add SBT plugin to help manage 3rd-party dependency versions

kyri-petrou:
With this PR, we publish a lightweight SBT plugin that contains the versions of the libraries that Caliban was built against.

The motivation behind adding this is that in many cases a user will need to include additional dependencies from one of the integrations (e.g., tapir). Currently, the user needs to specify the version of tapir / http4s / etc. manually.

In some cases, this might cause binary incompatibility issues if the 3rd-party library doesn't strictly adhere to their versioning scheme, which makes dependency management more difficult as they need to manually keep library versions in-sync with what Caliban uses.

Let me know what you think about adding this plugin. Also, happy to hear any recommendations for the plugin name!

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -97,6 +97,7 @@ lazy val root = project
     clientLaminext,
     tools,
     codegenSbt,
+    buildinfoSbt,
     federation,
     reporting,
     tracing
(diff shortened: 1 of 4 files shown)


ghostdogpr:
I am not really in favor of this for several reasons:
- I think people should be aware of the version they depend on, especially when it changes. Let's say we upgrade http4s in Caliban, users need to be aware of that when they upgrade so I don't think we should promote a way to depend on it seamlessly as in the example.
- Personally I find it simpler to just look at mvnrepository to know what a library I use depends on, for example https://mvnrepository.com/artifact/com.github.ghostdogpr/caliban-http4s_3/2.4.1
- One more module, especially a sbt plugin makes things more complex (to maintain, to use), even if this one is simple

Maybe we can do a better job at communicating which version we are using in the documentation. How about adding a table with the recent versions of caliban + versions of the libraries it depends on (tapir and http libs)?

I also noticed https://ghostdogpr.github.io/caliban/docs/#dependencies has old versions of tapir-json 😄

kyri-petrou:
After thinking about this a bit more and reading your comments, I think you're right. This starts to feel like we're doing the work SBT is meant to be doing.

Having said that, I do also agree we need to provide better information regarding our dependency tree (or keep the readme up to date whenever we release).

As a (very) side note, I feel like this might be one of the few things that Python's pip got right: being able to set an upper version on dependencies that the library is meant to work with

PR #1962, opened 2023-10-26
Play 3

kyri-petrou:
This PR updates play and play-json to 3.0.0 and tapir to 1.8.3 (which supports play 3). Starting with v3, `play-pekko-http-server` has substituted `play-akka-http-server` 🎉. However, play has dropped support for Scala 2.12 which means we have to make some changes to CI to make things work:

1. We're now publishing `caliban-play` for Scala 2.13 and Scala 3 (instead of 2.12 and 2.13)
2. Default scala version in `build.sbt` is now Scala 2.13, which is the only one that can compile all the modules in IntelliJ. I tested publishing locally using `+publishLocal` and the sbt plugin was being published, but I'm not too sure whether we might see issues with `ci-release` - we'll have to see once this PR is merged in.
3. The examples work only with Scala 2.13 now. If we removed the akka adapter from the examples they'll work with Scala 3 as well
4. I had to disable mima checks for the play adapter because too many things are not bincompat anymore due to the akka -> pekko change

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -8,8 +8,7 @@ jobs:
       - checkout
       - restore_cache:
           key: sbtcache
-      - run: sbt ++2.13.12 check
-      - run: sbt ++3.3.1 check
+      - run: sbt ++2.13.12 check ++3.3.1 check
       - save_cache:
           key: sbtcache
           paths:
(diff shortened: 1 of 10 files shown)


ghostdogpr (approved):

paulpdaniels:
I can't find any reference to play3 is it hosted somewhere else?

kyri-petrou:
> I can't find any reference to play3 is it hosted somewhere else?

@paulpdaniels The namespace changed from `com.typesafe.play` to `org.playframework`, is that what you mean?
https://central.sonatype.com/artifact/org.playframework/play_3

kyri-petrou:
I don't think there was an official release, but I can see the v3.0.0 tag in the repo 
image

paulpdaniels (commented):
I'm slightly concerned because there's no official release and this is a significant breaking change. Should we temporarily introduce a play3 module instead of breaking play compat?

ghostdogpr:
Tapir made the switch so I think we should too. I think the major version change in Play is due to the switch to Pekko but it doesn't change much apart from that (looking at the PR in Tapir, it's mostly changes in imports).

kyri-petrou:
@paulpdaniels I think the main issue would be maintaining modules with different versions of tapir. We've tried that in the past with zio-http but that caused issues. Also tapir will likely be updating zio-http to RC3 at some point soon which is another big breaking change.

I'm thinking perhaps we'll need to make a 1.5.0 release once that one is out as well

paulpdaniels:
I guess we have no choice if tapir is upgrading. I do see a difference though since zio-http is still in RC. Play has long been "stable"

paulpdaniels (approved):

Issue #1918, opened 2023-10-02
GraphQL instances get combined incorrectly if one or both root queries are somehow deferred


satorg:
Initial discussion in Discord: [link](https://discord.com/channels/629491597070827530/633200096393166868/1157359669795360969)

There's a test suite I've created in my fork of **caliban** that shows a bunch of various scenarios where the combine method does not work correctly:
https://github.com/satorg/ghostdogpr.caliban/blob/76f44085db19f619b3d5b6ac464ba5dd9382bec1/core/src/test/scala/caliban/execution/GraphqlCombineSpec.scala

In short, let's assume there are two root queries we would like to combine:
```scala
case class Query1(one: Int)
case class Query2(two: Int)

val query1 = Query1(1)
val query2 = Query2(2)
```

Basically, the `GraphQL#combine` only  works correctly, if both queries are provided as direct values:
```scala
val api = graphQL(RootResolver(query1)) |+| graphQL(RootResolver(query2))
```
However, if any of the root queries or both are somehow deferred, then even though the `combine` method completes successfully and pretends producing a correct combined schema, the result combined GraphQL instance won't not work correctly. It won't be failing, though, but simply returning `null` for the deferred source root query, which is very confusing.

For example, assume:
```scala
val api = graphQL(RootResolver(query1)) |+| graphQL(RootResolver(() => query2)) // watch `query2` here
```

Now if we print `api.render` then pretty much correct combined GraphQL schema will be rendered:
```graphql
schema {
  query: Query2
}

type Query2 {
  one: Int!
  two: Int!
}
```
Nothing suspicious at this point yet, is it? However, if we execute a valid GraphQL query on such an instance:
```scala
for {
  interpreter <- api.interpreter
  response    <- interpreter.execute("query test { one two }")
} yield response.data.toString
```
Then we get a pretty surprising response:
```json
{"one":1,"two":null}
```
Note: there will be no errors in the response, i.e. `response.errors` will be empty. It is just the `two` value will be set to `null` even though the rendered schema above claims it is a non-nullable one.

The same thing happens for `query1` as well as any other kind of deferring or indirection: `ZIO.succeed(query)`, `(_: Field) => query` or even `(_: Field) => ZIO.succeed(query)`. In all the cases, the combined GraphQL instance gets created without an error and renders the correct combined schema. It just does not work correctly.

ghostdogpr:
[...]

PR #1967, opened 2023-10-31
Resolve queried fields only in object schemas

kyri-petrou:
As discussed on Discord, in large schemas we currently see a big amount of CPU time being spent on resolving fields that are not queried. This PR fixes this (and by proxy, #1918) by using a `MetadataFieldStep` when resolving object schemas. This way, we only resolve fields that have been queried instead of every field in the object

### DIFF
--- a/core/src/main/scala-2/caliban/schema/SchemaDerivation.scala
+++ b/core/src/main/scala-2/caliban/schema/SchemaDerivation.scala
@@ -95,8 +95,6 @@ trait CommonSchemaDerivation[R] {
         )
     }
 
-    override private[schema] lazy val resolveFieldLazily: Boolean = !(ctx.isObject || _isValueType)
-
     override def resolve(value: T): Step[R] =
       if (ctx.isObject) PureStep(EnumValue(getName(ctx)))
       else if (_isValueType) resolveValueType(value)
--- a/core/src/main/scala-2/caliban/schema/SchemaDerivation.scala
+++ b/core/src/main/scala-2/caliban/schema/SchemaDerivation.scala
@@ -107,14 +107,12 @@ trait CommonSchemaDerivation[R] {
       head.typeclass.resolve(head.dereference(value))
     }
 
-    private def resolveObject(value: T): Step[R] = {
+    private def resolveObject(value: T): Step[R] = MetadataFunctionStep[R] { field =>
       val fieldsBuilder = Map.newBuilder[String, Step[R]]
-      fields.foreach { case (name, schema, dereference) =>
-        fieldsBuilder += name -> {
-          lazy val step = schema.resolve(dereference(value))
-          if (schema.resolveFieldLazily) FunctionStep(_ => step)
-          else step
-        }
+      fields.foreach {
+        case (name, schema, dereference) if field.fieldNames.contains(name) =>
+          fieldsBuilder += name -> schema.resolve(dereference(value))
+        case _                                                              => ()
       }
       ObjectStep(getName(ctx), fieldsBuilder.result())
     }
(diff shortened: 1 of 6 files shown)


core/src/main/scala/caliban/schema/Schema.scala
```diff
@@ -81,13 +81,7 @@ trait Schema[-R, T] { self =>
...
   def arguments: List[__InputValue] = Nil
 
-  /**
-   * Boolean flag indicating whether the [[resolve]] method should be wrapped in a [[FunctionStep]] when resolving
-   * object fields.
-   *
-   * Should be `false` except for objects or wrappers of objects. This prevents fields of case classes to be resolved when they're not
-   * requested in the query, which significantly improves performance in larger schemas
-   */
+  @deprecated("No longer needed")
```
kyri-petrou: Shall I just delete this method since it's package private? Next release will likely be 1.5.0 given the play and zio-http changes

ghostdogpr: ok with me

kyri-petrou:
Just pushed an amended multifield query benchmark to try and replicate the issue that I observed. These are the results based on the `multifieldBatchedQuery1000`:

Current main:
```
[info] Benchmark                                          Mode  Cnt    Score   Error  Units
[info] NestedZQueryBenchmark.multifieldBatchedQuery1000  thrpt    5  377.303 ± 3.436  ops/s
```

PR:
```
[info] Benchmark                                          Mode  Cnt    Score   Error  Units
[info] NestedZQueryBenchmark.multifieldBatchedQuery1000  thrpt    5  441.058 ± 3.671  ops/s
```

Seems that even with the addition of just 1 object where we query a single field, the difference becomes quite obvious. In real schemas where these cases occur for multiple fields the difference becomes very substantial

core/src/main/scala-3/caliban/schema/SchemaDerivation.scala
```diff
@@ -171,26 +171,22 @@ trait CommonSchemaDerivation {
       else if (isInput) mkInputObject[R](annotations, fields, info)(isInput, isSubscription)
       else mkObject[R](annotations, fields, info)(isInput, isSubscription)
 
-    override private[schema] lazy val resolveFieldLazily: Boolean = (!isValueType) && fields.isEmpty
-
     def resolve(value: A): Step[R] =
       if (fields.isEmpty) PureStep(EnumValue(name))
       else if (isValueType) resolveValueType(value)
-      else resolveObject(value)
+      else MetadataFunctionStep[R](f => resolveObject(value, f.fieldNames))
```
kyri-petrou, line 177: I ended up moving this here (instead of wrapping the code in `resolveObject`) for 2 reasons:

1. It's slightly more performant since `fieldNames` is a lazy val and there is a small overhead when evaluating if a lazy val was initialized. This can add up when it's evaluated for each field.
2. It makes profiling easier to follow since the logic in `resolveObject` won't be wrapped within an anonymous function

core/src/main/scala-2/caliban/schema/SchemaDerivation.scala
```diff
@@ -95,26 +95,22 @@ trait CommonSchemaDerivation[R] {
...
-      fields.foreach { case (name, schema, dereference) =>
-        fieldsBuilder += name -> {
-          lazy val step = schema.resolve(dereference(value))
-          if (schema.resolveFieldLazily) FunctionStep(_ => step)
-          else step
-        }
+      fields.foreach {
+        case (name, schema, dereference) if queriedFields.contains(name) =>
+          fieldsBuilder += name -> schema.resolve(dereference(value))
+        case _                                                           => ()
```
ghostdogpr:
```suggestion
        case (name, schema, dereference) =>
          if queriedFields.contains(name)
            fieldsBuilder += name -> schema.resolve(dereference(value))
```

ghostdogpr: Same suggestion in other files

kyri-petrou: Done 👍

ghostdogpr (approved):
Nice!

PR #1968, opened 2023-10-31
Update zio-http to RC4 & Play to 3.0.0

kyri-petrou:
Tapir has a new release with support for zio-http RC3. This required very few changes in the adapter, but had to change the examples to work with the new routes structure


kyri-petrou:
@ghostdogpr would it be possible to merge #1967 prior to this one? I'd like to try out those changes without migrating all the zio-http code 😅

examples/src/main/scala/example/ziohttp/ExampleApp.scala
```diff
@@ -36,7 +34,7 @@ object ExampleApp extends ZIOAppDefault {
         ZLayer.succeed(
           Server.Config.default
             .port(8088)
-            .withWebSocketConfig(ZHttpAdapter.defaultWebSocketConfig)
+            .webSocketConfig(ZHttpAdapter.defaultWebSocketConfig)
```
ghostdogpr: It's still necessary? I thought it wasn't

examples/src/main/scala/example/federation/v2/FederatedApp.scala
```diff
@@ -16,10 +16,9 @@ object FederatedApp extends ZIOAppDefault {
     _           <-
       Server
         .serve(
-          Http
-            .collectHttp[Request] { case _ -> Root / "api" / "graphql" =>
-              ZHttpAdapter.makeHttpService(HttpInterpreter(interpreter))
-            }
+          Routes(
+            Method.ANY / "api" / "graphql" -> ZHttpAdapter.makeHttpService(HttpInterpreter(interpreter)).toHandler
```
ghostdogpr: Should we return a `Handler` if we have to do `toHandler` to use it?

examples/src/main/scala/example/ziohttp/ExampleApp.scala
kyri-petrou: Unfortunately for now yes, until this makes it into a release: https://github.com/softwaremill/tapir/pull/3291

ghostdogpr: Ah hopefully it comes soon, let's merge this after?

examples/src/main/scala/example/federation/v2/FederatedApp.scala
kyri-petrou: The only thing about it is that I'm unsure whether there might be other methods of composition that I haven't figured out yet. E.g., the HttpApp is also convertible to Routes

examples/src/main/scala/example/ziohttp/ExampleApp.scala
kyri-petrou: Yeah sounds good to me 👍

kyri-petrou:
@ghostdogpr tapir 1.8.5 is out and it adds the ability to provide custom WS config for zio-http. Have a look at the changes in ZHttpAdapter and let me know what you think.

I also added the play update in this PR otherwise CI would fail since tapir 1.8.5 is not compatible with play 2.x

adapters/zio-http/src/main/scala/caliban/ZHttpAdapter.scala
```diff
@@ -3,27 +3,36 @@ package caliban
...
-      .withDefaultErrorResponse
+  ): HttpApp[R] = {
+
+    val opts = serverOptions.withCustomWebSocketConfig { req =>
+      val protocol = req.header(HeaderNames.SecWebSocketProtocol).fold(Protocol.Legacy: Protocol)(Protocol.fromName)
+      serverOptions.customWebSocketConfig(req) match {
+        case Some(existing) => existing.subProtocol(Some(protocol.name))
+        case _              => WebSocketConfig.default.subProtocol(Some(protocol.name))
+      }
+    }
```
kyri-petrou, line 34: I decided to patch the user-provided one instead of making the one that handles subprotocols the default. This way users that override the default server options won't have to do make changes to their code

adapters/zio-http/src/main/scala/caliban/ZHttpAdapter.scala
```diff
@@ -3,27 +3,36 @@ package caliban
...
-      .withDefaultErrorResponse
+
+  private def patchWsServerOptions[R](serverOptions: ZioHttpServerOptions[R]) =
+    serverOptions.withCustomWebSocketConfig { req =>
+      val protocol = req.header(HeaderNames.SecWebSocketProtocol).fold(Protocol.Legacy: Protocol)(Protocol.fromName)
+      serverOptions.customWebSocketConfig(req) match {
+        case Some(existing) => existing.subProtocol(Some(protocol.name))
+        case _              => WebSocketConfig.default.subProtocol(Some(protocol.name))
+      }
+    }
```
kyri-petrou, line 39: I decided to patch the user-provided one instead of making the one that handles subprotocols the default. This way users that override the default server options won't have to make changes to their code

paulpdaniels (approved):

kyri-petrou:
PR updated to migrate the QuickAdapter to RC4

examples/src/main/scala/example/federation/v2/FederatedApp.scala
ghostdogpr: But we're never going to use it as `Routes` because it doesn't have any path in it, right?

Maybe make `makeHttpService` return a `Handler` for the most common case but keep a variant that returns the `HttpApp`?

examples/src/main/scala/example/federation/v2/FederatedApp.scala
```diff
@@ -30,10 +29,9 @@ object FederatedApp extends ZIOAppDefault {
     _           <-
       Server
         .serve(
-          Http
-            .collectHttp[Request] { case _ -> Root / "api" / "graphql" =>
-              ZHttpAdapter.makeHttpService(HttpInterpreter(interpreter))
-            }
+          Routes(
```
ghostdogpr: Not directly related but how about using the quick adapter for federation examples (or any examples that are not about a specific adapter).

ghostdogpr (approved):

examples/src/main/scala/example/federation/v2/FederatedApp.scala
kyri-petrou: Yeah after thinking a bit about it you're right, I don't think there's much of a reason returning an HttpApp. I changed it to return Handlers instead

examples/src/main/scala/example/federation/v2/FederatedApp.scala
kyri-petrou: Ah good point. I'll change it

kyri-petrou: So much cleaner now 😃

PR #1975, opened 2023-11-02
Allow excluding fields from interfaces

### LINKED ISSUE #1829: Support a way to hide fields from interfaces
At the moment if you declare a field in a type and then compose it under an interface the interface will automatically expose all of the fields that appear in all implementations of the interface. There are however cases where this is not desirable behavior. Consider for instance a federated graph where different subgraphs expose different implementations of the interface:

```graphql
// Search graph
interface Product {
  id: ID!
}

type Query {
  search(): [Product!]!
}


// Store graph
type PhysicalProduct implements Product {
   id: ID!
   location: String!
   shippingCost: Int!
}

// Web graph
type WebProduct {
  id: ID!
  url: String!
}
```

Since each subgraph would not know about the implementations within the other graphs it would have no way of knowing which fields can be safely exposed.

I know magnolia doesn't support resolution of the fields within a scala interface, but we can probably go the other way and _exclude_ them from an interface (perhaps even a specific interface?) by using an annotation like `GQLExcludeInterface` to mark a field as not visible to the interface.

kyri-petrou:
Closes #1829.

@paulpdaniels what are your thoughts on this approach? I thought it might be better to reuse the same annotation and add the option to provide a list of field names that you want to hide from the interface. The change is not binary compatible but it's largely source compatible so users won't need to migrate any code

Alternatively, we could add a separate annotation that is applied on fields if we think it's more appropriate


ghostdogpr (approved):

paulpdaniels:
Would there be any value to having it separate? Like any scenarios we might want to allow it standalone? Otherwise I imagine that might have some footgun behavior if someone forgot to add both.

paulpdaniels (approved):

kyri-petrou:
> Would there be any value to having it separate? Like any scenarios we might want to allow it standalone? Otherwise I imagine that might have some footgun behavior if someone forgot to add both.

I can't think of any benefit in having it separate other than avoiding the binary incompatible change. Just thought to point it out just in case there was something I might have missed

kyri-petrou:
@ghostdogpr If I merge this in the next release will need to be 2.5.0 due since there are binary incompatible changes in this PR. Are we okay with that?

ghostdogpr:
In theory yes. I'd like to wait for the zio-http RC4 release to do all the breaking changes at once (with also the Play upgrade) so I hope it won't take too long. If we merge this we'll have to wait for the next release, but we can always do bug fixes in a branch if we really need to.

kyri-petrou:
> In theory yes. I'd like to wait for the zio-http RC4 release to do all the breaking changes at once (with also the Play upgrade) so I hope it won't take too long. If we merge this we'll have to wait for the next release, but we can always do bug fixes in a branch if we really need to.

Ok sounds good to me 👍

kyri-petrou:
By the way I wonder whether we should have a `Breaking change` (or similar) label just so that we know which PRs contain breaking changes without having to look through the code

ghostdogpr:
We could add that label 👍

PR #1998, opened 2023-11-10
Add zio-http adapter without dependency on Tapir (a.k.a. `QuickAdapter`)

kyri-petrou:
As we've seen with recent benchmarks, using Tapir as an "intermediary" adds some overhead to response latencies. This PR adds the `QuickAdapter`, which uses `zio-http` and `jsoniter-scala` as the server & json implementations to provide the best possible performance out-of-the-box.

This adapter is still fully compliant with the "GraphQL over HTTP" spec, but doesn't support websockets. I included multipart / defer support, but I'm happy to remove that if we think that's too much for the "slim" variant. Let me know what you think!

Note that unlike the tapir adapters, this one doesn't have an equivalent `intercepted` method. This is mainly because creating zio-http middleware is fairly straightforward, and since that it would require changes to users code anyways (since the request type is different) I thought it's better to fully omit it

TODO:
- [x] Add example
- [x] Documentation


frekw:
This is basically 95% similar to what we already use in prod! :D 👍

adapters/quick/src/main/scala/caliban/QuickAdapter.scala
kyri-petrou, line 1: Hmmm I think this class might be getting a bit crowder, mixing handling of the requests with creating the HTTP interfaces. I think I might split all the request handling logic in a separate class, wdyt?

kyri-petrou: Edit: I ended up making this change afterall as the code is much cleaner now

.circleci/config.yml
```diff
@@ -330,3 +330,5 @@ workflows:
             branches:
               only:
                 - series/2.x
+            tags:
```
ghostdogpr: Hmm... the problem is that I usually update the docs with the new version *after* it's published (to avoid the docs being up with a version that doesn't exist yet)

adapters/quick/src/main/scala/caliban/GraphiQLAdapter.scala
```diff
@@ -0,0 +1,87 @@
+package caliban
+
+import zio.http._
+
+object GraphiQLAdapter {
```
ghostdogpr: Maybe weird to call it "adapter" since it's not really what it does with our definition of adapters. Maybe just `GraphiQLHandler`?

adapters/quick/src/main/scala/caliban/quick/package.scala
```diff
@@ -0,0 +1,16 @@
+package caliban
+
+import zio.{ RIO, Trace, ZIO }
+import zio.http.{ RequestHandler, Response }
+
+package object quick {
+
+  implicit class GraphqlServerOps[R](val gql: GraphQL[R]) extends AnyVal {
+    def runServer(port: Int, api: String, graphiql: Option[String] = None)(implicit trace: Trace): RIO[R, Nothing] =
```
ghostdogpr: Can you add scaladocs? Not sure what `api` and `graphiql` are looking at types (are these paths? maybe it should be included in the names)

vuepress/docs/docs/adapters.md
```diff
@@ -3,9 +3,16 @@
 Once you have an interpreter able to execute GraphQL queries, you usually want to expose it using an HTTP API.
 Caliban comes with a few "ready-to-use" components (called "adapters") to expose your API with the most popular HTTP libraries.
 
+::: tip `QuickAdapter`
+Starting with v2.4.3, Caliban provides the opinionated [QuickAdapter](adapters.md#high-performance-quickadapter)
+that favours ease-of-use and performance at the expense of customizability.
+
+If you don't mind using `zio-http` as the server implementation, make sure to check it out!
```
ghostdogpr: How about instead "If you want the best possible performance, make sure to check it out!"

vuepress/docs/docs/README.md
```diff
@@ -21,6 +21,7 @@ To use `caliban`, add the following dependency to your `build.sbt` file:
 The following modules are optional:
 
 ```scala
+"com.github.ghostdogpr" %% "caliban-quick"      % "2.4.2" // quick adapter
```
ghostdogpr: Maybe add in the comment: "if you're just getting started, only import that one"

ghostdogpr: Should we add an example of running the quick server right on the Getting Started page? Sounds like it would be good.

.circleci/config.yml
kyri-petrou: Would requiring the `release` step to be run first solve this?

I can add this to the task:

```yaml
requires:
  - release
```

ghostdogpr: That would work, but then what about when we want to update docs without a release?

kyri-petrou: Hmm one way would be to run the `npm run docs:build` locally and merge it into main, but that might be a bit too cumbersome.

vuepress/docs/docs/README.md
kyri-petrou: I had to re-arrange a few things to make it coherent. Not sure if what I went with is the best approach - let me know

adapters/quick/src/main/scala/caliban/GraphiQLHandler.scala
```diff
@@ -0,0 +1,87 @@
+package caliban
+
+import zio.http._
+
+object GraphiQLHandler {
+
+  def handler(apiPath: String, graphiqlPath: String): RequestHandler[Any, Nothing] =
```
ghostdogpr: Maybe document that one since it's the one likely to be called

vuepress/docs/docs/README.md
```diff
@@ -10,50 +10,14 @@ The design principles of Caliban are the following:
...
-```scala
-"com.softwaremill.sttp.tapir" %% "tapir-json-circe"     % "1.2.11" // Circe
-"com.softwaremill.sttp.tapir" %% "tapir-jsoniter-scala" % "1.2.11" // Jsoniter
-"com.softwaremill.sttp.tapir" %% "tapir-json-play"      % "1.2.11" // Play JSON
-"com.softwaremill.sttp.tapir" %% "tapir-json-zio"       % "1.2.11" // ZIO JSON
-```
+## A simple example
 
-And then later in your code (you only need one!):
+First, add the following dependencies to your `build.sbt` file:
```
ghostdogpr:
```suggestion
First, add the following dependency to your `build.sbt` file:
```

vuepress/docs/docs/adapters.md
```diff
@@ -107,3 +114,67 @@ This adapter, available in the `caliban-tapir` dependency which has the same 3 m
...
+### Usage
+
+In order to use it, just add the following to your `build.sbt` file (no other dependencies required!):
+
+```scala
+libraryDependencies += "com.github.ghostdogpr" %% "caliban-quick" % "1.4.3"
+```
+
+By adding `import caliban.quick._`, we expose a few convenient extension methods on our `GraphQL` api. 
+e.g., we can serve our GraphQL api with minimal setup via a single command:
```
ghostdogpr:
```suggestion
For example, we can serve our GraphQL api with minimal setup via a single command: 
```

vuepress/docs/docs/README.md
ghostdogpr: Looks good!

.circleci/config.yml
ghostdogpr: I suggest for now, we put docs in a separate PR, and if we find a better solution for publishing docs we do it in a dedicated PR?

kyri-petrou: What about if we made the `microsite` workflow require manual approval? https://circleci.com/docs/workflows/#holding-a-workflow-for-a-manual-approval

We can "trial" it in a separate PR to see if we like it / works

ghostdogpr: I think the more fundamental issue with pushing "future" versions docs into the main branch is that until we release a new version, we can't update the docs anymore for other purposes. Any other change will have to wait a release. That's why I initially set it up this way, to have the freedom to update it anytime. In practice though, we usually update it more often when we change something and need to wait for the release. But the flexibility is nice 🤔

ghostdogpr: I guess an alternative would have to have docs versioning with an "unstable" version as well (like sttp/tapir), but it's work to set it up (not even sure vuepress can do that).

ghostdogpr (approved):

PR #2000, opened 2023-11-11
Fix incorrect description in interface server code gen

### LINKED ISSUE #1999: Codegen: Incorrect order of description in server sealed trait generation
I noticed a codegen problem when the result is a sealed trait:

```graphql
"""TypeNameDescription"""
interface TypeName {
  """fooDescription"""
  foo: AType!
}
```

```scala
@GQLInterface
@GQLDescription("traitDescription")
sealed trait TypeName extends scala.Product with scala.Serializable  {
         def @GQLDescription("description") foo : AType
}
```

And

```scala
def @GQLDescription("description") foo : AType
```
Isn't valid scala and should be : 

```scala
@GQLDescription("description") def status : AType
```

I suspect the problem comes from : https://github.com/ghostdogpr/caliban/blob/2a2479a4647eef592203df1538094a8d7deb84c9/tools/src/main/scala/caliban/tools/SchemaWriter.scala#L116

I can solve this but I'm afraid I don't know what is the correct way relatively to the rest of the codebase.

ghostdogpr:
Closes #1999

### DIFF
--- a/tools/src/test/scala/caliban/tools/SchemaWriterSpec.scala
+++ b/tools/src/test/scala/caliban/tools/SchemaWriterSpec.scala
@@ -2,7 +2,6 @@ package caliban.tools
 
 import caliban.parsing.Parser
 import zio.Task
-import zio.test.Assertion.equalTo
 import zio.test._
 
 object SchemaWriterSpec extends ZIOSpecDefault {
@@ -687,6 +686,7 @@ object SchemaWriterSpec extends ZIOSpecDefault {
         |union AllErrors = Bar | Foo | FooBar
         |
         |interface Error {
+        |  "description"
         |  message: String!
         |}
         |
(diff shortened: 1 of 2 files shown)


tools/src/test/scala/caliban/tools/SchemaWriterSpec.scala
```diff
@@ -721,6 +721,7 @@ object SchemaWriterSpec extends ZIOSpecDefault {
         |  sealed trait AllErrors extends scala.Product with scala.Serializable derives caliban.schema.Schema.SemiAuto
         |  @GQLInterface
         |  sealed trait Error extends scala.Product with scala.Serializable derives caliban.schema.Schema.SemiAuto {
```
kyri-petrou, line 723: As a side note, in Scala 3 we probably don't need the `extends scala.Product with scala.Serializable`. We can perhaps look into removing at some other time

kyri-petrou (approved):

PR #2005, opened 2023-11-13
Fix execution of nested `Sum` types

kyri-petrou:
With this PR, we resolve using the top-level schemas instead of the exploded leaf ones, otherwise the index returned by `.ordinal` doesn't match the index in the list


ghostdogpr (approved):

kyri-petrou:
@ghostdogpr just doing some final testing on something and I'll merge it in afterwards

PR #2007, opened 2023-11-15
Docs for QuickAdapter

kyri-petrou:


kyri-petrou:
Docs are expected to fail since mdoc fails to compile due to the missing methods

PR #2015, opened 2023-11-21
Don't include pure object fields in ZQuery collection

kyri-petrou:
As discussed on Discord, I gave this a go. Seems to offer a fairly good performance increase whenever we have a mix of pure and effectful fields.

series/2.x:
```shell
[info] Benchmark                                           Mode  Cnt    Score   Error  Units
[info] NestedZQueryBenchmark.multifieldBatchedQuery1000   thrpt    5  551.735 ± 8.117  ops/s
[info] NestedZQueryBenchmark.multifieldBatchedQuery10000  thrpt    5   45.115 ± 1.188  ops/s
```

PR:
```shell
[info] Benchmark                                           Mode  Cnt    Score   Error  Units
[info] NestedZQueryBenchmark.multifieldBatchedQuery1000   thrpt    5  633.503 ± 8.756  ops/s
[info] NestedZQueryBenchmark.multifieldBatchedQuery10000  thrpt    5   54.862 ± 0.919  ops/s
```

I also did some rough testing for cases that we don't have any pure fields, and it seems that the performance was roughly identical to what we had previously

### DIFF
--- a/benchmarks/src/main/scala/caliban/execution/NestedZQueryBenchmarkSchema.scala
+++ b/benchmarks/src/main/scala/caliban/execution/NestedZQueryBenchmarkSchema.scala
@@ -50,8 +50,8 @@ object NestedZQueryBenchmarkSchema {
   case class MultifieldRoot(entities: Query[List[MultifieldEntity]])
   case class MultifieldEntity(
     id: Int,
-    nested0: Query[Int],
-    nested1: Query[Int],
+    nested0: Int,
+    nested1: Int,
     nested2: Query[Int],
     nested3: Query[Int],
     nested4: Query[Int],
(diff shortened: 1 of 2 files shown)


ghostdogpr (approved):
It looks good!

I have one concern though (which is more general, and could be address in other PRs): we are adding more and more "optimized" code using mutable collections and so on, and I think it makes the code harder to read. I think it would be good to move this kind of code to dedicated functions so that the main "flow" stays short and readable. What do you think?

kyri-petrou:
> It looks good!
> 
> I have one concern though (which is more general, and could be address in other PRs): we are adding more and more "optimized" code using mutable collections and so on, and I think it makes the code harder to read. I think it would be good to move this kind of code to dedicated functions so that the main "flow" stays short and readable. What do you think?

Yeah I agree. I'll do it in a followup PR  in a bit 👍

PR #2016, opened 2023-11-21
Add `pre-commit` config with scalafmt

kyri-petrou:
Adds pre-commit config that runs scalafmt when committing new changes. I've started using everywhere that I can recently because I find it quite annoying to submit a PR only to find out that the checks failed due to formatting.

In case you've never used pre-commit before, it's [pretty easy to install](https://pre-commit.com/#installation). Once installed, you only need to run `pre-commit install` at the root of the repo to install the hooks.

Note that if the user doesn't have pre-commit installed or never runs `pre-commit install` in this repo, the hooks just won't be run (i.e., no errors or anything)

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,0 +1,12 @@
+repos:
+  - repo: local
+    hooks:
+      - id: scalafmt
+        name: scalafmt fix formatting
+        entry: scalafmt
+        args: [ --exclude=target, --mode=changed, --reportError ]
+        language: system
+        stages: [ commit, push ]
+        always_run: true
+        pass_filenames: false
+        minimum_pre_commit_version: '2.8.0'


kyri-petrou:
Thinking about it, instead of running `scalafmt` in the hook, we could run `sbt fmt` instead. That would be more correct but slower. What do you think?

ghostdogpr:
I have no strong opinion about it, I think I rarely have that problem (maybe because of my IDE settings) but I'm fine to merge it.

kyri-petrou:
> I have no strong opinion about it, I think I rarely have that problem (maybe because of my IDE settings) but I'm fine to merge it.

I usually have issues with formatting of imports. I tried playing around with different settings but couldn't get IntelliJ's "Optimize imports" feature to format them as expected by `scalafmt`.

PR #2019, opened 2023-11-23
Improved execution field merging

kyri-petrou:
In many cases, there's would be only a single condition on fields (either None or Set(Something)). In these cases, the field merging has already been handled within `Field.apply`, which means we don't need to do it again.

For these cases, we up to 15% improvement in execution time depending on the number of fields in the object

### DIFF
--- a/core/src/main/scala/caliban/execution/Executor.scala
+++ b/core/src/main/scala/caliban/execution/Executor.scala
@@ -305,29 +305,45 @@ object Executor {
   private[caliban] def fail(error: CalibanError): UIO[GraphQLResponse[CalibanError]] =
     ZIO.succeed(GraphQLResponse(NullValue, List(error)))
 
-  private[caliban] def mergeFields(field: Field, typeName: String): List[Field] =
-    field.fields match {
-      // Shortcut if all the fields have the same condition, which means we don't need to dedup
-      // as that's been handled in Field.apply
-      case head :: tail if tail.forall(_._condition == head._condition) =>
-        if (head._condition.forall(_.contains(typeName))) field.fields
-        else Nil
-      case Nil                                                          => Nil
-      case fields                                                       =>
-        val map = new java.util.LinkedHashMap[String, Field]()
-        fields.foreach { field =>
-          if (field._condition.forall(_.contains(typeName))) {
-            map.compute(
-              field.aliasedName,
-              (_, f) =>
-                if (f == null) field
-                else f.copy(fields = f.fields ::: field.fields)
-            )
-          }
-        }
-
-        map.values().asScala.toList
+  private[caliban] def mergeFields(field: Field, typeName: String): List[Field] = {
+    def haveSameCondition(head: Field, tail: List[Field]): Boolean = {
+      val condition = head._condition
+      var remaining = tail
+      while (!remaining.isEmpty) {
+        if (remaining.head._condition != condition) return false
+        remaining = remaining.tail
+      }
+      true
     }
+
+    def matchesTypename(f: Field): Boolean =
+      f._condition.isEmpty || f._condition.get.contains(typeName)
+
+    def mergeFields(fields: List[Field]) = {
+      val map       = new java.util.LinkedHashMap[String, Field]((fields.size / 0.75d).toInt + 1)
+      var remaining = fields
+      while (!remaining.isEmpty) {
+        val h = remaining.head
+        if (matchesTypename(h)) {
+          map.compute(
+            h.aliasedName,
+            (_, f) =>
+              if (f eq null) h
+              else f.copy(fields = f.fields ::: h.fields)
+          )
+        }
+        remaining = remaining.tail
+      }
+      map.values().asScala.toList
+    }
+
+    field.fields match {
+      // Shortcut if all the fields have the same condition, which means we don't need to merge as that's been handled in Field.apply
+      case h :: t if haveSameCondition(h, t) => if (matchesTypename(h)) field.fields else Nil
+      case Nil                               => Nil
+      case fields                            => mergeFields(fields)
+    }
+  }
 
   private def fieldInfo(field: Field, path: List[Either[String, Int]], fieldDirectives: List[Directive]): FieldInfo =
     FieldInfo(field.aliasedName, field, path, fieldDirectives, field.parentType)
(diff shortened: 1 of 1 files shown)


core/src/main/scala/caliban/execution/Executor.scala
```diff
@@ -305,29 +305,29 @@ object Executor {
...
+        else Nil
+      case Nil                                                          => Nil
+      case fields                                                       =>
+        val map = new java.util.LinkedHashMap[String, Field]()
+        fields.foreach { field =>
+          if (field._condition.forall(_.contains(typeName))) {
+            map.compute(
+              field.aliasedName,
+              (_, f) =>
+                if (f == null) field
```
paulpdaniels: Nit: just as a style preference I prefer `eq` when doing null checks

core/src/main/scala/caliban/execution/Executor.scala
```diff
@@ -305,29 +305,29 @@ object Executor {
...
+    field.fields match {
+      // Shortcut if all the fields have the same condition, which means we don't need to dedup
+      // as that's been handled in Field.apply
+      case head :: tail if tail.forall(_._condition == head._condition) =>
+        if (head._condition.forall(_.contains(typeName))) field.fields
+        else Nil
+      case Nil                                                          => Nil
+      case fields                                                       =>
+        val map = new java.util.LinkedHashMap[String, Field]()
+        fields.foreach { field =>
```
paulpdaniels: Naked iteration I think has about a 5% perf advantage over foreach, worth doing it here? I guesss May only be noticeable with _alot_ of fields.

paulpdaniels (approved):

core/src/main/scala/caliban/execution/Executor.scala
kyri-petrou: Hmm good question. I'm still a bit confused on when naked iterations provide a sizeable benefit, but so far these are the case that I think it really makes a difference:

1. When it contains reference to a `var`. The Scala compiler "wraps" `var`s into a `scala.runtime.ObjectRef` because it needs to pass it to the Function1, which causes a fair bit of overhead especially if the `var` is read / modified on each iteration
2. When the list contains tuples

Although, given that this is a very hot path when using multiple fragments, I'll change it.

PS: I'll also change the `.forall` iteration above because I just realised that one causes the boolean to be boxed

core/src/main/scala/caliban/execution/Executor.scala
```diff
@@ -306,27 +306,43 @@ object Executor {
...
+      true
     }
 
-    // Avoid conversions if no modification took place
-    if (modified) map.values().asScala.toList else field.fields
+    def matchesTypename(f: Field): Boolean =
+      f._condition.isEmpty || f._condition.get.contains(typeName)
+
+    def mergeFields(fields: List[Field]) = {
+      val map       = new java.util.LinkedHashMap[String, Field]((fields.size / 0.75d).toInt + 1)
```
paulpdaniels: Is this a known formula? Or just an estimate?

core/src/main/scala/caliban/execution/Executor.scala
```diff
@@ -306,27 +306,43 @@ object Executor {
...
-              f.copy(fields = f.fields ::: field.fields)
-            }
-        )
-      } else {
-        modified = true
+    def haveSameCondition(head: Field, tail: List[Field]): Boolean = {
+      val condition = head._condition
+      var remaining = tail
+      while (!remaining.isEmpty) {
+        if (remaining.head._condition != condition) return false
```
paulpdaniels, line 313: I was gonna suggest tail recursion but this will probably iterate tighter

core/src/main/scala/caliban/execution/Executor.scala
ghostdogpr: Would be good to explain what this is in a comment

kyri-petrou: The exact formula would be `Math.ceil(size / 0.75)` but this is close enough I believe. I can use that instead if you prefer.

The Map resizes when the number of items in it exceed the capacity * loading factor (defaults to 0.75). Using this formula guarantees that we have adequate capacity so that the Map won't need to resize. In JDK17+ there is a method that allows you to create the Maps taking the expected number of items as a parameter (instead of using the initial size) and it used this formula.

I'll add a comment for it.

core/src/main/scala/caliban/execution/Executor.scala
kyri-petrou: Hmm I think the compiler will rewrite the tailrec method likely to this implementation. I'll check, if it's the same I'll use tailrec

kyri-petrou: Looking at the generated code, it looks like `@tailrec` uses Java labels in an infinite while loop with return statements, whereas this one uses the `!remaining.isEmpty` method as a condition to terminate the `while` loop. Not sure which one would be more performant in theory

core/src/main/scala/caliban/execution/Executor.scala
paulpdaniels: I forget is scala list's size O(n) or is it stored on concat?

core/src/main/scala/caliban/execution/Executor.scala
paulpdaniels: Yeah we're probably getting into a grey area of compiler optimization. I think the conditionless while might be able to skip a branch if equal instruction, though for modern chips branches aren't really performance killers anymore

core/src/main/scala/caliban/execution/Executor.scala
kyri-petrou: It's O(n), although given that this List only contains the queried fields and that this method is only run for objects with multiple fragments, I think the overall impact of counting the number of elements in this list is extremely small

core/src/main/scala/caliban/execution/Executor.scala
```diff
@@ -370,4 +386,17 @@ object Executor {
...
+  /**
+   * The behaviour of mutable Maps (both Java and Scala) is to resize once the number of entries exceeds
+   * the capacity * loadFactor (default of 0.75d) threshold in order to prevent hash collisions.
+   *
+   * This method is a helper method to estimate the initial map size depending on the number of elements the Map is
+   * expected to hold
+   *
+   * NOTE: This method is the same as java.util.HashMap.calculateHashMapCapacity on JDK19+
+   */
+  private def calculateMapCapacity(nMappings: Int): Int =
```
ghostdogpr, line 399: After merging this you can use it in the other PR

ghostdogpr (approved):

PR #2020, opened 2023-11-23
Avoid unnecessary mapping of steps to queries

kyri-petrou:
By using ZQuery.foreach, we avoid the unnecessary extra `.map` to convert a list of steps to ZQueries


paulpdaniels (approved):

ghostdogpr:
Looks like it doesn't compile with 2.12

ghostdogpr (approved):

PR #2021, opened 2023-11-25
Iterate through queried fields instead of object fields

kyri-petrou:
I don't know why this only just now occurred to me, but iterating through the queried fields is much more efficient than iterating through all the fields when resolving objects. This way, we don't need to create the `fieldNames` Set, plus we don't need to iterate through every single possible object field just so that we only use the ones that have been queried.

Also, I managed to extract this logic and reuse it across all 3 places that we were using it.

### DIFF
--- a/core/src/main/scala/caliban/schema/ObjectFieldResolver.scala
+++ b/core/src/main/scala/caliban/schema/ObjectFieldResolver.scala
@@ -16,7 +16,7 @@ final private class ObjectFieldResolver[R, A](
     map
   }
 
-  def resolve(value: A): MetadataFunctionStep[R] = MetadataFunctionStep(resolveForField(value, _))
+  def resolve(value: A): Step[R] = MetadataFunctionStep(resolveForField(value, _))
 
   private def resolveForField(
     value: A,
--- a/core/src/main/scala/caliban/schema/ObjectFieldResolver.scala
+++ b/core/src/main/scala/caliban/schema/ObjectFieldResolver.scala
@@ -1,0 +1,35 @@
+package caliban.schema
+
+import caliban.execution.Field
+import caliban.schema.Step.{ MetadataFunctionStep, ObjectStep }
+
+import scala.collection.immutable.HashMap
+
+final private class ObjectFieldResolver[R, A](
+  objectName: String,
+  fields: Iterable[(String, A => Step[R])]
+) {
+
+  private val fieldsMap: java.util.HashMap[String, A => Step[R]] = {
+    val map = new java.util.HashMap[String, A => Step[R]]()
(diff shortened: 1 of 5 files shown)


core/src/main/scala/caliban/schema/ObjectFieldResolver.scala
```diff
@@ -0,0 +1,36 @@
...
+
+final private class ObjectFieldResolver[R, A](
+  objectName: String,
+  fields: Iterable[(String, A => Step[R])]
+) {
+
+  private val fieldsMap: java.util.HashMap[String, A => Step[R]] = {
+    val map = new java.util.HashMap[String, A => Step[R]]()
+    fields.foreach { case (name, resolve) => map.put(name, resolve) }
+    map
```
paulpdaniels, line 16: Could we use a standard Scala map here? Doesn't it have a constructor that takes an iterable?

core/src/main/scala/caliban/schema/ObjectFieldResolver.scala
```diff
@@ -0,0 +1,36 @@
...
+  }
+
+  def resolve(value: A): Step[R] = MetadataFunctionStep(resolveForField(value, _))
+
+  private def resolveForField(
+    value: A,
+    field: Field
+  ): Step[R] = {
+    val fieldsBuilder = HashMap.newBuilder[String, Step[R]]
```
paulpdaniels, line 25: Precompute size based on the distinct fields?

paulpdaniels (approved):

core/src/main/scala/caliban/schema/ObjectFieldResolver.scala
kyri-petrou: So the main driver for going with a Java HashMap here was to avoid the wrapping / unwrapping into `Option` when calling `.get`. I found that it offers a (small) performance increase and reduction in gc allocations.

Although I do think it's cleaner to just use `immutable.HashMap.from(fields)`, wdut?

core/src/main/scala/caliban/schema/ObjectFieldResolver.scala
kyri-petrou: Map builders don't expose a constructor to pre-compute their size. I took a look at using `fieldsBuilder.sizeHint(...)` but AFAICT the HashMap builder don't seem to override it (i.e., it's ignored).

Also it seems to me that immutable Maps in Scala don't use initial capacity or load factors. My guess is that it's because they don't need to mutate so their new capacity can be calculated on creation (`.result()`) or when updating them with new elements

ghostdogpr (approved):

PR #2023, opened 2023-11-28
Add `apply` method to Schema

kyri-petrou:
This PR adds helper methods to "summon" a schema of type A. This is particularly useful in Scala 3 when schemas are derived via the `derives` keyword. e.g.,:

```scala
case class Foo(x: Int) derives Schema.SemiAuto
case class Bar(x: Int)

// Before
given Schema[Any, Bar] = summon[Schema[Any, Foo]].contramap(???)

// Now
given Schema[Any, Bar] = Schema[Foo].contramap(???)
```

Also works with custom schemas:
```scala
trait Env
object MySchema extends SchemaDerivation[Env]
case class Foo(x: Int) derives MySchema.SemiAuto
case class Bar(x: Int)

given Schema[Env, Bar] = MySchema[Foo].contramap(???)
```

Another option would be to have the apply method defined on `object Schema` directly as below, but I thought it might be a bit annoying having to add the environment parameter each time

```scala
def apply[R, A](implicit ev: Schema[R, A]): Schema[R, A] = ev
```

### DIFF
--- a/core/src/main/scala-2/caliban/schema/SchemaDerivation.scala
+++ b/core/src/main/scala-2/caliban/schema/SchemaDerivation.scala
@@ -249,6 +249,7 @@ trait CommonSchemaDerivation[R] {
 }
 
 trait SchemaDerivation[R] extends CommonSchemaDerivation[R] {
+  def apply[A](implicit ev: Schema[R, A]): Schema[R, A] = ev
 
   /**
    * Returns an instance of `Schema` for the given type T.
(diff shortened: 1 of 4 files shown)


frekw (approved):
Nice improvement! 👍

ghostdogpr (approved):

kyri-petrou:
Agh, should probably add one for ArgBuilder while at it. I'll add it now

PR #2029, opened 2023-12-06
More object field resolution / reduction optimizations

kyri-petrou:
There are 2 main optimizations in this PR:

1. Avoid usages of FunctionX and the number of branches within `reduceObjectStep`
2. Use a `mutable.HashMap` in ObjectFieldResolver

Now, I'm somewhat conflicted about (2). For _some_ reason, Scala's immutable HashMapBuilder is much slower than using a mutable Map. I realised this inefficiency after looking at some profiling graphs, where a lot of time was spend on the `+=` method of the HashMapBuilder. It seems that even if the HashMapBuilder is mutable, the implementation copies the underlying arrays on each new entry (no idea why).

Since this is in our _very_ hot path and we probably can't go changing the Scala std library, one way to optimize for it is to use a mutable HashMap, and change the Map type on ObjectStep from Map (i.e., `immutable.Map`) to `collection.Map` which is the base type for both mutable and immutable collections. This change is binary incompatible but largely source-compatible so most users should not need to do any changes to their code.

As I said though, I'm fairly conflicted about this change. It does offer a fairly sizeable performance improvement over the current implementation, but it could potentially leak mutability if misused; even if `collection.Map` doesn't expose any mutable APIs a user _could potentially_ pattern match and get the mutable instance. Having said that, I don't know how much of an issue this actually is

series/2.x:
```
[info] Benchmark                                              Mode  Cnt    Score    Error  Units
[info] NestedZQueryBenchmark.multifieldSequentialQuery1000   thrpt    5  686.416 ± 30.976  ops/s
[info] NestedZQueryBenchmark.multifieldSequentialQuery10000  thrpt    5   61.166 ±  1.555  ops/s
```

PR (~10% improvement):
```
[info] Benchmark                                              Mode  Cnt    Score    Error  Units
[info] NestedZQueryBenchmark.multifieldSequentialQuery1000   thrpt    5  746.585 ± 21.516  ops/s
[info] NestedZQueryBenchmark.multifieldSequentialQuery10000  thrpt    5   65.220 ±  1.478  ops/s
```

### DIFF
--- a/core/src/main/scala/caliban/execution/Executor.scala
+++ b/core/src/main/scala/caliban/execution/Executor.scala
@@ -33,7 +33,9 @@ object Executor {
     queryExecution: QueryExecution = QueryExecution.Parallel,
     featureSet: Set[Feature] = Set.empty
   )(implicit trace: Trace): URIO[R, GraphQLResponse[CalibanError]] = {
-    val wrapPureValues = fieldWrappers.exists(_.wrapPureValues)
+    val wrapPureValues    = fieldWrappers.exists(_.wrapPureValues)
+    val isDeferredEnabled = featureSet(Feature.Defer)
+
     type ExecutionQuery[+A] = ZQuery[R, ExecutionError, A]
 
     val execution = request.operationType match {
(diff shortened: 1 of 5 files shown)


core/src/main/scala/caliban/execution/Executor.scala
```diff
@@ -57,32 +59,32 @@ object Executor {
...
           case f @ Field("__typename", _, _, _, _, _, _, directives, _, _, _)   =>
             Right((f.aliasedName, PureStep(StringValue(objectName)), fieldInfo(f, path, directives)))
           case f @ Field(name, _, _, _, _, _, args, directives, _, _, fragment) =>
-            val aliasedName = f.aliasedName
-            val field       = fields
-              .get(name)
-              .fold(NullStep: ReducedStep[R])(reduceStep(_, f, args, Left(aliasedName) :: path))
-
-            val info = fieldInfo(f, path, directives)
+            val field = fields.get(name) match {
```
ghostdogpr, line 68: Love the simplifications, the code is more readable

ghostdogpr (approved):
I would be against exposing a mutable map in the public API but the tricks of using `collection.Map` seems safe 🤔

PR #2031, opened 2023-12-08
Implement ObjectStep using a Function1

kyri-petrou:
It seems we've gone a long way around to finally end up here, but I find it very difficult to believe that there is a more optimized version of this 😅 

In short, this PR removes all the intermediate creation of `Map`s for us to create an ObjectStep, and instead uses a Function1. This allows us to:

1. Avoid wrapping `ObjectStep`s with a `MetadataFunctionStep`
2. Avoids creation of the intermediate `Map[String, Step[R]]`

In order to avoid breaking source-compatibility when creating ObjectStep, I also added a constructor that takes a `Map[String, Step[R]]`

The downside to this change is that in cases where we have aliased fields, the same field will be resolved more than once. However, in practise this is not really a problem because almost always aliased fields are then resolved to a FunctionStep. If we ever see that this _might_ end up being an issue, we can add an intermediate cache, but I highly doubt it will be the case

### DIFF
--- a/core/src/main/scala/caliban/schema/Schema.scala
+++ b/core/src/main/scala/caliban/schema/Schema.scala
@@ -176,7 +176,8 @@ trait GenericSchema[R] extends SchemaDerivation[R] with TemporalSchema {
     directives: List[Directive] = List.empty
   ): Schema[R1, A] =
     new Schema[R1, A] {
-      override def toType(isInput: Boolean, isSubscription: Boolean): __Type =
+      override def toType(isInput: Boolean, isSubscription: Boolean): __Type = {
+        val _ = resolver
         if (isInput) {
           makeInputObject(
             Some(customizeInputTypeName(name)),
@@ -187,9 +188,10 @@ trait GenericSchema[R] extends SchemaDerivation[R] with TemporalSchema {
             directives = Some(directives)
           )
         } else makeObject(Some(name), description, fields(isInput, isSubscription).map(_._1), directives)
+      }
 
       private lazy val resolver =
-        new ObjectFieldResolver[R1, A](name, fields(false, false).map(f => (f._1.name, f._2)))
+        ObjectFieldResolver[R1, A](name, fields(false, false).map(f => (f._1.name, f._2)))
 
       override def resolve(value: A): Step[R1] = resolver.resolve(value)
     }
--- a/core/src/main/scala/caliban/schema/Step.scala
+++ b/core/src/main/scala/caliban/schema/Step.scala
@@ -10,12 +10,16 @@ import zio.stream.ZStream
 sealed trait Step[-R]
 
 object Step {
-  case class ListStep[-R](steps: List[Step[R]])                                    extends Step[R]
-  case class FunctionStep[-R](step: Map[String, InputValue] => Step[R])            extends Step[R]
-  case class MetadataFunctionStep[-R](step: Field => Step[R])                      extends Step[R]
-  case class ObjectStep[-R](name: String, fields: collection.Map[String, Step[R]]) extends Step[R]
-  case class QueryStep[-R](query: ZQuery[R, Throwable, Step[R]])                   extends Step[R]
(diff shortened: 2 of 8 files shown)


core/src/main/scala/caliban/schema/Schema.scala
```diff
@@ -176,7 +176,8 @@ trait GenericSchema[R] extends SchemaDerivation[R] with TemporalSchema {
     directives: List[Directive] = List.empty
   ): Schema[R1, A] =
     new Schema[R1, A] {
-      override def toType(isInput: Boolean, isSubscription: Boolean): __Type =
+      override def toType(isInput: Boolean, isSubscription: Boolean): __Type = {
+        val _ = resolver
```
ghostdogpr, line 180: Why not just making `resolver` a `val`? It needs to be lazy?

core/src/main/scala/caliban/schema/Step.scala
```diff
@@ -10,12 +10,16 @@ import zio.stream.ZStream
...
-  case class ObjectStep[-R](name: String, fields: collection.Map[String, Step[R]]) extends Step[R]
-  case class QueryStep[-R](query: ZQuery[R, Throwable, Step[R]])                   extends Step[R]
-  case class StreamStep[-R](inner: ZStream[R, Throwable, Step[R]])                 extends Step[R]
+  case class ListStep[-R](steps: List[Step[R]])                         extends Step[R]
+  case class FunctionStep[-R](step: Map[String, InputValue] => Step[R]) extends Step[R]
+  case class MetadataFunctionStep[-R](step: Field => Step[R])           extends Step[R]
+  case class QueryStep[-R](query: ZQuery[R, Throwable, Step[R]])        extends Step[R]
+  case class StreamStep[-R](inner: ZStream[R, Throwable, Step[R]])      extends Step[R]
+
+  case class ObjectStep[-R](name: String, fields: String => Option[Step[R]]) extends Step[R]
```
ghostdogpr: I was thinking we could potentially remove `Option` and return `NullStep` directly to save the allocation. But you won't be able to use `orElse`, even though it can be worked around. Any thoughts?

kyri-petrou: Yeah I was thinking that as well but then hit the `.orElse` point. One way would be to check if the step is a NullValue but that could be a legit case of a field returning None. 

Another would be using a PartialFunction instead, although I don't think it'd be as performant. I can give it a go

core/src/main/scala/caliban/schema/Schema.scala
kyri-petrou: It needs to be lazy otherwise we get a stack-overflow error as the derivation for the field schemas hasn't been performed yet

core/src/main/scala/caliban/schema/Step.scala
kyri-petrou: Yeah PartialFunction seems too messy. Returning `NullStep` and then merging fields as below works, but has the caveat I mentioned above

```scala
case (ObjectStep(name, fields1), ObjectStep(_, fields2)) =>
      ObjectStep(
        name,
        s =>
          fields2(s) match {
            case NullStep => fields1(s)
            case step     => step
          }
      )
```

ghostdogpr: Maybe it's okay since merging arbitrarily keeps one of them anyway (we're not supposed to have the same field twice in theory)?

kyri-petrou: Yeah I think you're right. The chances that this is a legit usecase are extremely low anyways. Changed `fields` to `String => Step[R]` instead

ghostdogpr (approved):

core/src/main/scala/caliban/schema/Step.scala
ghostdogpr: Squeezing the performance to the last bit 😄

PR #2039, opened 2023-12-13
Optimize fragment validation

kyri-petrou:
Last stretch of optimizations before 2.5.0. Recently we focused very heavily on optimizing the Executor so I thought it'd be good to take a look at validation as well.

For the most of it, validation is fairly quick with the exception when fragments / recursive fields are involved, so I spent most of the time optimizing for them.

Main optimizations:
1. Avoiding usage of `ZPure.when(condition)(effect)` when `effect` is a method that relies on recursion (either direct or indirect).
2. "Caching" of the `hashCode` for `Selection` and `SelectionSet`. We use these objects in Sets and as keys in Maps, and due to their recursive nature calculating their hash code is fairly expensive

Other minor optimizations:
1. Change logic in Utils.cross so that we don't return the same tuple in mirror representation
2. Short-circuit `ZPure.foreachDiscard` for empty collections
3. (Executor) Avoid using `.partitionMap` when support for Defer is not enabled

Some of these optimizations require binary incompatible changes. Some of the methods are public, but I doubt that are used by any end-users.

series 2/x:
```
[info] Benchmark                             Mode  Cnt     Score    Error  Units
[info] GraphQLBenchmarks.fragmentsCaliban   thrpt    5   256.513 ±  8.843  ops/s
[info] GraphQLBenchmarks.introspectCaliban  thrpt    5  9000.317 ± 45.534  ops/s
```

PR:
```
[info] Benchmark                             Mode  Cnt      Score     Error  Units
[info] GraphQLBenchmarks.fragmentsCaliban   thrpt    5    482.216 ±   7.531  ops/s
[info] GraphQLBenchmarks.introspectCaliban  thrpt    5  12161.453 ± 139.028  ops/s
```

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -668,20 +668,13 @@ lazy val commonSettings = Def.settings(
   })
 )
 
-lazy val enforceMimaCompatibility = true // Enable / disable failing CI on binary incompatibilities
+lazy val enforceMimaCompatibility = false // Enable / disable failing CI on binary incompatibilities
 
 lazy val enableMimaSettingsJVM =
   Def.settings(
     mimaFailOnProblem     := enforceMimaCompatibility,
     mimaPreviousArtifacts := previousStableVersion.value.map(organization.value %% moduleName.value % _).toSet,
-    mimaBinaryIssueFilters ++= Seq(
-      ProblemFilters.exclude[IncompatibleMethTypeProblem]("caliban.schema.Step#ObjectStep*"),
-      ProblemFilters.exclude[IncompatibleResultTypeProblem]("caliban.schema.Step#ObjectStep*"),
-      ProblemFilters.exclude[DirectMissingMethodProblem]("caliban.schema.Annotations*"),
-      ProblemFilters.exclude[MissingTypesProblem]("caliban.schema.Annotations*"),
-      ProblemFilters.exclude[IncompatibleResultTypeProblem]("caliban.schema.Annotations*"),
-      ProblemFilters.exclude[IncompatibleMethTypeProblem]("caliban.QuickAdapter.toApp")
-    )
+    mimaBinaryIssueFilters ++= Seq()
   )
 
 lazy val enableMimaSettingsJS =
--- a/core/src/main/scala/caliban/execution/Executor.scala
+++ b/core/src/main/scala/caliban/execution/Executor.scala
@@ -67,20 +67,26 @@ object Executor {
     ): ReducedStep[R] = {
 
       def reduceObjectStep(objectName: String, getFieldStep: String => Step[R]): ReducedStep[R] = {
-        val filteredFields    = mergeFields(currentField, objectName)
-        val (deferred, eager) = filteredFields.partitionMap {
-          case f @ Field("__typename", _, _, _, _, _, _, directives, _, _, _)   =>
-            Right((f.aliasedName, PureStep(StringValue(objectName)), fieldInfo(f, path, directives)))
-          case f @ Field(name, _, _, _, _, _, args, directives, _, _, fragment) =>
-            val field = reduceStep(getFieldStep(name), f, args, Left(f.aliasedName) :: path)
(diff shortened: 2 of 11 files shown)


build.sbt
```diff
@@ -668,20 +668,13 @@ lazy val commonSettings = Def.settings(
   })
 )
 
-lazy val enforceMimaCompatibility = true // Enable / disable failing CI on binary incompatibilities
+lazy val enforceMimaCompatibility = false // Enable / disable failing CI on binary incompatibilities
```
kyri-petrou, line 672: PS: I disabled this since I had to add too many filters due to the removed methods

core/src/main/scala/caliban/execution/Executor.scala
```diff
@@ -67,20 +67,26 @@ object Executor {
...
-            val entry = (f.aliasedName, field, fieldInfo(f, path, directives))
-
-            fragment match {
-              // The defer spec provides some latitude on how we handle responses. Since it is more performant to return
-              // pure fields rather than spin up the defer machinery we return pure fields immediately to the caller.
-              case Some(IsDeferred(label)) if isDeferredEnabled && !field.isPure => Left((label, entry))
-              case _                                                             => Right(entry)
+        val (deferred, eager) = {
+          if (isDeferredEnabled) {
+            filteredFields.partitionMap { f =>
```
ghostdogpr, line 77: That one's pretty good! I ran the profiler a whole ago and I remember this `partitionMap` being quite visible in the flamegraph.

core/src/main/scala/caliban/validation/Validator.scala
```diff
@@ -69,6 +68,8 @@ object Validator {
       validateRootQuery(schema)
   }.runEither
 
+  private val zunit = ZPure.unit[Unit]
```
ghostdogpr, line 71: *Looks at the implementation of `ZPure.unit`*
😭

core/src/main/scala/caliban/validation/Validator.scala
```diff
@@ -1178,13 +1166,20 @@ object Validator {
...
         _ <- validateDirectives(t.directives, typeErrorContext)
         _ <- validateInputValueDirectives(t.allInputFields, typeErrorContext)
-        _ <- ZPure.foreachDiscard(t.allFields)(validateFieldDirectives(_, typeErrorContext))
+        _ <- validateAll(t.allFields)(validateFieldDirectives(_, typeErrorContext))
       } yield ()
     }
   }
+
+  // Pure's implementation doesn't check if the Iterable is empty, and this is causing some performance degradation
+  private def validateAll[R, A, B](
```
ghostdogpr, line 1180: How about adding this to `ZPure` so we ultimately don't need this?

ghostdogpr (approved):

core/src/main/scala/caliban/validation/Validator.scala
kyri-petrou: It's a bit tricky. The default implementation of `isEmpty` on an Iterable is to create an Iterator and call `.hasNext()` on it. In cases that the collection is non-empty, this can cause a big allocation. In our case, we're only using it with Lists and Maps, which override the default `.isEmpty` implementation.

Having said that, I don't know of any collections that don't override the isEmpty implementation so it might not be an issue 🤔

core/src/main/scala/caliban/execution/Executor.scala
kyri-petrou: Yeah besides avoiding wrapping the steps into an Either, `.map` is about twice as fast as manually creating a list with a ListBuilder (which is what partitionMap uses)

core/src/main/scala/caliban/validation/Validator.scala
paulpdaniels: Probably the lazy collections like Scala.Stream

PR #2040, opened 2023-12-14
Allow parallel execution of mutation non-top level fields

### LINKED ISSUE #2035: Allow executing mutations / subscriptions in parallel / batched mode
Came across this while optimizing the Executor. Currently we force the execution of mutations and subscriptions to sequential mode https://github.com/ghostdogpr/caliban/blob/078c28de39cc9fd9cdc4eb03a9aed014c63ae4b5/core/src/main/scala/caliban/execution/Executor.scala#L43-L47

According to the spec, [only the top-level fields are required to be executed sequentially](https://spec.graphql.org/October2021/#sec-Mutation) (at least for mutations, couldn't find anything about subscriptions requiring them to be executed serially). Ideally, we'd want to execute the top-level fields of mutations sequentially, but execute any child fields using the user-provided execution mode.

kyri-petrou:
Closes #2035

PS: I didn't find anything in the spec that specifies that subscriptions should be executed sequentially. @ghostdogpr @paulpdaniels do you happen to know how come and the execution mode for subscriptions was forced to `Sequential`?

### DIFF
--- a/core/src/main/scala/caliban/execution/Executor.scala
+++ b/core/src/main/scala/caliban/execution/Executor.scala
@@ -37,26 +37,29 @@ object Executor {
   )(implicit trace: Trace): URIO[R, GraphQLResponse[CalibanError]] = {
     val wrapPureValues    = fieldWrappers.exists(_.wrapPureValues)
     val isDeferredEnabled = featureSet(Feature.Defer)
+    val isMutation        = request.operationType == OperationType.Mutation
 
     type ExecutionQuery[+A] = ZQuery[R, ExecutionError, A]
 
-    val execution = request.operationType match {
-      case OperationType.Query        => queryExecution
-      case OperationType.Mutation     => QueryExecution.Sequential
-      case OperationType.Subscription => QueryExecution.Sequential
+    val executionMode = queryExecution match {
+      case QueryExecution.Sequential => 0
+      case QueryExecution.Parallel   => 1
+      case QueryExecution.Batched    => 2
     }
 
     def collectAll[E, A, B, Coll[+V] <: Iterable[V]](
-      in: Coll[A]
+      in: Coll[A],
+      isTopLevelField: Boolean
     )(
       as: A => ZQuery[R, E, B]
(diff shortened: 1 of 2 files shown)


paulpdaniels:
I don't think it makes a difference for subscriptions because we just produce a response value that contains all the streams

core/src/main/scala/caliban/execution/Executor.scala
```diff
@@ -37,26 +37,29 @@ object Executor {
...
       as: A => ZQuery[R, E, B]
     )(implicit bf: BuildFrom[Coll[A], B, Coll[B]]): ZQuery[R, E, Coll[B]] =
       if (in.sizeCompare(1) == 0) as(in.head).map(bf.newBuilder(in).+=(_).result())
+      else if (isTopLevelField && isMutation) ZQuery.foreach(in)(as)
       else
-        execution match {
-          case QueryExecution.Sequential => ZQuery.foreach(in)(as)
-          case QueryExecution.Parallel   => ZQuery.foreachPar(in)(as)
-          case QueryExecution.Batched    => ZQuery.foreachBatched(in)(as)
+        executionMode match {
```
ghostdogpr: Does that really make a difference to use ints? Code would be cleaner without 😄

ghostdogpr (approved):

core/src/main/scala/caliban/execution/Executor.scala
kyri-petrou: Pattern matching on Ints in increments compiles to a tableswitch, which has O(1) performance regardless of the number of elements in the pattern matching. Plus I do think that equality on ints is less complex than on objects since there's no need for type-checking.

Having said that, we only have 3 elements, and this code is probably a millionth of the overall execution complexity so I'll just go back to making it more readable 😅

kyri-petrou: Made the change. I placed Sequential last now though since I find it very hard to believe anyone would ever want to use that.. right?

ghostdogpr (approved):

core/src/main/scala/caliban/execution/Executor.scala
ghostdogpr: Thanks 😄

PR #2044, opened 2023-12-16
Optimize unnecessary .map when handling nullable fields

kyri-petrou:
There are a few micro-optimizations in this PR, but the main one is that we avoid the extra `.map` for nullable object fields. This results in roughly ~10% in our execution benchmarks, although that's fairly biased as all ZQuery fields are nullable

### DIFF
--- a/core/src/main/scala/caliban/execution/Executor.scala
+++ b/core/src/main/scala/caliban/execution/Executor.scala
@@ -11,7 +11,7 @@ import caliban.schema.Step._
 import caliban.schema.{ ReducedStep, Step, Types }
 import caliban.wrappers.Wrapper.FieldWrapper
 import zio._
-import zio.query.{ Cache, UQuery, URQuery, ZQuery }
+import zio.query._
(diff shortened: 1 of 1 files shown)


core/src/main/scala/caliban/execution/Executor.scala
```diff
@@ -177,14 +182,15 @@ object Executor {
       def objectFieldQuery(name: String, step: ReducedStep[R], info: FieldInfo) = {
         val q = wrap(loop(step), step.isPure, info)
 
-        if (info.details.fieldType.isNullable) q.catchAll(handleError).map((name, _))
-        else q.map((name, _))
+        if (info.details.fieldType.isNullable) {
+          q.foldQuery(
+            handleError(_).map((name, _)),
+            v => ZQuery.succeed((name, v))
+          )
```
kyri-petrou, line 189: TL;DR this is the main optimization

ghostdogpr (approved):

PR #2045, opened 2023-12-17
Add docs for caliban-quick uploads and derivation of fields from methods

kyri-petrou:


vuepress/docs/docs/adapters.md
```diff
@@ -155,14 +156,15 @@ import zio.http._
...
-    handler2 <- api.interpreter.map(QuickAdapter(_).handler)
+    handlers2 <- api.interpreter.map(QuickAdapter(_).handlers)
     // Creates a handler which serves the GraphiQL API from CDN
     graphiql = GraphiQLHandler.handler(apiPath = "/api/graphql", graphiqlPath = "/graphiql")
     app = Routes(
-            Method.ANY / "api" / "graphql" -> handler,
-            Method.GET / "graphiql"        -> graphiql
+            Method.ANY / "api" / "graphql" -> handlers.api,
+            Method.GET / "graphiql"        -> graphiql,
+            Method.POST / "upload" / "graphql" -> handlers.upload
```
ghostdogpr, line 167: Alignment looks a bit weird

vuepress/docs/docs/schema.md
```diff
@@ -242,6 +242,43 @@ given Schema[MyEnv, Origin] = EnvSchema.Auto.derived
   </code-block>
 </code-group>
 
+### Deriving fields from case class methods (Scala 3 only)
+
+In certain cases, your type might contain fields of which their value depends on other fields. For example, you might have a `Person` type with a `fullName` field that is derived from the `firstName` and `lastName` fields. In this case, you can use the `@GQLField` annotation to indicate that the field should be derived from the method with the same name.
```
ghostdogpr:
```suggestion
In certain cases, your type might contain fields whose value depends on other fields. For example, you might have a `Person` type with a `fullName` field that is derived from the `firstName` and `lastName` fields. In this case, you can use the `@GQLField` annotation to indicate that the field should be derived from the method with the same name.
```

vuepress/docs/docs/schema.md
```diff
@@ -242,6 +242,43 @@ given Schema[MyEnv, Origin] = EnvSchema.Auto.derived
...
+  lastName: String!
+  fullName: String!
+}
+```
+
+The methods annotated with `@GQLField` can return any type for which a `Schema` is defined for, including effects such as `ZIO` and `ZQuery`.
+In addition, you can use any other annotation that is supported for case class arguments, such as `@GQLName`, `@GQLDescription` and `@GQLDeprecated`.
+
+::: warning Caveats
+Derivation of fields via the `@GQLField` annotation be convenient in certain cases, but has the following limitations:
```
ghostdogpr:
```suggestion
Derivation of fields via the `@GQLField` annotation can be convenient in certain cases, but has the following limitations:
```

vuepress/docs/docs/adapters.md
kyri-petrou: Ahh yeah good catch 👍

PR #2046, opened 2023-12-17
Use `mapChunksZIO` on streams instead of `mapZIO`

kyri-petrou:


kyri-petrou:
@paulpdaniels I saw this the other day and wasn't sure whether it was the intended behaviour. Can you see any reason for not using `.mapChunksZIO` here?

ghostdogpr (approved):

PR #2048, opened 2023-12-17
Implement `path` via `PathValue` instead of Either[String, Int] for paths

kyri-petrou:
This is more of an experiment just to see what we think.

So, what's wrong with `Either[String, Int]`? There are 2 things that are kind of wrong with it:

1. From a UX / dev perspective, It can be unclear on what it means unless you understand the structure of a graphql path
2. From a performance perspective, it causes unnecessary object allocations and boxing of Ints in the Executor.

With this PR, we substitute the `Either[String, Int]` with `PathValue`, a sealed trait that is extended by `StringValue` and `IntValue.IntNumber`. In a world where we've all migrated to Scala 3, this could have easily been a `type PathValue = StringValue | IntValue.IntNumber` but unfortunately we have to make the code compatible with Scala 2 🥲.

I also made a few changes to some of the methods used to parse paths / IntValues from String to avoid wrapping them in `Try` (which causes boxing) and avoid throwing exceptions as a control flow mechanism

### DIFF
--- a/adapters/quick/src/main/scala/caliban/QuickRequestHandler.scala
+++ b/adapters/quick/src/main/scala/caliban/QuickRequestHandler.scala
@@ -14,7 +14,6 @@ import zio.stacktracer.TracingImplicits.disableAutoTrace
 import zio.stream.ZStream
 
 import java.nio.charset.StandardCharsets.UTF_8
-import scala.util.Try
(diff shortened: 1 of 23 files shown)


ghostdogpr (approved):
Love it!

PR #2052, opened 2023-12-20
More executor optimizations

kyri-petrou:
The main optimization in this PR is that we avoid an extra flatMap for each field when the field is not nullable.

### DIFF
--- a/core/src/main/scala/caliban/execution/Executor.scala
+++ b/core/src/main/scala/caliban/execution/Executor.scala
@@ -38,6 +38,7 @@ object Executor {
     val wrapPureValues    = fieldWrappers.exists(_.wrapPureValues)
     val isDeferredEnabled = featureSet(Feature.Defer)
     val isMutation        = request.operationType == OperationType.Mutation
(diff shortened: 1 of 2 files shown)


core/src/main/scala/caliban/execution/Executor.scala
```diff
@@ -111,7 +112,13 @@ object Executor {
           i += 1
           remaining = remaining.tail
         }
-        reduceList(lb.result(), Types.listOf(currentField.fieldType).fold(false)(_.isNullable))
+        reduceList(
```
ghostdogpr, line 115: Is this actually faster? `fold` is `inline`.

kyri-petrou: Inlining via the `@inline` annotation only works in Scala 2 and only if it's enabled via the `-opt-inline-from:` compiler flag (which we don't have enabled - I think we might want to enable that one).

ghostdogpr: ah :/

ghostdogpr (approved):

PR #2054, opened 2023-12-20
Enable Scala 2 inliner and update dependencies

kyri-petrou:
Subbing in for Scala Steward who seems to be on holidays 🤖🌴 .

Also enabling Scala 2's inliner. I've only enabled inlining from the standard library, but I'm happy to enable it globally if we don't see an issue with it


core/src/main/scala/caliban/parsing/parsers/StringParsers.scala
```diff
@@ -104,9 +104,9 @@ private[caliban] trait StringParsers {
       case _                           => l1
     }
     // remove start lines that are only whitespaces
-    val l3           = l2.dropWhile("[ \t]*".r.replaceAllIn(_, "").isEmpty)
+    val l3           = l2.dropWhile("[ \t]*".r.replaceAllIn(_, "").isEmpty): @noinline
```
ghostdogpr, line 108: Can you add a comment explaining why noinline?

kyri-petrou: Scala 2.12 (only) inlines the `dropWhile` method and a compiler error is raised because of it. I'll add that in

ghostdogpr (approved):

PR #2057, opened 2023-12-24
Create sbt submodules per scala version

kyri-petrou:
This allows us to run the compile / test all modules for a given Scala version which makes dev experience a bit better. e.g., to run all the test suites for Scala 3, all we need to run is `sbt ++3.3.1 rootJVM3/test`.

In addition to improving dev experience, it also makes CI run faster as SBT can parallelise the compiling of SBT modules & test execution

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -23,7 +23,7 @@ jobs:
       - checkout
       - restore_cache:
           key: sbtcache
-      - run: sbt ++2.12.18 core/test http4s/test akkaHttp/test pekkoHttp/test zioHttp/test quickAdapter/test catsInterop/test tools/test codegenSbt/test clientJVM/test monixInterop/compile tapirInterop/test federation/test reporting/test tracing/test
+      - run: sbt ++2.12.18 rootJVM212/Test/compile rootJVM212/test
       - save_cache:
           key: sbtcache
           paths:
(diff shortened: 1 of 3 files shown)


build.sbt
```diff
@@ -104,6 +100,51 @@ lazy val root = project
...
+  .aggregate(allProjects *)
+
+lazy val rootJVM212 = project
+  .in(file("target/rootJVM212"))
+  .settings(
+    crossScalaVersions := Nil,
+    publish / skip     := true,
+    ideSkipProject     := true
+  )
+  .settings(crossScalaVersions := Nil)
```
ghostdogpr: This is duplicated (below as well)

ghostdogpr (commented):
Ah that's nice. How about adding it to the instructions that show up when we start sbt? (for example the scala 3 case)

kyri-petrou:
> Ah that's nice. How about adding it to the instructions that show up when we start sbt? (for example the scala 3 case)

Done - Also made some small changes to it that I think make sense - let me know what you think

.circleci/config.yml
```diff
@@ -23,7 +23,7 @@ jobs:
       - checkout
       - restore_cache:
           key: sbtcache
-      - run: sbt ++2.12.18 core/test http4s/test akkaHttp/test pekkoHttp/test zioHttp/test quickAdapter/test catsInterop/test tools/test codegenSbt/test clientJVM/test monixInterop/compile tapirInterop/test federation/test reporting/test tracing/test
+      - run: sbt ++2.12.18 rootJVM212/compile rootJVM212/test
```
ghostdogpr: Minor: I don't think you need to specify both, `test` will trigger `compile`

ghostdogpr (approved):

.circleci/config.yml
ghostdogpr: I changed it

PR #2060, opened 2023-12-31
Fix content type performance issue

kyri-petrou:
Workaround for a performance issue related to https://github.com/zio/zio-http/issues/2573


ghostdogpr (approved):
Ideally revert that after they fix it 😄

PR #2062, opened 2024-01-02
Add `InjectEnv` implicit for cats.effect.IO

kyri-petrou:
When I started using Caliban, I was using it via the cats-effect interop. One of the things that I found confusing / frustrating at the time was that I couldn't use a concrete IO if I wanted to propagate anything in the environment (like a request context). I wasn't very experienced with Scala / ZIO at the time, so it didn't occur to me I could use an `IOLocal` and create my own `InjectEnv`.

With this PR, we make it easier for beginners / intermediate users to use concrete `cats.effect.IO` while also being able to propagate request context across the ZIO / CE boundary

### DIFF
--- a/examples/src/main/scala/example/interop/cats/ContextualCatsInteropIO.scala
+++ b/examples/src/main/scala/example/interop/cats/ContextualCatsInteropIO.scala
@@ -1,0 +1,111 @@
+package example.interop.cats
+
+import caliban._
+import caliban.interop.cats.CatsInterop
+import caliban.schema.GenericSchema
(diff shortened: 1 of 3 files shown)


ghostdogpr:
How about adding an example similar to https://github.com/ghostdogpr/caliban/blob/series/2.x/examples/src/main/scala/example/interop/cats/ContextualCatsInterop.scala ?

ghostdogpr (approved):

kyri-petrou:
@ghostdogpr sorry I must have missed the notification on this PR. Done 👍

ghostdogpr (approved):

PR #2064, opened 2024-01-03
codegen: Generate `@lazy` fields with abstract effect

johnspade:
Hello!

We use a heavily customized version of the Caliban code generator at $Work and would like to contribute back some of our changes. Caliban recently added support for the `@lazy` directive, the point of this PR is to improve its support for code with abstract effect (`F[_]`). Codegen currently generates code that won't compile due to a missing abstract type parameter in case classes:

```scala
type Foo {
  bar: String!
  baz: String! @lazy
}

object Types {
  final case class Foo(bar: String, baz: F[String])
}
```

This PR adds a couple tests and changes for object and nested lazy fields generation.

### DIFF
--- a/tools/src/main/scala/caliban/tools/SchemaWriter.scala
+++ b/tools/src/main/scala/caliban/tools/SchemaWriter.scala
@@ -9,6 +9,8 @@ import scala.collection.compat._
 
 object SchemaWriter {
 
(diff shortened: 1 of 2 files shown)


ghostdogpr (approved):
Thanks for the improvements!

PR #2075, opened 2024-01-14
Add annotations for forcing object fields as nullable / non-nullable

kyri-petrou:
This change is primarily useful for users that use cats-effect or Futures in their schemas, as they'll be able to force fallible effectual fields to be derived as non-nullable / nullable


ghostdogpr (approved):
Love how simple it was

PR #2077, opened 2024-01-15
Ensure fields are unique after merge

### LINKED ISSUE #2076: Duplicate keys when querying with a fragment compared to inlining
Had an issue beginning with 2.5 where nodes in a connection where not being returned. I can't seem to write a reproducer that does exactly that but given this schema:

```scala
sealed trait Widget
object Widget  {
  implicit val schema: Schema[Any, Widget] = Schema.gen

  case class Args(limit: Option[Int])
  object Args {
    implicit val schema: Schema[Any, Args]    = Schema.gen
    implicit val argBuilder: ArgBuilder[Args] = ArgBuilder.gen
  }

  @GQLName("WidgetA")
  case class A(
    name: String,
    children: Args => ZIO[Any, Throwable, A.Connection]
  ) extends Widget
  object A    {
    implicit val schema: Schema[Any, A] = Schema.gen

    @GQLName("WidgetAConnection")
    case class Connection(total: Int, nodes: Chunk[Child])
    object Connection {
      implicit val schema: Schema[Any, Connection] = Schema.gen
    }

    @GQLName("WidgetAChild")
    case class Child(name: String)
    object Child      {
      implicit val schema: Schema[Any, Child] = Schema.gen
    }
  }

  @GQLName("WidgetB")
  case class B(
    name: String,
    children: Args => ZIO[Any, Throwable, B.Connection]
  ) extends Widget
  object B    {
    implicit val schema: Schema[Any, B] = Schema.gen

    @GQLName("WidgetBConnection")
    case class Connection(total: Int, nodes: Chunk[Child])
    object Connection {
      implicit val schema: Schema[Any, Connection] = Schema.gen
    }

    @GQLName("WidgetBChild")
    case class Child(name: String)
    object Child      {
      implicit val schema: Schema[Any, Child] = Schema.gen
    }
  }
}

case class Queries(
  widget: Option[Widget]
)
object Queries {
  implicit val schema: Schema[Any, Queries] = Schema.gen
}
```

Querying with a fragment:
```graphql
query {
  widget {
    __typename
    ...Widgets
  }
}

fragment Widgets on Widget {
  ... on WidgetA {
    name
    children {
      total
      nodes { name }
    }
  }
  ... on WidgetB {
    name
    children {
      total
      nodes { name }
    }
  }
}
```

returns children with duplicate keys
```json
{
  "widget": {
    "__typename": "WidgetA",
    "name": "a",
    "children": {
      "total": 1,
      "nodes": [
        {"name": "1"}
      ],
      "total": 1,
      "nodes": [
        {"name": "1"}
      ]
    }
  }
}
```

Querying inline:
```graphql
query {
  widget {
    __typename
    ... on WidgetA {
      name
      children {
        total
        nodes { name }
      }
    }
    ... on WidgetB {
      name
      children {
        total
        nodes { name }
      }
    }
  }
}
```
gives
```json
{
  "widget": {
    "__typename": "WidgetA",
    "name": "a",
    "children": {
      "total": 1,
      "nodes": [
        {"name": "1"}
      ]
    }
  }
}
```

both queries produced no duplicates in 2.4.3

kyri-petrou:
Fixes #2076.

Some of the optimizations we did for 2.5.0 required us to rely more heavily on the deduplication happening during the construction of Field. Unfortunately, we missed an edge case where fields with the same name and condition still made it through in the list of fields. ~~To solve this, we need to ensure that fields being merged as part of `combine` are unique~~

EDIT: To solve this, we check for name & condition uniqueness within the Executor. To avoid the additional overhead, we "cache" the uniqueness flag as a lazy val on `Field`.

### DIFF
--- a/core/src/main/scala/caliban/execution/Executor.scala
+++ b/core/src/main/scala/caliban/execution/Executor.scala
@@ -349,16 +349,6 @@ object Executor {
     ZIO.succeed(GraphQLResponse(NullValue, List(error)))
 
   private[caliban] def mergeFields(field: Field, typeName: String): List[Field] = {
-    def haveSameCondition(head: Field, tail: List[Field]): Boolean = {
-      val condition = head._condition
-      var remaining = tail
-      while (remaining ne Nil) {
-        if (remaining.head._condition != condition) return false
-        remaining = remaining.tail
-      }
-      true
-    }
-
     def matchesTypename(f: Field): Boolean =
       f._condition.isEmpty || f._condition.get.contains(typeName)
 
@@ -380,12 +370,11 @@ object Executor {
       map.values().asScala.toList
     }
 
-    field.fields match {
-      // Shortcut if all the fields have the same condition, which means we don't need to merge as that's been handled in Field.apply
-      case h :: t if haveSameCondition(h, t) => if (matchesTypename(h)) field.fields else Nil
-      case Nil                               => Nil
-      case fields                            => mergeFields(fields)
-    }
+    val fields = field.fields
+    if (field.allFieldsUniqueNameAndCondition) {
+      if (fields.isEmpty || !matchesTypename(fields.head)) Nil
+      else fields
+    } else mergeFields(fields)
   }
 
   private def fieldInfo(field: Field, path: List[PathValue], fieldDirectives: List[Directive]): FieldInfo =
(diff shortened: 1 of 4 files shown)


kyri-petrou:
Eghh unfortunately I found another bug, need to take a different approach

kyri-petrou:
Ok now it seems that everything should be working as expected. The behaviour should be the same as that in versions prior to 2.5.0 but without the added overhead of deduplicating unnecessarily during execution

ghostdogpr (approved):

PR #2080, opened 2024-01-16
Re-enable mima for zio-http, play and quick modules

kyri-petrou:
Seems we missed these ones. We shouldn't have any breaking changes though so it 2.5.1 should be okay


kyri-petrou:
Error is "expected", seems that `2.5.1` isn't available in Maven Central yet so it's failing to fetch the artifact

PR #2081, opened 2024-01-16
Propagate type condition from nested fragments to parent

kyri-petrou:
Fix for sub-issue for #2076.

In short, when we have nested fragments with type conditions, we propagate the subtype selection to the top-level fragment.

@guymers, would you by any chance be able to publish this branch locally and test it against your schema to make sure I didn't accidentally cause something else to start failing. It shouldn't, but there might be some edge-case that we didn't account for

### DIFF
--- a/core/src/main/scala/caliban/execution/Field.scala
+++ b/core/src/main/scala/caliban/execution/Field.scala
@@ -11,7 +11,6 @@ import caliban.schema.{ RootType, Types }
 import caliban.{ InputValue, Value }
 
 import scala.collection.mutable
-import scala.jdk.CollectionConverters._
 
 /**
  * Represents a field used during the execution of a query
@@ -136,7 +135,7 @@ object Field {
     directives: List[Directive],
     rootType: RootType
   ): Field = {
-    val memoizedFragments      = new mutable.HashMap[String, (List[Field], Option[String])]()
+    val memoizedFragments      = new mutable.HashMap[String, (List[(Field, Option[String])])]()
     val variableDefinitionsMap = variableDefinitions.map(v => v.name -> v).toMap
 
     def loop(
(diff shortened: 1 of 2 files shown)


kyri-petrou:
@ghostdogpr PS: This change means that we no longer need to check for duplicates within the Executor (changes in the previous PR), but I'd rather play it safe for this release (even if it's ever so slightly less performant) and perhaps change it in the future if we're confident we can remove it

guymers:
This PR passes the test that I added for the issue on 2.5.0 and was failing on 2.5.1 due to the extra fields.

All our existing tests pass (but they pass on all versions).

core/src/main/scala/caliban/execution/Field.scala
```diff
@@ -186,47 +185,60 @@ object Field {
...
+              condition = subtypeNames0
+            ).foreach { case (f, c) =>
+              if (c.exists(isSubsetCondition)) addField(f, c)
+              else addField(f, typeName)
+            }
           }
       }
-      map.values().asScala.toList
+      val builder = List.newBuilder[(Field, Option[String])]
+      map.forEach { case ((_, cond), field) => builder += ((field, cond)) }
```
ghostdogpr, line 236: Is this zio-prelude's forEach? There's no `foreach`?

ghostdogpr (approved):

core/src/main/scala/caliban/execution/Field.scala
kyri-petrou: No this is Java's `LinkedHashMap#forEach`

ghostdogpr: Aha, was confused by the casing

Issue #1926, opened 2023-10-06
Codegen: Support union types as scala 3 union types


oyvindberg:
With namespacing gone after #1925, we're quite close to being able to output union types as union types.

The remaining issue is basically this:
            
  > The problem is that Scala 3 doesn't generate Mirror for union types, so we can't use typeclass derivation for it: https://github.com/lampepfl/dotty/pull/15279

That's true. But, and I'm on very thin ice here, I *think* we can write a macro which pattern matches on the union type, picks out all the members and `summonAll`s `Schema`s for them. 

Then we need to generate some code like this:
```scala
  type SearchResult = Human | Droid | Starship

  given Schema[Any, SearchResult] = Schema.typeUnion[SearchResult]

  // which would expand to something like this
  given Schema[Any, SearchResult] with {
    val _1: Schema[Any, Human] = summon[Schema[Any, Human]]
    val _2: Schema[Any, Droid] = summon[Schema[Any, Droid]]
    val _3: Schema[Any, Starship] = summon[Schema[Any, Starship]]
    val subTypes = List(_1, _2, _3)

    def resolve(value: SearchResult): caliban.schema.Step[Any] =
      value match {
        case x: Human => _1.resolve(x)
        case x: Droid => _2.resolve(x)
        case x: Starship => _3.resolve(x)
      }

    def toType(isInput: Boolean, isSubscription: Boolean): caliban.introspection.adt.__Type =
      caliban.schema.Types.makeUnion(Some("SearchResult"), None, subTypes.map(_.toType_(isInput, isSubscription)))
  }

```

_Originally posted by @oyvindberg in https://github.com/ghostdogpr/caliban/issues/1925#issuecomment-1750114106_

kyri-petrou:
I gave this a go some time ago and failed miserably, albeit my knowledge of macros goes as far as I can google things. Also things might have changed in the compiler since then. If you can make this work it'd be awesome :)

oyvindberg:
I think I can make it work. Progress so far:

Calling the macro like this:
```scala
  type A = Int | String | Types.QueryHeroArgs
  Foo.typeAliasSchema[A]

```

gives this data:

```scala
{
  starwars.generated.Foo.TypeAndSchema.apply[scala.Int]("scala.Int", caliban.schema.Schema.intSchema)
  starwars.generated.Foo.TypeAndSchema.apply[scala.Predef.String]("java.lang.String", caliban.schema.Schema.stringSchema)
[...]

PR #2091, opened 2024-01-21
Typesafe directive - for wrapping case classes on fields (incl. input types) and arguments

develeon:
By using @typesafe directive, the code generator will wrap your typesafe fields in case classes without
imposing further changes to GraphQl schema. Hence allowing type safety on important identifiers and fields.
Tested with Option and List as well.

Schema example:
```
directive @typesafe(name : String) on FIELD_DEFINITION | ARGUMENT_DEFINITION | INPUT_FIELD_DEFINITION

type Query {
  getUser(id: ID! @typesafe(name: "UserId")): User
}

type User {
  id: ID! @typesafe(name: "UserId")
  ...
}
```
with ID as scalar mapping to Int you would get something like this:
```
case class UserId(value: Int) extends AnyVal
object UserId {
  implicit val schema: Schema[Any, UserId] = summon[Schema[Any, Int]].contramap(_.value)
  implicit val argBuilder: ArgBuilder[UserId] = summon[ArgBuilder[Int]].map(UserId(_))
}

final case class QueryGetUserArgs(id: UserId)

final case class User(
  @GQLDirective(Directive("typesafe", Map("name" -> StringValue("UserId"))))
  id: UserId,
  ...
}
```

See SchemaWriterSpec.scala for more extensive type mapping.

### DIFF
--- a/core/src/main/scala/caliban/parsing/adt/Directive.scala
+++ b/core/src/main/scala/caliban/parsing/adt/Directive.scala
@@ -6,21 +6,20 @@ case class Directive(name: String, arguments: Map[String, InputValue] = Map.empt
 
 object Directives {
 
-  val LazyDirective      = "lazy"
-  val TypesafeDirective  = "typesafe"
-  val DeprecatedDiretive = "deprecated"
+  val LazyDirective       = "lazy"
+  val NewtypeDirective    = "newtype"
+  val DeprecatedDirective = "deprecated"
 
   def isDeprecated(directives: List[Directive]): Boolean =
-    directives.exists(_.name == DeprecatedDiretive)
+    directives.exists(_.name == DeprecatedDirective)
 
   def deprecationReason(directives: List[Directive]): Option[String] =
-    findDirective(directives, DeprecatedDiretive, "reason")
+    findDirective(directives, DeprecatedDirective, "reason")
 
-  def isTypesafe(directives: List[Directive]): Boolean          =
-    directives.exists(_.name == TypesafeDirective)
-  def typesafeName(directives: List[Directive]): Option[String] =
-    Directives
-      .findDirective(directives, TypesafeDirective, "name")
+  def isNewType(directives: List[Directive]): Boolean          =
+    directives.exists(_.name == NewtypeDirective)
+  def newTypeName(directives: List[Directive]): Option[String] =
+    findDirective(directives, NewtypeDirective, "name")
 
   private def findDirective(directives: List[Directive], name: String, argument: String): Option[String] =
     directives.collectFirst {
--- a/core/src/main/scala/caliban/parsing/adt/Directive.scala
+++ b/core/src/main/scala/caliban/parsing/adt/Directive.scala
@@ -5,17 +5,32 @@ import caliban.{ InputValue, Value }
 case class Directive(name: String, arguments: Map[String, InputValue] = Map.empty, index: Int = 0)
 
 object Directives {
+
+  val LazyDirective      = "lazy"
+  val TypesafeDirective  = "typesafe"
+  val DeprecatedDiretive = "deprecated"
+
   def isDeprecated(directives: List[Directive]): Boolean =
-    directives.exists(_.name == "deprecated")
+    directives.exists(_.name == DeprecatedDiretive)
 
   def deprecationReason(directives: List[Directive]): Option[String] =
+    findDirective(directives, DeprecatedDiretive, "reason")
+
+  def isTypesafe(directives: List[Directive]): Boolean          =
+    directives.exists(_.name == TypesafeDirective)
+  def typesafeName(directives: List[Directive]): Option[String] =
+    Directives
+      .findDirective(directives, TypesafeDirective, "name")
+
+  private def findDirective(directives: List[Directive], name: String, argument: String): Option[String] =
     directives.collectFirst {
-      case f if f.name == "deprecated" =>
+      case f if f.name == name =>
         f.arguments
-          .get("reason")
+          .get(argument)
           .flatMap(_ match {
             case Value.StringValue(value) => Some(value)
             case _                        => None
           })
     }.flatten
+
 }
--- a/core/src/main/scala/caliban/parsing/adt/Directive.scala
+++ b/core/src/main/scala/caliban/parsing/adt/Directive.scala
@@ -6,9 +6,9 @@ case class Directive(name: String, arguments: Map[String, InputValue] = Map.empt
 
 object Directives {
 
-  val LazyDirective       = "lazy"
-  val NewtypeDirective    = "newtype"
-  val DeprecatedDirective = "deprecated"
+  final val LazyDirective       = "lazy"
+  final val NewtypeDirective    = "newtype"
+  final val DeprecatedDirective = "deprecated"
 
   def isDeprecated(directives: List[Directive]): Boolean =
     directives.exists(_.name == DeprecatedDirective)
--- a/tools/src/main/scala/caliban/tools/SchemaWriter.scala
+++ b/tools/src/main/scala/caliban/tools/SchemaWriter.scala
@@ -2,14 +2,14 @@ package caliban.tools
 
 import caliban.parsing.adt.Definition.TypeSystemDefinition.TypeDefinition
 import caliban.parsing.adt.Definition.TypeSystemDefinition.TypeDefinition._
+import caliban.parsing.adt.Directives.{ LazyDirective, TypesafeDirective }
 import caliban.parsing.adt.Type.{ ListType, NamedType }
-import caliban.parsing.adt.{ Document, Type }
+import caliban.parsing.adt.{ Directive, Directives, Document, Type }
 
+import scala.annotation.tailrec
 import scala.collection.compat._
 
 object SchemaWriter {
-
-  private val LazyDirective = "lazy"
 
   def write(
     schema: Document,
@@ -142,8 +142,9 @@ object SchemaWriter {
         }
        """
 
-    def writeField(field: FieldDefinition, of: TypeDefinition, isMethod: Boolean): String = {
+    def writeField(inputField: FieldDefinition, of: TypeDefinition, isMethod: Boolean): String = {
       def containsNestedDirective(
+        field: FieldDefinition,
         directive: String
       ): Boolean =
         typeNameToDefinitionMap.get(Type.innerType(field.ofType)).fold(false) { t =>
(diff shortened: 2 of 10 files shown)


oyvindberg:
To give some background in case it's not obvious: We're working on improving the story for schema-first development. One thing we identified which we really wanted was to have statically typed IDs on the backend.

So given that preference, we identified a few paths forward:
- add a scalar for each ID type to our schema. As far as I understand, this pushes some complexity over to the clients of the API in that they need to provide a mapping for them when generating code and so on. We want the API to be easily consumable, so that's not optimal.
- add object types for each ID type in the schema. In this case clients will need to handle the wrapping themselves for all queries, and using the API will feel different and less convenient than a normal graphql API.
- This solution where we "tag" ID's as being of different flavours with directives. This way, clients will just see `ID` like in any other graphql API. They are free to read the annotated ID types in the schema as documentation, or customize their own codegen in the same manner as we suggest here. I think this is a very powerful and flexible technique to solve this use case.

Any thoughts on this @ghostdogpr?

ghostdogpr:
This looks good! I just need a little bit of time to review because it's code I'm not super familiar with (schema codegen was mostly contributed externally). Worst case scenario I'll take a look during the weekend 👍 

@kyri-petrou if you have time feel free to review as well

oyvindberg:
Nice! 

One very obvious todo is that it needs to be documented, likely alongside the `@lazy` changes from earlier @develeon . 

Can do that after we agree on the exact path forward here 👍

develeon:
> Nice!
> 
> One very obvious todo is that it needs to be documented, likely alongside the `@lazy` changes from earlier @develeon .
> 
> Can do that after we agree on the exact path forward here 👍

I'm all up for that. Also I think the @lazy directive seems very powerful, so it would be a pity if it where not to be used due to limited documentation.

oyvindberg:
Oh, it is! check https://ghostdogpr.github.io/caliban/docs/server-codegen.html#lazy-evaluation . I meant we should document this on the same page

kyri-petrou:
At a glance it looks good, but I wanna take some more time and understand the code better since I'm not too familiar with it either.

One question that comes to mind at a very high level, why `@typesafe`? Is this a directive already used by other schema-first libraries?

PS: I'm don't mind it either way, I'm just curious on whether there's any specific reasoning behind it :)

develeon:
There is no particular reason for the directive name, so open for suggestions.
Had some other working names, but decided to go back to using @typesafe as feel that more descriptive to what we want to achieve.

core/src/main/scala/caliban/parsing/adt/Directive.scala
```diff
@@ -5,17 +5,32 @@ import caliban.{ InputValue, Value }
 case class Directive(name: String, arguments: Map[String, InputValue] = Map.empty, index: Int = 0)
 
 object Directives {
+
+  val LazyDirective      = "lazy"
+  val TypesafeDirective  = "typesafe"
+  val DeprecatedDiretive = "deprecated"
```
ghostdogpr: typo: `Diretive` => `Directive`

core/src/main/scala/caliban/parsing/adt/Directive.scala
```diff
@@ -5,17 +5,32 @@ import caliban.{ InputValue, Value }
...
-    directives.exists(_.name == "deprecated")
+    directives.exists(_.name == DeprecatedDiretive)
 
   def deprecationReason(directives: List[Directive]): Option[String] =
+    findDirective(directives, DeprecatedDiretive, "reason")
+
+  def isTypesafe(directives: List[Directive]): Boolean          =
+    directives.exists(_.name == TypesafeDirective)
+  def typesafeName(directives: List[Directive]): Option[String] =
+    Directives
```
ghostdogpr: You can call `findDirective` directly

tools/src/main/scala/caliban/tools/SchemaWriter.scala
```diff
@@ -152,29 +153,60 @@ object SchemaWriter {
...
+        s"""$GQLTypesafeDirective${writeDescription(field.description)}${if (isMethod) "def " else ""}${safeName(
           field.name
         )} : $fieldType"""
       }
     }
 
-    def writeInputValue(value: InputValueDefinition): String =
-      s"""${writeDescription(value.description)}${safeName(value.name)} : ${writeType(value.ofType)}"""
+    def writeGQLTypesafeDirective(field: FieldDefinition) =
+      if (field.directives.exists(_.name == TypesafeDirective)) {
```
ghostdogpr: nitpick: rather than `exists` + `filter` + `head`, you could use `find` and then `fold` on the Option. Same below.

tools/src/main/scala/caliban/tools/SchemaWriter.scala
```diff
@@ -183,6 +215,42 @@ object SchemaWriter {
...
+    def writeTypesafeClasses(fieldType: Type, directive: Directive) = {
+      @tailrec def writeTypesafeType(fieldType: Type): String = fieldType match {
+        case NamedType(name, _) => scalarMappings.flatMap(_.get(name)).getOrElse(name)
+        case ListType(ftype, _) => writeTypesafeType(ftype)
+      }
+
+      val fnName       = directive.arguments("name").toInputString.replace("\"", "")
+      val typesafeType = writeTypesafeType(fieldType)
+
+      s"""case class $fnName(value : $typesafeType) extends AnyVal
```
ghostdogpr: Do we need to sanitize the name?

ghostdogpr (commented):
About the name `@typesafe` sounds a little bit too general for what it does here? How about something like `@typeAlias`?

kyri-petrou:
> How about something like `@typeAlias`

I was thinking the same. Another option would be `@newtype` but not sure how well known is that term

develeon:
> > How about something like `@typeAlias`
> 
> I was thinking the same. Another option would be `@newtype` but not sure how well known is that term

I refactor to use `@newtype`, think at least from world of Haskell it is established and fit well with our application. But if you prefer `@typeAlias` I will update it.

Also added documentation.

core/src/main/scala/caliban/parsing/adt/Directive.scala
develeon: Well spotted!

tools/src/main/scala/caliban/tools/SchemaWriter.scala
develeon: Thanks I even found some further optimization cause of this :)

core/src/main/scala/caliban/parsing/adt/Directive.scala
```diff
@@ -5,17 +5,31 @@ import caliban.{ InputValue, Value }
 case class Directive(name: String, arguments: Map[String, InputValue] = Map.empty, index: Int = 0)
 
 object Directives {
+
+  val LazyDirective       = "lazy"
+  val NewtypeDirective    = "newtype"
+  val DeprecatedDirective = "deprecated"
```
kyri-petrou: Very minor, but perhaps make these `final val` instead

tools/src/main/scala/caliban/tools/SchemaWriter.scala
```diff
@@ -183,6 +209,42 @@ object SchemaWriter {
...
+        case NamedType(name, _) => scalarMappings.flatMap(_.get(name)).getOrElse(name)
+        case ListType(ftype, _) => writeTypeOf(ftype)
+      }
+
+      val fnName  = safeName(directive.arguments("name").toInputString.replace("\"", ""))
+      val newtype = safeName(writeTypeOf(fieldType))
+
+      s"""case class $fnName(value : $newtype) extends AnyVal
+         |object $fnName {
+         |    implicit val schema: Schema[Any, $fnName] = summon[Schema[Any, $newtype]].contramap(_.value)
```
kyri-petrou: I'm a bit confused, does `summon` work for Scala 2?

Also, perhaps just use `Schema[$newtype].contramap(_.value)` instead, the `Schema.apply` method "summons" the implicit `Schema`

tools/src/main/scala/caliban/tools/SchemaWriter.scala
```diff
@@ -183,6 +209,42 @@ object SchemaWriter {
...
+         |    implicit val argBuilder: ArgBuilder[$fnName] = summon[ArgBuilder[$newtype]].map($fnName(_))
```
kyri-petrou: Same as above, I believe you can use `[ArgBuilder[$newtype].map($fnName(_))`

kyri-petrou (changes requested):
Looks great overall! My main comment / concern is that unless I'm missing something, the `summon` keyword wouldn't work when generating Scala 2 code

tools/src/main/scala/caliban/tools/SchemaWriter.scala
develeon: Could not get that to work, but used `implicitly` (Scala 2 version of summon).
Tested it on my Scala 3 setup and that worked, but have to double check my Scala options.

ghostdogpr: To catch these kind of issues it would be nice to add the new annotation to the sbt plug-in "scripted" test suite because it runs both on scala 2 and scala 3

kyri-petrou (approved):
I've never worked on this part of the codebase, but AFAICT it looks great :) The only thing that "saddens" me a bit is that we don't get to utilize Scala 3's opaque types, but I think it's likely for the best to have a unified Scala 2/3 implementation.

Thank you very much for the contribution, really highly appreciated!

develeon:
> I've never worked on this part of the codebase, but AFAICT it looks great :) The only thing that "saddens" me a bit is that we don't get to utilize Scala 3's opaque types, but I think it's likely for the best to have a unified Scala 2/3 implementation.
> 
> Thank you very much for the contribution, really highly appreciated!

If we are able to support opaque types in the future I think we can just add a type argument to the directive.
ex. `directive @newtype(name : String, type : String) `

But I did try bit with opaque types initially and ran into a issue, if I recall it was:
`Failed to synthesize an instance of type deriving.Mirror.Of`

Hence similar to the issue for union types: https://github.com/ghostdogpr/caliban/issues/1926

kyri-petrou:
> But I did try bit with opaque types initially and ran into a issue, if I recall it was: `Failed to synthesize an instance of type deriving.Mirror.Of`

With opaque types, you have to generate the Schema in the companion object of the opaque type for it to work, something along these lines:

```scala
import scala.compiletime.summonInline

opaque type Foo = Int
object Foo {
  inline given Schema[Any, Foo] = summonInline[Schema[Any, Int]]
}
```

Either way, I think let's steer off opaque types for now and we can revisit this if there is interest in having that option

ghostdogpr (approved):

vuepress/docs/docs/server-codegen.md
```diff
@@ -62,5 +62,79 @@ And you'll get a case class which looks something like this:
 case class MyType(myLazyField: zio.UIO[String], myField: String)
 ```
 
-When implementing this, `myLazyField` will only be avaluated if the client requested it in the query
+When implementing this, `myLazyField` will only be evaluated if the client requested it in the query
 
+### Newtype declaration
+
+`@Newtype` directive in caliban allows you to wrap your GraphQL fields into statically
```
ghostdogpr:
```suggestion
The `@newtype` directive in caliban allows you to wrap your GraphQL fields into statically
```

Fixed the casing.

Also, do you mind adding the docs change in another PR? We push docs on merge, but since this won't be available until the next release, if it's in a separate PR I can merge it when I do the release 🙏

develeon:
Removed documentation from this PR 
and created new PR with it here : https://github.com/ghostdogpr/caliban/pull/2101

ghostdogpr:
Thanks!

PR #2102, opened 2024-01-30
Optimize field parsing

kyri-petrou:
Improves performance ~15-20% of the parser. The main optimization comes from avoiding parsing the field name twice in cases that it's not aliased (which is almost always)

### DIFF
--- a/core/src/main/scala/caliban/parsing/parsers/SelectionParsers.scala
+++ b/core/src/main/scala/caliban/parsing/parsers/SelectionParsers.scala
@@ -10,7 +10,7 @@ import scala.annotation.nowarn
 
 @nowarn("msg=NoWhitespace") // False positive warning in Scala 2.x
 private[caliban] trait SelectionParsers extends ValueParsers {
-  def alias(implicit ev: P[Any]): P[String] = P(name ~ ":")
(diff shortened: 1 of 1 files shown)


core/src/main/scala/caliban/parsing/parsers/SelectionParsers.scala
```diff
@@ -27,23 +27,18 @@ private[caliban] trait SelectionParsers extends ValueParsers {
...
-        Field(
-          alias,
-          name,
-          args.getOrElse(Map()),
-          dirs.getOrElse(Nil),
-          sels.getOrElse(Nil),
-          index
-        )
+    P(Index ~ name ~ aliasOrName.? ~ arguments.? ~ directives.? ~ selectionSet.?).map {
+      case (index, alias, Some(name), args, dirs, sels) =>
```
kyri-petrou, line 46: This is the main optimization. If the 3rd parser succeeded (indicated by `Some(_)`), then it means we have an alias. This is much faster because the 3rd parser (`aliasOrName`) contains the `:` assertion at the start of the string rather than the end

ghostdogpr (approved):

PR #2103, opened 2024-01-30
Avoid redundant fragment validations

kyri-petrou:
The `FragmentValidator.findConflictsWithinSelectionSet` is by far our most expensive validation, accounting for ~5-10% of the execution time for queries returning a small number of fields even if they don't contain any fragments.

We can avoid it by checking whether the query contains any fragments / inline fragments, and skip it if it doesn't contain any. This should have negligible performance impact for queries containing fragments, and an extremely minor impact for queries that have inlined fragments but not fragment spreads

### DIFF
--- a/core/src/main/scala/caliban/validation/Validator.scala
+++ b/core/src/main/scala/caliban/validation/Validator.scala
@@ -444,14 +444,20 @@ object Validator {
     }
   }
 
+  private def containsFragments(selectionSet: List[Selection]): Boolean =
+    selectionSet.exists {
+      case f: Selection.Field => containsFragments(f.selectionSet)
+      case _                  => true
+    }
+
   private def validateSelectionSet(
     context: Context,
     selectionSet: List[Selection],
     currentType: __Type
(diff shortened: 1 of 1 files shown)


core/src/main/scala/caliban/validation/Validator.scala
```diff
@@ -444,14 +444,20 @@ object Validator {
     }
   }
 
+  private def containsFragments(selectionSet: List[Selection]): Boolean =
+    selectionSet.exists {
+      case f: Selection.Field => containsFragments(f.selectionSet)
+      case _                  => true
```
ghostdogpr: I think it would be nice to explicitly add the 2 cases here: first it's more readable and if some day we add a new case we'll have to update that.

ghostdogpr (approved):

core/src/main/scala/caliban/validation/Validator.scala
kyri-petrou: Funnily enough I had the same thought once I had a look at the PR again. Added 👍

PR #2104, opened 2024-01-30
Series/2.x - Code generation with derives fails for ZIO Environment != Any. 

develeon:
Code generation fails when using derives with other ZIO Environment than Any.

For instance:
`type MyEnv = Tracing with DataSource`

Solution is to alter derives with context for Operation types.
```'scala
  object EnvSchema extends caliban.schema.GenericSchema[MyEnv]

  final case class Query(
      ..
  ) derives EnvSchema.SemiAuto
```

Since code generation is not aware `MyEnv`, a `zioEnv` argument had been introduced to pass the context.


develeon:
What you think about this one @ghostdogpr? 

To us, schema generation unfortunately fails since we use ZQuery and Env != Any. 
This PR solves it as described above, but had to add a CLI argument to pass the Env type.

The mima test fails due to this, but I'm not sure if adding an argument is to be considered an actual breaking change?

Will add documentation on separate branch when we agrees on way forward here. 

Tried to add some helpful comments on changes below.

codegen-sbt/src/sbt-test/codegen/test-compile-newtype/build.sbt
```diff
@@ -15,14 +15,32 @@ lazy val root = project
...
+        case Some((3, _)) =>
+          Seq(
+            calibanSetting(file("src/main/graphql/schema.graphql"))(
+              _.genType(Codegen.GenType.Schema)
+                .clientName("GeneratedAPI")
+                .packageName("graphql")
+                .effect("MyZQuery")
+                .scalarMapping("ID" -> "String")
+                .addDerives(true)
+                .zioEnv("graphql.Env")
```
develeon: New CLI argument here.. passing the Env type. 
Since this is fix for derives, it only applies to Scala 3.

codegen-sbt/src/sbt-test/codegen/test-compile-newtype/modules/base/src/main/scala/graphql/package.scala
```diff
@@ -1,7 +1,8 @@
 import zio.query.ZQuery
+import javax.sql.DataSource
 
 package object graphql {
-  type Env         = Any
+  type Env         = DataSource
```
develeon, line 5: Env is of another type than Any.

tools/src/main/scala/caliban/tools/SchemaWriter.scala
```diff
@@ -19,13 +19,27 @@ object SchemaWriter {
...
-    addDerives: Boolean = false
+    addDerives: Boolean = false,
+    zioEnv: Option[String] = None
   ): String = {
-    val derivesSchema: String =
-      if (addDerives) " derives caliban.schema.Schema.SemiAuto" else ""
 
-    val derivesSchemaAndArgBuilder: String =
-      if (addDerives) " derives caliban.schema.Schema.SemiAuto, caliban.schema.ArgBuilder" else ""
+    val (derivesSchema, derivesSchemaAndArgBuilder, derivesEnvSchema, derivesOnQuery) = (addDerives, zioEnv) match {
```
develeon: Actual fixup is here.. pattern matching on derives and zioEnv..

tools/src/main/scala/caliban/tools/SchemaWriter.scala
```diff
@@ -82,7 +96,7 @@ object SchemaWriter {
       s"""
          |${writeDescription(op.description)}final case class ${op.name}${generic(op, isRootDefinition = true)}(
          |${op.fields.map(c => writeRootField(c, op)).mkString(",\n")}
-         |)$derivesSchema""".stripMargin
+         |)$derivesOnQuery""".stripMargin
```
develeon, line 100: We update all type Operation.

codegen-sbt/src/sbt-test/codegen/test-compile-newtype/build.sbt
ghostdogpr: I wonder if we should name it `envForDerives` or something like that, otherwise people may be confused that it's used in the effects returned or elsewhere.

ghostdogpr:
Looks good to me! For Mima, I think it's okay to just exclude the `tools` project for now (we've done it before). The impact should be zero and I don't think we need to do a major release for that. @kyri-petrou do you agree?

develeon:
> Looks good to me! For Mima, I think it's okay to just exclude the `tools` project for now (we've done it before). The impact should be zero and I don't think we need to do a major release for that. @kyri-petrou do you agree?

Added documentation in separate PR: https://github.com/ghostdogpr/caliban/pull/2108

tools/src/main/scala/caliban/tools/SchemaWriter.scala
```diff
@@ -19,13 +19,27 @@ object SchemaWriter {
...
 
-    val derivesSchemaAndArgBuilder: String =
-      if (addDerives) " derives caliban.schema.Schema.SemiAuto, caliban.schema.ArgBuilder" else ""
+    val (derivesSchema, derivesSchemaAndArgBuilder, derivesEnvSchema, derivesOnQuery) = (addDerives, zioEnv) match {
+      case (false, _)                                        => ("", "", "", "")
+      case (true, Some(env)) if !env.equalsIgnoreCase("Any") =>
+        (
+          " derives caliban.schema.Schema.SemiAuto",
+          " derives caliban.schema.Schema.SemiAuto, caliban.schema.ArgBuilder",
+          s"object EnvSchema extends caliban.schema.GenericSchema[${safeName(env)}]\n\n",
```
kyri-petrou: Prefer extending `SchemaDerivation` instead of `GenericSchema` for Scala 3. I had issues with extending `GenericSchema` in the past

kyri-petrou (commented):
Minor comment below, otherwise LGTM.

Re mima: I think it's okay to exclude it

ghostdogpr:
- did you see my comment about the naming? https://github.com/ghostdogpr/caliban/pull/2104#discussion_r1473650750
- to exclude mima you can comment it like we did here previously https://github.com/ghostdogpr/caliban/pull/2080

develeon:
> * did you see my comment about the naming? [Series/2.x - Code generation with derives fails for ZIO Environment != Any.  #2104 (comment)](https://github.com/ghostdogpr/caliban/pull/2104#discussion_r1473650750)
> 
>     * to exclude mima you can comment it like we did here previously [Re-enable mima for zio-http, play and quick modules #2080](https://github.com/ghostdogpr/caliban/pull/2080)

Sorry I missed that. Will fix

ghostdogpr (approved):

ghostdogpr:
Thanks!

PR #2107, opened 2024-01-30
codegen: Generate the GQLDeprecated annotation

johnspade:
When using the schema-first approach, it is useful to add annotations from the schema to the generated Caliban code, they will be used for introspection. It's helpful to see deprecated fields and types in tools like Graphiql.

The changes also include a small improvement following up on my previous PR #2064: root fields should also be generated with the parameterized effect type in the case of nested lazy fields.


ghostdogpr:
Nice! Could you add one case of this to the existing scripted tests (in the sbt plugin module). It will make sure the generated code compiles well.

johnspade:
> Nice! Could you add one case of this to the existing scripted tests (in the sbt plugin module). It will make sure the generated code compiles well.

Done, added these cases to the scripted tests

ghostdogpr (approved):
Thanks!

PR #2110, opened 2024-02-01
add-extensions-in-ServerError-print

harrylaou:
I think it is useful to have it included. 
In a case I am working it has info useful to the caller. 
eg 
```
"extensions": {
                "code": "BAD_REQUEST_ERROR",
                "information": [
                    "owner.id - Expected string, received null"
                ],
                "timestamp": "2024-02-01T09:02:37.910Z"
            }
```

### DIFF
--- a/client/src/main/scala/caliban/client/CalibanClientError.scala
+++ b/client/src/main/scala/caliban/client/CalibanClientError.scala
@@ -36,7 +36,7 @@ object CalibanClientError {
      * @param includeExtensions whether to include the extensions in the error message
      * @return the error message
      */
-    override def render(includeExtensions: Boolean = false) =
+    override def render(includeExtensions: Boolean = false): String =
       s"Server Error: ${errors.map(_.render(includeExtensions)).mkString("\n")}"
 
   }
--- a/client/src/main/scala/caliban/client/CalibanClientError.scala
+++ b/client/src/main/scala/caliban/client/CalibanClientError.scala
@@ -5,6 +5,7 @@ package caliban.client
  */
 sealed trait CalibanClientError extends Throwable with Product with Serializable {
   override def getMessage: String = toString
+  def getFullMessage: String   = toString
 }
 
 object CalibanClientError {
@@ -28,20 +29,9 @@ object CalibanClientError {
    */
   case class ServerError(errors: List[GraphQLResponseError]) extends CalibanClientError {
 
-    override def toString: String =
-      s"Server Error: ${errors
-        .map(e =>
-          s"${e.message} ${e.locations
-            .getOrElse(Nil)
-            .map(loc => s"at line ${loc.line} and column ${loc.column}")
-            .mkString(" ")}${e.path.fold("")(p =>
-            s" at path ${p.map {
-              case Left(value)  => s"/$value"
-              case Right(value) => s"[$value]"
-            }.mkString("")}${e.extensions.fold("")(ext => s" Extensions: $ext")}"
-          )}"
-        )
-        .mkString("\n")}"
-  }
+    override def toString: String = s"Server Error: ${errors.map(_.print).mkString("\n")}"
 
+    override def getFullMessage = s"Server Error: ${errors.map(_.fullPrint).mkString("\n")}"
+
+  }
 }
--- a/client/src/main/scala/caliban/client/CalibanClientError.scala
+++ b/client/src/main/scala/caliban/client/CalibanClientError.scala
@@ -38,7 +38,7 @@ object CalibanClientError {
             s" at path ${p.map {
               case Left(value)  => s"/$value"
               case Right(value) => s"[$value]"
-            }.mkString("")}"
+            }.mkString("")}${e.extensions.fold("")(ext => s" Extensions: $ext")}"
           )}"
         )
         .mkString("\n")}"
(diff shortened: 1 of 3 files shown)


client/src/main/scala/caliban/client/CalibanClientError.scala
```diff
@@ -38,7 +38,7 @@ object CalibanClientError {
             s" at path ${p.map {
               case Left(value)  => s"/$value"
               case Right(value) => s"[$value]"
-            }.mkString("")}"
+            }.mkString("")}${e.extensions.fold("")(ext => s" Extensions: $ext")}"
```
kyri-petrou: We need to be a bit careful here. If the server includes traces in the response, this could end up being a _very_ large error message.

harrylaou: Sure. 
What do you suggest? 
Some characters limit? Or something else?

My motivation for this is the above ☝🏼  error message. 

I consider it useful to be shown.

kyri-petrou: I think character limit makes sense, although we'll need to preserve the current behaviour so the character limit should be applied for extensions only, with a message as `< error message truncated due to size >` at the end (or something along these lines)

Also, I would suggest changing the method to use a `StringBuilder` instead of string interpolation, as it's more performant and makes things a bit more readable. With a StringBuilder you can keep checking its size as well as you append more to it

kyri-petrou: We could potentially have another method as well which doesn't truncate the error message

harrylaou: >We could potentially have another method as well which doesn't truncate the error message

Yes, sounds like a better idea. So existing behaviour doesn't change
I have updated the PR

client/src/main/scala/caliban/client/GraphQLResponseError.scala
```diff
@@ -17,7 +17,17 @@ case class GraphQLResponseError(
   locations: Option[List[Location]],
   path: Option[List[Either[String, Int]]],
   extensions: Option[__Value]
-)
+) {
+  def print: String =
```
kyri-petrou: Couple of recommendations here to improve UX a bit:

1. Change name to `render`
2. Add an argument with a default value `includeExtensions = false` (And remove `fullPrint` method below)
3. Add a short scaladoc

client/src/main/scala/caliban/client/CalibanClientError.scala
```diff
@@ -28,20 +29,9 @@ object CalibanClientError {
...
-              case Left(value)  => s"/$value"
-              case Right(value) => s"[$value]"
-            }.mkString("")}"
-          )}"
-        )
-        .mkString("\n")}"
-  }
+    override def toString: String = s"Server Error: ${errors.map(_.print).mkString("\n")}"
+
+    override def getFullMessage = s"Server Error: ${errors.map(_.fullPrint).mkString("\n")}"
```
kyri-petrou: Same as below, change this to `render` and add the `includeExtensions` argument. Also change the definition of `toString` to:

`override def toString: String = render(false)`

harrylaou: Sure. 
Please have another look.

client/src/main/scala/caliban/client/GraphQLResponseError.scala
```diff
@@ -17,7 +17,21 @@ case class GraphQLResponseError(
...
+   * @param includeExtensions whether to include the extensions in the error message
+   * @return the error message
+   */
+  def render(includeExtensions: Boolean): String =
+    s"${message} ${locations.getOrElse(Nil).map(loc => s"at line ${loc.line} and column ${loc.column}").mkString(" ")}${path.fold("")(p =>
+      s" at path ${p.map {
+        case Left(value)  => s"/$value"
+        case Right(value) => s"[$value]"
+      }.mkString("")}"
+    )}${if (includeExtensions) extensions.fold("")(ext => s" Extensions: $ext") else ""}"
```
paulpdaniels, line 33: We have a Renderer type already that you could use for this. I'm a little late in reviewing this so don't consider it a blocker. But it should also be more performant than string interpolation.

paulpdaniels: Ah forget I mentioned that, this is client  module, nvm

kyri-petrou (approved):

client/src/main/scala/caliban/client/CalibanClientError.scala
```diff
@@ -28,20 +29,15 @@ object CalibanClientError {
...
-        .mkString("\n")}"
-  }
+    override def toString: String = render(false)
+
+    /**
+     * Renders the error as a string
+     * @param includeExtensions whether to include the extensions in the error message
+     * @return the error message
+     */
+    override def render(includeExtensions: Boolean = false) =
```
ghostdogpr: Can you add the `String` return type explicitly?

harrylaou: Sure. Added.

ghostdogpr:
Thanks!

PR #2111, opened 2024-02-04
Parser & Validator optimizations

kyri-petrou:
Parser (~15% improvement based on our benchmark):

1. Avoid usage of named parser (i.e., `P( ... )`). Based on our benchmarks, they seem to perform a bit better. I think this is ok given that we don't do really use the parser name?
2. Introduce shortcuts cases where `.rep` is used. e.g., if we're parsing a selection set, we use the `!"}"` at the start of the parser to signal it that we reached the end of the selection set
3. Use `StringIn("caseA", "caseB")` in favour of `"caseA" | "caseB"`

Validator:

1. Change the `Validator.check` method to return an Either instead of a ZPure. It seems that past this method we don't need anything that is ZPure specific, and reducing the number of flatmaps improves performance particularly for very simple / small queries

General:
1. Added the `Validator.prepareEither` and `Parser.parseQueryEither` methods as a UX improvement for debugging / benchmarking

### DIFF
--- a/core/src/main/scala/caliban/execution/Executor.scala
+++ b/core/src/main/scala/caliban/execution/Executor.scala
@@ -244,7 +244,7 @@ object Executor {
 
         def collectMixed() = {
           val queries                       = new VectorBuilder[(String, ReducedStep[R], FieldInfo)]
-          val nil                           = Nil
+          val nil                           = Nil // Bring into stack memory to avoid fetching from heap on each iteration
           var names: List[String]           = nil
           var resolved: List[ResponseValue] = nil
           var remaining                     = steps
--- a/core/src/main/scala/caliban/execution/Executor.scala
+++ b/core/src/main/scala/caliban/execution/Executor.scala
@@ -109,8 +109,9 @@ object Executor {
       def reduceListStep(steps: List[Step[R]]) = {
         var i         = 0
         val lb        = ListBuffer.empty[ReducedStep[R]]
+        val nil       = Nil
         var remaining = steps
-        while (remaining ne Nil) {
+        while (remaining ne nil) {
           lb addOne reduceStep(remaining.head, currentField, arguments, PathValue.Index(i) :: path)
           i += 1
           remaining = remaining.tail
@@ -207,9 +208,10 @@ object Executor {
         def collectAllQueries() = {
           def combineQueryResults(results: List[ResponseValue]) = {
             val builder = ListBuffer.empty[(String, ResponseValue)]
+            val nil     = Nil
             var resps   = results
             var names   = steps
-            while (resps ne Nil) {
+            while (resps ne nil) {
               val (name, _, _) = names.head
               builder addOne ((name, resps.head))
               resps = resps.tail
@@ -224,11 +226,12 @@ object Executor {
         }
 
         def combineResults(names: List[String], resolved: List[ResponseValue])(fromQueries: Vector[ResponseValue]) = {
-          var results: List[(String, ResponseValue)] = Nil
+          val nil                                    = Nil
+          var results: List[(String, ResponseValue)] = nil
           var i                                      = fromQueries.length
           var remainingResponses                     = resolved
           var remainingNames                         = names
-          while (remainingResponses ne Nil) {
+          while (remainingResponses ne nil) {
             val name = remainingNames.head
             var resp = remainingResponses.head
             if (resp eq null) {
@@ -369,8 +372,9 @@ object Executor {
 
     def mergeFields(fields: List[Field]) = {
       val map       = new java.util.LinkedHashMap[String, Field](calculateMapCapacity(fields.size))
+      val nil       = Nil
       var remaining = fields
-      while (remaining ne Nil) {
+      while (remaining ne nil) {
         val h = remaining.head
         if (matchesTypename(h)) {
           map.compute(
@@ -408,9 +412,10 @@ object Executor {
     wrapPureValues: Boolean
   ): ReducedStep[R] = {
     var hasPures   = false
+    val nil        = Nil
     var hasQueries = wrapPureValues
     var remaining  = items
-    while ((remaining ne Nil) && !(hasPures && hasQueries)) {
+    while ((remaining ne nil) && !(hasPures && hasQueries)) {
       if (remaining.head._2.isPure) hasPures = true
       else hasQueries = true
       remaining = remaining.tail
--- a/core/src/main/scala/caliban/execution/Executor.scala
+++ b/core/src/main/scala/caliban/execution/Executor.scala
@@ -204,13 +204,10 @@ object Executor {
         isTopLevelField: Boolean
       ) = {
 
-        def collectAllQueries() =
-          collectAll(steps, isTopLevelField) { case (_, step, info) =>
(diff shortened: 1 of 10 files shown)


core/src/main/scala/caliban/execution/Executor.scala
```diff
@@ -107,10 +108,10 @@ object Executor {
 
       def reduceListStep(steps: List[Step[R]]) = {
         var i         = 0
-        val lb        = List.newBuilder[ReducedStep[R]]
+        val lb        = ListBuffer.empty[ReducedStep[R]]
```
ghostdogpr, line 111: Aren't they the same thing?

core/src/main/scala/caliban/execution/Executor.scala
```diff
@@ -207,7 +208,18 @@ object Executor {
           collectAll(steps, isTopLevelField) { case (_, step, info) =>
             // Only way we could have ended with pure fields here is if we wrap pure values, so we check that first as it's cheaper
             objectFieldQuery(step, info, wrapPureValues && step.isPure)
-          }.map(ls => ObjectValue(ls.lazyZip(steps).map { case (resp, (name, _, _)) => (name, resp) }))
+          }.map { ls =>
+            val builder = ListBuffer.empty[(String, ResponseValue)]
```
ghostdogpr, line 210: Move this to a separate function for readability?

core/src/main/scala/caliban/execution/Executor.scala
```diff
@@ -435,6 +448,11 @@ object Executor {
   private implicit class EnrichedListBufferOps[A](val lb: ListBuffer[A]) extends AnyVal {
     // This method doesn't exist in Scala 2.12 so we just use `.map` for it instead
     def mapInPlace[B](f: A => B): ListBuffer[B] = lb.map(f)
+    def addOne(value: A): ListBuffer[A]         = lb += value
+  }
+
+  private implicit class EnrichedVectorBuilderOps[A](private val lb: VectorBuilder[A]) extends AnyVal {
+    def addOne(value: A): VectorBuilder[A] = lb += value
```
ghostdogpr, line 462: Why?

core/src/main/scala/caliban/validation/Validator.scala
```diff
@@ -397,21 +416,24 @@ object Validator {
...
-          failValidation(
-            s"Fragment '${f.name}' forms a cycle.",
-            "The graph of fragment spreads must not form any cycles including spreading itself. Otherwise an operation could infinitely spread or infinitely execute on cycles in the underlying data."
-          )
-        else zunit
-      )
+      if (context.fragments.isEmpty) zunit
+      else {
+        val spreads     = collectFragmentSpreads(context.selectionSets)
+        val spreadNames = mutable.Set.from(spreads.map(_.name))
```
ghostdogpr, line 417: 😄

core/src/main/scala/caliban/wrappers/Wrapper.scala
```diff
@@ -160,31 +160,34 @@ object Wrapper {
...
-    }
-
-    def finalize[W <: Wrapper[R]](buffer: ListBuffer[W]): List[W] = buffer.sortBy(_.priority).result()
-
-    ZIO
-      .foreachDiscard(wrappers)(loop)
-      .as((finalize(o), finalize(p), finalize(v), finalize(e), finalize(f), finalize(i)))
-  }
+  ] =
+    if (wrappers.isEmpty) ZIO.succeed((Nil, Nil, Nil, Nil, Nil, Nil))
```
ghostdogpr: `ZIO.succeed((Nil, Nil, Nil, Nil, Nil, Nil))` could be saved into a val to prevent allocating it every time?

core/src/main/scala/caliban/execution/Executor.scala
kyri-petrou: Yes, although the ListBuffer is the specific implementation, whereas List.newBuilder returns an instance of GrowableBuilder. It's a micro-optimization so that subsequent `addOne` calls are called on the concrete class rather than on the interface

core/src/main/scala/caliban/execution/Executor.scala
kyri-petrou: Method doesn't exist in Scala 2.12 :(

ghostdogpr (approved):

core/src/main/scala/caliban/execution/Executor.scala
```diff
@@ -229,15 +243,16 @@ object Executor {
         }
 
         def collectMixed() = {
-          val queries   = Vector.newBuilder[(String, ReducedStep[R], FieldInfo)]
-          var names     = List.empty[String]
-          var resolved  = List.empty[ResponseValue]
-          var remaining = steps
-          while (remaining ne Nil) {
+          val queries                       = new VectorBuilder[(String, ReducedStep[R], FieldInfo)]
+          val nil                           = Nil
```
paulpdaniels: Is there some optimization reason to introduce this indirection?

core/src/main/scala/caliban/validation/Validator.scala
```diff
@@ -49,7 +50,9 @@ object Validator {
    * Verifies that the given document is valid for this type. Fails with a [[caliban.CalibanError.ValidationError]] otherwise.
    */
   def validate(document: Document, rootType: RootType)(implicit trace: Trace): IO[ValidationError, Unit] =
-    Configurator.configuration.map(_.validations).flatMap(check(document, rootType, Map.empty, _).unit.toZIO)
+    Configurator.configuration
+      .map(_.validations)
```
paulpdaniels: As long as you're simplifying maybe inline this into the flatMap as well?

core/src/main/scala/caliban/execution/Executor.scala
kyri-petrou: This is a small trick that I saw in jsoniter. It stores the Nil object into the stack memory so that it doesn't need to compare it from the heap on each iteration.

It's a very _micro_-optimization that's why I only did it here where we also initialize multiple values with Nil

ghostdogpr: I feel we should comment why we do such trick, otherwise it might get removed by someone else later.

paulpdaniels: "This one weird trick the compiler doesn't want you to know about" 🤡

core/src/main/scala/caliban/execution/Executor.scala
```diff
@@ -203,11 +204,24 @@ object Executor {
         isTopLevelField: Boolean
       ) = {
 
-        def collectAllQueries() =
+        def collectAllQueries() = {
+          def combineQueryResults(results: List[ResponseValue]) = {
+            val builder = ListBuffer.empty[(String, ResponseValue)]
+            var resps   = results
+            var names   = steps
+            while (resps ne Nil) {
```
paulpdaniels: Would the same trick be useful here?

paulpdaniels (approved):

core/src/main/scala/caliban/execution/Executor.scala
kyri-petrou: For consistency, I used the same pattern across all the methods in the Executor where we manually loop over the elements of a List

PR #2112, opened 2024-02-06
Speed up stream processing by using runForeach rather than mapZIO + runDrain

ghostdogpr:
Recently did some benchmarks and found that `mapZIO` + `runDrain` is much slower than `runForeach` (`mapZIO` should be avoided in general).

### DIFF
--- a/interop/tapir/src/main/scala/caliban/interop/tapir/ws/Protocol.scala
+++ b/interop/tapir/src/main/scala/caliban/interop/tapir/ws/Protocol.scala
@@ -75,7 +75,7 @@ object Protocol {
                                case GraphQLWSInput(Ops.ConnectionInit, id, payload)  =>
                                  val before     = ZIO.whenCase((webSocketHooks.beforeInit, payload)) {
                                    case (Some(beforeInit), Some(payload)) =>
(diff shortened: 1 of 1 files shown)


interop/tapir/src/main/scala/caliban/interop/tapir/ws/Protocol.scala
```diff
@@ -275,7 +274,9 @@ object Protocol {
                                    subscriptions.untrack(id)
                                  case GraphQLWSInput(Ops.ConnectionTerminate, _, _)   =>
                                    ZIO.interrupt
-                               }.runDrain.interruptible
+                                 case _                                               =>
+                                   ZIO.unit
+                               }.interruptible
                                  .catchAll(_ => output.offer(Right(connectionError)))
```
paulpdaniels: Technically we could use `orElse` here as well, which might be slightly more memory efficient

paulpdaniels (approved):

interop/tapir/src/main/scala/caliban/interop/tapir/ws/Protocol.scala
ghostdogpr: Updated it 😄

kyri-petrou (approved):
Nice!

PR #2118, opened 2024-02-13
Define an `innerThrowable` for `ExecutionError`s raised within provided `ArgBuilder`s

kyri-petrou:
As discussed on Discord: https://discord.com/channels/629491597070827530/633200096393166868/1204421751942152193

One downside is that for some ArgBuilders (UUID, temporal) we were already defining inner throwables. I don't think that's very necessary (given that for temporal ABs we add the message to the ExecutionError message. WDUT?

### DIFF
--- a/core/src/main/scala/caliban/schema/ArgBuilder.scala
+++ b/core/src/main/scala/caliban/schema/ArgBuilder.scala
@@ -217,5 +217,8 @@ case object InvalidInputArgument extends NoStackTrace {
   override def getMessage: String = "invalid input argument"
 
   private[caliban] def apply(expected: String, actual: String): ExecutionError =
-    ExecutionError(s"Can't an instance of '$expected' from '$actual''", innerThrowable = Some(InvalidInputArgument))
+    ExecutionError(
+      s"Can't build an instance of '$expected' from '$actual'",
+      innerThrowable = Some(InvalidInputArgument)
+    )
 }
--- a/core/src/main/scala/caliban/schema/ArgBuilder.scala
+++ b/core/src/main/scala/caliban/schema/ArgBuilder.scala
@@ -12,7 +12,6 @@ import java.time.format.DateTimeFormatter
 import java.time.temporal.Temporal
 import java.util.UUID
 import scala.annotation.implicitNotFound
-import scala.collection.immutable.BitSet
 import scala.util.Try
 import scala.util.control.{ NoStackTrace, NonFatal }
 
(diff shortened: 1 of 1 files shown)


core/src/main/scala/caliban/schema/ArgBuilder.scala
```diff
@@ -85,49 +89,50 @@ object ArgBuilder extends ArgBuilderInstances {
 }
 
 trait ArgBuilderInstances extends ArgBuilderDerivation {
+  private def error(msg: String): ExecutionError = ExecutionError(msg, innerThrowable = Some(InvalidInputArgument))
```
ghostdogpr: If we're always writing `Can't build an A from B`, how making A and B parameters and define the message once?

core/src/main/scala/caliban/schema/ArgBuilder.scala
```diff
@@ -205,6 +209,19 @@ trait ArgBuilderInstances extends ArgBuilderDerivation {
...
+case object InvalidInputArgument extends NoStackTrace {
+  override def getMessage: String = "invalid input argument"
+}
+
+private object MkInvalidInputArgument {
+  private val vowels = BitSet(List('a', 'e', 'i', 'o', 'u', 'A', 'E', 'I', 'O', 'U').map(_.toInt): _*)
+
+  def apply(expected: String, actual: String): ExecutionError = {
+    val prefix = if (expected.nonEmpty && vowels.contains(expected.codePointAt(0))) "an" else "a"
+    ExecutionError(s"Can't build $prefix $expected from $actual", innerThrowable = Some(InvalidInputArgument))
```
ghostdogpr: I'd just cheat and change to "build an instance of" 😅

kyri-petrou: Yeah good idea. I was thinking of preserving the error messages as much as possible but to be honest I don't think that's necessary

core/src/main/scala/caliban/schema/ArgBuilder.scala
```diff
@@ -205,6 +209,13 @@ trait ArgBuilderInstances extends ArgBuilderDerivation {
...
     case StringValue(v) => Right(Upload(v))
-    case other          => Left(ExecutionError(s"Can't build an Upload from $other"))
+    case other          => Left(InvalidInputArgument("Upload", other.toString))
   }
 }
+
+case object InvalidInputArgument extends NoStackTrace {
+  override def getMessage: String = "invalid input argument"
+
+  private[caliban] def apply(expected: String, actual: String): ExecutionError =
```
kyri-petrou: @ghostdogpr Is it too weird to have the `apply` method defined within the error itself? Mima complains if I create a private method within the `ArgBuilderInstances` for some reason

ghostdogpr: I think it's okay. Missing the word "build" in the message btw

core/src/main/scala/caliban/schema/ArgBuilder.scala
```diff
@@ -12,8 +12,9 @@ import java.time.format.DateTimeFormatter
 import java.time.temporal.Temporal
 import java.util.UUID
 import scala.annotation.implicitNotFound
+import scala.collection.immutable.BitSet
```
ghostdogpr: Can clean that import

core/src/main/scala/caliban/schema/ArgBuilder.scala
```diff
@@ -205,6 +209,16 @@ trait ArgBuilderInstances extends ArgBuilderDerivation {
...
     case StringValue(v) => Right(Upload(v))
-    case other          => Left(ExecutionError(s"Can't build an Upload from $other"))
+    case other          => Left(InvalidInputArgument("Upload", other.toString))
   }
 }
+
+case object InvalidInputArgument extends NoStackTrace {
+  override def getMessage: String = "invalid input argument"
+
+  private[caliban] def apply(expected: String, actual: String): ExecutionError =
```
ghostdogpr: `expected` and `actual` are a bit misleading because it sounds like they should be equal like in tests.
How about `argType` and `argValue`?

We could potentially accept `Any` for `ArgValue` to avoid doing `toString` multiple times too.

kyri-petrou: Yeah I think that's better 👍

ghostdogpr (approved):

PR #2120, opened 2024-02-17
Schema, ArgBuilder and Executor optimizations

kyri-petrou:
⚠️ Binary incompatible changes (for v2.6.0)

Main changes:
1. `Schema` and `ArgBuilder` are now abstract classes instead of traits. The benefit of this change is hard to quantify since it's highly dependent on the number of schemas that exist in an application 
2. `ReducedStep` is now an abstract class, which improves performance of the `isPure` method call
3. Reworked list reduction so that it's done in a single pass

Note on (3): Due to all the optimizations done in the Executor over the past few months, there was too much code which made things harder to read and follow. As an attempt to make it a bit more readable, I separated the logic used for step reduction and reduced step execution into separate classes. Let me know what you think about that - I can revert those changes if we don't like it


core/src/main/scala/caliban/execution/Executor.scala
```diff
@@ -106,23 +172,53 @@ object Executor {
...
+          PureStep(ListValue(lb.result()))
+        }
+
+        if (steps.isEmpty) PureStep(ListValue(Nil))
+        else {
+          val step = reduceStep(steps.head, currentField, arguments, PathValue.Index(0) :: path)
+          if (step.isPure) {
+            // In 99.99% of the cases, if the head is pure, all the other elements will be pure as well but we catch that error just in case
+            // NOTE: Our entire test suite passes without catching the error
+            try reducePures(step.asInstanceOf[PureStep], steps.tail)
```
paulpdaniels, line 220: Is this faster than doing an `isInstancePf` check first before attempting the cast?

core/src/main/scala/caliban/schema/Step.scala
```diff
@@ -77,8 +77,8 @@ object Step {
       }
 }
 
-sealed trait ReducedStep[-R] { self =>
-  def isPure: Boolean = false
+sealed abstract class ReducedStep[-R] { self =>
+  final def isPure: Boolean = self.isInstanceOf[PureStep]
```
paulpdaniels, line 81: Could we make it a Val?

paulpdaniels (approved):

core/src/main/scala/caliban/execution/Executor.scala
kyri-petrou: For the first step in the list, we already do the `isInstanceOf` check via the `isPure` method. The `try` block is for catching any cast errors in extremely unlikely cases where a list contains a mixture of pure and effectful steps. Because this is a very unlikely case (a list will usually contain either only pure or only effectful steps), we avoid type-checking every element in the list other than the first one

core/src/main/scala/caliban/schema/Step.scala
kyri-petrou: I thought about it, although in most cases, we call the `isPure` method on a subset of all the `ReducedStep`s (e.g., in a list we only call it on the first element). So the performance ends up slightly worse if we make it a `val`

ghostdogpr:
Is it worth potentially breaking user code if there's no clear performance gain? (talking about the first change)

kyri-petrou:
> Is it worth potentially breaking user code if there's no clear performance gain? (talking about the first change)

After thinking about it a bit more, I think that since it's difficult to prove that there will be any quantifiable benefit from changing Schema and ArgBuilder into abstract classes, we should leave them as is for now and perhaps re-consider it in the future if we have proof that it's indeed more performant. I'll revert the changes to those 2

paulpdaniels (approved):

core/src/main/scala/caliban/execution/QueryExecution.scala
```diff
@@ -3,29 +3,39 @@ package caliban.execution
 /**
  * Defines which type of parallelism to use when executing queries
  */
-sealed trait QueryExecution
+sealed trait QueryExecution {
+  def tag: Int
```
ghostdogpr, line 7: This is the pattern that was removed in ZIO and ZPure. Is it beneficial here?

kyri-petrou: Yes, but unlike in ZIO / ZPure here we call the `tag` only once per execution (instead of once per iteration). This way the penalty is negligible while utilizing the benefit of the tableswitch

ghostdogpr (approved):

ghostdogpr:
Let's merge and make the next release 2.6.0

PR #2121, opened 2024-02-18
Resolve empty union objects within Sum schema derivation & cleanups

kyri-petrou:
The main optimization in this PR is that we avoid matching on `PureStep(EnumValue(v))` for all steps by delegating this responsibility to the derived Sum schema.

Other cleanups & minor optimizations that I came across while working on this PR:
1. Moved handling of empty union objects into a common Scala 2 & 3 object to avoid code duplication
2. Added the `__DeprecatedArgs.include` since we've been using `__DeprecatedArgs(Some(true))` quite a lot
3. Moved StreamStep reduction into separate method so that the pattern matching doesn't contain any reduction logic
4. Improved performance of `Field#allFieldsUniqueNameAndCondition` by avoiding creation of the HashSet when there's only 1 queried field in the object

With these changes, we see ~10% improvement in execution of simple queries (where only the top-level field is effectful) and ~5% for cases that fields contain ZQueries

### DIFF
--- a/core/src/main/scala-2/caliban/schema/SchemaDerivation.scala
+++ b/core/src/main/scala-2/caliban/schema/SchemaDerivation.scala
@@ -1,10 +1,11 @@
 package caliban.schema
 
+import caliban.ResponseValue.ObjectValue
 import caliban.Value._
 import caliban.introspection.adt._
 import caliban.parsing.adt.Directive
 import caliban.schema.Annotations._
-import caliban.schema.Step.{ PureStep => _ }
+import caliban.schema.Step.{ MetadataFunctionStep, PureStep => _ }
 import caliban.schema.Types._
 import magnolia1._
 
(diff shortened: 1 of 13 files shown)


core/src/main/scala/caliban/introspection/adt/__DeprecatedArgs.scala
```diff
@@ -1,3 +1,13 @@
...
+
 case class __DeprecatedArgs(includeDeprecated: Option[Boolean] = None)
+
+// NOTE: This object extends AbstractFunction1 to maintain binary compatibility for Scala 2.13.
+// TODO: Remove inheritance in the next major version
+object __DeprecatedArgs extends AbstractFunction1[Option[Boolean], __DeprecatedArgs] {
+  val include: __DeprecatedArgs = __DeprecatedArgs(Some(true))
+
+  def apply(v: Option[Boolean] = None): __DeprecatedArgs = new __DeprecatedArgs(v)
+}
```
kyri-petrou, line 13: Fairly annoying we have to extend `AbstractFunction1` to make MiMA happy with Scala 2.13. We can remove this for 2.6.0

core/src/main/scala-2/caliban/schema/SchemaDerivation.scala
```diff
@@ -1,10 +1,11 @@
 package caliban.schema
 
+import caliban.ResponseValue.ObjectValue
```
ghostdogpr: Are imports right?

core/src/main/scala/caliban/schema/Step.scala
```diff
@@ -1,7 +1,8 @@
 package caliban.schema
 
 import caliban.CalibanError.ExecutionError
-import caliban.Value.NullValue
+import caliban.ResponseValue.ObjectValue
```
ghostdogpr: Unused imports ?

ghostdogpr (approved):

core/src/main/scala/caliban/schema/Step.scala
kyri-petrou: Removed. By the way shall we add a linter to check for unused imports and run it during CI only? It's pretty easy to miss them unfortunately

ghostdogpr: Yeah that'd be nice.

kyri-petrou: Ok I'll add it in a separate PR 👍

PR #2125, opened 2024-02-21
Derives fails for @lazy directive when Env != Any

develeon:
When using @lazy directive and derives. 
Generated objects might have fields with effect types and hence then there will be problem when Env != Any.

This merge request corrects this issue. 

What do you think? @ghostdogpr @kyri-petrou


ghostdogpr (approved):
Looks good, thanks!

PR #2126, opened 2024-02-21
SSE support

SvenW:
This adds support for subscriptions over SSE. Only the [Distinct connection mode](https://github.com/enisdenjo/graphql-sse/blob/master/PROTOCOL.md#distinct-connections-mode) part of the spec. I've verified it by running the zio-http example app and receiving updates there


interop/tapir/src/main/scala/caliban/interop/tapir/TapirAdapter.scala
```diff
@@ -114,6 +120,8 @@ object TapirAdapter {
         case _                                                              => false
       }
     )
+    val isSSE          =
+      request.header(HeaderNames.Accept).map(v => v.contains(MediaType.TextEventStream.toString())).getOrElse(false)
```
kyri-petrou: Consider extracting `MediaType.TextEventStream.toString()` into a `val` to avoid conversion on each request

kyri-petrou: Also maybe use `.exists` instead of `.map(...).getOrElse(false)`

interop/tapir/src/main/scala/caliban/interop/tapir/TapirAdapter.scala
```diff
@@ -142,17 +150,27 @@ object TapirAdapter {
...
-          MediaType.ApplicationJson,
-          code,
-          cacheDirective,
-          encodeSingleResponse(
-            resp,
-            keepDataOnErrors = true,
-            excludeExtensions = cacheDirective.map(_ => Set(Caching.DirectiveName))
-          )
-        )
+        val cacheDirective = computeCacheDirective(response.extensions)
```
kyri-petrou: Prefix is missing 
```suggestion
        val cacheDirective = HttpUtils.computeCacheDirective(response.extensions)
```

interop/tapir/src/main/scala/caliban/interop/tapir/TapirAdapter.scala
```diff
@@ -142,17 +150,27 @@ object TapirAdapter {
...
+              MediaType.ApplicationJson,
+              code,
+              cacheDirective,
+              encodeSingleResponse(
+                resp,
+                keepDataOnErrors = true,
+                excludeExtensions = cacheDirective.map(_ => Set(Caching.DirectiveName))
+              )
+            )
+        }
```
kyri-petrou: Couple of questions / comments here (@paulpdaniels let me know what you think as well)

1. I think the `isSSE` condition can be included in the pattern match above
2. According to the spec, SSE can also be used for `@defer` responses (which we currently support). Is it possible to handle deferred responses via SSE the same way we do we multipart responses?
3. I think we can pass `None` as the `cacheDirective` for SSE responses

interop/tapir/src/main/scala/caliban/interop/tapir/TapirAdapter.scala
```diff
@@ -199,6 +217,37 @@ object TapirAdapter {
     )
   }
 
+  private def encodeTextEventStreamResponse[E, BS](
+    resp: GraphQLResponse[E]
+  )(implicit streamConstructor: StreamConstructor[BS], responseCodec: JsonCodec[ResponseValue]): CalibanBody[BS] = {
+    val response: ZStream[Any, Throwable, ServerSentEvent] = resp.data match {
+      case ObjectValue(fields) =>
+        fields.foldLeft(ZStream.empty: ZStream[Any, Throwable, ServerSentEvent]) { case (_, v) =>
```
kyri-petrou, line 230: Won't this handling cause any top-level fields that are not streams to be discarded?

interop/tapir/src/main/scala/caliban/interop/tapir/TapirAdapter.scala
```diff
@@ -199,6 +217,37 @@ object TapirAdapter {
...
+                    responseCodec.encode(
+                      GraphQLResponse(
+                        ObjectValue(List(fieldName -> r)),
+                        resp.errors
+                      ).toResponseValue
+                    )
+                  ),
+                  Some("next")
+                )
+              }.orDie
```
kyri-petrou: Is the `.orDie` necessary since the method return type is `ZStream[Any, Throwable, ServerSentEvent]`?

interop/tapir/src/main/scala/caliban/interop/tapir/TapirAdapter.scala
```diff
@@ -199,6 +217,37 @@ object TapirAdapter {
...
+            case _                                =>
+              ZStream.succeed(ServerSentEvent(Some(responseCodec.encode(resp.toResponseValue)), Some("next")))
```
kyri-petrou, line 247: Shouldn't the last message be `complete`?

interop/tapir/src/main/scala/caliban/interop/tapir/TapirAdapter.scala
```diff
@@ -199,6 +217,37 @@ object TapirAdapter {
...
+          }
+        }
+      case _                   =>
+        ZStream.succeed(ServerSentEvent(Some(responseCodec.encode(resp.toResponseValue)), Some("next")))
```
kyri-petrou, line 251: Same as above

interop/tapir/src/main/scala/caliban/interop/tapir/TapirAdapter.scala
kyri-petrou: Should we append a `complete` message when the stream finishes?

kyri-petrou (commented):
Thank you very much for this contribution! Frankly I thought adding SSE support would be a lot more complicated 😅 

Other than my comments below, one of the things missing that we'll need is to add some tests in the `TapirSpec` suite.

interop/tapir/src/main/scala/caliban/interop/tapir/TapirAdapter.scala
SvenW: Don't think so. I've tested running regular queries with `text/event-stream` as accept header and that works as expected as well. But you mean if a subscription consists of both streams and regular ResponseValues? I can test that as well

interop/tapir/src/main/scala/caliban/interop/tapir/TapirAdapter.scala
SvenW: I think these should actually be a next event and then a complete event directly after. Data can only be sent in the next event according to the spec, right?

interop/tapir/src/main/scala/caliban/interop/tapir/TapirAdapter.scala
frekw: If my understanding of the spec is correct, we cannot ever get a mix of queries, mutations and/or subscriptions in a single GraphQLRequest. I don't think `GraphQLResponse` currently models this in a correct way so I think this is fine, but maybe changing it to more accurately representing what kind of query was executed would allow us to make less assumptions in the transport layer?

kyri-petrou: We can't get a mix of query / mutation / subscription, but isn't it possible to have multiple top-level subscription fields? e.g.,

```graphql
subscription {
  foo {
    A
  }
  bar {
    B
  }
}
```

If I'm reading the code correctly, the code above will not return the `foo` field

ghostdogpr: from the spec: `Subscription operations must have exactly one root field.` 
https://spec.graphql.org/October2021/#sec-Single-root-field

kyri-petrou: Ahh okay, easy then 👍

interop/tapir/src/main/scala/caliban/interop/tapir/TapirAdapter.scala
kyri-petrou: Ah yeah I think that's correct 👍

kyri-petrou (changes requested):
Really happy with the changes! I tested it locally and it works as expected, well done 🎉. There a few things missing though:

1. (Needed in this PR - sorry for this) We need to add at least 1 test to the `TapirAdapter` test suite. It doesn't need to be something too thorough, but we need to have some sort of an automated test that any future changes don't break this functionality
2. (Follow-up) We need to check if the operation type is a subscription in the `isSSE` check. Currently, if the client sends an `Accept: application/json, text/event-stream` header, we'll always respond with a streamed response even if the operation is a query / mutation.
3. (Follow-up) It'd be good to add this functionality also to the [`caliban-quick` adapter.](https://github.com/ghostdogpr/caliban/blob/25e45f59bdde727e23b7ec80a327f8386ebecf6f/adapters/quick/src/main/scala/caliban/QuickRequestHandler.scala#L20) Since `caliban-quick` doesn't support subscriptions at all currently, it'd be a great addition there since it'll support at least one type of subscriptions now

Regarding (2) and (3) I'm happy for it to be in a different PR or for me to work on those after this PR is merged in if you don't currently have the capacity. I  think (2) might a few changes in `core` so it might be a bit more involved. Just let me if you prefer to work on those @SvenW or whether you're happy for me to take it from there

SvenW:
> Really happy with the changes! I tested it locally and it works as expected, well done 🎉. There a few things missing though:
> 
> 1. (Needed in this PR - sorry for this) We need to add at least 1 test to the `TapirAdapter` test suite. It doesn't need to be something too thorough, but we need to have some sort of an automated test that any future changes don't break this functionality
> 2. (Follow-up) We need to check if the operation type is a subscription in the `isSSE` check. Currently, if the client sends an `Accept: application/json, text/event-stream` header, we'll always respond with a streamed response even if the operation is a query / mutation.
> 3. (Follow-up) It'd be good to add this functionality also to the [`caliban-quick` adapter.](https://github.com/ghostdogpr/caliban/blob/25e45f59bdde727e23b7ec80a327f8386ebecf6f/adapters/quick/src/main/scala/caliban/QuickRequestHandler.scala#L20) Since `caliban-quick` doesn't support subscriptions at all currently, it'd be a great addition there since it'll support at least one type of subscriptions now
> 
> Regarding (2) and (3) I'm happy for it to be in a different PR or for me to work on those after this PR is merged in if you don't currently have the capacity. I think (2) might a few changes in `core` so it might be a bit more involved. Just let me if you prefer to work on those @SvenW or whether you're happy for me to take it from there

Hi @kyri-petrou, I will definitely fix 1, battling the tapir client at the moment. But would love it if you could take 2 and 3 after that. My time has been a little limited this week but I hope to be able to focus more and fix number 1 during next week at least

interop/tapir/src/main/scala/caliban/interop/tapir/TapirAdapter.scala
```diff
@@ -89,6 +91,10 @@ object TapirAdapter {
       oneOfVariantValueMatcher[CalibanBody.Stream[stream.BinaryStream]](
         streamTextBody(stream)(CodecFormat.Json(), Some(StandardCharsets.UTF_8)).toEndpointIO
           .map(Right(_)) { case Right(value) => value }
+      ) { case Right(_) => true },
+      oneOfVariantValueMatcher[CalibanBody.Stream[stream.BinaryStream]](
+        streamBinaryBody(stream)(CodecFormat.TextEventStream()).toEndpointIO
+          .map(Right(_)) { case Right(value) => value }
```
paulpdaniels, line 97: I'm quite surprised this works since there is a `Right(_) => true` above this, does the selector take into account the Accepts headers when computing the match?

interop/tapir/src/main/scala/caliban/interop/tapir/TapirAdapter.scala
```diff
@@ -114,6 +120,8 @@ object TapirAdapter {
         case _                                                              => false
       }
     )
+    val isSSE          =
```
paulpdaniels: Could make this a `def` or `lazy` as well since we don't need to compute it if we match another case first

kyri-petrou: Adding to this, we could probably check the Accept header just a single time (we currently check it once for `acceptsGqlJson` and once for `isSSE`)

interop/tapir/src/main/scala/caliban/interop/tapir/TapirAdapter.scala
SvenW: I honestly don't know if it does take the Accept header into account but it does work. If we can do it in another way that is more readable or so I'm all for it

SvenW:
@kyri-petrou I added tests but now the QuickAdapter fails of course. Can we skip parts of the TapirAdapterSpec for that specific adapter or how should I handle it?

kyri-petrou:
> @kyri-petrou I added tests but now the QuickAdapter fails of course. Can we skip parts of the TapirAdapterSpec for that specific adapter or how should I handle it?

@SvenW Awesome! Yeah disable it for and I'll enable in my PR which will add the functionality for it

interop/tapir/src/main/scala/caliban/interop/tapir/TapirAdapter.scala
kyri-petrou: If memory serves well, I also came across this some time ago and I was also surprised about it. I think I also concluded that it takes the Accept header into account

kyri-petrou:
@SvenW sorry perhaps I wasn't very clear; What I meant above was to add an argument to `TapirAdapterSpec.makeSuite` so that we can enable / disable the SSE subscription tests (similar to how we optionally run WS and Upload tests)

kyri-petrou (approved):

kyri-petrou:
❤️

PR #2130, opened 2024-02-22
Improve UX for constructing a Caliban interpreter via interop

kyri-petrou:
Main changes:
1. Make `GraphQL#interpreterEither` public
2. Deprecate `interpreterAsync` methods in cats / monix interop in favour of `interpreterF` / `interpreterMonix` which don't have an implicit requirement on `ToEffect` / `Runtime`

Some notes:
- We could also add a `GraphQL#interpreterUnsafe` method which impurely throws the `Left` if we wish so to make testing simpler
- Regarding (2): I went with deprecating the methods to maintain binary compatibility. We could instead just change the implicit requirements on the `interpreterAsync` methods which won't need any user-facing changes but that means that the next version will need to be 2.6.0. Frankly, I prefer just doing this.

I also made changes to the documentation but will add them once we decide what to do with `interpreterAsync`

### DIFF
--- a/core/src/main/scala/caliban/GraphQL.scala
+++ b/core/src/main/scala/caliban/GraphQL.scala
@@ -14,7 +14,7 @@ import caliban.validation.Validator
 import caliban.wrappers.Wrapper
 import caliban.wrappers.Wrapper._
 import zio.stacktracer.TracingImplicits.disableAutoTrace
-import zio.{ IO, Trace, URIO, ZIO }
+import zio.{ IO, Trace, URIO, Unsafe, ZIO }
 
 /**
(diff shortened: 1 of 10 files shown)


ghostdogpr (approved):

satorg:
@kyri-petrou , regarding
> We could instead just change the implicit requirements on the `interpreterAsync`

Just my couple of cents: I personally like your decision to change the name to `interpreterF`, because I find the old `interpreterAsync` rather misleading. Because it is not somehow more "async" comparing to the native `interpreter` method that returns `ZIO`. They are both "async" in a similar way. Therefore, `interpreterF` makes more sense to differentiate it from the native one.

PR #2140, opened 2024-02-26
Update link to Auth example in FAQ

yoohaemin:
I fought with ZIO-http for 20 mins until I realized an example was available...


ghostdogpr (approved):
Thanks 😅

PR #2141, opened 2024-02-26
Add support for SSE-based subscriptions for `caliban-quick`

kyri-petrou:
Followup of #2126.

@SvenW can you take a look and let me know if everything looks okay to you?

### DIFF
--- a/adapters/quick/src/main/scala/caliban/QuickRequestHandler.scala
+++ b/adapters/quick/src/main/scala/caliban/QuickRequestHandler.scala
@@ -2,7 +2,8 @@ package caliban
 
 import caliban.Configurator.ExecutionConfiguration
 import caliban.HttpUtils.DeferMultipart
-import caliban.ResponseValue.StreamValue
+import caliban.ResponseValue.{ ObjectValue, StreamValue }
+import caliban.Value.NullValue
 import caliban.interop.jsoniter.ValueJsoniter
 import caliban.uploads.{ FileMeta, GraphQLUploadRequest, Uploads }
 import caliban.wrappers.Caching
@@ -148,13 +149,7 @@ final private class QuickRequestHandler[-R](interpreter: GraphQLInterpreter[R, A
     cacheDirective.fold(headers)(headers.addHeader(Header.CacheControl.name, _))
 
   private def transformResponse(httpReq: Request, resp: GraphQLResponse[Any])(implicit trace: Trace): Response = {
-
-    val acceptsGqlJson: Boolean =
-      httpReq.headers.get(Header.Accept.name).exists { h =>
-        // Better performance than having to parse the Accept header
-        h.length >= 33 && h.toLowerCase.contains("application/graphql-response+json")
-      }
-
+    val accepts        = new HttpUtils.AcceptsGqlEncodings(httpReq.headers.get(Header.Accept.name))
     val cacheDirective = HttpUtils.computeCacheDirective(resp.extensions)
 
     resp match {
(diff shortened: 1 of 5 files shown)


adapters/quick/src/main/scala/caliban/QuickRequestHandler.scala
```diff
@@ -205,11 +202,24 @@ final private class QuickRequestHandler[-R](interpreter: GraphQLInterpreter[R, A
       .mapConcatChunk(Chunk.fromArray)
   }
 
+  private def encodeTextEventStream(resp: GraphQLResponse[Any])(implicit trace: Trace) = {
+    val stream = (resp.data match {
+      case ObjectValue((fieldName, StreamValue(stream)) :: Nil) =>
+        stream.either.map {
+          case Right(r)  => GraphQLResponse(ObjectValue(List(fieldName -> r)), resp.errors)
+          case Left(err) => GraphQLResponse(ObjectValue(List(fieldName -> NullValue)), List(err))
+        }
```
kyri-petrou: @paulpdaniels Is this error handling sound? Also, is there a better way to merge the error and success channels? I couldn't for the life of me figure it out 😕

SvenW:
> Followup of #2126.
> 
> @SvenW can you take a look and let me know if everything looks okay to you?

Should follow the protocol from what I can tell. Nice rewrites of the accept header and the tapirspec as well. A lot more readable

interop/tapir/src/main/scala/caliban/interop/tapir/TapirAdapter.scala
```diff
@@ -149,12 +132,10 @@ object TapirAdapter {
             excludeExtensions = cacheDirective.map(_ => Set(Caching.DirectiveName))
           )
         )
-      case resp if accepts == GraphqlServerSentEvent.mediaType  =>
-        val code = response.errors.collectFirst { case HttpRequestMethod.MutationOverGetError => StatusCode.BadRequest }
-          .getOrElse(StatusCode.Ok)
```
kyri-petrou, line 154: @SvenW I missed this one in the previous PR. [Reading the spec](https://github.com/enisdenjo/graphql-sse/blob/master/PROTOCOL.md), I think we shouldn't do this for SSE (although I don't see how we would get a mutation over SSE anyways):

> Additionally, responding with a 400 (Bad Request) will cause the user agent to fail the connection. In some cases, like with the browser's native EventSource, the error event will hold no meaningful information helping to understand the validation issue(s).

SvenW: Yeah. Sounds like 200 is the only valid code to send

adapters/quick/src/main/scala/caliban/QuickRequestHandler.scala
```diff
@@ -205,11 +202,24 @@ final private class QuickRequestHandler[-R](interpreter: GraphQLInterpreter[R, A
       .mapConcatChunk(Chunk.fromArray)
   }
 
+  private def encodeTextEventStream(resp: GraphQLResponse[Any])(implicit trace: Trace) = {
+    val stream = (resp.data match {
+      case ObjectValue((fieldName, StreamValue(stream)) :: Nil) =>
+        stream.either.map {
+          case Right(r)  => GraphQLResponse(ObjectValue(List(fieldName -> r)), resp.errors)
```
paulpdaniels: Isn't this saying: for all response bodies return the original errors? Would we want errors to be null here perhaps?

kyri-petrou: Hmm I think you're right. It's probably better to send an initial one with all the errors (if they exist) and then the subsequent ones not containing errors

kyri-petrou: I made the the GraphQLResponse stream handling generic and placed it in HttpUtils in the core module so that both the caliban-quick and tapir adapters use the same one. Also the errors now are reported in the first event only

ghostdogpr (approved):

PR #2144, opened 2024-03-01
Return partial errors with application/graphql-response+json header

ghostdogpr:
`data` should be dropped only for parsing and validation errors, not field errors.
```
SHOULD not contain the data entry on document parsing failure when accepting application/graphql-response+json
SHOULD not contain the data entry on document validation failure when accepting application/graphql-response+json
```
Also see https://graphql.github.io/graphql-over-http/draft/#sec-application-graphql-response-json.Examples.Field-errors-encountered-during-execution and https://discord.com/channels/629491597070827530/633200096393166868/1212880877168889917

### DIFF
--- a/adapters/quick/src/main/scala/caliban/QuickRequestHandler.scala
+++ b/adapters/quick/src/main/scala/caliban/QuickRequestHandler.scala
@@ -2,8 +2,7 @@ package caliban
 
 import caliban.Configurator.ExecutionConfiguration
 import caliban.HttpUtils.{ DeferMultipart, ServerSentEvents }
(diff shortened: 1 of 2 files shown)


ghostdogpr:
Verified that we still pass all audits on https://graphql-http.com/ ✅

paulpdaniels (approved):

kyri-petrou:
Oof that was a pretty nasty bug. Thanks for fixing it. Should we create a patch release 5.5.3 release with this fix since we haven't merged in any binary incompatible changes yet?

ghostdogpr:
Yeah I think we can do that! Let's hold on merging the breaking ones.

PR #2145, opened 2024-03-01
Update Scala 3 version to 3.3.3

kyri-petrou:


kyri-petrou:
Going to merge this one in to avoid releasing a new version with 3.3.2 since it breaks binary compatibility:

https://github.com/lampepfl/dotty/releases/tag/3.3.3

PR #2150, opened 2024-03-04
WebSocket support for `caliban-quick`

kyri-petrou:
Main changes:
1. Moved `Protocol`, `WebSocketHooks` and `StreamTransformer` into the core module
2. Added request handling for WS requests in the `caliban-quick` module and exposed methods to configure hooks & keep-alive duration
3. Deprecated the `ZHttpAdapter` object, signalling that it will be removed in a future release

### DIFF
--- a/adapters/quick/src/main/scala/caliban/QuickRequestHandler.scala
+++ b/adapters/quick/src/main/scala/caliban/QuickRequestHandler.scala
@@ -6,9 +6,11 @@ import caliban.ResponseValue.StreamValue
 import caliban.interop.jsoniter.ValueJsoniter
 import caliban.uploads.{ FileMeta, GraphQLUploadRequest, Uploads }
 import caliban.wrappers.Caching
+import caliban.ws.Protocol
 import com.github.plokhotnyuk.jsoniter_scala.core._
 import com.github.plokhotnyuk.jsoniter_scala.macros.JsonCodecMaker
 import zio._
+import zio.http.ChannelEvent.UserEvent.HandshakeComplete
 import zio.http.Header.ContentType
 import zio.http._
 import zio.stacktracer.TracingImplicits.disableAutoTrace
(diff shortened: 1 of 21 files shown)


core/src/main/scala/caliban/ws/WebSocketHooks.scala
```diff
@@ -0,0 +1,125 @@
+package caliban.ws
+
+import caliban.{ GraphQLWSOutput, InputValue, ResponseValue }
+import zio.ZIO
+import zio.stream.ZStream
+
+trait StreamTransformer[-R, +E] {
+  def transform[R1 <: R, E1 >: E](stream: ZStream[R1, E1, GraphQLWSOutput]): ZStream[R1, E1, GraphQLWSOutput]
```
paulpdaniels: I’m wondering if we could use a `ZPipeline` for this instead of defining a new type?

kyri-petrou: Yeah I think that would be ideal, although my only concern is that it's a user-facing breaking change.

Having said that, the migration should be very straightforward by simply using `ZPipeline.fromFunction` so perhaps we should just do it and document the migration

adapters/quick/src/main/scala/caliban/QuickRequestHandler.scala
```diff
@@ -214,6 +235,28 @@ final private class QuickRequestHandler[-R](interpreter: GraphQLInterpreter[R, A
...
+                 .interruptWhen(ch.awaitShutdown)
+                 .map {
+                   case Right(output) => WebSocketFrame.Text(writeToString(output))
+                   case Left(close)   => WebSocketFrame.Close(close.code, Some(close.reason))
+                 }
+      _     <- ZIO.scoped(ch.receiveAll {
+                 case ChannelEvent.UserEventTriggered(HandshakeComplete) =>
+                   out.runForeach(frame => ch.send(ChannelEvent.Read(frame))).forkScoped
+                 case ChannelEvent.Read(WebSocketFrame.Text(text))       =>
+                   ZIO.suspend(queue.offer(readFromString[GraphQLWSInput](text)))
```
kyri-petrou, line 255: @paulpdaniels @frekw, do you happen to know how should the server handle invalid messages? Based on this logic, we are going to raise an error if the frame cannot be decoded to `GraphQLWSInput`, which AFAICT will close the stream.

Is this sound? Or should we be ignoring messages we can't decode?

core/src/main/scala/caliban/ws/WebSocketHooks.scala
paulpdaniels: This is already breaking right? So as long as we are doing that can probably also move this

kyri-petrou: It's not bin-compatible, but other than a deprecation warning the previous changes were source compatible.

Actually I think we might be able to keep it source-compatible (at least for the deprecated trait). Let me give it a go

kyri-petrou: Ok I overloaded the `message` method in the deprecated `WebSocketHooks` object to accept a `StreamTransformer`, that way it should be source-compatible for users that don't manually create an instance of `WebSocketHooks`

examples/src/main/scala/example/stitching/ExampleApp.scala
```diff
@@ -102,6 +104,7 @@ import zio.stream._
 import zio.http._
 import caliban.ZHttpAdapter
 
+@nowarn
```
ghostdogpr: How about using QuickAdapter here?

ghostdogpr (approved):

examples/src/main/scala/example/stitching/ExampleApp.scala
kyri-petrou: Ahh good catch, done 👍

kyri-petrou:
I'll merge this one in before we start getting merge conflicts, but also so that we have a snapshot version to try out / test.

PR #2159, opened 2024-03-13
Update `zio-http` to 3.0.0-RC6

kyri-petrou:


ghostdogpr (approved):
Nice! Any performance gain from RC5 or RC6?

kyri-petrou:
> Nice! Any performance gain from RC5 or RC6?

I'll need to run the tailcall benchmark to be able to tell since we don't have any benchmarks that also benchmark the HTTP server. I'll run them with the snapshot version later today

PR #2163, opened 2024-03-15
Add excludeDeprecated option to client codegen

### LINKED ISSUE #645: Add an option to codegen for including deprecated fields
Currently they are included, but we might want to omit them.

jeejeeone:
Add excludeDeprecated option to client codegen, when enabled excludes deprecated fields and enum values. Resolves https://github.com/ghostdogpr/caliban/issues/645

Note: not quite sure about codegen-sbt part 🔢

### DIFF
--- a/codegen-sbt/src/main/scala/caliban/codegen/CalibanSettings.scala
+++ b/codegen-sbt/src/main/scala/caliban/codegen/CalibanSettings.scala
@@ -28,6 +28,7 @@ sealed trait CalibanSettings {
   final def preserveInputNames(value: Boolean): Self              = withSettings(_.preserveInputNames(value))
   final def addDerives(value: Boolean): Self                      = withSettings(_.addDerives(value))
   final def envForDerives(value: String): Self                    = withSettings(_.envForDerives(value))
+  final def excludeDeprecated(value: Boolean): Self               = withSettings(_.excludeDeprecated(value))
 }
 
 final case class CalibanFileSettings(file: File, settings: CalibanCommonSettings) extends CalibanSettings {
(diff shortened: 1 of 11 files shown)


jeejeeone:
Just realised there is https://github.com/ghostdogpr/caliban/issues/1253 which relates to this since you can have enum with only deprecated fields + same applies for type. Perhaps it's enough to generate compiling code in these cases?

vuepress/docs/docs/client-codegen.md
```diff
@@ -83,13 +83,14 @@ The settings available on the `cs` (`CalibanSettings`) builder are:
 - `enableFmt`: Enable code formatting with scalafmt (default: `true`).
 - `extensibleEnums`: Generate a fallback case class for unknown enum values (default: `false`).
 - `headers` (only defined for `url` settings): Supply extra headers when fetching the schema from a URL.
+- `excludeDeprecated`: Exclude fields and enum values with deprecated directive (default: `false`).
```
ghostdogpr: Thanks for updating the docs! Do you mind doing it in a separate PR, because docs are published right away and it would be better to update them after this is released 🙏

ghostdogpr (commented):
LGTM!

vuepress/docs/docs/client-codegen.md
jeejeeone: No problem 👍 Probably on thu-sunday this week.

jeejeeone: PR for docs https://github.com/ghostdogpr/caliban/pull/2171

ghostdogpr:
Thanks!

PR #2167, opened 2024-03-18
Improve performance of `ApolloFederatedTracing` wrapper

kyri-petrou:
With these changes, we reduce the number of flatmaps performed in the ApolloFederatedTracing wrapper for the happy path (when fields are successes). Since this wrapper wraps pure values by default, this should see a noticeable improvement for users OOTB

### DIFF
--- a/core/src/main/scala/caliban/wrappers/Wrapper.scala
+++ b/core/src/main/scala/caliban/wrappers/Wrapper.scala
@@ -39,7 +39,7 @@ sealed trait Wrapper[-R] extends GraphQLAspect[Nothing, R] { self =>
     that.withWrapper(self)
 
   // Disables tracing only for wrappers in the caliban package
-  final private[caliban] def trace: Trace = Trace.empty
+  final private[caliban] implicit def trace: Trace = Trace.empty
 }
 
 object Wrapper {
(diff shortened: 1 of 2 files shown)


core/src/main/scala/caliban/wrappers/Wrapper.scala
```diff
@@ -39,7 +39,7 @@ sealed trait Wrapper[-R] extends GraphQLAspect[Nothing, R] { self =>
     that.withWrapper(self)
 
   // Disables tracing only for wrappers in the caliban package
-  final private[caliban] def trace: Trace = Trace.empty
+  final private[caliban] implicit def trace: Trace = Trace.empty
```
kyri-petrou, line 42: Just noticed this wasn't an implicit 🤦‍♂️

federation/src/main/scala/caliban/federation/tracing/ApolloFederatedTracing.scala
```diff
@@ -98,41 +98,50 @@ object ApolloFederatedTracing {
     wrapPureValues: Boolean
   ): FieldWrapper[Any] =
     new FieldWrapper[Any](wrapPureValues) {
+
+      private def updateState(startTime: Long, fieldInfo: FieldInfo)(result: Either[ExecutionError, ResponseValue]) =
+        ZQuery.succeed {
```
paulpdaniels: Do we even need to wrap this? We only execute it within a foldQuery. Could we just execute purely and then lift the result?

kyri-petrou: Yeah I think that's better actually. I'll do the change

kyri-petrou:
@ghostdogpr @paulpdaniels I made another small optimization that I missed initially. I'll merge it in once CI passes, but let me know if you have any concerns about it

ghostdogpr (approved):

PR #2180, opened 2024-04-05
Add SemanticNonNull support

XiNiHa:
Adds `SemanticNonNull` support for every effectful types.

Currently, the support is implemented to add `@semanticNonNull` directives to fields that are semantically non-null. However, as [the spec](https://github.com/graphql/graphql-spec/pull/1065) evolves and merges, it's very likely to have a separate syntax to annotate the semantic nullability of a field.

The implementation is very clear for Caliban since we know all the possibilities of errors and nulls. `@semanticNonNull` is only applied when 1. the resolver is faillible 2. the field is optional 3. the result type is not optional.

Although [the spec also supports more detailed configuration of semantic nullability for list fields](https://github.com/apollographql/specs/blob/cc241dc3759513f0fef3ec077ae5b4a84b45c66a/nullability/v0.3/nullability-v0.3.graphql#L14-L27), it was not applicable to Caliban's case since as it [currently doesn't resolve error inside stream to `null`](https://github.com/ghostdogpr/caliban/blob/7ddba86ea5881543df3eed69b74bc4eb0ead5983/core/src/main/scala/caliban/execution/Executor.scala#L248). (actually it feels a bit odd considering the decision of having effects with `E <: Throwable` as nullable)

While I think that the feature should be blocked by default with a feature flag, I have no idea how that would be implemented. Any ideas?

### DIFF
--- a/core/src/main/scala-2/caliban/schema/SchemaDerivation.scala
+++ b/core/src/main/scala-2/caliban/schema/SchemaDerivation.scala
@@ -95,24 +95,27 @@ trait CommonSchemaDerivation[R] {
           ctx.parameters
             .filterNot(_.annotations.exists(_ == GQLExcluded()))
             .map { p =>
-              val (isNullable, isNullabilityForced) = {
+              val (isNullable, isSemanticNonNull) = {
                 val hasNullableAnn = p.annotations.contains(GQLNullable())
                 val hasNonNullAnn  = p.annotations.contains(GQLNonNullable())
-                (!hasNonNullAnn && (hasNullableAnn || p.typeclass.nullable), hasNullableAnn || hasNonNullAnn)
+
+                if (hasNonNullAnn) (false, false)
+                else if (hasNullableAnn || p.typeclass.nullable) (true, false)
+                else if (p.typeclass.canFail) (true, true)
+                else (false, false)
               }
               Types.makeField(
                 getName(p),
                 getDescription(p),
                 p.typeclass.arguments,
                 () =>
-                  if (isNullable || (!isNullabilityForced && p.typeclass.canFail))
-                    p.typeclass.toType_(isInput, isSubscription)
+                  if (isNullable) p.typeclass.toType_(isInput, isSubscription)
                   else p.typeclass.toType_(isInput, isSubscription).nonNull,
                 p.annotations.collectFirst { case GQLDeprecated(_) => () }.isDefined,
                 p.annotations.collectFirst { case GQLDeprecated(reason) => reason },
                 Option(
                   p.annotations.collect { case GQLDirective(dir) => dir }.toList ++ {
-                    if (config.enableSemanticNonNull && !isNullable && p.typeclass.canFail)
+                    if (config.enableSemanticNonNull && isSemanticNonNull)
                       Some(SchemaUtils.SemanticNonNull)
                     else None
                   }
--- a/core/src/main/scala-2/caliban/schema/SchemaDerivation.scala
+++ b/core/src/main/scala-2/caliban/schema/SchemaDerivation.scala
@@ -94,7 +94,12 @@ trait CommonSchemaDerivation[R] {
                   else p.typeclass.toType_(isInput, isSubscription).nonNull,
                 p.annotations.collectFirst { case GQLDeprecated(_) => () }.isDefined,
                 p.annotations.collectFirst { case GQLDeprecated(reason) => reason },
-                Option(p.annotations.collect { case GQLDirective(dir) => dir }.toList).filter(_.nonEmpty)
+                Option(
+                  p.annotations.collect { case GQLDirective(dir) => dir }.toList ++ {
+                    if (isOptional && p.typeclass.semanticNonNull) Some(Directive("semanticNonNull"))
+                    else None
+                  }
+                ).filter(_.nonEmpty)
               )
             }
             .toList,
--- a/core/src/main/scala-2/caliban/schema/SchemaDerivation.scala
+++ b/core/src/main/scala-2/caliban/schema/SchemaDerivation.scala
@@ -106,7 +106,7 @@ trait CommonSchemaDerivation[R] {
                 Option(
                   p.annotations.collect { case GQLDirective(dir) => dir }.toList ++ {
                     if (enableSemanticNonNull && !isNullable && p.typeclass.canFail)
-                      Some(Directive("semanticNonNull"))
+                      Some(SchemaUtils.SemanticNonNull)
                     else None
                   }
                 ).filter(_.nonEmpty)
--- a/core/src/main/scala-2/caliban/schema/SchemaDerivation.scala
+++ b/core/src/main/scala-2/caliban/schema/SchemaDerivation.scala
@@ -12,6 +12,14 @@ import scala.language.experimental.macros
 trait CommonSchemaDerivation[R] {
 
   /**
+   * Enables the `SemanticNonNull` feature on derivation.
+   * It is currently disabled by default since it is not yet stable.
+   *
+   * Override this method and return `true` to enable the feature.
+   */
+  def enableSemanticNonNull: Boolean = false
+
+  /**
    * Default naming logic for input types.
    * This is needed to avoid a name clash between a type used as an input and the same type used as an output.
    * GraphQL needs 2 different types, and they can't have the same name.
@@ -96,7 +104,8 @@ trait CommonSchemaDerivation[R] {
                 p.annotations.collectFirst { case GQLDeprecated(reason) => reason },
                 Option(
                   p.annotations.collect { case GQLDirective(dir) => dir }.toList ++ {
-                    if (isOptional && p.typeclass.semanticNonNull) Some(Directive("semanticNonNull"))
+                    if (enableSemanticNonNull && isOptional && p.typeclass.semanticNonNull)
+                      Some(Directive("semanticNonNull"))
                     else None
                   }
                 ).filter(_.nonEmpty)
--- a/core/src/main/scala-2/caliban/schema/SchemaDerivation.scala
+++ b/core/src/main/scala-2/caliban/schema/SchemaDerivation.scala
@@ -76,7 +76,7 @@ trait CommonSchemaDerivation[R] {
                 getName(p),
                 getDescription(p),
                 () =>
-                  if (p.typeclass.optional || p.typeclass.canFail) p.typeclass.toType_(isInput, isSubscription)
+                  if (p.typeclass.optional) p.typeclass.toType_(isInput, isSubscription)
                   else p.typeclass.toType_(isInput, isSubscription).nonNull,
                 p.annotations.collectFirst { case GQLDefault(v) => v },
                 p.annotations.collectFirst { case GQLDeprecated(_) => () }.isDefined,
@@ -98,7 +98,7 @@ trait CommonSchemaDerivation[R] {
               val (isNullable, isNullabilityForced) = {
                 val hasNullableAnn = p.annotations.contains(GQLNullable())
                 val hasNonNullAnn  = p.annotations.contains(GQLNonNullable())
-                (!hasNonNullAnn && (hasNullableAnn || p.typeclass.optional), hasNullableAnn || hasNonNullAnn)
+                (!hasNonNullAnn && (hasNullableAnn || p.typeclass.nullable), hasNullableAnn || hasNonNullAnn)
               }
               Types.makeField(
                 getName(p),
--- a/core/src/main/scala-2/caliban/schema/SchemaDerivation.scala
+++ b/core/src/main/scala-2/caliban/schema/SchemaDerivation.scala
@@ -69,7 +69,7 @@ trait CommonSchemaDerivation[R] {
                 getName(p),
                 getDescription(p),
                 () =>
-                  if (p.typeclass.optional) p.typeclass.toType_(isInput, isSubscription)
+                  if (p.typeclass.optional || p.typeclass.canFail) p.typeclass.toType_(isInput, isSubscription)
                   else p.typeclass.toType_(isInput, isSubscription).nonNull,
                 p.annotations.collectFirst { case GQLDefault(v) => v },
                 p.annotations.collectFirst { case GQLDeprecated(_) => () }.isDefined,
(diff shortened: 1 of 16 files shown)


kyri-petrou:
I haven't looked into this PR in depth, but is this similar to this annotation?

https://github.com/ghostdogpr/caliban/blob/7ddba86ea5881543df3eed69b74bc4eb0ead5983/core/src/main/scala/caliban/schema/Annotations.scala#L77

XiNiHa:
Nope, it's about having 3 types of nullability instead of two - NonNull, SemanticNonNull, and Nullable. The directive you linked only (and correctly, even with the PR) overrides the nullability to be NonNull, while this PR separates out a new state, "it's NonNull unless it errors".

ghostdogpr:
> While I think that the feature should be blocked by default with a feature flag, I have no idea how that would be implemented. Any ideas?

Hmm, good question. We have a `Feature` concept already for `Defer` and we also `FiberRef` for configuration, but both are unapplicable here because this change affects schema derivation, which is a pure operation.

Maybe adding a method in `SchemaDerivation` that you can override and then requiring something like that:
```
object MySchema extends Schema.SchemaDerivation[Env] {
  override def enableSemanticNonNull: Boolean = true
}

// then use `MySchema.gen`
```

ghostdogpr (approved):
Looks okay to me.

@kyri-petrou can you check too?

kyri-petrou:
@XiNiHa thank you very much for your contribution! I'm trying to understand a bit the spec around `@semanticNonNull` and what is this PR meant to be implementing but I might need some help to understand what's going on

> Although [the spec also supports more detailed configuration of semantic nullability for list fields](https://github.com/apollographql/specs/blob/cc241dc3759513f0fef3ec077ae5b4a84b45c66a/nullability/v0.3/nullability-v0.3.graphql#L14-L27)

The spec linked here that comes from Apollo seems to use a field directive to add the `@semanticallyNonNull` directive to individual fields, irrespective of type. If I'm understanding this spec correctly, that means we would need to use field annotations for this instead of binding this to the type itself

> However, as https://github.com/graphql/graphql-spec/pull/1065 evolves and merges

Now this is the part that really confuses me. This RFC proposes an extension to the type system, for SemanticallyNonNull to be represented via syntax (e.g., `!String`). For this, using the type system like in the PR makes more sense, but then the PR doesn't do any of the changes proposed in the RFC but instead uses a field directive.

@XiNiHa could you please clarify this for me, which spec is it that this PR is implementing?

XiNiHa:
> The spec linked here that comes from Apollo seems to use a field directive to add the @semanticallyNonNull directive to individual fields, irrespective of type. If I'm understanding this spec correctly, that means we would need to use field annotations for this instead of binding this to the type itself

What I meant for the implementation was to add the `@semanticNonNull` annotation on `FIELD_DEFINITION`s, not for `OBJECT`s. Did you find anything that I did wrong? At least the tests are working correctly.

> Now this is the part that really confuses me. This RFC proposes an extension to the type system, for SemanticallyNonNull to be represented via syntax (e.g., !String). For this, using the type system like in the PR makes more sense, but then the PR doesn't do any of the changes proposed in the RFC but instead uses a field directive.

Fully implementing the spec (with the syntax addition) requires much more work for both Caliban and the ecosystem, and AFAIK none of the ecosystem members have implemented the syntax part. Instead, ecosystem members like Relay, Grats, and Apollo Kotlin, are experimenting with only the semantics, by using the directives instead of a separate syntax. Maybe [this document from Grats](https://grats.capt.dev/docs/guides/strict-semantic-nullability/) would be more helpful in understanding the surroundings.

kyri-petrou:
> Did you find anything that I did wrong? At least the tests are working correctly.

I'm a little bit worried about tying the implementation of a field-specific directive to the Schema / type system, as I fear it'll probably going to bite us down the line in some way. There is already some work underway to be able to do Schema transformations, and I feel this might add another dimension to the things we'll need to cater for.

I need to think this through a bit more, but instead of adding `def semanticNonNull: Boolean` to Schema, how about we have something like `def canFail: Boolean`? While it might be similar to what `semanticNonNull` is, it doesn't tie the Schema to field-specific features. Plus this opens up possibilities for other future features as it's much more generic.

In derivation, we can then decide whether we'll add the `@semanticNonNull` annotation to  fields depending on the combination of `canFail`, `optional` and `enableSemanticNonNull` from `SchemaDerivation`.

XiNiHa:
Sounds great. I'll adjust the implementation to reflect what you've proposed!

XiNiHa:
@kyri-petrou Done!

core/src/main/scala/caliban/schema/Schema.scala
```diff
@@ -546,31 +567,34 @@ trait GenericSchema[R] extends SchemaDerivation[R] with TemporalSchema {
   ): Schema[R1, ZStream[R1, Nothing, A]] =
     new Schema[R1, ZStream[R1, Nothing, A]] {
       override def optional: Boolean                                         = false
+      override def canFail: Boolean                                          = ev.canFail
```
kyri-petrou, line 576: Shouldn't this be `false`?

core/src/main/scala/caliban/schema/Schema.scala
```diff
@@ -522,22 +540,25 @@ trait GenericSchema[R] extends SchemaDerivation[R] with TemporalSchema {
   ): Schema[R0, ZQuery[R1, Nothing, A]] =
     new Schema[R0, ZQuery[R1, Nothing, A]] {
       override def optional: Boolean                                         = ev.optional
+      override def canFail: Boolean                                          = ev.canFail
```
kyri-petrou, line 549: Similarly, shouldn't this be `false`?

core/src/main/scala/caliban/schema/Schema.scala
```diff
@@ -496,22 +511,25 @@ trait GenericSchema[R] extends SchemaDerivation[R] with TemporalSchema {
   implicit def infallibleEffectSchema[R0, R1 >: R0, R2 >: R0, A](implicit ev: Schema[R2, A]): Schema[R0, URIO[R1, A]] =
     new Schema[R0, URIO[R1, A]] {
       override def optional: Boolean                                         = ev.optional
+      override def canFail: Boolean                                          = ev.canFail
```
kyri-petrou, line 520: `false`?

core/src/main/scala/caliban/schema/Schema.scala
```diff
@@ -699,43 +728,66 @@ abstract class PartiallyAppliedFieldBase[V](name: String, description: Option[St
...
   )(implicit ev: Schema[R, V1], ft: FieldAttributes): (__Field, V => Step[R]) =
     (makeField, (v: V) => FunctionStep(_ => fn(v).fold(ev.resolve, identity)))
 }
 
-case class PartiallyAppliedFieldWithArgs[V, A](name: String, description: Option[String], directives: List[Directive]) {
+case class PartiallyAppliedFieldWithArgs[V, A](
+  name: String,
+  description: Option[String],
+  directives: List[Directive],
+  enableSemanticNonNull: Boolean
```
kyri-petrou: I think that since users can provide directives here, we should probably leave that up to them to provide the directive without adding the `enableSemanticNonNull` argument.

Also, maybe we should store the directive somewhere instead of recreating it every time

kyri-petrou (commented):
Reviewed from my phone, but I want to test it a bit more on some services at $WORK later just to make sure we're not introducing any derivation regressions

core/src/main/scala/caliban/schema/Schema.scala
XiNiHa: IMO if `A` is a non-effectful faillible type, these should also be faillible. However since we have none of them, it'll always be false anyways. But semantically I see this more correct.

core/src/main/scala/caliban/schema/Schema.scala
XiNiHa: Done!

core/src/main/scala/caliban/schema/SchemaUtils.scala
```diff
@@ -4,8 +4,15 @@ import caliban.ResponseValue.ObjectValue
...
 import caliban.introspection.adt.{ __DeprecatedArgs, __Field, __Type }
 import caliban.schema.Step.MetadataFunctionStep
+import caliban.parsing.adt.Directive
 
 private[schema] object SchemaUtils {
+
+  /**
+   * Directive used to mark a field as semantically non-nullable.
+   */
+  val SemanticNonNull = Directive("semanticNonNull")
```
kyri-petrou, line 14: Perhaps this is better placed in `caliban.parsing.adt.Directive` as we have other directives defined in there

core/src/main/scala-3/caliban/schema/SchemaDerivation.scala
```diff
@@ -23,6 +23,14 @@ object PrintDerived {
 trait CommonSchemaDerivation {
   export DerivationUtils.customizeInputTypeName
 
+  /**
+   * Enables the `SemanticNonNull` feature on derivation.
+   * It is currently disabled by default since it is not yet stable.
+   *
+   * Override this method and return `true` to enable the feature.
+   */
+  def enableSemanticNonNull: Boolean = false
```
kyri-petrou: May I recommend we use a `DerivationConfig` case class instead? I can see us wanting to add more derivation customization options in the future

kyri-petrou (changes requested):
Looks great overall, although I just tested this PR locally with a service at $WORK (Scala 3), and it seems that there is a bug where top-level query and mutation fields that can fail are derived as non-null. For some reason, this doesn't seem to affect any other fields - just top-level ones

Can you please try and reproduce the issue in a test suite and fix the source of it?

Schema rendered with v2.6.0 vs this PR:

image

core/src/main/scala/caliban/schema/SchemaUtils.scala
XiNiHa: Since those are all name strings instead of actual directive instances, I'm worried that adding this causes some confusion :thinking:

kyri-petrou:
By the way before sending you off down some rabbithole, this might be just a rendering issue. So I'd check that first!

core/src/main/scala-3/caliban/schema/SchemaDerivation.scala
XiNiHa: https://github.com/ghostdogpr/caliban/pull/2180/commits/7d8e54627dd580445e5f82984be45c69826a2115

XiNiHa:
> it seems that there is a bug where top-level query and mutation fields that can fail are derived as non-null

I was not able to reproduce this ;(

kyri-petrou:
Ok I managed to track down the source of the issue. It seems that I had a custom schema defined in the application like this:

```scala
  given [A](using ev: Schema[R, A]): Schema[R, FieldOps => A] with {
    override def arguments: List[__InputValue]                    = ev.arguments
    override def optional: Boolean                                = ev.optional
    def toType(isInput: Boolean, isSubscription: Boolean): __Type = ev.toType_(isInput, isSubscription)

    def resolve(value: FieldOps => A): Step[R] = MetadataFunctionStep { f =>
      val fops = new FieldOps(f)
      ev.resolve(value(fops))
    }
  }
```
Since now the underlying `ev.optional` is different, that led to a different schema derivation

While this is not a common use-case, we can't assume that other users aren't doing the same thing. @ghostdogpr open to suggestions on naming, but I suggest we do something along these lines (although I'm kind of inclined towards just making it final and let it break source compatibility)

```scala
  @deprecatedOverriding("this method will be made final. Override canFail and nullable instead", "2.6.1")
  def optional: Boolean = canFail || nullable

  def canFail: Boolean  = false
  def nullable: Boolean = false // Happy for other name suggestions
```

Then the majority of the code should remain the same and use `optional` except for derivation. Thoughts?

XiNiHa:
@kyri-petrou applied your suggestion! (I named it `nullable` for now)

ghostdogpr:
Hmm there is a Mima failure, shall we just exclude it?

core/src/main/scala/caliban/schema/Schema.scala
```diff
@@ -73,9 +73,21 @@ trait Schema[-R, T] { self =>
   def resolve(value: T): Step[R]
 
   /**
-   * Defines if the type is considered optional or non-null. Should be false except for `Option`.
+   * Defines if the type can resolve to null or not.
+   * It is true if the type is nullable or if it can fail.
    */
-  def optional: Boolean = false
+  @deprecatedOverriding("this method will be made final. Override canFail and nullable instead", "2.6.1")
```
kyri-petrou: It looks like we're going for a major release next. Change version to `2.7.0` in the message

core/src/main/scala-2/caliban/schema/SchemaDerivation.scala
```diff
@@ -80,21 +95,28 @@ trait CommonSchemaDerivation[R] {
           ctx.parameters
             .filterNot(_.annotations.exists(_ == GQLExcluded()))
             .map { p =>
-              val isOptional = {
+              val (isNullable, isNullabilityForced) = {
                 val hasNullableAnn = p.annotations.contains(GQLNullable())
                 val hasNonNullAnn  = p.annotations.contains(GQLNonNullable())
-                !hasNonNullAnn && (hasNullableAnn || p.typeclass.optional)
+                (!hasNonNullAnn && (hasNullableAnn || p.typeclass.nullable), hasNullableAnn || hasNonNullAnn)
```
kyri-petrou: I think this should be `optional`?

```suggestion
                (!hasNonNullAnn && (hasNullableAnn || p.typeclass.optional), hasNullableAnn || hasNonNullAnn)
```

core/src/main/scala-3/caliban/schema/DerivationUtils.scala
```diff
@@ -120,27 +120,33 @@ private object DerivationUtils {
...
     getDescription(annotations),
     fields.map { (name, fieldAnnotations, schema) =>
-      val deprecatedReason = getDeprecatedReason(fieldAnnotations)
-      val isOptional       = {
+      val deprecatedReason                  = getDeprecatedReason(fieldAnnotations)
+      val (isNullable, isNullabilityForced) = {
         val hasNullableAnn = fieldAnnotations.contains(GQLNullable())
         val hasNonNullAnn  = fieldAnnotations.contains(GQLNonNullable())
-        !hasNonNullAnn && (hasNullableAnn || schema.optional)
+        (!hasNonNullAnn && (hasNullableAnn || schema.nullable), hasNullableAnn || hasNonNullAnn)
```
kyri-petrou, line 133: Same as above

kyri-petrou: Same as above

ghostdogpr:
I disabled Mima on the main branch so if you rebase that will solve that issue.

kyri-petrou (commented):
@XiNiHa could you please add a test for the use-case that I previously posted? I think the bug is still there at the moment

core/src/main/scala-2/caliban/schema/SchemaDerivation.scala
XiNiHa: I don't think the code was wrong, but I agree that was somewhat confusing to read. I've updated the logic to represent it better what I've meant.

kyri-petrou (approved):
Thanks for putting up with all my requests and delays in reviews! Happy to have this merged once CI is passing

XiNiHa:
Looks like CI is finally passing :tada:

core/src/main/scala/caliban/schema/Schema.scala
ghostdogpr: Yeah, shouldn't this one be false?

core/src/main/scala/caliban/schema/Schema.scala
ghostdogpr: Second this

core/src/main/scala/caliban/schema/Schema.scala
ghostdogpr: Ah, saw your comment below

ghostdogpr (approved):

ghostdogpr:
Thanks for adding this! Would you be able to submit a PR for adding a few docs about it?
Maybe a new section in https://github.com/ghostdogpr/caliban/blob/series/2.x/vuepress/docs/docs/schema.md

PR #2181, opened 2024-04-07
Improve jsoniter encoding performance

kyri-petrou:
TIL that the number of cases in when pattern-matching affects performance in an interesting way:

1. When the number of cases is relatively low (~less than 10), the cost of pattern-matching is relatively constant.
2. When using type-checking instead of unapply methods, we can increase the number of cases a bit while keeping the cost constant
3. When the number of cases increases above a threshold (which is not always constant!), then the cost of pattern-matching increases by a fair bit

I'm not sure exactly why this happens, but my guess is that the JVM manages to JIT/inline the pattern matching when the number of cases is relatively low.

Unfortunately, the only way to keep the number of cases below this threshold (12 type-checked cases in our case) was to remove the type-checking on StreamValue and use the `.toString` method. This should work OK as long as we don't need to extend ResponseValue with another kind of value, and forget to add it to the pattern matching. Given that the chances of adding another kind of ResponseValue type are very very small, I think this approach should be okay

With these changes, we see ~20% improvement in throughput when encoding `ResponseValue`s 🎉 🚀

PS: I also noticed that we were parsing floats into a `BigDecimalNumber` for all floats, so I changed it to first try and parse them into a `Double` which is more memory efficient, and fallback to `BigDecimalNumber` only in case of an error

### DIFF
--- a/core/src/main/scala/caliban/interop/jsoniter/jsoniter.scala
+++ b/core/src/main/scala/caliban/interop/jsoniter/jsoniter.scala
@@ -80,44 +80,31 @@ private[caliban] object ValueJsoniter {
     out.writeObjectEnd()
   }
 
-  private def encodeResponseValue(x: ResponseValue, out: JsonWriter, depth: Int): Unit = {
+  /**
+   * Note on performance:
+   *
+   * We keep the number of cases low (less than 12) for the most commonly used types and avoid using unapply methods in the pattern-matching.
+   * To keep the number of cases below this value, we handle the `FloatValue` type separately.
+   */
+  private def encodeResponseValue(x: ResponseValue, out: JsonWriter, depth: Int): Unit = x match {
+    case v: StringValue               => out.writeVal(v.value)
+    case v: BooleanValue              => out.writeVal(v.value)
+    case v: IntValue.IntNumber        => out.writeVal(v.value)
+    case v: IntValue.LongNumber       => out.writeVal(v.value)
+    case v: ResponseValue.ObjectValue => writeResponseObject(v.fields, out, decrDepth(depth, out))
+    case v: ResponseValue.ListValue   => writeResponseArray(v.values, out, decrDepth(depth, out))
+    case NullValue                    => out.writeNull()
+    case v: EnumValue                 => out.writeVal(v.value)
+    case v: FloatValue.DoubleNumber   => out.writeVal(v.value)
+    case v: IntValue.BigIntNumber     => out.writeVal(v.value)
+    case v: FloatValue                => encodeFloatValue(v, out)
+    case v: ResponseValue.StreamValue => out.writeVal(v.toString)
+  }
 
-    /**
-     * Note on performance:
-     *
-     * We keep the number of cases low (less than 12) for the most commonly used types and avoid using unapply methods in the pattern-matching.
-     * This way we can better utilize JVM inlining, while still preserving match safety via the compiler if we ever extend ResponseValue with more cases.
-     *
-     * If our value is not one of the common ones, we fallback to the full encoder.
-     */
-    def fullEncoder(): Unit = x match {
-      case v: StringValue                 => out.writeVal(v.value)
-      case v: BooleanValue                => out.writeVal(v.value)
-      case v: IntValue.IntNumber          => out.writeVal(v.value)
-      case v: IntValue.LongNumber         => out.writeVal(v.value)
-      case v: IntValue.BigIntNumber       => out.writeVal(v.value)
-      case v: FloatValue.FloatNumber      => out.writeVal(v.value)
-      case v: FloatValue.DoubleNumber     => out.writeVal(v.value)
-      case v: FloatValue.BigDecimalNumber => out.writeVal(v.value)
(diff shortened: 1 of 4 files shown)


core/src/main/scala/caliban/interop/jsoniter/jsoniter.scala
```diff
@@ -5,6 +5,7 @@ import caliban._
 import caliban.parsing.adt.LocationInfo
 import com.github.plokhotnyuk.jsoniter_scala.core._
 
+import scala.annotation.nowarn
```
ghostdogpr: What makes a warning?

core/src/main/scala/caliban/interop/jsoniter/jsoniter.scala
```diff
@@ -78,20 +80,28 @@ private[caliban] object ValueJsoniter {
...
+    case v: EnumValue                   => out.writeVal(v.value)
+    case v: ResponseValue.ObjectValue   => writeResponseObject(v.fields, out, decrDepth(depth, out))
+    case v: ResponseValue.ListValue     => writeResponseArray(v.values, out, decrDepth(depth, out))
+    // Not very common, so we leave them last
+    case v: IntValue.LongNumber         => out.writeVal(v.value)
+    case v: IntValue.BigIntNumber       => out.writeVal(v.value)
+    case v: FloatValue.FloatNumber      => out.writeVal(v.value)
+    case v: FloatValue.DoubleNumber     => out.writeVal(v.value)
+    case v: FloatValue.BigDecimalNumber => out.writeVal(v.value)
+    case v                              => out.writeVal(v.toString)
```
ghostdogpr: This is going to bite us some day 🥲 
Couldn't we group the remaining `IntValue` (and `FloatValue`) into one case and do a second pattern matching?

core/src/main/scala/caliban/interop/jsoniter/jsoniter.scala
kyri-petrou: Oops, dev leftover

core/src/main/scala/caliban/interop/jsoniter/jsoniter.scala
kyri-petrou: I was thinking the same thing, but then I thought it's very unlikely we'll add more ResponseValue types. But the more I think of it, I think you're right, better be safe than sorry.

kyri-petrou: I tried a couple of different ways in order to fully preserve type-safety, and the only feasible way to do it was to have a "full encoder" to fallback to. Encoding non-common types won't be too great performance-wise, but at least we get really good performance when for common types, and we get to preserve type-safety

ghostdogpr: What's the issue with 
```scala
case v: IntValue    => some method that only pattern match on IntValue
case v: FloatValue  => some method that only pattern match on FloatValue
```
at the end and no `case _ =>`?
We can still keep `IntNumber` at the top.

core/src/main/scala/caliban/interop/jsoniter/jsoniter.scala
```diff
@@ -78,20 +80,44 @@ private[caliban] object ValueJsoniter {
...
+     * If our value is not one of the common ones, we fallback to the full encoder.
+     */
+    def fullEncoder(): Unit = x match {
+      case v: StringValue                 => out.writeVal(v.value)
+      case v: BooleanValue                => out.writeVal(v.value)
+      case v: IntValue.IntNumber          => out.writeVal(v.value)
+      case v: IntValue.LongNumber         => out.writeVal(v.value)
+      case v: IntValue.BigIntNumber       => out.writeVal(v.value)
+      case v: FloatValue.FloatNumber      => out.writeVal(v.value)
+      case v: FloatValue.DoubleNumber     => out.writeVal(v.value)
```
paulpdaniels: Shouldn’t we put the cases that are _not_ included in the fast encoder first? Or does it not make a difference to the compiler?

paulpdaniels (approved):

core/src/main/scala/caliban/interop/jsoniter/jsoniter.scala
kyri-petrou: So the issue with `IntValue` is that both `IntNumber` and `LongNumber` are commonly used (`LongNumber` is used a fair bit in some wrappers like `ApolloTracing`), so handling it generically doesn't make sense cause the only case left is `BigIntNumber`.

I ended up handling only `FloatValue` generically, while leaving `DoubleNumber` in the main pattern match cause I think that's something that's somewhat commonly used. Running the benchmarks shows that we're still OK with the number of cases that we have this way

core/src/main/scala/caliban/interop/jsoniter/jsoniter.scala
kyri-petrou: I ended up only handling `FloatValue` in a separate method, seems that the number of cases is still low enough to get the ~20% improvement

ghostdogpr (approved):

PR #2182, opened 2024-04-07
Avoid unnecessary `.unit`s in Validator

kyri-petrou:
In many places in the Validator, we use `.unit` after `ZPure.when` to make the compiler happy. However, this adds an unnecessary flatMap each time we do this.

To avoid this, we add a `failWhen` method which unlike `ZPure.when`, it returns an `EReader[Any, ValidationError, Unit]`, so we don't need to call `.unit` afterwards

### DIFF
--- a/core/src/main/scala/caliban/validation/Validator.scala
+++ b/core/src/main/scala/caliban/validation/Validator.scala
@@ -931,9 +931,11 @@ object Validator {
         val interfaceFieldNames = supertype.map(fieldNames).toSet.flatten
         val isMissingFields     = objectFieldNames.union(interfaceFieldNames) != objectFieldNames
 
-        lazy val missingFields = interfaceFieldNames.diff(objectFieldNames).toList.sorted
         failWhen(interfaceFieldNames.nonEmpty && isMissingFields)(
-          s"$objectContext is missing field(s): ${missingFields.mkString(", ")}",
+          {
+            val missingFields = interfaceFieldNames.diff(objectFieldNames).toList.sorted
+            s"$objectContext is missing field(s): ${missingFields.mkString(", ")}"
+          },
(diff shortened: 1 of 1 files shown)


core/src/main/scala/caliban/validation/Validator.scala
```diff
@@ -1265,7 +1228,7 @@ object Validator {
   )(f: A => EReader[R, ValidationError, B]): EReader[R, ValidationError, Unit] =
     in.sizeCompare(1) match {
       case -1 => zunit
-      case 0  => f(in.head).unit
+      case 0  => f(in.head).asInstanceOf[EReader[Any, ValidationError, Unit]]
```
ghostdogpr: Could we change f to `f: A => EReader[R, ValidationError, Unit]` to avoid these shenanigans? I think with the usage of `failWhen`, almost all (if not all) return `Unit` now

kyri-petrou: Ahh yeah good point. I had to change a usage of `ZPure.whenCase` but it works

core/src/main/scala/caliban/validation/Validator.scala
```diff
@@ -1262,22 +1225,27 @@ object Validator {
    */
   private def validateAll[R, A, B](
     in: Iterable[A]
-  )(f: A => EReader[R, ValidationError, B]): EReader[R, ValidationError, Unit] =
+  )(f: A => EReader[R, ValidationError, Unit]): EReader[R, ValidationError, Unit] =
```
ghostdogpr, line 1228: You can remove the B type param

kyri-petrou: Oops 😅 Done

ghostdogpr (approved):
Love it!

PR #2187, opened 2024-04-12
Allow providing a custom `zio.query.Cache` when executing queries

kyri-petrou:
Initially I wanted to add a param to configure the initial size of the ZQuery cache (e.g `Cache.empty(128)`), but then I realised it might be better to allow providing a custom Cache implementation altogether.

Having said that, I'm slightly conflicted about this change as it could potentially be a huge foot-gun if users were to provide an effect that uses a globally initialised `Cache`. I'm _hoping_ that people won't go on and do that without understanding the potential implications of it.

### DIFF
--- a/core/src/main/scala/caliban/Configurator.scala
+++ b/core/src/main/scala/caliban/Configurator.scala
@@ -26,7 +26,7 @@ object Configurator {
     allowMutationsOverGetRequests: Boolean = false,
     queryExecution: QueryExecution = QueryExecution.Parallel,
     validations: List[QueryValidation] = AllValidations,
-    queryCache: UIO[Cache] = Cache.empty
+    queryCache: UIO[Cache] = Cache.empty(Trace.empty)
   )
 
   private val configRef: FiberRef[ExecutionConfiguration] =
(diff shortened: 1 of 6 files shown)


core/src/main/scala/caliban/Configurator.scala
```diff
@@ -13,20 +16,24 @@ object Configurator {
...
+   *                   Default: The default empty cache implementation from zio-query
    */
   case class ExecutionConfiguration(
     skipValidation: Boolean = false,
     enableIntrospection: Boolean = true,
     allowMutationsOverGetRequests: Boolean = false,
     queryExecution: QueryExecution = QueryExecution.Parallel,
-    validations: List[QueryValidation] = AllValidations
+    validations: List[QueryValidation] = AllValidations,
+    queryCache: UIO[Cache] = Cache.empty
```
paulpdaniels: Should this also have the explicit trace?

paulpdaniels (approved):

core/src/main/scala/caliban/execution/Executor.scala
```diff
@@ -30,13 +30,15 @@ object Executor {
    * @param plan an execution plan
    * @param fieldWrappers a list of field wrappers
    * @param queryExecution a strategy for executing queries in parallel or not
+   * @param makeCache the initial size to use for the ZQuery cache
```
ghostdogpr: the comment doesn't match the parameter

ghostdogpr (approved):

Issue #2185, opened 2024-04-11
Allow overriding Tapir endpoint inputs created by Caliban's HttpInterpreter


olisikh:
Hi,

I want to override the path prefix of Caliban's Tapir endpoints made by `HttpInterpreter`, but it seems there's no way to do it with existing HttpInterpreter implementation.

Ultimately I want to run a `NettyZioServer` providing tapir endpoints directly and control the paths my APIs go into.

Example:
```scala
val api: List[ZServerEndpoint[...]] = ???
val graphql: List[ZServerEndpoint[...]] = ???
val status: List[ZServerEndpoint[...]] = ???

NettyZioServer[R]()
  .addEndpoints(api)     // path /api
  .addEndpoints(graphql) // path /graphql (desired), but currently it's "any path" due to .in(extractFromRequest(identity))
  .addEndpoints(status)  // path /status
  .host(...)
  .port(...)
  .start()
```

I've originally asked this question in Caliban's Discord channel here: https://discord.com/channels/629491597070827530/633200096393166868/1227766362714275922

Pierre answered that we could implement some intermediate case class within HttpInterpreter to allow overriding endpoint inputs:
https://discord.com/channels/629491597070827530/633200096393166868/1227772627758612550

I'm happy to do this myself once I get time (probably over the weekend), I initially thought of feeding in the user's EndpointInput replacing the `.in(extractFromRequest(identitiy))` to allow users to decide the rest of the inputs they want.

Please let me know if you otherwise have any ideas or suggestions. 

Appreciate your time 🙏

PR #2188, opened 2024-04-13
Allow prepending input path to HttpInterpreter generated Tapir endpoints

olisikh:
Hi,

I did not manage to implement full endpoints mapping because mapping over endpoints would depend on `Streams` and I can't figure out the type parameters 😞, since streams instance is provided only when `.serverEndpoints` is called (last step in the builder).

But I managed to expose Tapir's `.prependIn` endpoint combinator to add an extra `EndpointInput[Unit]` for the `PublicEndpoint` that `HttpInterpreter` makes.

It does the job, I can use `HttpInterpreter` like this now (tested with real project):

```scala
val status = ...
val api = ...
val graphql = HttpInterpreter(gqlInterpreter)
  .prependInput("graphql" / "api")
  .intercept[Any](myInterceptor)
  .serverEndpoints[Env, ZioStreams](ZioStreams)
  
 NettyZioSever[Env]()
   .addEndpoints(status)
   .addEndpoints(api)
   .addEndpoints(swagger)
   ...
   .start()
```

Any help / feedback is very welcome 🙏 

Issue #2185

Oleksii

### DIFF
--- a/interop/tapir/src/main/scala/caliban/interop/tapir/HttpInterpreter.scala
+++ b/interop/tapir/src/main/scala/caliban/interop/tapir/HttpInterpreter.scala
@@ -8,6 +8,7 @@ import sttp.tapir.Codec.JsonCodec
 import sttp.tapir.model.ServerRequest
 import sttp.tapir._
 import zio._
+import sttp.tapir.EndpointInput.FixedPath
 
 sealed trait HttpInterpreter[-R, E] { self =>
   protected def endpoints[S](streams: Streams[S]): List[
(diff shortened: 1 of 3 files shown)


ghostdogpr:
I gave it a try and I see what you mean about the streams type. This solution is a little bit specific but the implementation is short and simple so it seems ok. @kyri-petrou @paulpdaniels any better idea or shall we just ship this?

ghostdogpr (approved):

paulpdaniels:
Is there anything we can do with this other than append path prefixes? Perhaps it would be better in that case to just expose a `prependPath(path: String)` then?

olisikh:
Hi @paulpdaniels,

I think it has to be `EndpointInput[Unit]`.
 
The path with multiple segments and without output into `serverLogic` is encoded as `EndpointInput[Unit]` in Tapir, if we were to make it a String, it would make 1-segment paths and that is not very useful.

To my knowledge this is how it works:
```
.in("hello")             // makes it a /hello
.in("hello" / "world")   // makes it a /hello/world
.in("hello/world")       // makes it a hello%2Fworld (Tapir would encode slashes within a String)
```


PS I can think of making the input as `def prependPath(pathPrefix: List[String])` instead and then combine the list elements with `/` to make it `EndpointInput[Unit]`

Example:
```scala
HttpInterpreter(graphql)
  .prependPath("graphql" :: "api" :: Nil)
  ...
  .serverEndpoints[Env, ZioStreams](ZioStreams)
```

olisikh:
Narrowed it down to just `.prependPath(path: List[String])`, and I also like it more. Please let me know if it looks good for you and feel free to change anything the way you like 🙏 

PS I tried to make input as `EndpointInput.FixedPath[Unit]` but at the call-site compiler can't infer that `"hello" / "world"` is a `FixedPath[Unit]` 🤷‍♂️

paulpdaniels (approved):

ghostdogpr:
How about the WebSocketInterpreter? Can it be added there too?

ghostdogpr:
Thanks!

PR #2211, opened 2024-05-05
Add Scala 3-specific method for building custom schemas

kyri-petrou:
Improves UX of building custom schemas significantly for Scala 3:

Scala 2:
```scala
case class Author(id: String, firstName: String, lastName: String)

implicit val authorSchema: Schema[Any, Author] = Schema.obj("Author"){ implicit attrs =>
  List(
    field("id")(_.id),
    field("fullName")(_ => "foo"),
  )
}
```

Scala 3:
```scala
case class Author(id: String, firstName: String, lastName: String)

given Schema[Any, Author] = Schema.custom("Author")(
  field("id")(_.id),
  field("fullName")(_ => "foo"),
)
```

Not too sure on the name though, `custom` implies that it can be used to build other kinds of schemas, but `customObject` feels a bit too verbose. Thoughts @ghostdogpr?

### DIFF
--- a/core/src/main/scala/caliban/schema/Schema.scala
+++ b/core/src/main/scala/caliban/schema/Schema.scala
@@ -142,7 +142,7 @@ trait Schema[-R, T] { self =>
   }
 }
 
-object Schema extends GenericSchema[Any]
+object Schema extends GenericSchema[Any] with SchemaVersionSpecific
 
 trait GenericSchema[R] extends SchemaDerivation[R] with TemporalSchema {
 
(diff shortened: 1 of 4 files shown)


ghostdogpr:
Since the original was obj, how about customObj?

kyri-petrou:
> Since the original was obj, how about customObj?

Done 👍

core/src/main/scala/caliban/schema/Schema.scala
```diff
@@ -311,6 +311,8 @@ trait GenericSchema[R] extends SchemaDerivation[R] with TemporalSchema {
    *  )
    *
    * }}}
+   *
+   * @see `custom` for an improved variant using context functions (Scala 3 only)
```
ghostdogpr: Rename here too

ghostdogpr (approved):

PR #2212, opened 2024-05-06
Update sbt to 1.10.0

kyri-petrou:
Cleaned up the config file a bit as well.

Also, do we need the `sbt` script in the repo or is that some old-time artifact? I deleted it but can put it back if it's needed


ghostdogpr (approved):
Ah that change is nice, no need to update the config.yml every time!

PR #2215, opened 2024-05-07
Add derivation of `Schema` for union types (towards #1926)

oyvindberg:
Finally found time to implement a first iteration of this this. 
A patch for the codegen will probably be forthcoming, but this at least allows us to use union types with a bit more ease than before.

Usage looks like this:
```scala
final case class Foo(value: String) derives Schema.SemiAuto
final case class Bar(foo: Int) derives Schema.SemiAuto
type Payload = Foo | Bar
given Schema[Any, Payload] = Schema.unionType[Payload]
```

This will produce a schema like this:
```graphql
schema {
  query: Query
}

union Payload = Foo | Bar

type Bar {
  foo: Int!
}

type Foo {
  value: String!
}

type Query {
  testQuery(isFoo: Boolean!): Payload!
}

```

### Limitations:
- in order to capture the name of the type alias, users will need to explicitly provide a type name to `Schema.unionType`. The compile will fail otherwise.
- ~~No annotations are picked up yet. It probably works the same as for the other cases, but I didn't have time to implement it now.~~

### DIFF
--- a/core/src/main/scala-3/caliban/schema/SchemaDerivation.scala
+++ b/core/src/main/scala-3/caliban/schema/SchemaDerivation.scala
@@ -140,6 +140,8 @@ trait SchemaDerivation[R] extends CommonSchemaDerivation {
 
   inline def genDebug[R, A]: Schema[R, A] = PrintDerived(derived[R, A])
 
+  inline def unionType[T]: Schema[R, T] = ${ TypeUnionDerivation.typeUnionSchema[R, T] }
+
   final lazy val auto = new AutoSchemaDerivation[Any] {}
 
   final class SemiAuto[A](impl: Schema[R, A]) extends Schema[R, A] {
--- a/core/src/main/scala-3/caliban/schema/TypeUnionDerivation.scala
+++ b/core/src/main/scala-3/caliban/schema/TypeUnionDerivation.scala
@@ -1,0 +1,76 @@
+package caliban.schema
+
+import caliban.introspection.adt.__Type
+
+import scala.quoted.*
+
+object TypeUnionDerivation {
+  inline def derived[R, T]: Schema[R, T] = ${ typeUnionSchema[R, T] }
+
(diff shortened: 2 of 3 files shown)


kyri-petrou:
Awesome! Love it! Do you know whether it also works with opaque types?

Also, totally out of the scope of this PR, but would a similar approach also work for using intersection types for interfaces?

oyvindberg:
> Awesome! Love it! Do you know whether it also works with opaque types?

My understanding is that opaque types really just need to fetch an instance of the underlying type and expose it in its companion object where it's known that the two types are the same. I haven't tried it yet, but there is a chance it could be easy. 

> Also, totally out of the scope of this PR, but would a similar approach also work for using intersection types for interfaces?

Interesting idea. This code here follows `Or` AST nodes, I suppose a variant of that code could follow `And`-nodes or similar instead and do something productive with it.

I rebased the branch again now, so hopefully it's good to go.

ghostdogpr:
> - No annotations are picked up yet. It probably works the same as for the other cases, but I didn't have time to implement it now.

Is this still missing?

ghostdogpr:
For annotations it would just be calls to `getName`, `getDescription` and `getDirectives` in `DerivationUtils`, similar to https://github.com/ghostdogpr/caliban/blob/63d3e1d15399e81d49c87bd0164c68ecb05c86aa/core/src/main/scala-3/caliban/schema/SumSchema.scala#L40

Do you think you'll be able to add them? If not, we can probably merge it and someone else opens an other PR.

ghostdogpr:
Btw I checked that the magnolia macro we use to get annotations works on type aliases: https://scastie.scala-lang.org/rI5LUKi5TrKCCbZ3tkrZNQ

oyvindberg:
Thanks for the pointers. Added annotation handling now 👍

ghostdogpr (approved):
Thanks!

guizmaii:
Thanks guys for your work! Was really helpful to implement: https://github.com/zio/zio/pull/9723 😘

PR #2218, opened 2024-05-08
Schema transformations

### LINKED ISSUE #130: Support for schema transformations

ghostdogpr:
Closes https://github.com/ghostdogpr/caliban/issues/130

Adds the ability to transform types of an existing schema:
- rename types, fields and arguments
- filter interfaces, fields and arguments
- custom transformations

Example usage:
```scala
api
  .transform(Transformer.RenameType("InnerObject" -> "Renamed"))
  .transform(Transformer.ExcludeField("Query" -> "b"))
```


kyri-petrou:
As a side note, I think we should either document for users to avoid this pattern, or we should optimize for it. With a small number of cases it won't make a big difference, but it'll add up if they start doing this with a lot of them.

```scala
api
  .transform(Transformer.RenameType("Foo" -> "Bar"))
  .transform(Transformer.RenameType("Bar" -> "Baz"))
  .transform(Transformer.RenameType("Bla" -> "Bra"))
```

EDIT: For completeness, this is the most performant implementation:

```scala
api
  .transform(Transformer.RenameType(
    "Foo" -> "Bar",
    "Bar" -> "Baz",
    "Bla" -> "Bra",
))
```

kyri-petrou (approved):
Love it ❤️

I have a bunch of places I can use this at $WORK so can start using it once we merge in and report any potential issues.

I think given that this is a really great feature, we should just mark the next release as 2.7.0 and disable MiMa. 2.6.0 doesn't seem to have any reported issues so far so I'm hopeful we won't need to do any patch releases soon

ghostdogpr:
The PR about nullability also breaks mima so we can ship these 2 together. Let's disable mima as you said. 

Let me give it a final review before merging 🙏

PR #2219, opened 2024-05-09
Allow excluding fields on input objects via transformers

kyri-petrou:
~~Gave it a go with being able to exclude nested arguments. Added and existing tests seem to be passing.~~

~~@ghostdogpr what do you think of this approach? The one downside is that we're not doing the same for execution; validation should take care of it. It's only a problem in cases that validation isn't enabled~~

EDIT: Went with a different approach afterall. Instead of excluding nested arguments (which is really tricky because you have to make sure to exclude them from everywhere that the input type is used), instead I added a new transformer that allows excluding fields from input objects. The code is much cleaner and easier to reason with


kyri-petrou:
Changing to draft as I want to experiment with another approach as well

ghostdogpr (approved):

PR #2233, opened 2024-05-14
Add SnapshotTest

oyvindberg:
This reads `CI` environment variable (same as sbt) to determine if we're running in CI.

This variable determines which behaviour the tests have
- If we're running in CI it checks that the contents of all the snapshot files are up to date
- Otherwise, it'll update the files with new contents.

The point is that a given change will showcase the effective results right in the PR, and otherwise be convenient to work with

Extracted from #2216


oyvindberg:
a cool thing here for later would be to verify that the code compiles as well

ghostdogpr:
> a cool thing here for later would be to verify that the code compiles as well

There is the `typeCheck` assertion in ZIO Test.

ghostdogpr:
> There is the `typeCheck` assertion in ZIO Test.

Ah never mind, I think it's for strings known at compile time only.

PR #2234, opened 2024-05-15
add subgraph compatibility check

### LINKED ISSUE #1730: Setup Apollo Federation spec compliance in CI
See https://www.apollographql.com/blog/backend/testing/new-apollo-federation-spec-compliance-tools-for-graphql-server-maintainers/

Let's run it on branches that modify the code in the federation submodule.

paulpdaniels:
Adding the subgraph compatibility check that should allow PRs to be check for breaking changes to federation.

Resolves #1730


apollo-compatibility/Dockerfile
```diff
@@ -0,0 +1,12 @@
+FROM hseeberger/scala-sbt:17.0.2_1.9.9_3.1.1 AS build
```
kyri-petrou: Stupid question, but any reason that we can't use the same one used by CI? I believe it has `sbt` and all that installed as well

.github/workflows/compatibility.yaml
kyri-petrou, line 1: Shouldn't this workflow be added to the circleci config instead? Or do we need to use github actions?

kyri-petrou:
@paulpdaniels disregard my review, only just realised the PR is still in draft 🤦

apollo-compatibility/Dockerfile
paulpdaniels: TBH I just copied this almost wholesale from https://github.com/sangria-graphql/sangria-federated/pull/287

.github/workflows/compatibility.yaml
paulpdaniels: Not sure, step one: get this working, step two figure out if it can run in circle CI. They have a convenient action so that seems the most accessible path forward.

kyri-petrou: Ohh I totally missed that part 🤦🏻‍♂️ Yeah I don't know if it's even worth to try migrating it to CircleCI

ghostdogpr:
You can mention it resolves https://github.com/ghostdogpr/caliban/issues/1730

docs/docs/optimization.html
```diff
@@ -37,7 +37,9 @@
...
...ery optimization</h1> <p>A GraphQL query may request multiple fields that are using the same resolver. It's not a problem if the resolver is a simple value, but it can be less than optimal when the resolver runs an effect (such as reading from a database).</p> <p>We might want to:</p> <ul><li><strong>cache</strong> identical queries (deduplication)</li> <li><strong>batch</strong> queries to the same source</li></ul> <p>This is possible in Caliban using the <a href="https://github.com/zio/zquery" target="_blank" rel="noopener noreferrer"><code>ZQuery</code><span><svg xmlns="http://www.w3.org/2000/svg" aria-hidden="true" focusable="false" x="0px" y="0px" viewBox="0 0 100 100" width="15" height="15" class="icon outbound"><path fill="currentColor" d="M18.8,85.1h56l0,0c2.2,0,4-1.8,4-4v-32h-8v28h-48v-48h28v-8h-32l0,0c-2.2,0-4,1.8-4,4v56C14.8,83.3,16.6,85.1,18.8,85.1z"></path> <polygon fill="currentColor" points="45.7,48.7 51.3,54.3 77.2,28.5 77.2,37.2 85.2,37.2 85.2,14.9 62.8,14.9 62.8,22.9 71.5,22.9"></polygon></svg> <span class="sr-only">(opens new window)</span></span></a> data type.</p> <p>Additionally, one may want to perform optimizations based on the fields selected by the client.
```
ghostdogpr: This file isn't meant to be modified

paulpdaniels: Ah an Intellij "smart" refactoring did this

paulpdaniels:
@ghostdogpr @kyri-petrou ready for re-review

build.sbt
```diff
@@ -140,7 +141,8 @@ lazy val rootJVM3 = project
     ideSkipProject     := true
   )
   .aggregate({
-    val excluded: Set[ProjectReference] = Set(clientJS, clientNative, clientLaminext, codegenSbt, akkaHttp)
+    val excluded: Set[ProjectReference] =
```
kyri-petrou, line 144: I'm hoping one day we'll have Scala 3 as the base version for the repo. 

If it's something that's quick to add, could we also have this running for Scala 3 just to reduce migration efforts later on? If it takes more than 10 mins to setup then no worries we can do it at a later time

kyri-petrou (approved):

build.sbt
```diff
@@ -149,6 +151,7 @@ lazy val macros = project
   .settings(name := "caliban-macros")
   .settings(commonSettings)
   .settings(enableMimaSettingsJVM)
+  .disablePlugins(AssemblyPlugin)
```
ghostdogpr, line 154: I wonder if there is a way to disable it globally and enable it only where we need instead.

apollo-compatibility/README.md
```diff
@@ -0,0 +1,18 @@
+# federated subgraph to test apollo federation spec compatibility
```
ghostdogpr:
```suggestion
# Federated subgraph to test apollo federation spec compatibility
```

apollo-compatibility/src/main/scala/Main.scala
```diff
@@ -0,0 +1,35 @@
+import caliban.CalibanError
```
ghostdogpr, line 1: Can you sort imports?

ghostdogpr (commented):
Pretty cool! Not only can we use this for compatibility check, but it's also a good demo project for using Federation with Caliban. I think we should even add it to the docs 😄

build.sbt
paulpdaniels: I couldn't find anything related to that since it's an auto plugin. The only idea I had was to wrap it in a method:

`def standardProject = project.disablePlugins(AssemblyPlugin)`

build.sbt
paulpdaniels: I'll try to enable it, it's a somewhat unique module because it's getting wrapped into a fat jar so I need to make sure that it doesn't get confused about how to do that. I was actually thinking this would have been good practice for writing something in scala3 so I might go back through in a later PR and update the source too.

paulpdaniels: Looks like enabling it works

apollo-compatibility/src/main/scala/Main.scala
paulpdaniels: I forgot to push the commit with this change, I'll do another with just the import formatting

PR #2241, opened 2024-05-20
Use auto-blocking executor for zio-http tests

kyri-petrou:
Attempt no2 at fixing flaky tests. Will re-trigger CI a few times, hopefully issue goes away


kyri-petrou:
Closing this one as it seems the issue is with zio-http and not with blocking code (see https://github.com/zio/zio-http/pull/2852)

PR #2246, opened 2024-05-23
Parser microoptimizations

kyri-petrou:
I was implementing a parser using fastparse today at $WORK and found out that `CharsWhileIn("xx", minChars)` is _much_ more efficient than using `.repX`. With these changes we get ~5-10% performance increase in our ParserBenchmark.

Note that I also changed the `fullIntrospectionQuery` val to strip the margins of the query. I realised that we were benchmarking too heavily the whitespace parser

### DIFF
--- a/benchmarks/src/main/scala/caliban/ParserBenchmark.scala
+++ b/benchmarks/src/main/scala/caliban/ParserBenchmark.scala
@@ -1,22 +1,18 @@
 package caliban
 
 import caliban.parsing.Parser
-import cats.effect.IO
-import io.circe.Json
 import org.openjdk.jmh.annotations._
-import sangria.execution._
-import sangria.marshalling.circe._
+import org.openjdk.jmh.infra.Blackhole
 import sangria.parser.QueryParser
 
 import java.util.concurrent.TimeUnit
-import scala.concurrent.duration._
-import scala.concurrent.{ Await, ExecutionContextExecutor, Future }
+import scala.concurrent.ExecutionContextExecutor
 
 @State(Scope.Thread)
 @BenchmarkMode(Array(Mode.Throughput))
 @OutputTimeUnit(TimeUnit.SECONDS)
-@Warmup(iterations = 5, time = 3, timeUnit = TimeUnit.SECONDS)
-@Measurement(iterations = 5, time = 3, timeUnit = TimeUnit.SECONDS)
+@Warmup(iterations = 5, time = 1, timeUnit = TimeUnit.SECONDS)
+@Measurement(iterations = 5, time = 1, timeUnit = TimeUnit.SECONDS)
 @Fork(1)
 class ParserBenchmark {
   import ComplexQueryBenchmark._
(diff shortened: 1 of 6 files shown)


core/src/main/scala/caliban/parsing/parsers/Parsers.scala
```diff
@@ -348,7 +348,7 @@ object Parsers extends SelectionParsers {
     schemaExtension | typeExtension
 
   def definition(implicit ev: P[Any]): P[Definition] =
-    executableDefinition | typeSystemDefinition | typeSystemExtension
+    typeSystemDefinition | typeSystemExtension
```
ghostdogpr: It doesn't match the spec `definition` (keeping the code close to the spec is good for maintainability), how about changing `document` instead?

kyri-petrou: I reverted the change. _tbh_ it only makes a difference for invalid queries really

benchmarks/src/main/scala/caliban/ParserBenchmark.scala
```diff
@@ -1,51 +1,45 @@
...
   import ComplexQueryBenchmark._
 
   implicit val executionContext: ExecutionContextExecutor = scala.concurrent.ExecutionContext.global
 
   @Benchmark
-  def runCaliban(): Unit = {
-    val io = Parser.parseQuery(fullIntrospectionQuery)
-    Caliban.run(io)
+  def runCaliban(bh: Blackhole): Unit = {
+    bh.consume(Parser.parseQueryEither(fullIntrospectionQuery).fold(throw _, identity))
```
plokhotnyuk: No need to use an extra blackhole parameter, just returning of results will be enough to use more efficient compiler blackholes, see: https://shipilev.net/jvm/anatomy-quarks/27-compiler-blackholes/

kyri-petrou: Ah nice, thanks for this! I'll change it

core/src/main/scala/caliban/parsing/parsers/Parsers.scala
ghostdogpr: But how about changing `document` to `(Start ~ definition.rep ~ End)`? I agree with the fix, just wanted to do it somewhere else 😄

kyri-petrou: `document` is currently this:

```scala
((Start ~ executableDefinition.rep ~ End) | (Start ~ definition.rep ~ End)).map(seq => ParsedDocument(seq.toList))
```

The reason for having `Start ~ executableDefinition.rep ~ End` separately is so that when when we reach the end of the file after `executableDefinition` has successfully parsed (which is the most common thing we parse), we shortcut to the end and don't try `typeSystemDefinition` prior to exiting. I'm not sure if it's even possible, but in the case that there is a typeSystem definition in the query then we'll fallback to the full parser.

I actually realised that the "fix" I did earlier wouldn't work for cases that there is a mix of them; although I don't know if that's a valid query 🤔. If yes I should add a test for it

ghostdogpr: Ah, didn't know we were trying to be clever 🤣

ghostdogpr (approved):

PR #2247, opened 2024-05-23
Fix IDE settings for apollo-compatibility module

kyri-petrou:
I realised that code highlighting wasn't working for the `apollo-compatibility` module; seems that it was accidentally excluded for Scala 2.13 and 3


paulpdaniels:
Damn negatives

paulpdaniels (approved):

PR #2259, opened 2024-06-01
Fill in error path and location when a field wrapper fails

ghostdogpr:
I realized we were not filling error path and location when a failure happened in a field wrapper.

### DIFF
--- a/core/src/main/scala/caliban/execution/Executor.scala
+++ b/core/src/main/scala/caliban/execution/Executor.scala
@@ -386,20 +386,20 @@ object Executor {
             case wrapper :: tail =>
               val q =
                 if (isPure && !wrapper.wrapPureValues) query
-                else
-                  wrapper
-                    .wrap(query, fieldInfo)
-                    .mapErrorCause(e =>
-                      effectfulExecutionError(
-                        PathValue.Key(fieldInfo.name) :: fieldInfo.path,
-                        Some(fieldInfo.details.locationInfo),
-                        e
-                      )
-                    )
-
+                else wrapper.wrap(query, fieldInfo)
               loop(q, tail)
           }
-        loop(query, fieldWrappers)
+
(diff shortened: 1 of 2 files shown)


core/src/main/scala/caliban/execution/Executor.scala
```diff
@@ -394,7 +384,19 @@ object Executor {
...
+                  wrapper
+                    .wrap(query, fieldInfo)
+                    .mapErrorCause(e =>
+                      effectfulExecutionError(
+                        PathValue.Key(fieldInfo.name) :: fieldInfo.path,
+                        Some(fieldInfo.details.locationInfo),
+                        e
+                      )
+                    )
+
```
kyri-petrou: I think we should re-write the entire `wrap` method to something like this, to avoid applying the error handler on every field wrapper:

```scala
      def wrap(query: ExecutionQuery[ResponseValue], isPure: Boolean, fieldInfo: FieldInfo) = {
        @tailrec
        def loop(query: ExecutionQuery[ResponseValue], wrappers: List[FieldWrapper[R]]): ExecutionQuery[ResponseValue] =
          wrappers match {
            case Nil             => query
            case wrapper :: tail =>
              val q =
                if (isPure && !wrapper.wrapPureValues) query
                else wrapper.wrap(query, fieldInfo)
              loop(q, tail)
          }

        if ((isPure && !wrapPureValues) || (fieldWrappers eq Nil)) query
        else {
          loop(query, fieldWrappers).mapErrorCause(e =>
            effectfulExecutionError(
              PathValue.Key(fieldInfo.name) :: fieldInfo.path,
              Some(fieldInfo.details.locationInfo),
              e
            )
          )
        }
      }
```

What do you think?

ghostdogpr: Good point!

PR #2261, opened 2024-06-01
Delete deprecated rendering class

paulpdaniels:
This class has been deprecated since 2.3.1, I think we can safely delete it now.


ghostdogpr (approved):

ghostdogpr:
@paulpdaniels ah sorry, I merged it then reverted it because I am going to do a minor release and I think it's better to include this the next major instead. Could you reopen it? 😄

PR #2273, opened 2024-06-08
Add federation 2.8, fix federation v2 introspection

paulpdaniels:
Adds support for the new federation [2.8 directives](https://www.apollographql.com/docs/federation/federated-types/federated-directives#saving-and-referencing-data-with-contexts)

Also fixes an issue where the `_FieldSet` type was getting added to the schema in federation v2


ghostdogpr (commented):
LGTM. Can you check the scala 3 CI failure?

PR #2276, opened 2024-06-09
build.sbt cleanup

guymers:
Remove explicit ZIO test framework as it is built into sbt.

Remove snapshot resolvers to avoid depending on snapshots.

Remove scala-xml v1 workaround as it is no longer required.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -44,7 +44,6 @@ inThisBuild(
     organization             := "com.github.ghostdogpr",
     homepage                 := Some(url("https://github.com/ghostdogpr/caliban")),
     licenses                 := List(License.Apache2),
-    resolvers += "Sonatype OSS Snapshots" at "https://oss.sonatype.org/content/repositories/snapshots",
     Test / parallelExecution := false,
     scmInfo                  := Some(
       ScmInfo(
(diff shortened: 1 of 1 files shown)


paulpdaniels (approved):

ghostdogpr:
I am not sure what happened but both your PRs seem to not have triggered the CI 🤔

kyri-petrou:
I'm a bit conflicted about removing the snapshot repository from `built.sbt`. On one hand I agree with the reasoning, but on the other hand I often test zio / zio-query / zio-http snapshot versions via the caliban repo. Is there any way we can disable it only for CI when merging in main? Maybe via an envvar?

ghostdogpr:
> I'm a bit conflicted about removing the snapshot repository from `built.sbt`. On one hand I agree with the reasoning, but on the other hand I often test zio / zio-query / zio-http snapshot versions via the caliban repo. Is there any way we can disable it only for CI when merging in main? Maybe via an envvar?

How about using [Global Settings](https://www.scala-sbt.org/1.x/docs/Global-Settings.html#Global+Settings)?

guymers:
> I am not sure what happened but both your PRs seem to not have triggered the CI 🤔

Yeah dunno, pushed again and now it seems to be running

kyri-petrou:
> How about using [Global Settings](https://www.scala-sbt.org/1.x/docs/Global-Settings.html#Global+Settings)?

It's a nice workaround but in some cases it's useful to be able to use it for draft PRs as well (e.g., https://github.com/ghostdogpr/caliban/pull/2251). I'm not against removing it though, but if we can have a fairly decent way to keep it for dev/PRs and disable it for the main branch that'd be ideal AFAICT

ghostdogpr:
> > How about using [Global Settings](https://www.scala-sbt.org/1.x/docs/Global-Settings.html#Global+Settings)?
> 
> It's a nice workaround but in some cases it's useful to be able to use it for draft PRs as well (e.g., #2251). I'm not against removing it though, but if we can have a fairly decent way to keep it for dev/PRs and disable it for the main branch that'd be ideal AFAICT

How about leaving it commented out then?

kyri-petrou:
> How about leaving it commented out then?

Yeah I think that should be fine 👍

PR #2279, opened 2024-06-14
Make `allFieldsMap` private on __Type and use an accessor method

kyri-petrou:
In order to avoid mutability leaking, we made the type of the `allFieldsMap` the generic `collection.Map`. The issue with this is that all the methods on the Map are now megamorphic and add a small overhead.

Since we use this map multiple times in the lifetime of a request, and we usually end up getting all the requested fields from it, I think we should try and optimize a bit for it by making it private and using a getter method which returns `null` in cases that it's not found. That way we avoid the megamorphic call + the allocation of `Option` (which is for all fields except `__typename`)

### DIFF
--- a/core/src/main/scala/caliban/transformers/Transformer.scala
+++ b/core/src/main/scala/caliban/transformers/Transformer.scala
@@ -314,7 +314,7 @@ object Transformer {
           step.copy(fields =
             fieldName =>
               if (inner.contains(fieldName)) {
-                val args = field.fieldType.allFieldsMap(fieldName).allArgNames
+                val args = field.fieldType.getFieldOrNull(fieldName).allArgNames
                 mapFunctionStep(fields(fieldName))(_.filterNot { case (argName, _) => !args.contains(argName) })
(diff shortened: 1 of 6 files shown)


core/src/main/scala/caliban/transformers/Transformer.scala
```diff
@@ -314,7 +314,7 @@ object Transformer {
           step.copy(fields =
             fieldName =>
               if (inner.contains(fieldName)) {
-                val args = field.fieldType.allFieldsMap(fieldName).allArgNames
+                val args = field.fieldType.getFieldOrNull(fieldName).allArgNames
```
ghostdogpr: Seems unsafe?

kyri-petrou: AFAICT it shouldn't be possible to get here without the field being present on the `__Type`, but I guess it's better to be safe than sorry. I'll add a check

ghostdogpr (approved):

PR #2282, opened 2024-06-16
Make it easier to use tapir with impure effects

ghostdogpr:
Add `serverEndpointsFuture` and `serverEndpointsIdentity` to `HttpInterpreter` and `HttpUploadInterpreter` so that it's easier to use any tapir backend that uses `Future` or `Identity` (used with ox).

Also use these in Akka, Pekko and Play interpreters instead of duplicating code.

I didn't implement such a thing for the `WebSocketInterpreter` because it's way more complex but I might give it a try later.


ghostdogpr:
Mima complains about methods that I don't think anyone should have used. I think we can ignore them as long we release this in a major release.

interop/tapir/src/main/scala/caliban/interop/tapir/HttpInterpreter.scala
```diff
@@ -31,6 +35,24 @@ sealed trait HttpInterpreter[-R, E] { self =>
...
+  }
+
+  def serverEndpointsIdentity[S](streams: Streams[S])(runtime: Runtime[R])(implicit
+    streamConstructor: StreamConstructor[streams.BinaryStream]
+  ): List[ServerEndpoint[S, Identity]] = {
+    implicit val r: Runtime[R] = runtime
+    serverEndpoints(streams).map(
+      convertHttpEndpointToIdentity[R, streams.BinaryStream, S, (GraphQLRequest, ServerRequest)]
+    )
+  }
```
kyri-petrou, line 54: Should we make these methods package-private so that we don't have to worry about mima in the future? Or do we need them public?

interop/tapir/src/main/scala/caliban/interop/tapir/TapirAdapter.scala
```diff
@@ -199,24 +200,27 @@ object TapirAdapter {
...
-      endpoint.PRINCIPAL,
-      endpoint.INPUT,
-      endpoint.ERROR_OUTPUT,
-      endpoint.OUTPUT,
-      ZioStreams,
-      Future
-    ](
+  private val rightUnit: Right[Nothing, Unit] = Right(())
+
+  def convertHttpEndpointToFuture[R, BS, S, I](
```
kyri-petrou: As above, can we make this package-private?

paulpdaniels (approved):

interop/tapir/src/main/scala/caliban/interop/tapir/HttpInterpreter.scala
ghostdogpr: They need to be public, for anyone who wants to use another adapter than the ones we provide (e.g. ox).

interop/tapir/src/main/scala/caliban/interop/tapir/TapirAdapter.scala
ghostdogpr: Yeah I think you'd always use `serverEndpointsFuture`/`serverEndpointsIdentity` rather than these.

kyri-petrou (approved):

PR #2285, opened 2024-06-17
Update tapir and zio json

kyri-petrou:
There seem to be some new implicits provided by zio-json which are messing up with implicit resolution. Workaround is to use `JsonCodec` as the implicit we provide and let the new implicits of `zio-json` extract the Encoder / Decoder from there


core/src/main/scala/caliban/GraphQLRequest.scala
```diff
@@ -47,6 +45,13 @@ object GraphQLRequest {
   implicit def playJsonWrites[F[_]: IsPlayJsonWrites]: F[GraphQLRequest] =
     caliban.interop.play.json.GraphQLRequestPlayJson.graphQLRequestWrites.asInstanceOf[F[GraphQLRequest]]
 
+  @deprecated("kept for compatibility purposes only", "1.7.2")
```
ghostdogpr, line 48: Maybe next release will be 1.8.0? We have a bunch of breaking changes pending to be merged. Maybe we wait for the `oneof` support?

kyri-petrou: To be honest if the next release is a major we can just delete these methods.

Let's merge it in as-is and I'll create another PR to delete the methods and merge it when we're certain we don't need to create a patch release

ghostdogpr (approved):

PR #2287, opened 2024-06-17
Change `caliban-zio-http` to depend on `caliban-quick`

kyri-petrou:
As discussed on Discord with @ghostdogpr, the `caliban-zio-http` module has been deprecated for some time now but it's still used according to the monthly package downloads.

Since we don't want to be maintaining the zio-http <-> tapir interop and having to deal with any potential version conflicts, I think for the time being we can change `caliban-zio-http` to depend on `caliban-quick` (instead of tapir) and use it under the hood.

For most users, this requires very few API changes to essentially pass the interpreter and config directly to the method rather than using the `HttpInterpreter` / `WebSocketInterpreter`. The only "real" issue is with users that use `RequestInterceptor`s. However, these interceptors are not really necessary for zio-http and can somewhat easily be re-written as `Middleware`.


ghostdogpr (approved):

paulpdaniels (approved):
Let’s just make sure to include something in the release notes about how to migrate from request interceptors

PR #2288, opened 2024-06-17
Disable Netty's leak detection in `QuickAdapter#runServer`

kyri-petrou:
Since we're not creating netty responses directly, I don't think there's any reason for us to have leak detection enabled by default

### DIFF
--- a/adapters/quick/src/main/scala/caliban/QuickAdapter.scala
+++ b/adapters/quick/src/main/scala/caliban/QuickAdapter.scala
@@ -3,6 +3,8 @@ package caliban
 import caliban.Configurator.ExecutionConfiguration
 import zio._
 import zio.http._
+import zio.http.netty.NettyConfig
+import zio.http.netty.NettyConfig.LeakDetectionLevel
 import zio.stacktracer.TracingImplicits.disableAutoTrace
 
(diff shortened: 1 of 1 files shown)


adapters/quick/src/main/scala/caliban/QuickAdapter.scala
```diff
@@ -80,7 +82,11 @@ final class QuickAdapter[R] private (requestHandler: QuickRequestHandler[R]) {
   )(implicit trace: Trace, tag: Tag[R]): RIO[R, Nothing] =
     Server
       .serve[R](routes(apiPath, graphiqlPath = graphiqlPath, uploadPath = uploadPath, webSocketPath = webSocketPath))
-      .provideSomeLayer[R](Server.defaultWithPort(port))
+      .provideSomeLayer[R](
+        ZLayer.succeed(Server.Config.default.port(port))
+          ++ ZLayer.succeed(NettyConfig.default.leakDetection(LeakDetectionLevel.DISABLED))
+          >+> Server.customized
+      )
```
kyri-petrou, line 89: PS: I wanted to use `provideSome` but I got slapped with a different variant of [this issue](https://github.com/zio/zio/issues/8936), due to an ambiguity in the implicit `trace`

ghostdogpr (approved):

PR #2290, opened 2024-06-17
Improve introspection

kyri-petrou:
I noticed that the `IntrospectionDerivation` was using auto-derivation, which is generating a bit more bytecode than necessary and increases compilation time. Unfortunately I was only able to implement semi-auto derivation for Scala 3 as I was getting SO errors with Scala 2 (even with `implicit lazy val` 😕).

In either way, the `core` module compiles in just over 4 seconds on a warm compiler on my M2 macbook now.

While I was at it, I saw some low hanging fruits to optimize / cleanup the Introspector a bit (sorry not sorry for the mutable code 😅)

Also I gave up on adding exclusions to Mima and just disabled it altogether.


core/src/main/scala-3/caliban/introspection/IntrospectionDerivation.scala
```diff
@@ -5,14 +5,24 @@ import caliban.Value.StringValue
...
+  private given Schema[Any, __Directive]      = Schema.derived
+  private given Schema[Any, __DeprecatedArgs] = Schema.derived
 
-  implicit lazy val inputValueSchema: Schema[Any, InputValue]   = Schema.gen
-  implicit lazy val typeSchema: Schema[Any, __Type]             = Schema.gen
-  implicit lazy val __directiveSchema: Schema[Any, __Directive] = Schema.gen
-  val introspectionSchema: Schema[Any, __Introspection]         = Schema.gen
+  // Unions, so we can auto-derive them cheaply
+  private given Schema[Any, __TypeKind]          = Schema.Auto.derived
+  private given Schema[Any, __DirectiveLocation] = Schema.Auto.derived
```
kyri-petrou, line 22: Love that we can do this in Scala 3. It was only by accident that I discovered it was possible without adding an import

ghostdogpr (approved):

PR #2293, opened 2024-06-18
Add transformer that excludes fields / inputs based on directives

kyri-petrou:
Kind of experimental work based on transformers to allow excluding fields / input arguments based on the `@GQLTag` annotation.

This can be particularly helpful for schema versioning, e.g., having a `/beta` and a `/v1` endpoint, and excluding fields from `/v1` until we're happy with them without breaking compatibility in the APIs


core/src/main/scala/caliban/introspection/adt/__Field.scala
```diff
@@ -12,7 +12,8 @@ case class __Field(
   `type`: () => __Type,
   isDeprecated: Boolean = false,
   deprecationReason: Option[String] = None,
-  @GQLExcluded directives: Option[List[Directive]] = None
+  @GQLExcluded directives: Option[List[Directive]] = None,
+  @GQLExcluded tags: Set[String] = Set.empty
```
kyri-petrou: I'm not crazy happy about adding another argument to the case class here, but I tried to use directives for it but they ended up showing in the rendered schema. Happy to consider alternatives if we don't like this

ghostdogpr: Actually that's one thing I was not happy with in that blog post: https://blog.pierre-ricadat.com/graphql-in-scala-role-based-access-control: the directives show up in the schema and you might not want that. In that particular example I could now use those tags instead of the directives, but I am wondering if just String is powerful enough (you might want more, for example the roles were typed in my example). How about using directives but making it configurable whether they appear in the introspection/rendering?

kyri-petrou: I see your point. I think adding a way to hide directives from rendering/introspection makes sense.

However,how about we control it at the directive definition instead? Perhaps by adding a boolean argument `isIntrospectable` to the class that defaults to `false`?

ghostdogpr: Yeah I think that makes sense.

kyri-petrou: > How about using directives but making it configurable

Oh wait, this is what you recommended here. For some reason I thought you recommended to add some sort of config that can be passed to introspection / rendering

ghostdogpr: Yeah I think that makes sense = Yeah I think my idea makes sense 🤣

core/src/main/scala/caliban/schema/Annotations.scala
```diff
@@ -81,4 +82,16 @@ object Annotations extends AnnotationsVersionSpecific {
    */
   case class GQLOneOfInput() extends StaticAnnotation
 
+  /**
+   * Compile-time annotation that can be used in conjunction with [[caliban.transformers.Transformer]] to
+   * customize schema generation.
+   */
+  case class GQLTag(tags: String*)
```
paulpdaniels: We have an existing `GQLTag` annotation in federation since it's part of that spec already. Could we just move that instead?

kyri-petrou:
Following @paulpdaniels comment [here](https://github.com/ghostdogpr/caliban/pull/2293#discussion_r1645413030), I realised we were missing an opportunity to make the transformer work on _any_ directive instead of just the caliban-provided annotations.

So now users can do the following:

```scala

final case class Experimental extends GQLDirective("experimental")
final case class Internal extends GQLDirective("internal")

val apiA: GraphQL[Any] = api.transform(Transformer.ExcludeDirectives(Experimental()))
val apiB: GraphQL[Any] = api.transform(Transformer.ExcludeDirectives(Experimental(), Internal()))
```

core/src/main/scala-3/caliban/schema/AnnotationsVersionSpecific.scala
```diff
@@ -12,4 +14,14 @@ trait AnnotationsVersionSpecific {
    */
   case class GQLField() extends StaticAnnotation
 
+  /**
+   * Annotation used to provide directives to a schema type
+   */
+  open class GQLDirective(val directive: Directive) extends StaticAnnotation
```
kyri-petrou, line 29: Made this an `open class` in Scala 3, so that users can extend it without having to add the `import scala.language.adhocExtensions` import. More info on open classes [here](https://docs.scala-lang.org/scala3/reference/other-new-features/open-classes.html)

core/src/main/scala/caliban/rendering/DocumentRenderer.scala
```diff
@@ -572,11 +572,14 @@ object DocumentRenderer extends Renderer[Document] {
...
-      writer append d.name
-      inputArgumentsRenderer.unsafeRender(d.arguments, indent, writer)
-    }
+    override def unsafeRender(d: Directive, indent: Option[Int], writer: StringBuilder): Unit =
+      if (d.isIntrospectable) {
+        writer append '@'
+        writer append d.name
+        inputArgumentsRenderer.unsafeRender(d.arguments, indent, writer)
+      } else {
+        writer.deleteCharAt(writer.size - 1)
```
paulpdaniels: Is this to delete the potential space before the directive?

kyri-petrou: Yeap

kyri-petrou: I think I might have been a bit lazy here, there might be a better way to do it perhaps (by not adding the space to begin with?). I couldn't find where the space is added though :/

ghostdogpr (approved):

core/src/main/scala/caliban/rendering/DocumentRenderer.scala
paulpdaniels: I would use the `when` combinator instead since this isn't handling the "compact" mode case

kyri-petrou: Hmm seems that `when` doesn't work because we still get the empty strings. I used `filter` within the `contramap` method to filter out the directives from the List instead 👍

core/src/main/scala/caliban/transformers/Transformer.scala
```diff
@@ -326,20 +328,80 @@ object Transformer {
...
+     * A transformer that allows excluding fields and inputs with specific directives.
+     *
+     * {{{
+     *   case object Experimental extends GQLDirective(Directive("experimental"))
+     *   case object Internal extends GQLDirective(Directive("internal"))
+     *
+     *   ExcludeDirectives(Experimental, Internal)
+     * }}}
+     */
+    def apply(directives: GQLDirective*): Transformer[Any] =
```
paulpdaniels, line 343: I wonder if there should also be a filter variant `GQLDirective => Boolean`, since this would require exact match of arguments as well

kyri-petrou: I was actually thinking that as well. I'll add it 👍

kyri-petrou: Done

kyri-petrou:
@paulpdaniels I'll merge this PR in, in case there is anything else you'd like me to address let me know and I can do it in a followup PR

PR #2296, opened 2024-06-19
Fix cats interop schema bug

### LINKED ISSUE #2295: Broken backward compatibility in CatsInterop since v2.7.0
Consider a reproducer:
```scala
import caliban.interop.cats.implicits._
import caliban.schema.Schema.auto._
import caliban.{RootResolver, graphQL}
import cats.effect.std.{Console, Dispatcher}
import cats.effect.{IO, IOApp}
import zio.Runtime

case class Queries(one: IO[String])

object ExampleApp extends IOApp.Simple {
  implicit val zioRuntime: Runtime[Any] = Runtime.default

  override def run: IO[Unit] =
    Dispatcher.parallel[IO].use { implicit dispatcher =>
      val rendered = graphQL(RootResolver(Queries(IO.pure("")))).render

      Console[IO].println(rendered)
    }
}
```
With Caliban versions up to v2.6.0 (inclusive), the output is the following:
```
schema {
  query: Queries
}
type Queries {
  one: String!
}
```
Starting with Caliban v2.7.0 the output changes to this one:
```
schema {
  query: Queries
}
type Queries {
  one: String
}
```
I.e. The `one` query suddenly becomes _nullable_ (even though no changes to the model was made).
Works the same for both Scala 2.13 and 3.3.

kyri-petrou:
/fixes #2295

### DIFF
--- a/interop/cats/src/main/scala/caliban/interop/cats/CatsInterop.scala
+++ b/interop/cats/src/main/scala/caliban/interop/cats/CatsInterop.scala
@@ -190,9 +190,8 @@ object CatsInterop {
       override def toType(isInput: Boolean, isSubscription: Boolean): __Type =
         ev.toType_(isInput, isSubscription)
 
-      override def nullable: Boolean =
-        ev.nullable
-      override def canFail: Boolean  = true
+      override def nullable: Boolean = ev.nullable
+      override def canFail: Boolean  = ev.canFail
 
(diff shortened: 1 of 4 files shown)


ghostdogpr (approved):

satorg:
Thank you @kyri-petrou for the fix! I really appreciate your prompt support!

PR #2298, opened 2024-06-23
Return Exit for methods in Pagination

kyri-petrou:
We could easily also return `Either` as well, but at least Exit is source-compatible


ghostdogpr (approved):

ghostdogpr:
Looks good but there are some compile errors

ghostdogpr (approved):

PR #2304, opened 2024-06-24
Fix codegen path on windows

OlegYch:


tools/src/main/scala/caliban/tools/Codegen.scala
```diff
@@ -66,14 +66,14 @@ object Codegen {
...
                                     val path =
-                                      if (splitFiles) s"${arguments.toPath.reverse.dropWhile(_ != '/').reverse}$objectName.scala"
-                                      else arguments.toPath
+                                      if (splitFiles) new File(new File(arguments.toPath).getParentFile(), s"$objectName.scala")
+                                      else new File(arguments.toPath)
 
                                     ZIO.blocking(
                                       ZIO
-                                        .attempt(new PrintWriter(new File(path)))
+                                        .attempt(new PrintWriter(path))
```
ghostdogpr: Could you rename `path` to `file` so that the name matches the type? 🙏

ghostdogpr:
Thanks!

PR #2306, opened 2024-06-25
[Scala 3] Add annotation to derive all case class methods as graphql fields

kyri-petrou:
This PR adds the `@GQLMethodsAsFields` annotation that can be used to decorate a case class / case object to have all its public methods derived as fields (unless a method is annotated with `@GQLExcluded`.

We could potentially reuse the `@GQLField` annotation, although after staring at it for too long it really didn't make sense to me. I think we should be quite explicit and make it very clear to the user when looking at a case class definition that the public methods are going to be derived as fields. Let me know what you think @ghostdogpr!

/cc @paulpdaniels since I know you're migrating service(s) to Scala 3 you might be interested in this


ghostdogpr (approved):
Awesome!

I hate to be that guy but can you make a documentation PR? 😄

kyri-petrou:
> can you make a documentation PR? 😄

Is it too late to close this PR? 🥲

PR #2308, opened 2024-06-25
Use method handlers in `AllValidations`

kyri-petrou:
I don't really fully understand why this makes such a big difference, but for some reason it does 🤷 My guess is that we're no longer relying on the ZEnvironment to access the `Context`, but I'm struggling to believe that alone explains _this much_ difference.

Even introspection seems to be getting some improvement, although that one could be due to noise

series/2.x
```
[info] Benchmark                             Mode  Cnt        Score        Error  Units
[info] ValidationBenchmark.deep             thrpt    3   993038.766 ± 172816.842  ops/s
[info] ValidationBenchmark.introspection    thrpt    3    68131.837 ±  15473.702  ops/s
[info] ValidationBenchmark.multifield       thrpt    3  1120473.149 ± 198296.763  ops/s
[info] ValidationBenchmark.simple           thrpt    3  1587383.592 ±  68291.852  ops/s
[info] ValidationBenchmark.variableCoercer  thrpt    3    80349.172 ±   8376.846  ops/s
```

PR
```
[info] Benchmark                             Mode  Cnt        Score        Error  Units
[info] ValidationBenchmark.deep             thrpt    3  1317845.769 ±  29982.619  ops/s
[info] ValidationBenchmark.introspection    thrpt    3    70385.037 ±  14097.810  ops/s
[info] ValidationBenchmark.multifield       thrpt    3  1592332.948 ±  61986.333  ops/s
[info] ValidationBenchmark.simple           thrpt    3  2517365.378 ± 227198.856  ops/s
[info] ValidationBenchmark.variableCoercer  thrpt    3    80016.091 ±   2607.627  ops/s
```

PS: The PR has very minimal number of changes. Most of them are due to whitespaces, make sure to use the "Hide whitespace" option in Github PR view when reviewing :)

### DIFF
--- a/core/src/main/scala/caliban/validation/Validator.scala
+++ b/core/src/main/scala/caliban/validation/Validator.scala
@@ -34,10 +34,10 @@ object Validator {
   /**
    * A QueryValidation is a pure program that can access a Context, fail with a ValidationError or succeed with Unit.
    */
-  type QueryValidation            = EReader[Context, ValidationError, Unit]
+  type QueryValidation            = Context => EReader[Any, ValidationError, Unit]
   private type OptionalValidation = Option[EReader[Any, ValidationError, Unit]]
 
-  lazy val AllValidations: List[QueryValidation] =
+  val AllValidations: List[QueryValidation] =
     List(
       validateFragmentSpreads,
       validateOperationNameUniqueness,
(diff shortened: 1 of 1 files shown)


core/src/main/scala/caliban/validation/Validator.scala
```diff
@@ -34,10 +34,10 @@ object Validator {
   /**
    * A QueryValidation is a pure program that can access a Context, fail with a ValidationError or succeed with Unit.
    */
-  type QueryValidation            = EReader[Context, ValidationError, Unit]
+  type QueryValidation            = Context => EReader[Any, ValidationError, Unit]
```
kyri-petrou, line 37: Technically a _very_ user-facing breaking change, but I'll be extremely surprised if we have any users that define custom validations and provide them to the config. In the very unlikely case that someone does this, it should be very straightforward to convert the validation to a function

kyri-petrou:
Going to reopen this as a separate PR

PR #2309, opened 2024-06-25
support the @stream directive

paulpdaniels:
Adds support for the `@stream` directive which seems to have stabilized a bit. I attempted to merge the two concepts of defer and stream but they are just too different conceptually.

TODO:

- [x] Add validation for the stream and defer directives per spec
- [x] Add more unit tests for stream

### DIFF
--- a/core/src/main/scala/caliban/GraphQL.scala
+++ b/core/src/main/scala/caliban/GraphQL.scala
@@ -311,6 +311,14 @@ trait GraphQL[-R] { self =>
     override protected val transformer: Transformer[R]             = self.transformer
   }
 
+  final def enableAll(features0: Set[Feature]): GraphQL[R] = new GraphQL[R] {
+    override protected val schemaBuilder: RootSchemaBuilder[R]     = self.schemaBuilder
+    override protected val wrappers: List[Wrapper[R]]              = self.wrappers
+    override protected val additionalDirectives: List[__Directive] = self.additionalDirectives
+    override protected val features: Set[Feature]                  = self.features ++ features0
+    override protected val transformer: Transformer[R]             = self.transformer
+  }
+
   /**
    * Transforms the schema using the given transformer.
    * This can be used to rename or filter types, fields and arguments.
--- a/core/src/main/scala/caliban/execution/Executor.scala
+++ b/core/src/main/scala/caliban/execution/Executor.scala
@@ -44,13 +44,13 @@ object Executor {
     transformer: Transformer[R] = Transformer.empty[R]
   )(implicit trace: Trace): URIO[R, GraphQLResponse[CalibanError]] = {
     val wrapPureValues      = fieldWrappers.exists(_.wrapPureValues)
+    val featureFlags        = featureSet.foldLeft(0)(_ | _.mask)
     val stepReducer         =
       new StepReducer[R](
         transformer,
         request.operationType eq OperationType.Subscription,
-        isDeferredEnabled = featureSet(Feature.Defer),
-        isStreamEnabled = featureSet(Feature.Stream),
-        wrapPureValues = wrapPureValues
+        wrapPureValues = wrapPureValues,
+        featureFlags
       )
     val reducedStepExecutor =
       new ReducedStepExecutor[R](
@@ -168,9 +168,8 @@ object Executor {
   private final class StepReducer[R](
     transformer: Transformer[R],
     isSubscription: Boolean,
-    isDeferredEnabled: Boolean,
-    isStreamEnabled: Boolean,
-    wrapPureValues: Boolean
+    wrapPureValues: Boolean,
+    flags: Feature.Flags
   )(implicit trace: Trace) {
 
     def reduceStep(
@@ -192,7 +191,7 @@ object Executor {
 
         val filteredFields    = mergeFields(currentField, objectName)
         val (deferred, eager) =
-          if (isDeferredEnabled) {
+          if (Feature.isDeferEnabled(flags)) {
             filteredFields.partitionMap { f =>
               val entry = reduceField(f)
               f.fragment match {
@@ -292,7 +291,7 @@ object Executor {
           )
         } else {
           currentField match {
-            case IsStream(label, Some(initialCount)) if initialCount > 0 && isStreamEnabled =>
+            case IsStream(label, Some(initialCount)) if initialCount > 0 && Feature.isStreamEnabled(flags) =>
               ReducedStep.QueryStep(
                 ZQuery.fromZIONow(
                   for {
@@ -317,7 +316,7 @@ object Executor {
                   }
                 )
               )
-            case IsStream(label, _) if isStreamEnabled                                      =>
+            case IsStream(label, _) if Feature.isStreamEnabled(flags)                                      =>
               ReducedStep.DeferStreamStep(
                 reduceStep(PureStep(ListValue(Nil)), currentField, arguments, path),
                 ReducedStep.StreamStep(
--- a/core/src/main/scala/caliban/execution/Executor.scala
+++ b/core/src/main/scala/caliban/execution/Executor.scala
@@ -14,7 +14,7 @@ import caliban.wrappers.Wrapper.FieldWrapper
 import zio._
 import zio.query._
 import zio.stacktracer.TracingImplicits.disableAutoTrace
-import zio.stream.ZStream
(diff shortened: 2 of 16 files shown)


core/src/main/scala/caliban/GraphQL.scala
```diff
@@ -30,6 +30,7 @@ trait GraphQL[-R] { self =>
   protected val additionalDirectives: List[__Directive]
   protected val features: Set[Feature]
   protected val transformer: Transformer[R]
+  protected val additionalValidations: List[Validator.QueryValidation]
```
kyri-petrou: I had a quick look through the PR and couldn't find any usages of this. What is it being used for?

paulpdaniels: Still in progress, debating the best way to add new validations through an aspect. Either through this mechanism or by using an execution wrapper to manipulate Configuration. The latter seems like it would add more overhead per run

examples/src/main/scala/example/ExampleApi.scala
```diff
@@ -91,13 +92,13 @@ object ExampleApi {
...
-      printErrors @@                  // wrapper that logs errors
-      apolloTracing() @@              // wrapper for https://github.com/apollographql/apollo-tracing
-      DeferSupport.defer              // wrapper that enables @defer directive support
+      maxFields(300) @@                         // query analyzer that limit query fields
+      maxDepth(30) @@                           // query analyzer that limit query depth
+      timeout(3 seconds) @@                     // wrapper that fails slow queries
+      printSlowQueries(500 millis) @@           // wrapper that logs slow queries
+      printErrors @@                            // wrapper that logs errors
+      apolloTracing() @@                        // wrapper for https://github.com/apollographql/apollo-tracing
+      IncrementalDelivery.aspect(Feature.Defer) // wrapper that enables @defer directive support
```
ghostdogpr: How about making some nice aliases `IncrementalDelivery.defer` and `IncrementalDelivery.stream`?

ghostdogpr: And maybe `IncrementalDelivery.all`

kyri-petrou:
@paulpdaniels the mima checks are disabled in the main branch, so if you rebase the PR it'll pass now.

By the way let's try and get this one in the upcoming release since it contains breaking changes

core/src/main/scala/caliban/execution/Executor.scala
```diff
@@ -141,6 +169,7 @@ object Executor {
     transformer: Transformer[R],
     isSubscription: Boolean,
     isDeferredEnabled: Boolean,
+    isStreamEnabled: Boolean,
     wrapPureValues: Boolean
```
paulpdaniels: @kyri-petrou what do you think about collapsing these into a single int with bit masking?

examples/src/main/scala/example/ExampleApi.scala
paulpdaniels: Will do today

core/src/main/scala/caliban/execution/Executor.scala
kyri-petrou: Yeah I think that's a good idea 👍

paulpdaniels:
@kyri-petrou thanks, rebased

examples/src/main/scala/example/ExampleApi.scala
paulpdaniels: Done

ghostdogpr: I think you can use `IncrementalDelivery.defer` which is shorter

interop/tapir/src/test/scala/caliban/interop/tapir/TestApi.scala
```diff
@@ -65,6 +66,6 @@ object TestApi extends GenericSchema[TestService with Uploads] {
       timeout(3 seconds) @@           // wrapper that fails slow queries
       printSlowQueries(500 millis) @@ // wrapper that logs slow queries
       apolloTracing() @@              // wrapper for https://github.com/apollographql/apollo-tracing
-      DeferSupport.defer @@
+      IncrementalDelivery.aspect(Feature.Defer) @@
```
ghostdogpr: Here too (and everywhere else)

core/src/main/scala/caliban/execution/Feature.scala
```diff
@@ -1,8 +1,62 @@
...
+}
 
 object Feature {
-  case object Defer  extends Feature
-  case object Stream extends Feature
+  type Flags = Int
+
+  private def isEnabled(flags: Flags, feature: Feature): Boolean = (flags & feature.mask) != 0
+  def isDeferEnabled(flags: Flags): Boolean                      = isEnabled(flags, Defer)
+  def isStreamEnabled(flags: Flags): Boolean                     = isEnabled(flags, Stream)
```
kyri-petrou: In this case it doesn't really matter since there are only 2 objects extending `Feature`, but to avoid the virtual call to `.mask` I usually do this:

```suggestion
  private def isEnabled(flags: Flags, mask: Int): Boolean = (flags & mask) != 0
  def isDeferEnabled(flags: Flags): Boolean                      = isEnabled(flags, Defer.mask)
  def isStreamEnabled(flags: Flags): Boolean                     = isEnabled(flags, Stream.mask)
```

core/src/main/scala/caliban/execution/Fragment.scala
```diff
@@ -7,7 +7,7 @@ case class Fragment(name: Option[String], directives: List[Directive]) {}
 
 object Fragment {
   object IsDeferred {
-    def unapply(fragment: Fragment): Option[Option[String]] =
+    def unapply(fragment: Fragment): Option[(Option[String])] =
```
kyri-petrou: Out of curiosity, is there a reason for this change or is it a dev leftover?

core/src/main/scala/caliban/execution/Executor.scala
```diff
@@ -163,7 +191,7 @@ object Executor {
 
         val filteredFields    = mergeFields(currentField, objectName)
         val (deferred, eager) =
-          if (isDeferredEnabled) {
+          if (Feature.isDeferEnabled(flags)) {
```
kyri-petrou, line 194: Should we also be checking whether `isStreamEnabled` here?

paulpdaniels: No this is only checking for defer at the fragment level, stream can only apply to list types so no need to check here

core/src/main/scala/caliban/execution/Fragment.scala
paulpdaniels: Artifact of dev, I think I was playing around with adding more context and compiler apparently doesn't complain about a tuple of one

kyri-petrou (approved):
LGTM

interop/tapir/src/test/scala/caliban/interop/tapir/TestApi.scala
```diff
@@ -1,13 +1,14 @@
 package caliban.interop.tapir
 
 import caliban._
+import caliban.execution.Feature
```
ghostdogpr: Not needed

interop/tapir/src/test/scala/caliban/interop/tapir/TestApi.scala
```diff
@@ -1,13 +1,14 @@
...
 import caliban.interop.tapir.TestData._
 import caliban.schema.Annotations.{ GQLDeprecated, GQLDescription }
 import caliban.schema.{ GenericSchema, Schema }
 import caliban.schema.ArgBuilder.auto._
 import caliban.uploads.{ Upload, Uploads }
 import caliban.wrappers.ApolloTracing.apolloTracing
-import caliban.wrappers.{ Caching, DeferSupport }
+import caliban.wrappers.{ Caching, DeferSupport, IncrementalDelivery }
```
ghostdogpr: I think DeferSupport is not needed either?

examples/src/main/scala/example/ExampleApi.scala
```diff
@@ -2,12 +2,13 @@ package example
 
 import example.ExampleData._
 import caliban._
+import caliban.execution.Feature
```
ghostdogpr: Same 2 imports can be removed

ghostdogpr (approved):

PR #2310, opened 2024-06-26
Use `Either` instead of `ZPure` in `Validator` / `VariableCoercer`

kyri-petrou:
So far we've been using a ZPure in the Validator & VariableCoercer mainly to ensure stack-safety when validating queries. However, for a stack-overflow error to occur we need to either:
1. Have fragments that reference themselves or,
2. An extremely deep query (modern JVMs can handle thousands of recursive calls before throwing a SO error)

Since we already validate (1), the only "real" problem that remains is (2). However, I think that this is not a real issue because the real case where a query would be deep enough to trigger a SO error is for malicious purposes / to DDoS the server.

To guard against this possibility, users _should_ be using the `maxDepth` wrapper. But in the case they're not, we catch the SO error as a last resort and fail gracefully via a Validation error.

This approach is much more performant but also allows us to become "leaner", reducing our reliance on ZIO for most core operations

### Benchmark results:

series/2.x
```
[info] Benchmark                             Mode  Cnt        Score        Error  Units
[info] ValidationBenchmark.deep             thrpt    3   993038.766 ± 172816.842  ops/s
[info] ValidationBenchmark.introspection    thrpt    3    68131.837 ±  15473.702  ops/s
[info] ValidationBenchmark.multifield       thrpt    3  1120473.149 ± 198296.763  ops/s
[info] ValidationBenchmark.simple           thrpt    3  1587383.592 ±  68291.852  ops/s
[info] ValidationBenchmark.variableCoercer  thrpt    3    80349.172 ±   8376.846  ops/s
```

PR:
```
[info] Benchmark                             Mode  Cnt        Score        Error  Units
[info] ValidationBenchmark.deep             thrpt    6  2259244.590 ±  42814.904  ops/s
[info] ValidationBenchmark.introspection    thrpt    6   111222.161 ±  19679.459  ops/s
[info] ValidationBenchmark.multifield       thrpt    6  2663080.586 ±  32436.908  ops/s
[info] ValidationBenchmark.simple           thrpt    6  5988742.244 ± 239073.495  ops/s
[info] ValidationBenchmark.variableCoercer  thrpt    6   181029.042 ±  18279.451  ops/s
```

PS: I'll followup with another PR to remove `zio-prelude` as a dependency since it's not _really_ needed anymore

### DIFF
--- a/core/src/main/scala/caliban/validation/Validator.scala
+++ b/core/src/main/scala/caliban/validation/Validator.scala
@@ -52,14 +52,14 @@ object Validator {
    * Fails with a [[caliban.CalibanError.ValidationError]] otherwise.
    */
   def validate(document: Document, rootType: RootType)(implicit trace: Trace): IO[ValidationError, Unit] =
-    Configurator.ref.getWith(v => Exit.fromEither(check(document, rootType, Map.empty, v.validations).map(_ => ())))
+    Configurator.ref.getWith(v => Exit.fromEither(check(document, rootType, Map.empty, v.validations).unit))
 
   /**
    * Verifies that the given document is valid for this type for all available validations.
    * Fails with a [[caliban.CalibanError.ValidationError]] otherwise.
    */
   def validateAll(document: Document, rootType: RootType): Either[ValidationError, Unit] =
-    check(document, rootType, Map.empty, AllValidations).map(_ => ())
+    check(document, rootType, Map.empty, AllValidations).unit
 
   /**
    * Verifies that the given schema is valid. Fails with a [[caliban.CalibanError.ValidationError]] otherwise.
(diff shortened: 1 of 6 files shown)


core/src/main/scala/caliban/validation/Validator.scala
```diff
@@ -67,29 +66,31 @@ object Validator {
...
+      _      <- validateAllDiscard(types.sorted)(validateType)
+      _      <- validateClashingTypes(types)
+      _      <- validateDirectives(types)
+      _      <- validateRootMutation(schema)
+      _      <- validateRootSubscription(schema)
+      schema <- validateRootQuery(schema)
+    } yield schema
+  }
+
+  private val unit: Either[ValidationError, Unit] = Right(())
```
ghostdogpr: Isn't it same as `unitR`? I prefer `unit` though.

core/src/main/scala/caliban/validation/Validator.scala
```diff
@@ -53,7 +52,7 @@ object Validator {
    * Fails with a [[caliban.CalibanError.ValidationError]] otherwise.
    */
   def validate(document: Document, rootType: RootType)(implicit trace: Trace): IO[ValidationError, Unit] =
-    Configurator.ref.getWith(v => Exit.fromEither(check(document, rootType, Map.empty, v.validations).unit))
+    Configurator.ref.getWith(v => Exit.fromEither(check(document, rootType, Map.empty, v.validations).map(_ => ())))
```
ghostdogpr: How about adding a small extension method for that?

ghostdogpr (approved):

PR #2315, opened 2024-06-27
Avoid repeating validations on Fragments

kyri-petrou:
One of the pesky things about Fragments is that it's very "easy" to end up repeating doing the same work, especially when it comes to validation. This happens as a field's selection set will contain a fragment, and we end up re-validating the fragment all-over again.

This PR adds caching for fragment validation in a few places in the Validator. Note that we've already been doing this in some other parts.

With these changes, there's ~60% throughput increase for validating fragments based on the `introspection` query from this very "simple" optimization :)


ghostdogpr (approved):
Brilliant 😄

PR #2318, opened 2024-06-28
World's smallest micro-optimization

kyri-petrou:
Full disclosure. This will probably have 0 meaningful impact to anyone's application, but I saw it the other day and couldn't help myself 🥲 I wouldn't be doing this is the previous code was intuitive or more readable, but I think they're both equally bad when it comes to readability.

Anyways, in APQs, instead of converting each byte into a string and extracting its 4-bit representation (hex), we just do the comparison on the fly avoiding materialising anything in memory.


ghostdogpr:
Hum... 🙄 
I think the old one is definitely more readable 😅

PR #2321, opened 2024-06-29
Add method to "suspend" wrappers and reimplement APQs

kyri-petrou:
Should have made this PR prior to the 2.8.0 release, but I guess it'll have to wait till 2.9.0 🥲

Main changes in this PR:

- Added `Wrapper.suspend` which allows "suspending" side effects used in the construction of a Wrapper. This works the same way as `EffectfulWrapper` but without having to use ZIO for it
- Rewrote a few wrappers to use suspended wrappers instead. Unfortunately, since we used `EffectfulWrapper` as the type returned from the different methods, we can't change them without breaking binary compatibility.
- Rewrote APQs to return a `Wrapper[Any]` instead of `Wrapper[ApolloPersistence]`. There are a few reasons for this, but mainly this is because it makes it quite annoying to integrate this wrapper in non-ZIO applications, especially if the users are not familiar with ZLayers and how to eliminate dependencies from the environment. Users now have 2 ways to use this wrapper:
  - `ApolloPersistedQueries.wrapper`, which will create a new instance of `ApolloPersistence.Live` and reuse it for all queries
  - `ApolloPersistedQueries.wrapper(???)` which allows users to provide a custom `ApolloPersistence` implementation


core/src/main/scala/caliban/wrappers/FieldMetrics.scala
```diff
@@ -56,16 +56,13 @@ object FieldMetrics {
...
-    Wrapper.EffectfulWrapper(
-      for {
-        timings  <- ZIO.succeed(new AtomicReference(List.empty[Timing]))
-        failures <- ZIO.succeed(new AtomicReference(List.empty[String]))
-        clock    <- ZIO.clock
-        metrics   = new Metrics(totalLabel, durationLabel, buckets, extraLabels)
-      } yield overallWrapper(timings, failures, metrics) |+|
-        Unsafe.unsafe(implicit us => fieldWrapper(clock.unsafe.nanoTime(), timings, failures))
-    )
+  )(implicit clock: Clock = Clock.ClockLive): Wrapper[Any] =
```
ghostdogpr: I quite prefer the old way, that let you customize the clock with `ZIO.withClock`, implicit parameter is a bit annoying.

federation/src/test/scala/caliban/federation/tracing/FederationTracingSpec.scala
```diff
@@ -33,20 +33,22 @@ object FederationTracingSpec extends ZIOSpecDefault {
...
-  def api(excludePureFields: Boolean) = graphQL(
-    RootResolver(
-      Queries(
-        me = ZQuery.succeed(
-          User(
-            "abc123",
-            ZIO.sleep(100.millis) as Name("my_first", Some("my_last")),
-            age = 42,
-            parents = ZIO.succeed(List(Parent("my_parent")))
+  def api(excludePureFields: Boolean) = ZIO.clock.map { implicit clock =>
```
ghostdogpr: Yeah definitely not sold on this 😄  I don't think running that effect once per query is significant at all 🤔

kyri-petrou: Ok I'll change these ones back

kyri-petrou:
I reverted the changes to the methods with Clock implicits, and also reverted the deletion of `ApolloCache`. This way the PR is fully binary compatible (I removed the label as well)

core/src/test/scala/caliban/wrappers/WrappersSpec.scala
```diff
@@ -272,51 +268,6 @@ object WrappersSpec extends ZIOSpecDefault {
           }
         )
       },
-      test("Apollo Caching") {
```
kyri-petrou, line 275: I won't revert this one, just to make it easier to delete the file in the next major release

ghostdogpr (approved):

PR #2329, opened 2024-07-06
Add param to `GraphQLRequest` to identify GET requests

kyri-petrou:
In short, I would like to reduce our reliance on `FiberRef` to propagate information across different parts in Caliban for various reasons but primarily because updating FiberRefs continuously will result in degraded performance. With this PR, we use the GraphQLRequest extensions to propagate that information as it's much cheaper to do so.

@ghostdogpr @paulpdaniels ~~Do you see an issue with this approach? Alternatively, we could add a parameter to the `executeRequest` method, but that would require breaking the API, and that might be too soon after the last major version release~~

EDIT: Based on @paulpdaniels recommendation, and since we're updating the minor version, I opted for the bincompat breaking change by adding the param to GraphQLRequest


paulpdaniels:
I'm not really in favor of this approach. I'd move toward pushing the value into the graphql request directly instead of mixing it with user-land code. Even if that means waiting for 2.8.0

kyri-petrou:
Yeah I was worried about that part as well. I'll make the change and mark the PR as breaking

core/src/main/scala/caliban/GraphQLRequest.scala
```diff
@@ -13,18 +13,21 @@ case class GraphQLRequest(
   query: Option[String] = None,
   operationName: Option[String] = None,
   variables: Option[Map[String, InputValue]] = None,
-  extensions: Option[Map[String, InputValue]] = None
-) {
+  extensions: Option[Map[String, InputValue]] = None,
+  @transient isHttpGetRequest: Boolean = false
```
paulpdaniels, line 17: What does this do in the context of a case class member?

paulpdaniels (approved):

core/src/main/scala/caliban/GraphQLRequest.scala
kyri-petrou: This one comes from `jsoniter` (it's not the Scala one). It excludes the case class member from serialization and always uses the default value for deserialization

ghostdogpr (approved):

PR #2332, opened 2024-07-08
Attempt at a fix to #2331

### LINKED ISSUE #2331: Since 2.8.0 compilation take forever
Unfortunately, the code I am working on is closed source. But since 2.8.0 (same issue 2.8.1) compilation take forever and never finish.

kyri-petrou:
This is a shot-in-the-dark attempt to help #2331 because I'm really out of ideas on what might be going wrong.
I doubt it'll help much but it definitely can't make things worse 🤷


ghostdogpr:
I think we can wait to have more information

kyri-petrou:
@ghostdogpr I think we can merge this one in. It's an improvement even if it doesn't fix the issue since we can do a 1-pass iteration over the methods instead of 2 for the common case (no `@GQLFieldsFromMethods` annotation)

ghostdogpr (approved):

PR #2338, opened 2024-07-10
Optimize execution of pure and single-field object queries

kyri-petrou:
This PR optimizes the execution of the following:
1. Queries which query a single field from objects, and that field is effectful
2. Queries that do not contain any effectful fields. While this might seem like a very specific / non-realistic optimization, it is quite relevant for Loom-based frameworks where we don't need to use effects for async/blocking tasks


ghostdogpr (approved):

core/src/main/scala/caliban/execution/Executor.scala
```diff
@@ -126,10 +126,10 @@ object Executor {
...
     }
 
-    for {
-      cache    <- makeCache
-      response <- runQuery(stepReducer.reduceStep(plan, request.field, Map.empty, Nil), cache)
-    } yield response
+    stepReducer.reduceStep(plan, request.field, Map.empty, Nil) match {
+      case PureStep(resp) => Exit.succeed(GraphQLResponse(resp, Nil))
+      case reducedStep    => makeCache.flatMap(runQuery(reducedStep, _))
+    }
```
paulpdaniels, line 134: Don't we still need to suspend this? Returning an Exit means we lose referential transparency if it's now getting strictly evaluated.

kyri-petrou: Technicaly, you're right, this is a bit of an interesting because we execute this method on each request, and it's not really a user-facing method (even if it's public). So it doesn't _need_ to be referentially transparent. But since it does return a `ZIO`, I think we should make it referentially transparent to avoid any potential issues in the future

core/src/main/scala/caliban/wrappers/Wrapper.scala
```diff
@@ -175,46 +175,47 @@ object Wrapper {
...
-            case Nil => None
-            case fs  => Some(ZIO.collectAllDiscard(fs))
-          }
-        case EffectfulWrapper(wrapper)        =>
-          Some(wrapper.flatMap {
-            loop(_) match {
-              case None    => Exit.unit
-              case Some(w) => w
+    else
+      ZIO.suspendSucceed {
```
kyri-petrou, line 179: Added this here in order to align with https://github.com/ghostdogpr/caliban/pull/2338/files#r1673288713. We don't need this method to be referentially transparent because it's executed on each request, but better to have it referentially transparent for consistency

paulpdaniels (approved):

PR #2341, opened 2024-07-17
Make `jsoniter` a required dependency and use it for enc/dec in tapir adapters

kyri-petrou:
As discussed on Discord, this is an attempt to improve/simplify UX by using jsoniter as the only JSON encoding/decoding implementation for requests/responses.

Since the users are not expected to provide custom codecs for requests/responses (in fact, most of the time they shouldn't even have to know of the existence of `GraphQLRequest` / `GraphQLResponse`), giving them an option of providing custom codecs is highly redundant and increases complexity.

With the changes in this PR, users no longer need to add a `tapir-json-*` dependency, and they no longer need to add any imports where they create the `HttpInterpreter`. Plus, this makes it easier for us to maintain this project in the longer term, as evident by this 👇 

image


paulpdaniels (approved):

ghostdogpr (approved):
Can you make a separate PR for updating the docs?

kyri-petrou:
> Can you make a separate PR for updating the docs?

Sure will do

PR #2347, opened 2024-07-25
Support introspecting and client codegen of deprecated arguments

### LINKED ISSUE #2345: Deprecated arguments are missing in introspection query
Since we upgraded from 2.02 to 2.6.0 introspection queries no longer return arguments marked as deprecated.
This doesn't seem to affect all definitions that are marked as deprecated. 

I managed to reproduce the issues with new tests here:
https://github.com/hwielenberg/caliban/commit/9a18e0bd42a9f5abb1bc4774f8ea7c3c42db2d59

Our main problem is the `ArgumentAdded` error in the `IntrospectionClientSpec` spec. But the tests show other inconsistencies as well.

Here are the failures:
```
sbt:caliban-tools> testOnly caliban.tools.RemoteSchemaSpec
+ RemoteSchemaSpec
  - is isomorphic
...
      -  getObject(name: String! @deprecated(reason: "Use nameV2"), nameV2: String!): Object!
      +  getObject(name: String!, nameV2: String!): Object!
...
```

```
sbt:caliban-tools> testOnly caliban.tools.IntrospectionClientSpec
+ IntrospectionClientSpec
...
  - is isomorphic
    ✗ List was not empty (size 4)
    res.isEmpty
    res = List(
        ArgumentAdded(
            argName = "name",
            target = Field(name = "getObject", typeName = "Queries"),
            optional = true
        ), 
        DirectiveDefinitionDeleted(
            directiveName = "skip"
        ), 
        DirectiveDefinitionDeleted(
            directiveName = "include"
        ),
        DirectiveDefinitionDeleted(
            directiveName = "specifiedBy"
        )
    )
    at /Users/henning.wielenberg/caliban/tools/src/test/scala/caliban/tools/IntrospectionClientSpec.scala:48 

```

kyri-petrou:
/closes #2345

Couple of remarks:
1. I used `true` as the default for this option for both codegen and introspection. Should we be using `false`?
2. I used a class instead of case class for `IntrospectionClient.Config`. This is because classes are much easier to extend without breaking binary compatibility in the future. Should we care about this since it's in tools?

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -218,7 +218,7 @@ lazy val tools = project
       "dev.zio"                       %% "zio-json"            % zioJsonVersion % Test
     )
   )
-  .dependsOn(core, clientJVM)
+  .dependsOn(core, clientJVM, quickAdapter % Test)
 
 lazy val tracing = project
   .in(file("tracing"))
--- a/client/src/main/scala/caliban/client/IntrospectionClient.scala
+++ b/client/src/main/scala/caliban/client/IntrospectionClient.scala
@@ -150,8 +150,17 @@ object IntrospectionClient {
         arguments = List(Argument("includeDeprecated", includeDeprecated, "Boolean"))
       )
     def inputFields[A](innerSelection: SelectionBuilder[__InputValue, A]): SelectionBuilder[__Type, Option[List[A]]] =
-      Field("inputFields", OptionOf(ListOf(Obj(innerSelection))))
-    def ofType[A](innerSelection: SelectionBuilder[__Type, A]): SelectionBuilder[__Type, Option[A]]                  =
+      inputFields(None)(innerSelection)
+    def inputFields[A](includeDeprecated: Option[Boolean])(
+      innerSelection: SelectionBuilder[__InputValue, A]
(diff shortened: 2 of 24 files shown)


ghostdogpr:
> 1. I used `true` as the default for this option for both codegen and introspection. Should we be using `false`?

I think it should be `false` to not break anything.

> 2. I used a class instead of case class for `IntrospectionClient.Config`. This is because classes are much easier to extend without breaking binary compatibility in the future. Should we care about this since it's in tools?

I know there's a caliban-mill library that uses `tools`, maybe some people have custom CI scripts as well. Maybe we just avoid breaking the codegen interface and it should be enough.

ghostdogpr:
Wait, I made `supportRepeatable` true by default? 🤣 Let's be consistent with old me then.

kyri-petrou:
> Wait, I made `supportRepeatable` true by default? 🤣 Let's be consistent with old me then.

Yeah I actually had it set to `false` initially and changed it afterwards to be consistent 😆

ghostdogpr (approved):

PR #2366, opened 2024-08-17
Optimize `Function0` allocations

kyri-petrou:
This PR optimizes (mostly) `Function0` allocations that originate from `map.getOrElse(key, ???)`. In Scala 2, the optimization is fairly straight-forward; we just need to store the default value in a `Function0`. Since `=> A` compiles to `() => A`, this means that invocations of the `getOrElse(key, DefaultValueFn())` don't create a `Function0`.

In Scala 3 however, it's slightly more complex. For some reason (might be a bug, I'll open an issue in Dotty about it), this works only if the val is annotated with `@static`, which compiles to a `final static` field. To avoid having to do this everywhere, I added a package-private `syntax` object with the `getOrElseNull` method.

I've also noticed some small optimizations that could be done here and there so I did them as part of this PR

### DIFF
--- a/core/src/main/scala/caliban/validation/Validator.scala
+++ b/core/src/main/scala/caliban/validation/Validator.scala
@@ -27,6 +27,7 @@ import scala.collection.mutable.ListBuffer
 
 object Validator extends SchemaValidator {
   import ValidationOps._
+  import caliban.syntax._
 
   /**
    * A QueryValidation is a pure program that can access a Context, fail with a ValidationError or succeed with Unit.
@@ -349,12 +350,7 @@ object Validator extends SchemaValidator {
           )
 
           val v2 =
-            if (
-              t match {
-                case Some(t0) => t0._isOneOfInput
-                case _        => false
-              }
-            )
+            if (t.isDefined && t.get._isOneOfInput)
               Some(
                 failWhen(v.variableType.nullable)(
                   s"Variable '${v.name}' cannot be nullable.",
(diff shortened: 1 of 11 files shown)


core/src/main/scala/caliban/validation/Validator.scala
```diff
@@ -349,12 +350,7 @@ object Validator extends SchemaValidator {
           )
 
           val v2 =
-            if (
-              t match {
-                case Some(t0) => t0._isOneOfInput
-                case _        => false
-              }
-            )
+            if (t.isDefined && t.get._isOneOfInput)
```
ghostdogpr: `exists`

ghostdogpr: Or did you avoid it to prevent the allocation under Scala 3?

ghostdogpr (approved):

core/src/main/scala/caliban/validation/Validator.scala
kyri-petrou: Sorry forgot to respond, yeah that was the reason 👍

ghostdogpr: I saw your thumb up 😄

ghostdogpr: It's quite sad we have to do that though 😭

kyri-petrou: Actually, I think here we can change it. This is only executed for input variables which are usually very few

kyri-petrou: Done 👍

ghostdogpr (approved):

PR #2368, opened 2024-08-19
Optimize `Field#allFieldsUniqueNameAndCondition`

kyri-petrou:
This method exists in the _very_ hot path of execution, so even micro-optimizations to it are quite nice.

Basically this PR does the following:
1. Avoids initializing the `lazy val` when we can check if the conditions are unique without iterating through the fields
2. Removes the unique check on the `aliasedName`, which also removes the need for allocating a HashSet. It just occurred to me that we if the conditions are all the same, then the names are guaranteed to be unique because otherwise we would have deduplicated them when we create the Field

### DIFF
--- a/core/src/main/scala/caliban/execution/Field.scala
+++ b/core/src/main/scala/caliban/execution/Field.scala
@@ -48,9 +48,9 @@ case class Field(
 
   // TODO: Change the name to `allConditionsUnique` in the next minor version
   private[caliban] def allFieldsUniqueNameAndCondition: Boolean =
-    fields.isEmpty || fields.tail.isEmpty || allConditionsUniqueLzy
+    fields.isEmpty || fields.tail.isEmpty || allConditionsUniqueLazy
 
-  private lazy val allConditionsUniqueLzy: Boolean = {
+  private lazy val allConditionsUniqueLazy: Boolean = {
     val headCondition = fields.head._condition
     var rem           = fields.tail
     var res           = true
--- a/core/src/main/scala/caliban/execution/Field.scala
+++ b/core/src/main/scala/caliban/execution/Field.scala
@@ -46,25 +46,21 @@ case class Field(
   private[caliban] val aliasedName: String =
     if (alias.isEmpty) name else alias.get
 
-  private[caliban] lazy val allFieldsUniqueNameAndCondition: Boolean = {
-    def inner(fields: List[Field]): Boolean = {
-      val headCondition = fields.head._condition
+  // TODO: Change the name to `allConditionsUnique` in the next minor version
+  private[caliban] def allFieldsUniqueNameAndCondition: Boolean =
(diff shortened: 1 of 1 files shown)


core/src/main/scala/caliban/execution/Field.scala
```diff
@@ -46,25 +46,21 @@ case class Field(
...
-      var rem = fields.tail
-      while (rem ne Nil) {
-        val f        = rem.head
-        val continue = seen.add(f.aliasedName) && f._condition == headCondition
-        if (!continue) return false
+  // TODO: Change the name to `allConditionsUnique` in the next minor version
+  private[caliban] def allFieldsUniqueNameAndCondition: Boolean =
+    fields.isEmpty || fields.tail.isEmpty || allConditionsUniqueLzy
+
+  private lazy val allConditionsUniqueLzy: Boolean = {
```
ghostdogpr: Lzy -> Lazy?

core/src/main/scala/caliban/execution/Field.scala
```diff
@@ -46,25 +46,21 @@ case class Field(
...
-    def inner(fields: List[Field]): Boolean = {
-      val headCondition = fields.head._condition
-
-      val seen = new mutable.HashSet[String]
-      seen.add(fields.head.aliasedName)
-
-      var rem = fields.tail
-      while (rem ne Nil) {
-        val f        = rem.head
-        val continue = seen.add(f.aliasedName) && f._condition == headCondition
```
ghostdogpr, line 59: Is there a test to verify that? I don't remember the logic well enough to know if this is correct to remove the condition 😄

core/src/main/scala/caliban/execution/Field.scala
```diff
@@ -46,25 +46,21 @@ case class Field(
...
-        if (!continue) return false
+  // TODO: Change the name to `allConditionsUnique` in the next minor version
+  private[caliban] def allFieldsUniqueNameAndCondition: Boolean =
+    fields.isEmpty || fields.tail.isEmpty || allConditionsUniqueLazy
+
+  private lazy val allConditionsUniqueLazy: Boolean = {
+    val headCondition = fields.head._condition
+    var rem           = fields.tail
+    var res           = true
+    while ((rem ne Nil) && res) {
```
paulpdaniels: Wasn't there some optimization to assign Nil to a local variable that we used elsewhere?

paulpdaniels (approved):

core/src/main/scala/caliban/execution/Field.scala
kyri-petrou: Yes there are a few tests for it. It's also not possible to get a duplication of the name here when the conditions are all the same since we already deduplicate by using `(aliasedName, condition)` as the deduplication key

ghostdogpr (approved):

PR #2373, opened 2024-08-25
use zio-config 4.x

### LINKED ISSUE #2371: Upgrade or remove zio-config
Currently we are using zio-config 3.x even though the latest version is 4.x. The problem is that 4.x removed the `CommandLine` source, which is the main reason we were using it (see https://github.com/zio/zio-config/releases/tag/v4.0.0-RC2).

We should either find a way to upgrade or remove that dependency so that we are not blocked in the future.

paulpdaniels:
Upgrade zio-config to latest

closes #2371


ghostdogpr:
Can you run fmt?

ghostdogpr (approved):

PR #2375, opened 2024-08-25
Provide more examples of tapir interop

calvinlfer:
- Caliban GraphQL to Tapir Endpoint (added)
- Tapir Endpoint to Caliban GraphQL (existing)

image


examples/src/main/scala/example/calibantotapir/tapir/GraphiqlEndpoint.scala
```diff
@@ -0,0 +1,82 @@
...
+import sttp.capabilities.zio.ZioStreams
+
+object GraphiQLEndpoint {
+  val graphiql: ZServerEndpoint[Any, ZioStreams] =
+    infallibleEndpoint.get
+      .in("graphiql")
+      .out(htmlBodyUtf8)
+      .serverLogic(_ => ZIO.right(html(apiPath = "graphql", uiPath = "graphiql")))
+
+  private def html(apiPath: String, uiPath: String): String =
```
kyri-petrou: Is this the same html that's served via the `quick` module? If yes, should we move it in a common module so that it can be used by the tapir modules as well?

calvinlfer: yessir! it's the exact same

calvinlfer: Should we put it in tapir-interop? Looks like quick can access it then

kyri-petrou: `quick` depends on `tapir-interop` only for tests. It'll have to go into core unfortunately, which isn't a very suitable place for it but it's the only common module.

Maybe let's open a separate issue / PR for it (supporting graphiql OOTB for tapir-based adapters).

calvinlfer: Created https://github.com/ghostdogpr/caliban/issues/2376#issue-2485501938 to track this

examples/src/main/scala/example/calibantotapir/README.md
```diff
@@ -0,0 +1,11 @@
+# Caliban to Tapir
+
+This example shows how to convert a Caliban GraphQL datatypes and turn them into Tapir endpoints.
+This is useful when you have both REST APIs and GraphQL endpoints and you want to use Tapir to expose a unified API.
+
+The example also shows how to mount the GraphQL endpoint to `POST` and `GET` `/graphql` as well as provide `/graphiql` for the GraphiQL UI.
```
ghostdogpr: You can remove the part about Graphiql since it was removed

ghostdogpr (approved):

examples/src/main/scala/example/calibantotapir/README.md
calvinlfer: oops

ghostdogpr:
Thanks!

PR #2379, opened 2024-08-27
Graphiql endpoints for tapir adapters

### LINKED ISSUE #2376: Support creating graphiql endpoints for tapir based adapters
Currently only the quick adapter provides the graphiql endpoint but the other (tapir based) adapters are left to the user to implement. 

It would be nice to provide graphiql endpoints to more adapters

kyri-petrou:
/closes #2376

@calvinlfer was there something in the examples you wanted to add with graphiql?


interop/tapir/src/main/scala/caliban/interop/tapir/HttpInterpreter.scala
```diff
@@ -214,4 +214,24 @@ object HttpInterpreter {
...
+    val apiPath0 = apiPath.split("/").filter(_.nonEmpty).mkString("/", "/", "")
+    infallibleEndpoint.get
+      .in(extractFromRequest(_.pathSegments))
+      .out(htmlBodyUtf8)
+      .mapIn { segments =>
+        val uiPath = segments.mkString("/", "/", "")
+        Right(HttpUtils.graphiqlHtml(apiPath = apiPath0, uiPath = uiPath))
+      }(
+        _.value.split("/").filter(_.nonEmpty).toList
+      )
```
kyri-petrou: Does anyone think of a better way to do this? It works but I'm not too happy with this

ghostdogpr (approved):

interop/tapir/src/main/scala/caliban/interop/tapir/HttpInterpreter.scala
ghostdogpr: When is this even used?

paulpdaniels: I mean a less clean way but more efficient way would be to combine the steps into a single operation. But if it's not a hot path I wouldn't worry about it

calvinlfer: @ghostdogpr this would be used for the caliban-to-tapir adapter use case 

im actually curious @kyri-petrou, would it be easier to expose the tapir data type itself (`EndpointInput[Unit]`) so you didn’t have to rebuild the path segment itself? Or is that exposing too much tapir details?

kyri-petrou: I ended up reimplementing the endpoint properly, requiring an implicit MonadError (from sttp) to create it. It should be possible now to customize the endpoint using the `.endpoint` method on the `ServerEndpoint.Full`

ghostdogpr (approved):

PR #2383, opened 2024-08-31
Update ZIO Http

kyri-petrou:
Due to the removal of deprecated methods in ZIO Http, we have to remove the deprecated methods. The issue now is that if we want to strictly follow the versioning scheme we have to bump the version to 2.9.0.

Note that users can't update to ZIO Http RC10 until we release a new version


ghostdogpr:
There is already a bunch of pending breaking change PRs so let's make 2.9 the next one.

paulpdaniels:
Should we wait for next week? They were saying on the channel that 3.0 is targeted for 9/6 barring any large issues

ghostdogpr:
> Should we wait for next week? They were saying on the channel that 3.0 is targeted for 9/6 barring any large issues 

Oh nice, finally! Yeah I think we're not in a hurry so if 3.0 is soon let's wait!

kyri-petrou:
Updated version to 3.0.0

paulpdaniels (approved):

PR #2389, opened 2024-09-05
Added test for render validation of PostService

develeon:
Added basic test for render validation of PostService.

Proposing to add tests for render validation of compiletime generated services.

This will allow us to verify and better track behavior of new implementations. when using code first approach for GraphQl.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -290,6 +290,8 @@ lazy val codegenSbt = project
           "-Xmx1024M",
           "-Xss4M",
           "-Dplugin.version=" + version.value,
+          "-Dzio.test.version=" + ScriptedDependency.Version.zioTest,
+          "-Dsttp.version=" + ScriptedDependency.Version.sttp,
           s"-Dproject.dir=${baseDirectory.value.getAbsolutePath}"
         )
     },
--- a/build.sbt
+++ b/build.sbt
@@ -1,6 +1,8 @@
 import com.typesafe.tools.mima.core.*
 import org.scalajs.linker.interface.ModuleSplitStyle
 import sbtcrossproject.CrossPlugin.autoImport.{ crossProject, CrossType }
+import sbt.*
+import Keys.*
 
 val scala212 = "2.12.20"
 val scala213 = "2.13.16"
@@ -216,7 +218,24 @@ lazy val tools = project
       "dev.zio"                       %% "zio-test"            % zioVersion     % Test,
       "dev.zio"                       %% "zio-test-sbt"        % zioVersion     % Test,
       "dev.zio"                       %% "zio-json"            % zioJsonVersion % Test
-    )
+    ),
+    Test / publishArtifact := true,
+
+    // Include test artifact for publishLocal
+    publishLocalConfiguration := {
+      val config        = publishLocalConfiguration.value
+      val testArtifacts = (Test / packagedArtifacts).value
+      config.withArtifacts(config.artifacts ++ testArtifacts).withOverwrite(true)
+    },
+    // Exclude test artifact from publish
+    publishConfiguration      := {
+      val config = publishConfiguration.value
+      config
+        .withArtifacts(config.artifacts.filterNot { case (artifact, _) =>
+          artifact.configurations.exists(_.name == "test")
+        })
+        .withOverwrite(true)
+    }
   )
   .dependsOn(core, clientJVM, quickAdapter % Test)
 
@@ -267,18 +286,25 @@ lazy val codegenSbt = project
   .settings(
     scriptedLaunchOpts   := {
       scriptedLaunchOpts.value ++
-        Seq("-Xmx1024M", "-Xss4M", "-Dplugin.version=" + version.value)
+        Seq(
+          "-Xmx1024M",
+          "-Xss4M",
+          "-Dplugin.version=" + version.value,
+          s"-Dproject.dir=${baseDirectory.value.getAbsolutePath}"
+        )
     },
     scriptedBufferLog    := false,
-    scriptedDependencies := {
-      (macros / publishLocal).value
-      (core / publishLocal).value
-      (clientJVM / publishLocal).value
-      (tools / publishLocal).value
-      publishLocal.value
-    }
+    scriptedDependencies := scriptedDependencies
+      .dependsOn(
+        macros / publishLocal,
+        core / publishLocal,
+        clientJVM / publishLocal,
+        tools / publishLocal,
+        publishLocal
+      )
+      .value
   )
-  .dependsOn(tools)
+  .dependsOn(tools % "compile->compile;test->test")
 
 lazy val catsInterop = project
   .in(file("interop/cats"))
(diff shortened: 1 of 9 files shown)


ghostdogpr:
How about using zio-test? It's better to stick with one library.

tools/src/test/scala/caliban/tools/SnapshotTest.scala
```diff
@@ -16,51 +18,64 @@ trait SnapshotTest extends ZIOSpecDefault {
...
+    }
+
 }
 
 object SnapshotTest {
   val `.git`: Path = Path.of(".git")
   val cwd: Path    = Path.of(sys.props("user.dir"))
 
-  val projectRoot: Path = {
+  lazy val projectRoot: Path = {
```
develeon, line 78: For scripted test this will fail so better not evaluate it.

tools/src/test/scala/caliban/tools/SnapshotTest.scala
```diff
@@ -16,51 +18,64 @@ trait SnapshotTest extends ZIOSpecDefault {
...
-      str.map { str =>
-        val isCi = SnapshotTest.isCi
+      str.map { content =>
         val path = SnapshotTest.projectRoot.resolve(s"tools/src/test/resources/snapshots/$testName/${label + ".scala"}")
+        writeAndCompare(path, content, label)
+      }
+    }
+  }
 
-        def write(): TestResult = {
```
develeon, line 23: Moving helper function out for reuse

tools/src/test/scala/caliban/tools/SnapshotTest.scala
```diff
@@ -16,51 +18,64 @@ trait SnapshotTest extends ZIOSpecDefault {
...
-            assertTrue(false).label(
-              s"Could not read snapshot file '$path'. Rerun with environment `CI` not set to 'true' to create and then check in the file"
-            )
-          case Failure(_)                =>
-            write()
+    var exitCode = 0
+    // at least don't take the git lock multiple times from same process. this can still fail if concurrent processes try to take it.
+    GitLock.synchronized {
+      // allow failing external command, but complain to stderr
+      exitCode =
```
develeon, line 37: execution can fail with exitCode so we need to check it as it won`t be throwing exception

tools/src/test/scala/caliban/tools/SnapshotTest.scala
```diff
@@ -16,51 +18,64 @@ trait SnapshotTest extends ZIOSpecDefault {
...
+  def writeAndCompare(path: Path, content: String, label: String): TestResult =
+    Try(Files.readString(path)) match {
+      case Success(existing) if SnapshotTest.isCi        =>
+        assertTrue(content == existing).label(
+          s"generated result for test '$label' did not match snapshot contents in file '$path. Rerun with environment `CI` not set to 'true' to update and then check in the file"
+        )
+      case Success(existing) if existing.equals(content) =>
+        assert(())(Assertion.anything)
+      case Success(_)                                    =>
+        write(path, content)
```
develeon, line 63: Only if it changed we attempt to update it and add it to git.
In case of scripted test, it will be updated, but test will fail on adding to git. (running from /tmp ) - as warning, 
but snapshot will be updated for second run.

build.sbt
```diff
@@ -216,7 +216,8 @@ lazy val tools = project
       "dev.zio"                       %% "zio-test"            % zioVersion     % Test,
       "dev.zio"                       %% "zio-test-sbt"        % zioVersion     % Test,
       "dev.zio"                       %% "zio-json"            % zioJsonVersion % Test
-    )
+    ),
+    Test / publishArtifact := true
```
develeon: For local builds we need to provide test classes of project tools (SnapshotTest)

build.sbt
```diff
@@ -267,18 +268,25 @@ lazy val codegenSbt = project
   .settings(
     scriptedLaunchOpts   := {
       scriptedLaunchOpts.value ++
-        Seq("-Xmx1024M", "-Xss4M", "-Dplugin.version=" + version.value)
+        Seq(
+          "-Xmx1024M",
+          "-Xss4M",
+          "-Dplugin.version=" + version.value,
+          s"-Dproject.dir=${baseDirectory.value.getAbsolutePath}"
```
develeon, line 295: Since sbt-test run from /tmp we add project.dir property to be able to resolve and update graphql snapshot file.

build.sbt
ghostdogpr: It's not going to try to publish something to Maven when we do a new release?

ghostdogpr (approved):

build.sbt
```diff
@@ -216,7 +218,24 @@ lazy val tools = project
...
+    },
+    // Exclude test artifact from publish
+    publishConfiguration      := {
+      val config = publishConfiguration.value
+      config
+        .withArtifacts(config.artifacts.filterNot { case (artifact, _) =>
+          artifact.configurations.exists(_.name == "test")
+        })
+        .withOverwrite(true)
+    }
```
develeon, line 238: This should prevent test compile classes from being included on publish, but allow for it for publishLocal.

ghostdogpr (approved):
LGTM, @kyri-petrou you wanna check it?

codegen-sbt/src/sbt-test/compiletime-codegen/test-compile/README.md
```diff
@@ -1,3 +1,14 @@
...
\ No newline at end of file
+This test project has been copied from:
+https://github.com/guizmaii/poc_compile_time_caliban_client_generation
+
+### Running locally
+You can run these tests using following sbt commandos:
+
+```sbt
+project codegenSbt
+++2.12.20
```
kyri-petrou: Perhaps change this to `+2.12` instead so that we don't have to update it on scala version updates

codegen-sbt/src/sbt-test/compiletime-codegen/test-compile/build.sbt
```diff
@@ -26,18 +26,20 @@ ThisBuild / crossScalaVersions := allScala
...
+)
 
 lazy val sttp = Seq(
   "com.softwaremill.sttp.client3" %% "core" % "3.9.8",
   "com.softwaremill.sttp.client3" %% "zio"  % "3.9.8"
 )
 
+lazy val zioTest = Seq(
+  "dev.zio" %% "zio-test" % "2.1.9" % Test,
+  "dev.zio" %% "zio-test-sbt" % "2.1.9" % Test
```
kyri-petrou: I don't think Scala Steward will update this. Any chance that we can place versions somewhere where we can reuse them? Or simpler (less changes) perhaps include them in the BuildInfo object for the publishLocal configuration?

Also, I don't think these lines are correctly formatted

kyri-petrou (commented):
Sorry for the delay in reviewing, I've been quite busy the past week

codegen-sbt/src/sbt-test/compiletime-codegen/test-compile/build.sbt
develeon: Updated formatting + README.

Did not see any .scala-steward.conf, is it part of github setup?
Could it be better to append sub projects to .scala-steward`s buildRoots?

ghostdogpr:
Sorry I forgot about this.
Could you rebase and pull the latest tags to your fork? It makes mima checks fail.

PR #2390, opened 2024-09-05
Interface error when combined with union types

develeon:
Adds validation check for logrocket service exampled derived from
* [Handling GraphQL errors like a champ with unions and interfaces](https://blog.logrocket.com/handling-graphql-errors-like-a-champ-with-unions-and-interfaces/ )

Seems interface are not generated when combined with union types.
image


develeon:
Do not thing need this, the provided workaround in FAQ solves this sufficiently

PR #2408, opened 2024-09-17
Fix ArgBuilder derivation for case classes containing only optional fields

### LINKED ISSUE #2407: Incorrect behaviour of `ArgBuilder` containing only nullable fields when input is null
Given the following case class:

```scala
case class Foo(value1: Option[String], value2: Option[String]) derives ArgBuilder
```

I expect that the derived `ArgBuilder#build` should work in the following ways:
1. Input: `{}`, output: `Right(Foo(None, None))`
2. Input: `null`, output `Left(ExecutionError("bla"))`

However, currently both cases succeed with `Foo(None, None)`. Can be reproduced with the following test that should pass but fails since the first condition is not `Left`

```scala
    suite("derived build")(
      test("should fail when null is provided for case class with optional fields") {
        case class Foo(value: Option[String])
        val ab = ArgBuilder.gen[Foo]
        assertTrue(
          ab.build(NullValue).isLeft,
          ab.build(ObjectValue(Map())).isRight
        )
      }
    )
```

PS: This is not a problem when validation is enabled as an error will be raised during the validation phase

kyri-petrou:
/fixes #2407

Issue was that we were treating all case classes as potential value types. Fix is to determine during compile-time whether the case class is a value type (either using annotations or checking if the case class extends `AnyVal`)

@guymers I also added the tests that were in #2403 to ensure that `oneOf` input derivation is working as expected now.

### DIFF
--- a/core/src/main/scala-2/caliban/schema/ArgBuilderDerivation.scala
+++ b/core/src/main/scala-2/caliban/schema/ArgBuilderDerivation.scala
@@ -3,7 +3,7 @@ package caliban.schema
 import caliban.CalibanError.ExecutionError
 import caliban.InputValue
 import caliban.Value._
-import caliban.schema.Annotations.{ GQLDefault, GQLName, GQLOneOfInput }
+import caliban.schema.Annotations.{ GQLDefault, GQLName, GQLOneOfInput, GQLValueType }
 import magnolia1._
 
 import scala.collection.compat._
@@ -27,7 +27,6 @@ trait CommonArgBuilderDerivation {
   }
 
   def join[T](ctx: CaseClass[ArgBuilder, T]): ArgBuilder[T] = new ArgBuilder[T] {
-
     private val params = {
       val arr = Array.ofDim[(String, EitherExecutionError[Any])](ctx.parameters.length)
       ctx.parameters.zipWithIndex.foreach { case (p, i) =>
@@ -40,6 +39,9 @@ trait CommonArgBuilderDerivation {
 
     private val required = params.collect { case (label, default) if default.isLeft => label }
 
+    private val isValueType =
+      ctx.isValueClass || (ctx.annotations.exists(_.isInstanceOf[GQLValueType]) && params.length == 1)
+
     override private[schema] val partial: PartialFunction[InputValue, Either[ExecutionError, T]] = {
       case InputValue.ObjectValue(fields) if required.forall(fields.contains) => fromFields(fields)
     }
@@ -47,7 +49,8 @@ trait CommonArgBuilderDerivation {
     def build(input: InputValue): Either[ExecutionError, T] =
       input match {
         case InputValue.ObjectValue(fields) => fromFields(fields)
-        case value                          => ctx.constructMonadic(p => p.typeclass.build(value))
+        case value if isValueType           => ctx.constructMonadic(p => p.typeclass.build(value))
+        case _                              => Left(ExecutionError("expected an input object"))
       }
 
     private[this] def fromFields(fields: Map[String, InputValue]): Either[ExecutionError, T] =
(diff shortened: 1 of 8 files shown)


core/src/main/scala-2/caliban/schema/ArgBuilderDerivation.scala
```diff
@@ -40,14 +39,18 @@ trait CommonArgBuilderDerivation {
...
+
     override private[schema] val partial: PartialFunction[InputValue, Either[ExecutionError, T]] = {
       case InputValue.ObjectValue(fields) if required.forall(fields.contains) => fromFields(fields)
     }
 
     def build(input: InputValue): Either[ExecutionError, T] =
       input match {
         case InputValue.ObjectValue(fields) => fromFields(fields)
-        case value                          => ctx.constructMonadic(p => p.typeclass.build(value))
+        case value if isValueType           => ctx.constructMonadic(p => p.typeclass.build(value))
```
kyri-petrou: @ghostdogpr do you recon we should change the line above to `case InputValue.ObjectValue(fields) if !isValueType` as well? Or are we risking to create issues with edge cases?

core/src/main/scala-2/caliban/schema/ArgBuilderDerivation.scala
```diff
@@ -40,14 +39,18 @@ trait CommonArgBuilderDerivation {
...
+        case _                              => Left(ExecutionError("expected an input object"))
```
ghostdogpr:
```suggestion
        case _                              => Left(ExecutionError("Expected an input object"))
```

core/src/main/scala-2/caliban/schema/ArgBuilderDerivation.scala
ghostdogpr: Do you mean the line below? I think it's okay to fail 🤔

core/src/main/scala-3/caliban/schema/ArgBuilderDerivation.scala
```diff
@@ -143,7 +153,8 @@ trait CommonArgBuilderDerivation {
     def build(input: InputValue): Either[ExecutionError, A] =
       input match {
         case InputValue.ObjectValue(fields) => fromFields(fields)
-        case value                          => fromValue(value)
+        case value if isValueType           => fromValue(value)
+        case _                              => Left(ExecutionError("expected an input object"))
```
ghostdogpr:
```suggestion
        case _                              => Left(ExecutionError("Expected an input object"))
```

core/src/main/scala-2/caliban/schema/ArgBuilderDerivation.scala
kyri-petrou: No I mean the line above (i.e., `case InputValue.ObjectValue(fields) => fromFields(fields)`). I'm wondering whether it should be changed to:

```scala
case InputValue.ObjectValue(fields) if !isValueType => fromFields(fields)
```

A case where this can be "problematic" is this:

```scala
        @GQLValueType
        case class Foo(value: Option[String])

        val ab = ArgBuilder.gen[Foo]
        assertTrue(
          ab.build(ObjectValue(Map())).isLeft
        )
```
Currently this test fails because `Map()` is a valid value if `Foo` wasn't a value type, but since it is a value type it should fail as `{}` is not a valid input for `Option[String]`

ghostdogpr: Yeah I think it makes sense to change as you suggested.

kyri-petrou:
@ghostdogpr I noticed that we had some very similar logic for determining value types in SchemaDerivation. In order to avoid conflicts, I reused the same logic for both deriving schemas & arg builders

ghostdogpr (approved):

PR #2426, opened 2024-10-04
SchemaComparator support for type extensions

### LINKED ISSUE #2417: SchemaComparison ignores TypeExtension:s
`caliban.tools.SchemaComparison.compare` is a very useful tool for writing unit tests for code that produces and transforms schemas. However, it is currently limited by not checking extension definitions. For example:
```scala
  def run =
    val first = SchemaLoader.fromString( "extend type Query {a: Int}")
    val second = SchemaLoader.fromString("extend type Query {b: Int}")
    for
      diff <-
        SchemaComparison
          .compare(first, second)
          .mapError(SchemaComparisonErrorException.apply)
      _ <- ZIO.log(s"Expected non-empty diff: $diff")
    yield ()
```

Diff is empty but would expect two diffs on the `Query` type: 
*  field `b` has been added
*  field `a` has been deleted.

joheriks:
Compare type extensions. Some refactoring of function parameters (passing `List[T]` instead of `Map[String, T]`) to reduce duplicate code.
Fixes #2417.

### DIFF
--- a/tools/src/main/scala/caliban/tools/SchemaComparison.scala
+++ b/tools/src/main/scala/caliban/tools/SchemaComparison.scala
@@ -1,10 +1,14 @@
 package caliban.tools
 
 import caliban.parsing.adt.Definition.TypeSystemDefinition.TypeDefinition._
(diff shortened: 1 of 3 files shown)


ghostdogpr (approved):
Thanks!

PR #2428, opened 2024-10-04
Added test for fragment on root query

develeon:
We had some problem with fragment in root query in our environment where fragments where not included in output.
So added this minor test for it, but it worked fine. So decided to add it.

What do you think, could it be in that in a more complex scenario with ZQuery this does not work as expected.


ghostdogpr (approved):

ghostdogpr:
Hmm it's hard to tell 🤔 Could you try to minimize your use case?

develeon:
> Hmm it's hard to tell 🤔 Could you try to minimize your use case?

Will try to setup a simple but more elaborated test see if can figure it out.

PR #2436, opened 2024-10-11
Make jsoniter codecs public

ghostdogpr:
Requested on Discord.


paulpdaniels:
Was just thinking what if instead of exposing this, we wrapped it in a `Renderer` instead?

PR #2446, opened 2024-10-21
add support for federation 2.9 and 2.10

paulpdaniels:
Federation 2.9 and 2.10 have been released and include a set of [new directives](https://www.apollographql.com/docs/graphos/reference/federation/directives).

Adding support here, though I can't test them because they require enterprise subscriptions 🤡 .

### DIFF
--- a/federation/src/main/scala/caliban/federation/package.scala
+++ b/federation/src/main/scala/caliban/federation/package.scala
@@ -1,24 +1,28 @@
 package caliban
 
 import caliban.federation.v2x.{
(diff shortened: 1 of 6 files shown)


ghostdogpr (approved):
Is the Federation Specification Compatibility Test testing that part?

paulpdaniels:
@ghostdogpr it's testing for compatibility violations, but I think it only goes up to v2.3 or 2.4. Everything after that point is basically gated behind the enterprise wall, and they don't offer library implementers a way to test it.

Issue #2458, opened 2024-11-06
Invalid value too long string exceeded 'maxCharBufSize'


kaaveland:
I should probably be able to figure this one out on my own, but I haven't been able to. I'm getting this error for very long strings in graphql requests to my graphql api -- it's simply being returned as a 400 Bad Request. I would like to adjust jsoniters `maxCharBufSize`, which as far as I can tell, can be set in `ReaderConfig`:

https://github.com/plokhotnyuk/jsoniter-scala/blob/master/jsoniter-scala-core/shared/src/main/scala/com/github/plokhotnyuk/jsoniter_scala/core/ReaderConfig.scala#L38

I could also provide by own instance to `JsonReader`: 

https://github.com/plokhotnyuk/jsoniter-scala/blob/master/jsoniter-scala-core/js/src/main/scala/com/github/plokhotnyuk/jsoniter_scala/core/JsonReader.scala#L36

Alternatively, perhaps set up caliban to use a different json library, although https://github.com/ghostdogpr/caliban/pull/2341 makes it seem like the project wants to avoid this in the future. 

Is there an easy way to convince caliban to accept strings that are a little longer?

ghostdogpr:
Ah actually we did this for Caliban Client: https://github.com/ghostdogpr/caliban/pull/2085
I think we could do the same on the server side too. Wanna make a PR?

kaaveland:
I'll take a stab at this 👍 

What's the right place to start, should I just start to looking for usages of `readFromString` in `core`?

ghostdogpr:
So for the quick adapter that would be in `QuickRequestHandler` (look for `readFromString` and `readFromArray`). For the tapir-based adapters it's trickier because the decoding is done by Tapir, and I am not sure if they let you configure it. It requires some research (I see that when we import `sttp.tapir.json.jsoniter._`, it calls `TapirJsonJsoniter` so maybe we have to make a custom version of that).

kaaveland:
It looks like doing something like this is fairly simple for the tapir adapters:
```scala

object MaxCharBufSizeJsonJsoniter extends TapirJsonJsoniter {
  override lazy val readerConfig: ReaderConfig =
    ReaderConfig
      .withAppendHexDumpToParseException(false)
      .withMaxCharBufSize(2147483645)
      .withMaxBufSize(2147483645)
}

```

[...]

PR #2459, opened 2024-11-07
Set the max buffersize for jsoniter

kaaveland:
This is a first take on https://github.com/ghostdogpr/caliban/issues/2458

Let me know if there's something missing, or if I should try another way to do this.

### DIFF
--- a/adapters/quick/src/main/scala/caliban/QuickRequestHandler.scala
+++ b/adapters/quick/src/main/scala/caliban/QuickRequestHandler.scala
@@ -75,7 +75,8 @@ final private class QuickRequestHandler[R](
 
     def decodeQueryParams(queryParams: QueryParams): Either[Response, GraphQLRequest] = {
       def extractField(key: String) =
-        try Right(queryParams.getAll(key).headOption.map(readFromString[InputValue.ObjectValue](_).fields))
+        try
+          Right(queryParams.getAll(key).headOption.map(readFromString[InputValue.ObjectValue](_, readerConfig).fields))
         catch { case NonFatal(_) => Left(badRequest(s"Invalid $key query param")) }
 
       for {
@@ -101,7 +102,7 @@ final private class QuickRequestHandler[R](
         body.asArray.foldZIO(
           _ => Exit.fail(BodyDecodeErrorResponse),
           arr =>
-            try checkNonEmptyRequest(readFromArray[GraphQLRequest](arr))
+            try checkNonEmptyRequest(readFromArray[GraphQLRequest](arr, readerConfig))
             catch { case NonFatal(_) => Exit.fail(BodyDecodeErrorResponse) }
         )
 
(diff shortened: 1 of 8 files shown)


adapters/quick/src/main/scala/caliban/QuickRequestHandler.scala
```diff
@@ -25,6 +25,10 @@ final private class QuickRequestHandler[R](
   import QuickRequestHandler._
   import ValueJsoniter.stringListCodec
 
+  private val readerConfig: ReaderConfig = ReaderConfig
+    .withMaxBufSize(2147483645)
```
ghostdogpr: Any reason to not use `Int.MaxValue`?

kyri-petrou: From memory I think the max that it can be set to is `Int.MaxValue - 2`. Maybe we can use that instead though

adapters/quick/src/main/scala/caliban/QuickRequestHandler.scala
```diff
@@ -25,6 +25,10 @@ final private class QuickRequestHandler[R](
   import QuickRequestHandler._
   import ValueJsoniter.stringListCodec
 
+  private val readerConfig: ReaderConfig = ReaderConfig
```
kyri-petrou: Very minor, but maybe place this in the companion object instead of the class

adapters/quick/src/main/scala/caliban/QuickRequestHandler.scala
```diff
@@ -25,6 +25,10 @@ final private class QuickRequestHandler[R](
...
+    .withMaxBufSize(2147483645)
+    .withMaxCharBufSize(2147483645)
```
kyri-petrou: Probably good to also add `.withAppendHexDumpToParseException(false)` (for consistency and UX)

kyri-petrou (commented):
Could we also add a test case in `TapirAdapterSpec` that would fail using the pre-PR code? That way we can ensure that the implicits are all properly used across all adapters.

adapters/quick/src/main/scala/caliban/QuickRequestHandler.scala
kaaveland: It's `Int.MaxValue - 2`, yeah. We can use that, if you'd prefer.

ghostdogpr: Yeah I think that's a bit better

kaaveland:
I added a test case, but it currently also fails on this branch -- we're getting `413 Request Entity Too Large` from `PlayAdapterSpec` and `QuickAdapterSpec`. I'm working on figuring out how to make these two servers accept larger payloads. I did verify that it reproduces the error on `series/2.x`.

kaaveland:
@ghostdogpr want to take a new look? There's a test case that fails on main and works on the branch now. I tried to avoid using `application.conf` for the `PlayAdapterSpec`, but I was unable to make the settings stick by passing them as properties. 🤔

ghostdogpr (approved):
Thanks!

PR #2486, opened 2024-12-16
[Bug] Nested fragments don't work as expected

lervag:
Fragment validation fails for the following nested fragment with the `ValidationError` `"foo has conflicting types: Baz.foo and Query.foo. Try using an alias."`. Notice that `Query.foo` is not part of the actual query.

```graphql
query {
  bar {
    ...Outer
  }
}

fragment Outer on Bar {
  baz {
    foo {
      id
    }
    ...Inner
  }
}

fragment Inner on Baz {
  foo {
    id
  }
}
```

The underlying schema is:

```graphql
case class Query(
  bar: Option[Bar],
  foo: List[Foo]
)

case class Foo(
  id: String
)
case class Bar(
  id: String,
  baz: Option[Baz]
)

case class Baz(
  id: String,
  foo: Option[Foo]
)
```

Notice that there is _no_ error if either

1. we remove `Query.foo`, or
2. we remove the `foo { id }` from the `Outer` fragment.

We (me and @develeon) believe this is a bug with Caliban and specifically within [FragmentValidator.scala](https://github.com/lervag/caliban/blob/series/2.x/core/src/main/scala/caliban/validation/FragmentValidator.scala), but we have not been able to figure out exactly where it goes wrong. We notice that @frekw have worked on this file, so he is probably relevant for this discussion.

This PR includes a test for this bug that will fail. Do you want us to create an issue as well?

### DIFF
--- a/core/src/test/scala/caliban/FragmentSchema.scala
+++ b/core/src/test/scala/caliban/FragmentSchema.scala
@@ -1,0 +1,32 @@
+package caliban
+
+import caliban.schema.Annotations._
+
+object FragmentSchema {
+
+  @GQLDescription("Queries")
+  case class Query(
+    bar: Option[Bar],
+    foo: List[Foo]
+  )
+
+  case class Foo(
+    id: String
+  )
+  case class Bar(
+    id: String,
+    baz: Option[Baz]
+  )
+
+  case class Baz(
+    id: String,
+    foo: Option[Foo]
+  )
+
+  val resolverFooBar: RootResolver[Query, Unit, Unit] = RootResolver(
+    Query(
+      bar = Some(Bar("1", Some(Baz("2", Some(Foo("3")))))),
+      foo = List(Foo("1"), Foo("2"))
+    )
+  )
+}
--- a/core/src/test/scala/caliban/execution/FragmentSpec.scala
+++ b/core/src/test/scala/caliban/execution/FragmentSpec.scala
@@ -50,6 +50,35 @@ object FragmentSpec extends ZIOSpecDefault {
           res <- int.execute(query)
         } yield assertTrue(res.data.toString == """{"amos":{"name":"Amos Burton"}}""")
       },
+      test("fragment with inner and outer") {
+        val interpreter = graphQL(caliban.FragmentSchema.resolverFooBar).interpreter
+        val query       = gqldoc("""
+           query {
+             bar {
+               ...Outer
+             }
+           }
+
+           fragment Outer on Bar {
+             baz {
+               foo {
+                id
+               }
+               ...Inner
+             }
+           }
+
+           fragment Inner on Baz {
+             foo {
+               id
+            }
+           }
+           """)
+        for {
+          int <- interpreter
+          res <- int.execute(query)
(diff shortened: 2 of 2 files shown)


kyri-petrou:
AFAICT the fact that we're raising an error is correct, but the error message is wrong. After inlining of the fragments, your query becomes the following:

```gql
query {
  bar {
    baz {
      foo {
        id
      }
      foo { # <--- This field is queried twice without an alias
        id
      }
    }
  }
}
```

alf:
It is perfectly legal to fetch the same field multiple times in a GraphQL Query, so your example should work and not cause an error.

alf:
See https://spec.graphql.org/October2021/#sec-Field-Selection-Merging for details.

core/src/test/scala/caliban/FragmentSchema.scala
```diff
@@ -0,0 +1,32 @@
...
+    id: String,
+    baz: Option[Baz]
+  )
+
+  case class Baz(
+    id: String,
+    foo: Option[Foo]
+  )
+
+  val resovlerFooBar: RootResolver[Query, Unit, Unit] = RootResolver(
```
alf:
```suggestion
  val resolverFooBar: RootResolver[Query, Unit, Unit] = RootResolver(
```

core/src/test/scala/caliban/execution/FragmentSpec.scala
```diff
@@ -50,6 +50,35 @@ object FragmentSpec extends ZIOSpecDefault {
           res <- int.execute(query)
         } yield assertTrue(res.data.toString == """{"amos":{"name":"Amos Burton"}}""")
       },
+      test("fragment with inner and outer") {
+        val interpreter = graphQL(caliban.FragmentSchema.resovlerFooBar).interpreter
```
alf:
```suggestion
        val interpreter = graphQL(caliban.FragmentSchema.resolverFooBar).interpreter
```

lervag:
> After inlining of the fragments, your query becomes the following:
> 
> ```graphql
> query {
>   bar {
>     baz {
>       foo {
>         id
>       }
>       foo { # <--- This field is queried twice without an alias
>         id
>       }
>     }
>   }
> }
> ```

I think @alf is right; there is nothing wrong with querying the same field twice. And if we pass that inlined query to Caliban it will return the correct result.

alf (commented):
Typos in tests are annoying.

ghostdogpr:
Could you confirm if this is only a validation issue? If you disable validation with `Configurator.setSkipValidation`, does it return the proper result? (probably yes since you said it works with the inlined query)

lervag:
> Could you confirm if this is only a validation issue? If you disable validation with `Configurator.setSkipValidation`, does it return the proper result? (probably yes since you said it works with the inlined query)

Yes, I just checked this on our project and can confirm that it seems to be only a validation issue.

ghostdogpr:
Opened https://github.com/ghostdogpr/caliban/pull/2491 to fix this

lervag:
Thanks! It's been holiday season, so I've been offline for a while. But I'll test the latest updates and let you know if something doesn't work as expected now.

PR #2488, opened 2024-12-18
Enhance federated rendering

paulpdaniels:
Adding some utilities to rendering to make it easier when rendering federated schemas for schema checks


ghostdogpr:
There are some weird compile errors 🤔

paulpdaniels:
Hmm strange too the 2.13 build was working for me locally. Let me try a clean and try again on my machine

paulpdaniels:
@ghostdogpr Ok compile errors finally addressed 😅

federation/src/main/scala/caliban/federation/FederationDirectivesV1.scala
```diff
@@ -0,0 +1,15 @@
...
+
+trait FederationDirectivesV1 {
+
+  case class GQLKey(fields: String) extends GQLDirective(Key(fields))
+
+  object Key {
+    def apply(fields: String, name: String = "key"): Directive =
+      Directive(name, Map("fields" -> StringValue(fields)))
+  }
+}
```
paulpdaniels, line 15: v1 `@key` does not allow the `resolvable` argument, which could be a source of confusion to users. Splitting this so that using v1 no longer gives that option.

core/src/main/scala/caliban/GraphQL.scala
```diff
@@ -37,7 +37,17 @@ trait GraphQL[-R] { self =>
   /**
    * Returns a string that renders the API types into the GraphQL SDL.
    */
-  final def render: String = DocumentRenderer.render(toDocument)
+  final def render: String =
+    renderWith(DocumentRenderer)
+
+  final def renderCompact: String =
```
ghostdogpr, line 50: Can you add scaladoc for this one too?

ghostdogpr (approved):

core/src/main/scala/caliban/GraphQL.scala
paulpdaniels: Done

PR #2491, opened 2024-12-20
Fix fragment validation when a field is both in a fragment and outside a fragment

ghostdogpr:
Fixes https://github.com/ghostdogpr/caliban/pull/2486

`parentType` was always equal to the top parent, even in nested calls, and it was being compared with the fragment's type, which caused a mismatch. I used `f1` because at that point we know that f1 and f2 are compatible.

### DIFF
--- a/core/src/test/scala/caliban/execution/FragmentSpec.scala
+++ b/core/src/test/scala/caliban/execution/FragmentSpec.scala
@@ -50,6 +50,33 @@ object FragmentSpec extends ZIOSpecDefault {
           res <- int.execute(query)
         } yield assertTrue(res.data.toString == """{"amos":{"name":"Amos Burton"}}""")
       },
+      test("fragment with inner and outer") {
+        val interpreter = graphQL(caliban.FragmentSchema.resolverFooBar).interpreter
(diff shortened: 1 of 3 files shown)


core/src/test/scala/caliban/execution/FragmentSpec.scala
```diff
@@ -50,6 +51,35 @@ object FragmentSpec extends ZIOSpecDefault {
...
+               foo {
+                id
+               }
+               ...Inner
+             }
+           }
+           fragment Inner on Baz {
+             foo {
+               id
+            }
```
paulpdaniels, line 75: Do we have tests for the failing case as well? Where the fragments don’t match exactly?

paulpdaniels (approved):

core/src/test/scala/caliban/execution/FragmentSpec.scala
ghostdogpr: Yep, we have 2: "inline fragment selection with different field types" and "conflicting different responses".

lervag:
Thanks! I can confirm that this fixes the issue on our side! 🚀

PR #2493, opened 2024-12-21
fix(codegen): abstract effect not propagated to union siblings

### LINKED ISSUE #2492: bug: codegen generates incorrect types for sealed trait with lazy directive and abstractEffectType enabled
```graphql
directive @lazy on FIELD_DEFINITION

schema {
  query: Query
}

type Query {
    u: U
}


type A {
  field: Int @lazy
}

type B {
  field: Int
}

union U = A | B


```

```scala
import Types._

object Types {

  final case class A[F[_]](field: F[scala.Option[Int]]) extends U
  final case class B(field: scala.Option[Int]) extends U

  sealed trait U extends scala.Product with scala.Serializable

}

object Operations {

  final case class Query[F[_]](
      u: F[scala.Option[U[F]]]
                     // ^ Types.U does not take type parameters
  )

}
```

This is probably because SchemaWriter does not properly handle union that contains types with `F[_]` here.

https://github.com/ghostdogpr/caliban/blob/636b2bd1884e9a209ca12ce0bff88f4b0c98df49/tools/src/main/scala/caliban/tools/SchemaWriter.scala#L185

i10416:
fix https://github.com/ghostdogpr/caliban/issues/2492

### DIFF
--- a/tools/src/test/scala/caliban/tools/SchemaWriterSpec.scala
+++ b/tools/src/test/scala/caliban/tools/SchemaWriterSpec.scala
@@ -80,6 +80,32 @@ object SchemaWriterSpec extends SnapshotTest {
         isEffectTypeAbstract = true
       )
     ),
+    snapshotTest("union, abstracted effect type and lazy combination")(
+      gen(
+        """
+         type Query {
+           effectful: U0!
+           pure: U1!
+         }
+         union U0 = A | B | C
+         union U1 = C | D
+         type A {
+           lazyField: Int @lazy
+         }
+         type B {
+           lazyFieldWithArgs(int: Int): Int @lazy
+         }
+         type C {
+           field: Int
+         }
+         type D {
+          fiend: Int
+         }
+         """,
+        effect = "F",
+        isEffectTypeAbstract = true
+      )
+    ),
     snapshotTest("simple mutation with abstracted effect type")(
       gen(
         """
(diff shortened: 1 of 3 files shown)


i10416:
@ghostdogpr Hello, thank you for maintaining this project. Could you be kind to review this change?

tools/src/test/scala/caliban/tools/SchemaWriterSpec.scala
```diff
@@ -80,6 +80,32 @@ object SchemaWriterSpec extends SnapshotTest {
...
+           lazyField: Int @lazy
+         }
+         type B {
+           lazyFieldWithArgs(int: Int): Int @lazy
+         }
+         type C {
+           field: Int
+         }
+         type D {
+          fiend: Int
```
ghostdogpr: it's very minor but there's a small typo 😄

ghostdogpr (approved):
Looks good!

tools/src/test/scala/caliban/tools/SchemaWriterSpec.scala
i10416: Thanks. I've just pushed the fix.

ghostdogpr:
Thanks for fixing this! Not all edge cases and combinations have been tested. Will merge when the CI is green!

i10416:
ref:  https://github.com/ghostdogpr/caliban/pull/2064

PR #2495, opened 2024-12-21
fix(codegen): interface effect propagation

### LINKED ISSUE #2494: bug: codegen generates incorrect interface types when field is effectful and effect type is abstract
related to https://github.com/ghostdogpr/caliban/issues/2492

## Problem

```graphql
directive @lazy on FIELD_DEFINITION

type Query {
    main: Main!
}
interface I {
    fa: Int! @lazy
}

type Main implements I {
  fa: Int! @lazy
}
```

```scala
import Types._

import caliban.schema.Annotations._

object Types {

  final case class Main[F[_]](fa: F[Int]) extends I

  @GQLInterface
  sealed trait I extends scala.Product with scala.Serializable {
    def fa: F[Int] // Not found: type F
  }

}

object Operations {

  final case class Query[F[_]](
      main: F[Main[F]]
  )

}
```

## Expectation

Interface should track `F[_]` in  fields.
It means codegen should generate `sealed trait I[F[_]] extends ...` and `final case class Main[F[_]](fa: F[Int]) extends I[F]` in the example above.

## Note
There are two scenarios.

- field is marked by lazy
- field is transitively effectful

i10416:
fix https://github.com/ghostdogpr/caliban/issues/2494

### DIFF
--- a/core/src/main/scala/caliban/parsing/adt/Definition.scala
+++ b/core/src/main/scala/caliban/parsing/adt/Definition.scala
@@ -77,13 +77,17 @@ object Definition {
       }
     }
 
-    sealed trait TypeDefinition extends TypeSystemDefinition {
+    sealed trait TypeDefinition                             extends TypeSystemDefinition {
       def name: String
       def description: Option[String]
       def directives: List[Directive]
 
       final def isDeprecated: Boolean             = Directives.isDeprecated(directives)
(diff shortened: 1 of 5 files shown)


i10416:
@ghostdogpr This PR is ready for review now. Would you review this PR if you have time?

ghostdogpr (approved):
Great, thank you!

PR #2508, opened 2025-01-10
Fix encoding of json keys that contain utf-8 characters

valentin-panalyt:
As discussed in discord, this PR aims to fix the case where a json key contains UTF-8 characters, which are then not correctly encoded by jsoniter, leading to invalid json in the (http) response.


kyri-petrou (approved):
Bit sad we have to do this, but unless we can differentiate between `ObjectValue`s that originate from custom scalars to those that originate from the schema itself, I can't think of another way. While differentiating is possible, it will make the code more complex and I don't think we should do it simply for improving json encoding performance.

kyri-petrou:
@ghostdogpr happy to merge in if you're also happy with it

PR #2510, opened 2025-01-12
implement global identifier spec

### LINKED ISSUE #2507: Support for "global object identification" and the Node interface
I want to implement support for [global object identification](https://graphql.org/learn/global-object-identification/) in my service. In particular, I want to implement the `Node` interface and the `Query.node` field. The latter is non-trivial; we currently have an implementation that requires us to manually add each type that implements Node to a custom data structure that we use to resolve the `ID`. So, each type will `extend Node[TYPE]` and we add the `TYPE` to a slightly complex structure manually to handle the resolver.

I am curious if anyone has implemented this in a more general manner that is more straight forward to use?

The dream scenario would be to have something where we simply do `case class Foo(...) extends Node`. Or perhaps `case class Foo(...) extends Node[Foo]` where everything else required for `Query.node` is handled automatically.

paulpdaniels:
Closes #2507 

Attempts to implement the global identifier specification from Relay. Similar strategy to how federation is implemented, but the `Node` type has to be created as an `interface` instead of as a type union.


li-nkSN:
I wonder if this feature will work with server codegen. I am not sure if there would be any issue supporting that with the implementation.

core/src/test/scala/caliban/relay/GlobalIdentifierSpec.scala
```diff
@@ -0,0 +1,131 @@
...
+        ) == "schema{query:Query}interface Node{id:ID!}type Character implements Node{id:ID! name:String!}type Query{characters:[Character!]! ships:[Ship!]! node(id:ID!):Node}type Ship implements Node{id:ID! name:String! purpose:String!}"
+      )
+    },
+    test("resolve node field") {
+      val augmented = api @@
+        RelaySupport.withGlobalIdentifiers[ID](shipResolver, characterResolver)
+
+      val query =
+        """{
+          | character: node(id:"Character:2"){
```
li-nkSN, line 97: I'm not sure if it's part of the spec. But by at least convention I believe the ID is usually stored as a base64 encoded string. Do you plan to employ the convention  ??

core/src/main/scala/caliban/relay/TypeResolver.scala
```diff
@@ -0,0 +1,11 @@
+package caliban.relay
+
+import caliban.CalibanError.ExecutionError
+
+/**
+ * Used in GlobalIds to resolve the type of the returned object based on the ID.
+ * This is needed in order to correctly derive the subtype of the object.
+ */
+trait TypeResolver[ID] {
+  def resolve(a: ID): Either[ExecutionError, String]
```
li-nkSN, line 10: Should you provide a def extractLocalId as part of the TypeResolver trait ?

li-nkSN:
Will this be backported to Caliban 2.X ??

addl feedback:

Should this include

Documentation explaining the default behavior and how to override
Migration Guide for users who might have existing non-base64 IDs
Performance Consideration: Maybe cache decoded values for frequently accessed ID

core/src/test/scala/caliban/relay/GlobalIdentifierSpec.scala
paulpdaniels: It's left up to the consumer of this how they want to encode their ids, so the TypeResolver is flexible to allow users to implement it to their spec. Maybe we could include a helper that assumes that convention

li-nkSN:
> Will this be backported to Caliban 2.X ??


Testing against https://github.com/ghostdogpr/caliban/pull/2696

ghostdogpr:
@paulpdaniels i rebased and fixed some issue, is it ok to merge?

PR #2517, opened 2025-01-16
Fix minor error in schema#schema-transformations

lervag:
`GQLDirective` takes a `Directive` as argument, not a string.


ghostdogpr (approved):
Thanks!

lervag:
No problem! Minor thing, but since I noticed it I thought it better to contribute immediately ;)

PR #2547, opened 2025-02-12
Spelling

jsoref:
This PR corrects misspellings identified by the [check-spelling action](https://github.com/marketplace/actions/check-spelling)

The misspellings have been reported at https://github.com/jsoref/caliban/actions/runs/13278582458#summary-37072536513

The action reports that the changes in this PR would make it happy: https://github.com/jsoref/caliban/actions/runs/13278582569#summary-37072536820

PR #2548, opened 2025-02-12
Fix compatibility workflow to run on ubuntu-latest

jsoref:
This should fix https://github.com/ghostdogpr/caliban/actions/runs/13278600366/job/37073018618?pr=2547
- #2547

### DIFF
--- a/.github/workflows/compatibility.yaml
+++ b/.github/workflows/compatibility.yaml
@@ -8,6 +8,7 @@ on:
       - apollo-compatibility/**
       - federation/**
       - adapters/quick/**
+      - .github/workflows/compatibility.yaml
 
 jobs:
   compatibility:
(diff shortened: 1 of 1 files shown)


.github/workflows/compatibility.yaml
```diff
@@ -8,6 +8,7 @@ on:
       - apollo-compatibility/**
       - federation/**
       - adapters/quick/**
+      - .github/workflows/compatibility.yaml
```
jsoref, line 11: This enables validating changes to the workflow itself which is a generally useful thing to have.

I validated it by adding a similar block with `push:` in https://github.com/check-spelling-sandbox/caliban/actions/runs/13278865960/job/37073242437

ghostdogpr:
Thanks!

Issue #2580, opened 2025-03-10
Can't derive ArgBuilder for recursive types


ghostdogpr:
(reported in Discord)

```scala
object TestApp extends App {
  case class RecursiveArgs(
    id: String,
    rec: Option[RecursiveArgs],
  )

  lazy implicit val recArgBuilder: ArgBuilder[RecursiveArgs] = ArgBuilder.gen

  println(recArgBuilder)
}
```

This leads to a stack overflow on Scala 2. It seems to hang on Scala 3. `Schema` derivation works on the other hand.

ghostdogpr:
Closed by https://github.com/ghostdogpr/caliban/pull/2581

PR #2581, opened 2025-03-10
Fix recursion for ArgBuilder

WesselVS:
Fixed issue https://github.com/ghostdogpr/caliban/issues/2580 for both scala 2 and 3


ghostdogpr (approved):
Thanks! I thing it was a regression at some point when `default` was introduced

PR #2598, opened 2025-03-29
Add support for subscriptionUrl in GraphiQL

retriku:
This pull request adds support for testing subscriptions from Graphiql tool

### DIFF
--- a/adapters/quick/src/main/scala/caliban/GraphiQLHandler.scala
+++ b/adapters/quick/src/main/scala/caliban/GraphiQLHandler.scala
@@ -18,10 +18,30 @@ object GraphiQLHandler {
   def handler(apiPath: String, wsPath: Option[String]): RequestHandler[Any, Nothing] = {
     val headers = Headers(Header.ContentType(MediaType.text.html).untyped)
     zio.http.handler { (req: Request) =>
+      val body = wsPath.fold(html(apiPath, req.path.encode))(html(apiPath, req.path.encode, _))
       Response(
         Status.Ok,
         headers,
-        Body.fromString(html(apiPath, req.path.encode, wsPath))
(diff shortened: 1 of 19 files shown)


adapters/quick/src/main/scala/caliban/GraphiQLHandler.scala
```diff
@@ -11,9 +11,30 @@ object GraphiQLHandler {
...
+  }
+
+  /**
+   * Creates a handler which serves the GraphiQL UI from CDN.
+   *
+   * @param apiPath The path at which the API can be introspected.
+   *
+   * @see [[https://github.com/graphql/graphiql/tree/main/examples/graphiql-cdn]]
+   */
+  @deprecated("Use overloaded method also providing optional graphql subscription param", since = "2.10.1")
```
ghostdogpr, line 37: To reduce code duplication, how about making this one call the other one with `wsPath = None`?

core/src/main/scala/caliban/HttpUtils.scala
```diff
@@ -128,9 +128,14 @@ private[caliban] object HttpUtils {
        |  <body>
        |    <div id="graphiql">Loading...</div>
        |    <script>
+       |      console.log("window.location.href=")
```
ghostdogpr: do we need to keep that?

retriku: no we do not.. removed

adapters/quick/src/main/scala/caliban/GraphiQLHandler.scala
retriku: you are right! fixed

ghostdogpr (approved):
Looks good, thank you!

PR #2604, opened 2025-04-05
Fix validation for lists containing `oneOf` inputs

### LINKED ISSUE #2603: Cannot run mutation that has `oneOf` input objects inside a list
```scala
test("oneOf input inside array") {

  case class AddPets(pets: List[Pet.Wrapper])
  case class Queries(addPets: AddPets => List[Pet])

  val api: GraphQL[Any] = graphQL(
    RootResolver(
      Queries(_.pets.map(_.pet))
    )
  )

  val cases = List(
    gqldoc("""{
      addPets(pets: [
        { cat: { name: "a" } },
        { dog: { name: "b" } }
      ]) {
        __typename
        ... on Cat { name }
        ... on Dog { name }
      }
    }""") -> """{"addPets":[{"__typename":"Cat","name":"a"},{"__typename":"Dog","name":"b"}]}"""
  )

  ZIO.foldLeft(cases)(assertCompletes) { case (acc, (query, expected)) =>
    api.interpreter
      .flatMap(_.execute(query))
      .map(response => acc && assertTrue(response.data.toString == expected))
  }
}
```

Fails with `Argument 'pets' on field 'addPets' is not a valid OneOf Input Object` due to https://github.com/ghostdogpr/caliban/blob/v2.10.0/core/src/main/scala/caliban/validation/Validator.scala#L594 returning true for a list that contains input objects.

guymers:
/fixes #2603

### DIFF
--- a/core/src/main/scala/caliban/validation/ValueValidator.scala
+++ b/core/src/main/scala/caliban/validation/ValueValidator.scala
@@ -76,31 +76,20 @@ private object ValueValidator {
                 validateType(inputType.ofType.getOrElse(inputType), other, context, s"List item in $errorContext")
             }
 
+          case INPUT_OBJECT if inputType._isOneOfInput =>
+            Validator.validateOneOfInputValue(argValue, errorContext)
+
           case INPUT_OBJECT =>
             argValue match {
               case ObjectValue(fields) =>
-                if (inputType._isOneOfInput && fields.size != 1) {
-                  failValidation(
-                    s"$errorContext is not a valid OneOf Input Object",
-                    "OneOf Input Object arguments must specify exactly one non-null key"
-                  )
-                } else if (
-                  inputType._isOneOfInput && fields.size == 1 && fields.headOption.map(_._2).contains(NullValue)
-                ) {
-                  failValidation(
-                    s"$errorContext is not a valid OneOf Input Object",
-                    "OneOf Input Object arguments must specify exactly one non-null key"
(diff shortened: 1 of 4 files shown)


ghostdogpr:
The mima failure is very weird, it's comparing against an old version 🤔

guymers:
Looks like it was using the latest tag that was in my fork, I pushed 2.10.0 to it, passing now.

ghostdogpr (approved):
LGTM

@kyri-petrou do you wanna take a look since you added `oneOf` in the first place?

core/src/main/scala/caliban/validation/ValueValidator.scala
```diff
@@ -79,14 +79,28 @@ private object ValueValidator {
...
+                    "OneOf Input Object arguments must specify exactly one non-null key"
+                  )
+                } else if (
+                  inputType._isOneOfInput && fields.size == 1 && fields.headOption.map(_._2).contains(NullValue)
+                ) {
+                  failValidation(
+                    s"$errorContext is not a valid OneOf Input Object",
+                    "OneOf Input Object arguments must specify exactly one non-null key"
+                  )
+                } else {
```
kyri-petrou: Can't we make `Validator.validateOneOfInputValue` package-private (within `validation`) and reuse it here?

guymers: Done

PR #2612, opened 2025-04-10
Adding ability to partially mutate objects

SuperIzya:
Added flag `partialMutations` for server code-generation. With this flag, the server will generate all fields of input types as `Either[Unit, T]`, where `Left(())` means that the field is missing. This allows to distinguish the case when the field is set explicitly to `null` from the case when the field is not passed to mutation and it's value should not change.


SuperIzya:
Wrong branch

PR #2613, opened 2025-04-10
Add ability to partially mutate objects.

SuperIzya:
Added flag `partialMutations` for server code-generation. With this flag, the server will generate all fields of input types as `Either[Unit, T]`, where `Left(())` means that the field is missing. This allows to distinguish the case when the field is set explicitly to `null` from the case when the field is not passed to mutation and it's value should not change.


ghostdogpr:
A couple notes before I review further:
- I don't think you need to introduce `Undefined`. `ArgBuilder` already supports constructing a different value based on whether the input is null or missing (see `buildMissing` and [this example](https://github.com/ghostdogpr/caliban/blob/b507b783108f32b62093ebcda3948c7e29755ee1/core/src/test/scala/caliban/TriState.scala))
- The server codegen already has a lot of options and this one feels very specific. I wonder if you couldn't use `OneOf` instead? It is supported already and allows a clearly defined behavior compared to relying on null/missing.

SuperIzya:
My use case is the following: I have a `schema.graphql` file with `input` types and mutations like
```
type Entity {
  id: Int!
  name: String
  value: Int
}

input UpdateEntity {
  name: String
  value: Int
}

type Mutations {
  updateEntity(entityId: Int!, updateEntity: UpdateEntity!): Entity!
}
```
From this definition I am building a typescript client, scala server and have some users making "manual" updates. For the sake of those "manual" users I need to distinguish the cases (as in subj), when I don't want to change the name from when I want to set it to `null` explicitly. If you can at least hint me how can I do that with existing tools, I will create a new PR with documentation change, because I tried everything (I could find) and failed...

P.S.: currently we don't have the ability to set a field to `null` - if the field is `null` we use the old value.

ghostdogpr:
So, the question is: do you have to absolutely expose this schema or are you flexible? If you are flexible, using `@oneof` could be a good idea. See an example here: https://github.com/ghostdogpr/caliban/blob/series/2.x/tools/src/test/scala/caliban/tools/SchemaWriterSpec.scala#L251

If you must adhere to that schema, I think a more general option would be to allow for a custom `F[_]` instead of `Either[Unit, _]` (like we do for effect types) for all "optional input" then users can use what they want and they can provide an `ArgBuilder` for their custom type.

SuperIzya:
Unfortunately I am not flexible with the schema. Also, isn't missing fields on input types part of the standard?

+1 to `F[_]` for "optional input"

core/src/main/scala/caliban/Value.scala
```diff
@@ -13,6 +13,9 @@ sealed trait InputValue extends Serializable { self =>
   def toInputString: String = ValueRenderer.inputValueRenderer.renderCompact(self)
 }
 object InputValue {
+  case object UndefinedValue                              extends InputValue {
```
ghostdogpr: As I mentioned we don't need that as we can use `buildMissing` in `ArgBuilder`

tools/src/main/scala/caliban/tools/SchemaWriter.scala
```diff
@@ -260,11 +261,14 @@ object SchemaWriter {
     def writeInputValue(value: InputValueDefinition): String = {
       val GQLNewTypeInputDirective = writeGQLNewTypeDirective(value.directives)
 
-      val inputDef = resolveNewTypeInputDef(value).getOrElse(value)
+      val inputDef    = resolveNewTypeInputDef(value).getOrElse(value)
```
ghostdogpr, line 264: This makes sense only if the field is nullable, right? I guess we shouldn't apply it for nullable fields?

Also what happens with the Option? Does it create `Either[Unit, Option[A]]`? If that's the case, it's not great if you want to use something like `TriState[A]`, I think that F should replace the `Option`.

core/src/main/scala/caliban/Value.scala
SuperIzya: What default value are you suggesting for a missing field? IMHO, the case of a missing field should be solved on the level of the domain-logic, not in deserializer.

tools/src/main/scala/caliban/tools/SchemaWriter.scala
SuperIzya: It makes sense for both nullable and not-nullable fields:
```
input UpdateValue {
  value: Int!
  description: String
}
```
`description` is nullable and a `null` value is expected, whereas 
`value` is not and we expect a number there. But the field still can me missing, in case the user doesn't want to change the value. Right or am I missing something?
If the above is correct, then `TriState[A]` won't work unfortunately as and other effect containing `Option`. Hence `F[Option[A]]` in case of nullable fields.

core/src/main/scala/caliban/Value.scala
ghostdogpr: Domain logic does not use `InputValue` either. `ArgBuilder` allows transforming the input into whatever type is used in business logic.

Your `missingInput` function can be implemented without `UndefinedValue`:
```scala
def missingInput[A, F[_]](onMissing: F[A], provided: A => F[A])(implicit ev: ArgBuilder[A]): ArgBuilder[F[A]] =
  new ArgBuilder[F[A]] {
    override def buildMissing(default: Option[String]): Either[ExecutionError, F[A]] = Right(onMissing)
    def build(input: InputValue): Either[ExecutionError, F[A]]                       = ev.build(input).map(provided)
  }
```

tools/src/main/scala/caliban/tools/SchemaWriter.scala
ghostdogpr: > value is not and we expect a number there. But the field still can me missing, in case the user doesn't want to change the value. Right or am I missing something?

If the field is not nullable, it can't be null. The query validation will fail if you send null.

core/src/main/scala/caliban/introspection/adt/__InputValue.scala
```diff
@@ -13,6 +13,7 @@ case class __InputValue(
   description: Option[String],
   `type`: () => __Type,
   defaultValue: Option[String],
+  isOptional: Boolean = false,
```
ghostdogpr: This class must adhere to the spec: https://spec.graphql.org/October2021/#sec-The-__InputValue-Type
The optionality is already expressed by `type` (see https://spec.graphql.org/October2021/#sec-The-__Type-Type.Non-Null).

tools/src/main/scala/caliban/tools/SchemaWriter.scala
SuperIzya: Non-nullable field still can be missed in the input. Hence `F[A]`, where `A` may (or may not) be `Option[B]`.

And that's also the reason why `TriState[A]` won't work: it's always `F[Option[A]]`.

ghostdogpr: > Non-nullable field still can be missed in the input.

No, this is an invalid query then, unless it has a default value but then it will have that value.

SuperIzya: I've found it in the [GraphQL spec](https://spec.graphql.org/draft/#sec-Input-Objects.Input-Coercion). What should I do with the PR?

ghostdogpr: It says the same thing I was saying (see highlights in bold):

> If no value is provided for a defined input object field and that field definition provides a default value, the default value should be used. If **no default value** is provided and the input object field’s type is **non-null**, **an error should be raised**. Otherwise, if the field is not required, then no entry is added to the coerced unordered map.

> If the value null was provided for an input object field, **and the field’s type is not a non-null type**, an entry in the coerced unordered map is given the value null. In other words, there is a semantic difference between the explicitly provided value null versus having not provided a value.

SuperIzya:
Breaking GraphQL standard

PR #2619, opened 2025-04-12
Feat: GraphQL schema visualization

i10416:
Hello, maintainers. Thank you for the great GraphQL library.
To make developer experience slightly better,  I write a draft for GraphQL schema visualizer(Caliban Document type to `.dot` language converter). For now, It can transform the following schema into the figure below.

If you think it is useful and you are happy with adding this feature to caliban tools module, I will continue implementing the remaining part of visualization such as union type, mutation operation and subscription operation. Otherwise, let me know. I will withdraw this PR.

```graphql
type Root {
    account(id: String!): Account
    allAccounts: [Account!]!
    search(input: SearchInput!): [Account!]!
}

interface Node {
  id: String!
}
interface Statusable {
  status: Status!
}

enum Status {
  ACTIVE
  INACTIVE
}

input SearchInput {
  text: String!
}

type Account implements & Node {
  id: String!
  email: String
  status: Status!
}
```

image

### DIFF
--- a/core/src/main/scala/caliban/parsing/adt/Document.scala
+++ b/core/src/main/scala/caliban/parsing/adt/Document.scala
@@ -6,6 +6,7 @@ import caliban.parsing.adt.Definition.ExecutableDefinition.{ FragmentDefinition,
 import caliban.parsing.adt.Definition.TypeSystemDefinition.TypeDefinition._
 import caliban.parsing.adt.Definition.TypeSystemDefinition.{ DirectiveDefinition, SchemaDefinition, TypeDefinition }
 import caliban.parsing.adt.OperationType.{ Mutation, Query, Subscription }
+import caliban.parsing.adt.Definition.TypeSystemDefinition.AggregationTypeDefinition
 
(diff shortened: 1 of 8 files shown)


ghostdogpr:
Hi! Thanks for the proposal.

I'm wondering if this is worth implementing and maintaining when users can simply use something like https://github.com/sheerun/graphqlviz ?

@kyri-petrou @paulpdaniels @frekw you guys have any opinion?

i10416:
> I'm wondering if this is worth implementing and maintaining when users can simply use something like graphviz

Actually, it is inspired by graphqlviz.
One advantage over the original one is that it is easier to use from other Scala tools because it is implemented in Scala and does not require other dependencies.

paulpdaniels:
I haven't really used the graphviz tools, so I'm curious what workflows would you use these in that we would benefit from having it natively in scala.

i10416:
> I haven't really used the graphviz tools, so I'm curious what workflows would you use these in ...

At least, It reduces third party dependencies and simplifies tool installation.
Anyway, I understand that there is not big benefit compared to maintainance burden.
I withdraw this PR.

PR #2621, opened 2025-04-14
Optimize `FederationHelpers` and `ValidationOps`

kyri-petrou:
Noticed some minor things that could be optimized / cleaned up a bit while working on some other things.

I thought to target `series/3.x` since one of the changes Mima complains about it

### DIFF
--- a/core/src/main/scala/caliban/validation/ValidationOps.scala
+++ b/core/src/main/scala/caliban/validation/ValidationOps.scala
@@ -17,8 +17,9 @@ private[caliban] object ValidationOps {
   def validateAllDiscard[A](
     in: List[A]
   )(f: A => Either[ValidationError, Unit]): Either[ValidationError, Unit] = {
+    val nil = Nil
     var rem = in
-    while (rem ne Nil) {
+    while (rem ne nil) {
       val res = f(rem.head)
       if (res.isLeft) return res
       else rem = rem.tail
@@ -58,10 +59,11 @@ private[caliban] object ValidationOps {
   def validateAll[A, B](
     in: List[A]
   )(f: A => Either[ValidationError, B]): Either[ValidationError, List[B]] = {
+    val nil     = Nil
     var rem     = in
     val builder = ListBuffer.empty[B]
 
-    while (rem ne Nil) {
+    while (rem ne nil) {
       f(rem.head) match {
         case Right(v) => builder.addOne(v)
         case left     => return left.asInstanceOf[Either[ValidationError, List[B]]]
@@ -83,10 +85,10 @@ private[caliban] object ValidationOps {
         case _                                   => other
       }
 
-    def as[B](a: B): Either[E, B] =
+    def as[B](b: B): Either[E, B] =
       (self: @unchecked) match {
         case l: Left[E @unchecked, B @unchecked] => l
-        case _                                   => Right(a)
+        case _                                   => Right(b)
       }
 
     def unit: Either[E, Unit] =
--- a/core/src/main/scala/caliban/validation/ValidationOps.scala
+++ b/core/src/main/scala/caliban/validation/ValidationOps.scala
@@ -6,6 +6,8 @@ import scala.collection.mutable
 import scala.collection.mutable.ListBuffer
 
 private[caliban] object ValidationOps {
+  import caliban.syntax._
+
   val unit: Either[Nothing, Unit] = Right(())
 
   def when(cond: Boolean)(f: => Either[ValidationError, Unit]): Either[ValidationError, Unit] =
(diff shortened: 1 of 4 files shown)


ghostdogpr (approved):

federation/src/main/scala/caliban/federation/FederationHelpers.scala
```diff
@@ -6,23 +6,24 @@ import caliban.Value.{ NullValue, StringValue }
...
+    var rem    = list
 
-    while (error.isEmpty && iterator.hasNext) {
-      val b = iterator.next()
-      b match {
-        case Left(value)  =>
-          result.clear()
-          error = Some(value)
-        case Right(value) => result += value
+    while (rem ne Nil) {
```
paulpdaniels: Micro: Should we use the `val nil = Nil` trick here?

core/src/main/scala/caliban/validation/ValidationOps.scala
```diff
@@ -80,26 +79,20 @@ private[caliban] object ValidationOps {
   final implicit class EitherOps[E, A](private val self: Either[E, A]) extends AnyVal {
     def *>[B](other: Either[E, B]): Either[E, B] =
       (self: @unchecked) match {
-        case _: Right[?, ?]                      => other
         case l: Left[E @unchecked, B @unchecked] => l
+        case _                                   => other
       }
 
     def as[B](a: B): Either[E, B] =
```
paulpdaniels: Can we make the parameter name `b`?

federation/src/main/scala/caliban/federation/FederationHelpers.scala
```diff
@@ -6,23 +6,24 @@ import caliban.Value.{ NullValue, StringValue }
...
+import scala.collection.mutable.ListBuffer
+
 private[federation] object FederationHelpers {
-  def traverseEither[A, B](list: List[Either[A, B]]): Either[A, List[B]] = {
-    val iterator = list.iterator
-    val result   = List.newBuilder[B]
-    var error    = Option.empty[A]
+  import caliban.syntax._
+
+  private def traverseEither[A, B](list: List[Either[A, B]]): Either[A, List[B]] = {
```
paulpdaniels: I'm wondering if this should actually be an iterable. Then we could use a lazily evaluated view over the incoming collection to improve perf for cases where it fails early in the traverse

core/src/main/scala/caliban/validation/ValidationOps.scala
```diff
@@ -50,26 +52,23 @@ private[caliban] object ValidationOps {
...
   )(f: A => Either[ValidationError, B]): Either[ValidationError, List[B]] = {
-    var i       = 0
     var rem     = in
-    var err     = null.asInstanceOf[ValidationError]
     val builder = ListBuffer.empty[B]
-    while ((rem ne Nil) && (err eq null)) {
-      val value = rem.head
-      f(value) match {
+
+    while (rem ne Nil) {
```
paulpdaniels: Local `nil` or will it matter?

federation/src/main/scala/caliban/federation/FederationHelpers.scala
```diff
@@ -39,26 +40,24 @@ private[federation] object FederationHelpers {
...
 
   case class RepresentationsArgs(representations: List[_Any])
 
   implicit val representationsArgBuilder: ArgBuilder[RepresentationsArgs] = {
     case InputValue.ObjectValue(fields) =>
-      fields.get("representations").toRight(ExecutionError("_Any must contain a __typename value")).flatMap {
+      fields.getOrElseNull("representations") match {
         case InputValue.ListValue(values) =>
           traverseEither(values.map(anyArgBuilder.build)).map(RepresentationsArgs.apply)
+        case null                         => Left(ExecutionError("_Any must contain a __typename value"))
```
paulpdaniels: Error message seems off

federation/src/main/scala/caliban/federation/FederationHelpers.scala
kyri-petrou: Done

core/src/main/scala/caliban/validation/ValidationOps.scala
kyri-petrou: Done

federation/src/main/scala/caliban/federation/FederationHelpers.scala
kyri-petrou: I kept it as a list but changed it to `List[InputValue]` and moved `anyArgBuilder.build` within the method instead. Since this method is used only in a single place where we already have the `List[InputValue]` materialized I think this makes the most sense performance-wise.

federation/src/main/scala/caliban/federation/FederationHelpers.scala
kyri-petrou: Changed it, not sure if it's the best but kept it aligned with `_Any`

paulpdaniels (approved):

PR #2625, opened 2025-04-16
Avoid unnecessary schema renaming

kyri-petrou:
This might come as a surprise, but I've seen usages of `rename` in the wild where it renames the Schema to the same name as the one that is inferred. This PR optimizes such path so that we avoid going through the renaming loop during query execution.

We also avoid creating a new `__Type` and simply return the existing one in such paths which makes better use of the cached lazy vals


ghostdogpr (approved):
Why were they doing that? Any idea?

kyri-petrou:
AFAICT it was used as an alternative to `@GQLName` in a codegen tool to ensure that names don't accidentally change, unknowing that `rename` is something that it's evaluated during runtime whereas the annotation is a compile-time operation.

PR #2626, opened 2025-04-17
Change EntityResolver#resolve to return a Step

ghostdogpr:
This improves performance as well as allows simpler constructors that take a pure function or a ZIO.

### DIFF
--- a/federation/src/main/scala/caliban/federation/EntityResolver.scala
+++ b/federation/src/main/scala/caliban/federation/EntityResolver.scala
@@ -2,16 +2,17 @@ package caliban.federation
 
 import caliban.execution.Field
 import caliban.introspection.adt.__Type
+import caliban.schema.Step.QueryStep
 import caliban.schema.{ ArgBuilder, Schema, Step }
 import caliban.{ CalibanError, InputValue }
-import zio.IO
+import zio.ZIO
 import zio.query.ZQuery
 
 /**
  * A resolver which is used when attempting to materialize types from their "any" representation
  */
 trait EntityResolver[-R] {
-  def resolve(value: InputValue): ZQuery[R, CalibanError, Step[R]]
+  def resolve(value: InputValue): Step[R]
   def toType: __Type
 }
 
@@ -20,10 +21,42 @@ object EntityResolver {
     resolver: A => ZQuery[R, CalibanError, Option[T]]
   )(implicit schema: Schema[R, T]): EntityResolver[R] =
     new EntityResolver[R] {
-      override def resolve(value: InputValue): ZQuery[R, CalibanError, Step[R]] =
-        ZQuery.fromEither(implicitly[ArgBuilder[A]].build(value)).flatMap { arg =>
-          resolver(arg).map(_.fold[Step[R]](Step.NullStep)(schema.resolve))
-        }
+      override def resolve(value: InputValue): Step[R] =
+        QueryStep(
+          ZQuery.fromEither(implicitly[ArgBuilder[A]].build(value)).flatMap { arg =>
+            resolver(arg).map(_.fold[Step[R]](Step.NullStep)(schema.resolve))
+          }
+        )
+
+      override def toType: __Type = schema.toType_()
+    }
+
+  def fromFunction[A: ArgBuilder, T](
+    resolver: A => Option[T]
+  )(implicit schema: Schema[Any, T]): EntityResolver[Any] =
+    new EntityResolver[Any] {
+      override def resolve(value: InputValue): Step[Any] =
+        QueryStep(
+          ZQuery.fromEither(implicitly[ArgBuilder[A]].build(value)).map { arg =>
+            resolver(arg).fold[Step[Any]](Step.NullStep)(schema.resolve)
+          }
+        )
+
+      override def toType: __Type = schema.toType_()
+    }
+
+  def fromZIO[R, A: ArgBuilder, T](
+    resolver: A => ZIO[R, CalibanError, Option[T]]
+  )(implicit schema: Schema[R, T]): EntityResolver[R] =
+    new EntityResolver[R] {
+      override def resolve(value: InputValue): Step[R] =
+        QueryStep(
+          ZQuery.fromZIO(
+            ZIO.fromEither(implicitly[ArgBuilder[A]].build(value)).flatMap { arg =>
+              resolver(arg).map(_.fold[Step[R]](Step.NullStep)(schema.resolve))
+            }
+          )
+        )
 
       override def toType: __Type = schema.toType_()
     }
@@ -39,12 +72,12 @@ object EntityResolver {
       resolver: Field => A => ZQuery[R, CalibanError, Option[T]]
     )(implicit schema: Schema[R, T], argBuilder: ArgBuilder[A]): EntityResolver[R] =
       new EntityResolver[R] {
-        override def resolve(value: InputValue): ZQuery[R, CalibanError, Step[R]] =
-          ZQuery.fromEither(argBuilder.build(value)).map { arg =>
+        override def resolve(value: InputValue): Step[R] =
+          QueryStep(ZQuery.fromEither(argBuilder.build(value)).map { arg =>
             Step.MetadataFunctionStep(field =>
               Step.QueryStep(resolver(field)(arg).map(_.fold[Step[R]](Step.NullStep)(schema.resolve)))
             )
-          }
+          })
 
         override def toType: __Type = schema.toType_()
       }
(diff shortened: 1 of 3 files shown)


ghostdogpr:
@kyri-petrou is that what you had in mind? Feel free to change more things if you wanted to.

federation/src/main/scala/caliban/federation/EntityResolver.scala
```diff
@@ -20,10 +21,42 @@ object EntityResolver {
...
+          ZQuery.fromEither(implicitly[ArgBuilder[A]].build(value)).map { arg =>
+            resolver(arg).fold[Step[Any]](Step.NullStep)(schema.resolve)
+          }
+        )
+
+      override def toType: __Type = schema.toType_()
+    }
+
+  def fromZIO[R, A: ArgBuilder, T](
+    resolver: A => ZIO[R, CalibanError, Option[T]]
```
paulpdaniels, line 74: For consistency, I think this should be either `fromFunctionZIO` or the other one should be `fromOption`

I think we could also have an `either` version and perhaps a an alias to `apply` like `fromQuery` for discoverability?

federation/src/main/scala/caliban/federation/EntityResolver.scala
```diff
@@ -20,10 +21,42 @@ object EntityResolver {
...
+        QueryStep(
+          ZQuery.fromEither(implicitly[ArgBuilder[A]].build(value)).flatMap { arg =>
+            resolver(arg).map(_.fold[Step[R]](Step.NullStep)(schema.resolve))
+          }
+        )
+
+      override def toType: __Type = schema.toType_()
+    }
+
+  def fromFunction[A: ArgBuilder, T](
```
paulpdaniels: I guess these would need fully ascribed types or do `(a: A) =>`. Should we do a partially applied signature instead?

federation/src/main/scala/caliban/federation/EntityResolver.scala
```diff
@@ -20,10 +21,42 @@ object EntityResolver {
...
-      override def resolve(value: InputValue): ZQuery[R, CalibanError, Step[R]] =
-        ZQuery.fromEither(implicitly[ArgBuilder[A]].build(value)).flatMap { arg =>
-          resolver(arg).map(_.fold[Step[R]](Step.NullStep)(schema.resolve))
-        }
+      override def resolve(value: InputValue): Step[R] =
+        QueryStep(
+          ZQuery.fromEither(implicitly[ArgBuilder[A]].build(value)).flatMap { arg =>
+            resolver(arg).map(_.fold[Step[R]](Step.NullStep)(schema.resolve))
+          }
+        )
```
kyri-petrou: Avoids the extra flatmap and `ZQuery.fromEither`:

```suggestion
        ArgBuilder[A].build(value) match {
          case Right(arg)  =>
            val q = resolver(arg).map {
              case Some(value) => schema.resolve(value)
              case _           => Step.NullStep
            }
            Step.QueryStep(q)
          case Left(error) => Step.FailureStep(error)
        }
```

federation/src/main/scala/caliban/federation/EntityResolver.scala
```diff
@@ -20,10 +21,42 @@ object EntityResolver {
...
+  def fromFunction[A: ArgBuilder, T](
+    resolver: A => Option[T]
+  )(implicit schema: Schema[Any, T]): EntityResolver[Any] =
+    new EntityResolver[Any] {
+      override def resolve(value: InputValue): Step[Any] =
+        QueryStep(
+          ZQuery.fromEither(implicitly[ArgBuilder[A]].build(value)).map { arg =>
+            resolver(arg).fold[Step[Any]](Step.NullStep)(schema.resolve)
+          }
+        )
```
kyri-petrou: We can avoid wrapping into an effect altogether
```suggestion
        ArgBuilder[A].build(value) match {
          case Right(arg)  =>
            resolver(arg) match {
              case Some(value) => schema.resolve(value)
              case _           => Step.NullStep
            }
          case Left(error) => Step.FailureStep(error)
        }
```

federation/src/main/scala/caliban/federation/EntityResolver.scala
```diff
@@ -20,10 +21,42 @@ object EntityResolver {
...
+  )(implicit schema: Schema[R, T]): EntityResolver[R] =
+    new EntityResolver[R] {
+      override def resolve(value: InputValue): Step[R] =
+        QueryStep(
+          ZQuery.fromZIO(
+            ZIO.fromEither(implicitly[ArgBuilder[A]].build(value)).flatMap { arg =>
+              resolver(arg).map(_.fold[Step[R]](Step.NullStep)(schema.resolve))
+            }
+          )
+        )
```
kyri-petrou: Same for ZQuery above but for ZIO:

```suggestion
        ArgBuilder[A].build(value) match {
          case Right(arg)  =>
            val q = resolver(arg).map {
              case Some(value) => schema.resolve(value)
              case _           => Step.NullStep
            }
            Step.QueryStep(ZQuery.fromZIONow(q))
          case Left(error) => Step.FailureStep(error)
        }
```

federation/src/main/scala/caliban/federation/EntityResolver.scala
```diff
@@ -39,12 +72,12 @@ object EntityResolver {
       resolver: Field => A => ZQuery[R, CalibanError, Option[T]]
     )(implicit schema: Schema[R, T], argBuilder: ArgBuilder[A]): EntityResolver[R] =
       new EntityResolver[R] {
-        override def resolve(value: InputValue): ZQuery[R, CalibanError, Step[R]] =
-          ZQuery.fromEither(argBuilder.build(value)).map { arg =>
+        override def resolve(value: InputValue): Step[R] =
```
kyri-petrou, line 107: Can be optimized as above

federation/src/main/scala/caliban/federation/FederationSupport.scala
```diff
@@ -82,7 +81,7 @@ abstract class FederationSupport(
       override def resolve(value: _Entity): Step[R] =
         _entityMap
           .get(value.__typename)
-          .fold[Step[R]](Step.NullStep)(resolver => QueryStep(resolver.resolve(value.value)))
+          .fold[Step[R]](Step.NullStep)(_.resolve(value.value))
```
kyri-petrou: Super minor but we can avoid the Option wrapping and allocating functions by:

```scala
getOrElse(value.__typename, null) match {
  case  null => Step.NullStep
  case resolver => resolver.resolve(value.value)
}
```

kyri-petrou (commented):
Yeap this is what I had in mind 👍 Left a few recommendations below to optimize them a bit more (particularly the `A => Option[T]` one)

federation/src/main/scala/caliban/federation/EntityResolver.scala
ghostdogpr: Applied all the suggestions 👍

federation/src/main/scala/caliban/federation/EntityResolver.scala
ghostdogpr: Renamed to `fromOption` and added `fromEither`

federation/src/main/scala/caliban/federation/EntityResolver.scala
ghostdogpr: It's a bit annoying because there would be so many of them (for ZQuery there are two: `EntityResolverPartiallyApplied` and `MetadataEntityResolverPartiallyApplied`).

But actually I found that those do not infer very well when the schema requires an `R` that is different from the `R` of the ZIO/ZQuery (that doesn't matter for Option/Either cases).

federation/src/main/scala/caliban/federation/FederationSupport.scala
```diff
@@ -81,9 +80,10 @@ abstract class FederationSupport(
        */
       override def resolve(value: _Entity): Step[R] =
         _entityMap
-          .get(value.__typename)
-          .fold[Step[R]](Step.NullStep)(resolver => QueryStep(resolver.resolve(value.value)))
-
+          .getOrElse(value.__typename, null) match {
+          case null     => Step.NullStep
+          case resolver => resolver.resolve(value.value)
```
paulpdaniels, line 85: This preserves the behavior which should be fine for this PR but I just realized the null case here is more of a defect since it means either the gateway or the entity resolver map is wrong. I'll look into how Apollo does this in subgraphs. It's not part of the spec which makes me think it's undefined behavior that we should perhaps throw to avoid silent errors

kyri-petrou (approved):

paulpdaniels (approved):

PR #2629, opened 2025-04-21
Make `Schema`, `ArgBuilder` and `Wrapper` abstract classes

kyri-petrou:
Since traits don't contain implementations of methods or lazy vals, by using traits it means they're reimplemented each time we create a new anonymous instance of these traits. Using abstract classes is much more suitable for such usecases. The downside is that a class can't extend more than 1 abstract class, but that shouldn't be an issue since none of these are meant to be mixed-in.

Also did the following while I was at it that results in a more optimized JVM code:
- Marked all methods / overrides of anonymous Schema instances as final since for some reason anonymous classes are not marked as `final` automatically by the compiler
- Made some methods on `Schema` and `ArgBuilder` (e.g., `map`, `contramap`) final as these are not meant to be overwritten
- Made some case classes extending abstract classes / traits final

### DIFF
--- a/core/src/main/scala/caliban/schema/Schema.scala
+++ b/core/src/main/scala/caliban/schema/Schema.scala
@@ -36,7 +36,7 @@ If you use a custom type as an argument, you also need to provide an implicit Ar
 See https://ghostdogpr.github.io/caliban/docs/schema.html for more information.
 """
 )
-trait Schema[-R, T] { self =>
+abstract class Schema[-R, T] { self =>
 
   private lazy val asType: __Type             = toType(): @nowarn("msg=deprecated")
   private lazy val asInputType: __Type        = toType(isInput = true): @nowarn("msg=deprecated")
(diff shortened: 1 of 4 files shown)


ghostdogpr (approved):

core/src/main/scala/caliban/schema/Schema.scala
```diff
@@ -413,16 +413,18 @@ trait GenericSchema[R] extends SchemaDerivation[R] with TemporalSchema {
     val description: String = s"Either $typeAName or $typeBName"
 
     implicit val leftSchema: Schema[RA, A]  = new Schema[RA, A] {
-      override def nullable: Boolean                                         = true
-      override def canFail: Boolean                                          = evA.canFail
-      override def toType(isInput: Boolean, isSubscription: Boolean): __Type = evA.toType_(isInput, isSubscription)
-      override def resolve(value: A): Step[RA]                               = evA.resolve(value)
+      final override def nullable: Boolean                                         = true
```
paulpdaniels, line 416: We could make this a `val` no?

paulpdaniels (approved):

core/src/main/scala/caliban/schema/Schema.scala
ghostdogpr: @kyri-petrou you wanna change that or shall we merge?

kyri-petrou: Sorry guys missed the notifications on this one.

There's no real benefit in storing primitives that are simple expressions as vals. In fact, Scala vals are compiled to methods that point to private static fields, so likely more performant to have it as a `def` (if there's any difference at all)

PR #2644, opened 2025-05-14
Added option for forwarding upload request

develeon:
Proposal to extend the `QuickAdapter` with a boolean `forwardUploadRequests` which can be set true
in order to forward upload request/multipart form-data to the upload handler.

In this way clients do not have to change endpoints for upload requests.

@ghostdogpr what do you think?

### DIFF
--- a/adapters/quick/src/main/scala/caliban/QuickRequestHandler.scala
+++ b/adapters/quick/src/main/scala/caliban/QuickRequestHandler.scala
@@ -42,14 +42,25 @@ final private class QuickRequestHandler[R](
   def configureWebSocket[R1](config: quick.WebSocketConfig[R1]): QuickRequestHandler[R & R1] =
     new QuickRequestHandler[R & R1](interpreter, config)
 
-  def handleHttpRequest(request: Request)(implicit trace: Trace): URIO[R, Response] = ZIO.suspendSucceed {
-    transformHttpRequest(request)
-      .flatMap(executeRequest(request.method, _))
-      .foldZIO(
-        Exit.succeed,
-        resp => Exit.succeed(transformResponse(request, resp))
+  def handleHttpRequest(forwardUploadRequests: Boolean = false)(request: Request)(implicit
+    trace: Trace
+  ): URIO[R, Response] =
+    if (
+      forwardUploadRequests && request.headers.exists(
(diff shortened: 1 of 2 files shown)


ghostdogpr:
How about removing the parameter and apply that behavior if `apiPath == uploadPath` (building a single route)?

develeon:
> How about removing the parameter and apply that behavior if `apiPath == uploadPath` (building a single route)?

That sounds like a better approach. 
I want to add, that Graphql Multipart seem to have some issues with cross site forgery,
so we are going to moving away from it and think we will rather do the signed URL approach.

Found [this for best practises](https://www.apollographql.com/blog/file-upload-best-practices  )
Maybe could be worth adding a word of caution..

adapters/quick/src/main/scala/caliban/QuickRequestHandler.scala
```diff
@@ -42,14 +42,25 @@ final private class QuickRequestHandler[R](
...
-      .flatMap(executeRequest(request.method, _))
-      .foldZIO(
-        Exit.succeed,
-        resp => Exit.succeed(transformResponse(request, resp))
+  def handleHttpRequest(forwardUploadRequests: Boolean = false)(request: Request)(implicit
+    trace: Trace
+  ): URIO[R, Response] =
+    if (
+      forwardUploadRequests && request.headers.exists(
+        _.renderedValue.contains(MediaType.multipart.`form-data`.fullType)
```
kyri-petrou:
```suggestion
      forwardUploadRequests &&
      request.body.mediaType.exists(MediaType.multipart.`form-data`.matches(_, ignoreParameters = true))
```

I think this should be a lot more performant as it doesn't require iterating through all the headers. AFAIK the content type should automatically be attached to the body by `zio-http`

paulpdaniels:
> Found [this for best practises](https://www.apollographql.com/blog/file-upload-best-practices) Maybe could be worth adding a word of caution..

Don't read _too_ much into that Apollo really only considered it a bad practice until they could put it behind an enterprise flag: https://www.apollographql.com/docs/graphos/routing/configuration/yaml#preview_file_uploads

ghostdogpr (approved):

PR #2649, opened 2025-05-19
Add argbuilder instance for maps

WesselVS:
Add ArgBuilder instance for Map[K, V]

### DIFF
--- a/core/src/main/scala/caliban/schema/ArgBuilder.scala
+++ b/core/src/main/scala/caliban/schema/ArgBuilder.scala
@@ -215,18 +215,18 @@ trait ArgBuilderInstances extends ArgBuilderDerivation {
     case other                       => ev.build(other).map(List(_))
   }
 
-  implicit def map[V](implicit ev: ArgBuilder[V]): ArgBuilder[Map[String, V]] = {
+  implicit def map[K, V](implicit keyEv: ArgBuilder[K], valueEv: ArgBuilder[V]): ArgBuilder[Map[K, V]] = {
     case InputValue.ListValue(kvs) =>
       kvs
-        .foldLeft[Either[ExecutionError, Map[String, V]]](Right(Map.empty)) {
+        .foldLeft[Either[ExecutionError, Map[K, V]]](Right(Map.empty)) {
           case (res @ Left(_), _)                       => res
           case (Right(res), InputValue.ObjectValue(kv)) =>
             for {
-              key   <- if (kv.contains("key")) { string.build(kv("key")) }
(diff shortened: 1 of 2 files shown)


core/src/main/scala/caliban/schema/ArgBuilder.scala
```diff
@@ -215,6 +215,27 @@ trait ArgBuilderInstances extends ArgBuilderDerivation {
     case other                       => ev.build(other).map(List(_))
   }
 
+  implicit def map[K, V](implicit keyEv: ArgBuilder[K], valueEv: ArgBuilder[V]): ArgBuilder[Map[K, V]] = {
+    case InputValue.ListValue(kvs) =>
+      kvs
+        .foldLeft[Either[ExecutionError, Map[K, V]]](Right(Map.empty)) {
```
kyri-petrou: 2 comments here:
1. Since the input values are sequenced, we should use a Map implementation that preserves this sequence. I think `VectorMap` is likely a better option as it has a better lookup performance
2. For performance reasons, I think we should iterate over kvs (using `kvs.iterator`) and a map builder (i.e., `VectorMap.newBuilder`)

core/src/main/scala/caliban/schema/ArgBuilder.scala
```diff
@@ -215,6 +215,27 @@ trait ArgBuilderInstances extends ArgBuilderDerivation {
...
+            for {
+              key   <- if (kv.contains("key")) { keyEv.build(kv("key")) }
+                       else { Left(InvalidInputArgument("key", kv)) }
+              value <-
+                if (kv.contains("value")) {
+                  valueEv.build(kv("value"))
+                } else {
+                  Left(InvalidInputArgument("value", kv))
+                }
+            } yield res + (key -> value)
```
hearnadam: If you continue to use an immutable map, prefer `updated` to avoid the tuple allocation.
```suggestion
            } yield res.updated(key, value)
```

core/src/main/scala/caliban/schema/ArgBuilder.scala
WesselVS: 2.12 does not seem to have VectorMap.

kyri-petrou: What about `ListMap`?

WesselVS: Yes, it uses ListMap now.

ghostdogpr (approved):

PR #2690, opened 2025-07-01
support extend in DocumentRenderer

camarena:
This PR adds support for extend syntax in the DocumentRenderer, addressing scenarios where extended types, enums, unions, or schema blocks should be correctly rendered in the generated SDL output.
* Added logic to handle extend declarations for:
  * type
  * interface
  * enum
  * union
  * schema
* Updated tests to cover new extend rendering cases.
* Minor refactoring for clarity and maintainability.

** Notes **
* When comparing DocumentRenderer.render output in tests or reviews, make sure to ignore trailing and leading whitespace, as it does not affect valid SDL but may differ slightly due to formatting. Ideally tests should ignore whitespace everywhere
* **About schemaRenderer**: I did not modify the schemaRenderer method, even though its current logic is somewhat of a workaround. In a valid GraphQL schema there should only be one schema definition — additional schema entries should properly be extend schema blocks. The existing implementation treats subsequent schema definitions as if they were extensions.
I added what I believe is the correct logic for this behavior but left it commented out to avoid breaking existing legacy use cases for now

Fixes: #2688


ghostdogpr:
I updated your branch because there was a flaky test that I marked as ignored. However your changes seem to make some other tests fail, can you check?

camarena:
I fixed the tests.
@ghostdogpr I would like to know your thoughts about my comment `About schemaRenderer` ^. Thanks

core/src/main/scala/caliban/rendering/DocumentRenderer.scala
```diff
@@ -322,6 +335,64 @@ object DocumentRenderer extends Renderer[Document] {
...
+      override protected[caliban] def unsafeRender(
+        extension: TypeSystemExtension.SchemaExtension,
+        indent: Option[Int],
+        write: StringBuilder
+      ): Unit = {
+        import extension._
+        unsafeRenderImpl("extend schema", None, directives, query, mutation, subscription, indent, write)
+      }
+    }
+  // TODO: The implementation of schemaRenderer is wrong. Two schema definations is invalid. The current implementation
```
ghostdogpr, line 387: @paulpdaniels could you check that?

core/src/test/scala-3/caliban/Scala3SpecificSpec.scala
```diff
@@ -34,19 +34,19 @@ object Scala3SpecificSpec extends ZIOSpecDefault {
...
-                                 |enum FooEnum {
-                                 |  A
-                                 |  B
-                                 |  C
-                                 |}
-                                 |
-                                 |type Queries {
-                                 |  item: FooEnum!
-                                 |}""".stripMargin)
+        assertTrue(rendered.trim == """schema {
```
ghostdogpr: what changed and why?

paulpdaniels:
> The existing implementation treats subsequent schema definitions as if they were extensions.

This isn't correct, there can only be a single `schemaDefinition` within a document, the `extend` is derived from the presence of certain features of the graph and something we landed on in !2433 to address federated graphs where the subgraph definition isn't actually a proper schema on it's own because it doesn't have a `Query` definition. So we need to have an `extend schema` to be able to define the schema level directives.

To make this work I think we'd need to move where we detect this to when we create a document to render the graph. If we detect this type of schema we'd want to push a `SchemaTypeExtension` instead of a `SchemaDefinition`

core/src/main/scala/caliban/rendering/DocumentRenderer.scala
```diff
@@ -383,6 +456,23 @@ object DocumentRenderer extends Renderer[Document] {
...
+    typeExtensionRenderer.list.contramap(_.sorted)
+
+  private lazy val typeExtensionRenderer: Renderer[TypeSystemExtension] = new Renderer[TypeSystemExtension] {
+    override def unsafeRender(extension: TypeSystemExtension, indent: Option[Int], writer: StringBuilder): Unit =
+      extension match {
+        case t: TypeSystemExtension.SchemaExtension    => schemaExtensionRenderer.unsafeRender(t, indent, writer)
+        case t: TypeExtension.EnumTypeExtension        => enumTypeExtensionRenderer.unsafeRender(t, indent, writer)
+        case t: TypeExtension.InputObjectTypeExtension => inputObjectExtensionRenderer.unsafeRender(t, indent, writer)
+        case t: TypeExtension.InterfaceTypeExtension   => interfaceTypeExtensionRenderer.unsafeRender(t, indent, writer)
+        case t: TypeExtension.ScalarTypeExtension      => () // Scalar can't be extended
```
paulpdaniels: https://spec.graphql.org/October2021/#sec-Scalar-Extensions

> Scalar type extensions are used to represent a scalar type which has been extended from some original scalar type. For example, this might be used by a GraphQL tool or service which adds directives to an existing scalar.

We would need to be able to support adding directives to them for instance

core/src/main/scala/caliban/rendering/DocumentRenderer.scala
```diff
@@ -357,7 +428,9 @@ object DocumentRenderer extends Renderer[Document] {
               renderOp("subscription", subscription)
               newlineOrEmpty(indent, writer)
               writer append '}'
-            }
+              newlineOrEmpty(indent, writer)
+            } else
+              newlineOrEmpty(indent, writer)
```
paulpdaniels, line 434: I don't think we need these newlines do we? Generally it's the trailing element that adds the whitespace

core/src/test/scala-3/caliban/Scala3SpecificSpec.scala
camarena: The intent in the documentation is that when you render an entry you leave the buffer in the state for the next one to render properly. That adds an extra newline at the end. I made sure that once you render an entry you are always at the beginning of the next line. Without those changes I was getting something like

```
schema {
   ...
}extend schema {
}
```

core/src/main/scala/caliban/rendering/DocumentRenderer.scala
camarena: I'll add this

core/src/test/scala-3/caliban/Scala3SpecificSpec.scala
camarena: I could redo the logic to assume that you are at the end of the last entry but that will add a newline at the very beginning. That was the reason the last couple of tests broke. It was assuming the output would be:
```
schema {
   ...
}
scalar Unit
```

but with this changes it was
```
schema{
   ...
}

scalar Unit
```

core/src/main/scala/caliban/rendering/DocumentRenderer.scala
camarena: The convention was not consistent in all the code. The changes I made to make it uniform is to assume that when a new entry is about to be rendered the state is to be at the beginning of a new line. I made the changes to guarantee that. For example we were getting
```
schema {
  ...
}
scalar Foo
```
now we get
```
schema{
 ...
}

scalar Foo
```
as I mentioned elsewhere that implies that there is always a newline at the end. I made sure all the tests now pass

camarena:
> > The existing implementation treats subsequent schema definitions as if they were extensions.
> 
> This isn't correct, there can only be a single `schemaDefinition` within a document, the `extend` is derived from the presence of certain features of the graph and something we landed on in !2433 to address federated graphs where the subgraph definition isn't actually a proper schema on it's own because it doesn't have a `Query` definition. So we need to have an `extend schema` to be able to define the schema level directives.
> 
> To make this work I think we'd need to move where we detect this to when we create a document to render the graph. If we detect this type of schema we'd want to push a `SchemaTypeExtension` instead of a `SchemaDefinition`

The parser does not do any semantic validation so it will parse a schema with more than one schemaDefinition. The parser code already does the right thing processing extend schema. In the context of the renderer I believe the code that I have commented is the correct one. It allows you to add to features to the schema using extend. I just wanted to get the thumbs up to remove the incorrect code that I think was added there because the extend support was not in the renderer.

core/src/main/scala/caliban/rendering/DocumentRenderer.scala
```diff
@@ -322,6 +336,64 @@ object DocumentRenderer extends Renderer[Document] {
...
+    }
+  // TODO: The implementation of schemaRenderer is wrong. Two schema definations is invalid. The current implementation
+  // treats it as an extension. The correct implementation should be the follwoing one but I'm keeping the old one for
+  // backward compatibility for now.
+  //  private lazy val schemaRenderer: Renderer[SchemaDefinition] = new Renderer[SchemaDefinition] with SchemaOps {
+  //    override def unsafeRender(definition: SchemaDefinition, indent: Option[Int], writer: StringBuilder): Unit = {
+  //      import definition._
+  //      unsafeRenderImpl("schema", description, directives, query, mutation, subscription, indent, writer)
+  //    }
+  //  }
```
paulpdaniels, line 395: Im ok with this change generally, as long as the current behavior (for partial schemas such as in federation) still works. The issue with the change as is, is that when going from GraphQL to renderer schema we don't do any extension types. I'd suggest that we instead then update the GraphQL#toDocument method to output the schema as an extension type if it's subgraph variant.

core/src/main/scala/caliban/rendering/DocumentRenderer.scala
paulpdaniels: I just checked again, the convention we use is pretty standard with the exception of descriptions which have pretty strict formatting requirements. We instead use the overload for `.list` that takes a renderer which will handle the trailing space correctly

paulpdaniels (changes requested):

camarena:
@paulpdaniels can you clarify what are you asking me to change? The code as I updated it, not just this method but changes elsewhere pass all existing and new tests of formatting with the exception that it ends always in a newline. If I restore to the old logic some of the new tests look not very nice since I added the extend and the previous tests were not exhaustive. For example a Document with no schema entry leaves a blank line at the beginning and scalars do not leave a newline before the previous entry.  If you point me to the strict formatting requirements I can go from there

ghostdogpr:
I would prefer to avoid the usage of `trim` so we can see exactly what's the changed output. `trim` would hide if for example we add too many spaces.

core/src/main/scala/caliban/rendering/DocumentRenderer.scala
ghostdogpr: Looking at the diff the formatting looks ok?

camarena:
@ghostdogpr I removed the use of trim on tests

ghostdogpr:
LGTM but I'd like @paulpdaniels to confirm because I've never worked with extends myself.

core/src/main/scala/caliban/rendering/DocumentRenderer.scala
paulpdaniels: I guess it's fine for now, it's difficult to audit these so if it's already passing then ✅ , and anyway I have some changes I plan to make to this interface for 3.x so I'll address that then.

core/src/main/scala/caliban/rendering/DocumentRenderer.scala
paulpdaniels: Do we want to keep the todo and address it in a follow up then?

paulpdaniels (approved):
I think we can merge this as is, I need it anyway for my work on schema extensions

ghostdogpr:
thanks @camarena !

PR #2691, opened 2025-07-01
Fix nullable `oneOf` root not allowing null

guymers:
Currently validation is too strict and rejects `null` being passed to a `oneOf` that is nullable.


ghostdogpr:
Looks good, thanks!

PR #2696, opened 2025-07-04
Test global id support against series/2.x

li-nkSN:


ghostdogpr:
Will be released on 3.x

PR #2700, opened 2025-07-08
Fix inconsistent hashCode when caching hash code

### LINKED ISSUE #2699: Inconsistent hashCode of caliban.parsing.adt.Selection
Trait `caliban.parsing.adt.Selection` [has suspicious hashCode declaration](https://github.com/ghostdogpr/caliban/blob/43e0d719b0e8ce5d0abdde454a427edb1a30c3a6/core/src/main/scala/caliban/parsing/adt/Selection.scala#L7):

```scala
sealed trait Selection extends Serializable {
  @transient final override lazy val hashCode: Int = super.hashCode()
}
```

It seems that this declaration breaks hashCode implementation for case classes that extend this trait. For example, we have `Selection.Field` class [declared](https://github.com/ghostdogpr/caliban/blob/43e0d719b0e8ce5d0abdde454a427edb1a30c3a6/core/src/main/scala/caliban/parsing/adt/Selection.scala#L12) as follows:

```scala
  case class Field(
    alias: Option[String],
    ...
  ) extends Selection
```

Since it is a "case class" one might expect that it is safe to put those classes into Set. But in fact this case class has broken hashCode behaviour, and therefore Scala `Set[Selection.Field]` is broken too.

Below is the `zio-test` snippet that reproduces the problem:

```scala
      test("check hashCode") {
        val field1 = Selection.Field(
          alias = None,
          name = "id",
          arguments = Map.empty,
          directives = List.empty,
          selectionSet = List.empty,
          index = 0
        )

        val field2 = Selection.Field(
          alias = None,
          name = "id",
          arguments = Map.empty,
          directives = List.empty,
          selectionSet = List.empty,
          index = 0
        )

        assertTrue(
          field1 == field2,
          field1.hashCode == field2.hashCode
        )
      }
```

The test output is:
```
  - check hashCode
✗ 1899428643 was not equal to 1978377980
    field1.hashCode == field2.hashCode
    .hashCode = 1899428643
    field1 = Field(
      alias = None,
      name = "id",
      arguments = Map(
        
      ),
      directives = Nil,
      selectionSet = Nil,
      index = 0
    )
```

kyri-petrou:
/fixes #2699

This is a bit weird that this issue wasn't surfaced until now. I'm really not sure what I was thinking when I wrote the code, this is totally on me :/

### DIFF
--- a/core/src/main/scala/caliban/parsing/adt/Selection.scala
+++ b/core/src/main/scala/caliban/parsing/adt/Selection.scala
@@ -3,8 +3,11 @@ package caliban.parsing.adt
 import caliban.InputValue
 import caliban.parsing.adt.Type.NamedType
 
+import scala.util.hashing.MurmurHash3
+
 sealed trait Selection extends Serializable {
-  @transient final override lazy val hashCode: Int = super.hashCode()
+  // TODO: Kept for binary compatibility
+  @transient override def hashCode: Int = super.hashCode()
 }
 
 object Selection {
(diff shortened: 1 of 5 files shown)


core/src/main/scala/caliban/parsing/adt/Selection.scala
```diff
@@ -3,8 +3,11 @@ package caliban.parsing.adt
 import caliban.InputValue
 import caliban.parsing.adt.Type.NamedType
 
+import scala.util.hashing.MurmurHash3
+
 sealed trait Selection extends Serializable {
-  @transient final override lazy val hashCode: Int = super.hashCode()
+  // TODO: Kept for binary compatibility
+  @transient override def hashCode: Int = super.hashCode()
```
kyri-petrou: Another option is to make `Selection` extend `Product` in series/3.x and implement this here using `productHash` instead

PR #2701, opened 2025-07-08
Fixed warnings produced by client codegen under Scala 3.7

### LINKED ISSUE #2687: Generated code for client produces warning under Scala 3.7.x
The generated code for client contains implicits passed explicitly which now requires the `using` keyword from Scala 3.7.0 or a warning is triggered. Unfortunately we can't just add that because then the code will not compile under older Scala versions. It seems we need something like the `scala3_sources` in ScalaPB. See similar workaround: https://github.com/scalapb/ScalaPB/pull/1854

ghostdogpr:
Closes #2687

### DIFF
--- a/tools/src/main/scala/caliban/tools/ClientWriter.scala
+++ b/tools/src/main/scala/caliban/tools/ClientWriter.scala
@@ -6,6 +6,7 @@ import caliban.parsing.adt.Type.{ ListType, NamedType }
 import caliban.parsing.adt.{ Directives, Document, Type }
 
 import scala.annotation.tailrec
+import scala.collection.compat._
 
 object ClientWriter {
 
--- a/tools/src/main/scala/caliban/tools/ClientWriter.scala
+++ b/tools/src/main/scala/caliban/tools/ClientWriter.scala
@@ -1,6 +1,5 @@
 package caliban.tools
 
-import caliban.Value.StringValue
 import caliban.parsing.adt.Definition.TypeSystemDefinition.TypeDefinition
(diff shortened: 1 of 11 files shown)


tools/src/main/scala/caliban/tools/ClientWriter.scala
```diff
@@ -235,16 +235,20 @@ object ClientWriter {
       val argBuilder                                       = filteredArgs match {
         case Nil  => ""
         case list =>
-          s", arguments = List(${list.zipWithIndex.map { case (arg, idx) =>
-              s"""Argument("${arg.name}", ${safeName(arg.name)}, "${arg.ofType.toString}")(encoder$idx)"""
```
ghostdogpr, line 239: Stop passing implicits explicitly

tools/src/main/scala/caliban/tools/ClientWriter.scala
```diff
@@ -235,16 +235,20 @@ object ClientWriter {
...
             }.mkString(", ")})"
       }
       val implicits                                        = filteredArgs match {
         case Nil  => ""
         case list =>
-          s"(implicit ${list.zipWithIndex.map { case (arg, idx) =>
-              s"""encoder$idx: ArgEncoder[${writeType(arg.ofType)}]"""
-            }.mkString(", ")})"
+          s"(implicit ${list
+              .distinctBy(_.ofType)
```
ghostdogpr, line 246: In case 2 parameters have the same type.

PR #2703, opened 2025-07-08
Use `@threadUnsafe` for cached hash codes in Scala 3

kyri-petrou:
`lazy val` is quite heavyweight in different ways because it needs to guarantee "exactly once" semantics when called. In almost all of our lazy val definitions we don't care about exactly once as we use them for caching.

In Scala 3 we can use a much more lightweight implementation using `@threadUnsafe` that doesn't guarantee "exactly once" semantics. The downside of course is that concurrent thread access to the same lazy val might be computed more than once, but since the computation of `hashCode` is pure it doesn't really matter. Also, we mostly dedup these objects during validation, which is not done in parallel. 

This way our objects have a smaller memory footprint and access to `hashCode` doesn't require a volatile read / write plus unboxing of the `Int` (yeah, "normal" lazy vals box primitives......)

Note that we have a lot more lazy vals that can be treated this way, but I thought to keep it simple for the time being as this change is binary compatible so we can add this annotation to lazy vals as needed

### DIFF
--- a/core/src/main/scala/caliban/parsing/adt/Selection.scala
+++ b/core/src/main/scala/caliban/parsing/adt/Selection.scala
@@ -1,13 +1,15 @@
 package caliban.parsing.adt
 
 import caliban.InputValue
+import caliban.Scala3Annotations.threadUnsafe
 import caliban.parsing.adt.Type.NamedType
 
 import scala.util.hashing.MurmurHash3
 
-sealed trait Selection extends Serializable {
-  // TODO: Kept for binary compatibility
-  @transient override lazy val hashCode: Int = super.hashCode()
+sealed trait Selection extends Product with Serializable {
+
+  @transient @threadUnsafe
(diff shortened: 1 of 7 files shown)


core/src/main/scala/caliban/parsing/adt/Selection.scala
```diff
@@ -1,13 +1,15 @@
...
 
 import scala.util.hashing.MurmurHash3
 
-sealed trait Selection extends Serializable {
-  // TODO: Kept for binary compatibility
-  @transient override lazy val hashCode: Int = super.hashCode()
+sealed trait Selection extends Product with Serializable {
+
+  @transient @threadUnsafe
+  override final lazy val hashCode: Int = MurmurHash3.productHash(this)
```
kyri-petrou, line 12: Also made `Selection` extend Product here since this targets the `series/3.x` branch

ghostdogpr (approved):
Very nice that we can cross-compile it

PR #2709, opened 2025-07-13
account for supported and unsupported scalars, as well as scalarMappings, when generating ArgEncoders

dontgitit:
fixes https://github.com/ghostdogpr/caliban/pull/2708

### DIFF
--- a/tools/src/main/scala/caliban/tools/ClientWriter.scala
+++ b/tools/src/main/scala/caliban/tools/ClientWriter.scala
@@ -252,9 +252,14 @@ object ClientWriter {
           // when generating implicits, we have to be careful to not generate duplicates for any scalar types
           // otherwise, we'll get compilation errors due to multiple available implicits
           s"(implicit ${list.distinctBy {
-              case value if isTypeScalar(value.ofType) && !isScalarSupported(getSafeTypeName(value.ofType)) =>
-                writeType(typeAsString(value.ofType))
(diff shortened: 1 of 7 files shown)


tools/src/main/scala/caliban/tools/ClientWriter.scala
```diff
@@ -242,12 +249,23 @@ object ClientWriter {
...
-              .map { case (arg, idx) =>
-                s"""encoder$idx: ArgEncoder[${writeType(arg.ofType)}]"""
-              }
+          // when generating implicits, we have to be careful to not generate duplicates for any scalar types
+          // otherwise, we'll get compilation errors due to multiple available implicits
+          s"(implicit ${list.distinctBy {
+              case value if isTypeScalar(value.ofType) =>
+                val rawScalar = getTypeName(value.ofType)
+                val scalar    = safeTypeName(rawScalar)
+                val scalaType = if (isScalarSupported(scalar) || scalarMappingsWithDefaults.contains(rawScalar)) {
```
dontgitit, line 258: this can also be
```scala
val scalar = KnownPackagePrefix.replaceFirstIn(safeTypeName(getTypeName(value.ofType)), "")
val scalaType = if (isScalarSupported(scalar)) {
```

but will probably lead to more confusing errors if we missed any packages to strip...

ghostdogpr (approved):
Thanks!

PR #2711, opened 2025-07-14
No sub-directory with client name should be created in case of split-files.

alexdupre:
The client name is ignored when split-files is used.


ghostdogpr (approved):
Thanks!

PR #2640, opened 2025-05-10
Split tools module

paulpdaniels:
Discussed on Discord. Splitting the tools module into three separate modules:

- tools (contains generic tooling like SchemaLoader, SchemaComparison, IntrospectionClient)
- stitching (contains stitching related code)
- codegen (contains utils for generating scala code from graphql)

PR #2720, opened 2025-07-24
Add support for directives at the query level.

alexdupre:
Directives can be set at the query level, but caliban currently doesn't allow it.

This patch is necessary for example to support the `@inContext` directive of the Shopify Storefront API: https://shopify.dev/changelog/storefront-api-incontext-directive-supports-languages

### DIFF
--- a/client/src/main/scala/caliban/client/SelectionBuilder.scala
+++ b/client/src/main/scala/caliban/client/SelectionBuilder.scala
@@ -97,17 +97,20 @@ sealed trait SelectionBuilder[-Origin, +A] { self =>
    * Transforms a root selection into a GraphQL request.
    * @param useVariables if true, all arguments will be passed as variables (default: false)
    * @param queryName if specified, use the given query name
+   * @param directives a list of directives to be added to the query
    * @param dropNullInputValues if true, drop all null values from input object arguments (default: false)
    */
   def toGraphQL[A1 >: A, Origin1 <: Origin](
     useVariables: Boolean = false,
     queryName: Option[String] = None,
+    directives: List[Directive] = Nil,
     dropNullInputValues: Boolean = false
   )(implicit ev: IsOperation[Origin1]): GraphQLRequest = {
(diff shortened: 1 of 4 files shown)


alexdupre:
The change is API compatible, but not ABI compatible, because I've added a new last argument with default value to an existing function. Not sure what you'd like to do.

ghostdogpr:
Do you need this urgently? If not how about targeting series/3.x?

alexdupre:
No rush, I'm ok with merging it into the 3.x branch.

ghostdogpr:
I think it might take a couple more months before 3.x? There's https://github.com/ghostdogpr/caliban/pull/2640 and I think @paulpdaniels had another thing he wanted to add.

alexdupre:
That's fine. Will you take care of merging to the new branch or should I do it?

ghostdogpr:
if you can rebase to the other branch and make the PR target it that would be great!

alexdupre:
Done. The only test failure seems a spurious connectivity issue.

client/src/main/scala/caliban/client/SelectionBuilder.scala
```diff
@@ -145,13 +156,14 @@ sealed trait SelectionBuilder[-Origin, +A] { self =>
     uri: Uri,
     useVariables: Boolean = false,
     queryName: Option[String] = None,
+    directives: List[Directive] = Nil,
```
ghostdogpr, line 153: Can you modify the scaladoc above and in the other functions too?

ghostdogpr:
Actually https://github.com/ghostdogpr/caliban/pull/2723 changes things a bit how we build the query string so may be better to apply it after.

paulpdaniels:
Yeah was gonna say you may want to wait until I get my PR in first since I rewrote the query rendering

alexdupre:
Rebased.

ghostdogpr (approved):
Thanks!

PR #2723, opened 2025-07-25
Support Client fragments

paulpdaniels:
Adds support for client fragments when building `SelectionBuilder`. A large part of this PR includes a rewrite of the logic that renders a selection to a request. When adding the fragment logic, it became difficult to keep track of how the rendering worked, so I reimplemented it based on the `Renderer` we have in core. This implementation was more complex because it required a state value be propagated along with the renderer to track arguments


ghostdogpr (approved):
Very nice! Hopefully our test suite was large enough that it caught any issues 😄

PR #2771, opened 2025-10-06
Cross build for sbt 2.0.0-RC7

mrdziuban:
Sets up the `codegen-sbt` project to cross-build for sbt 2.0.0-RC7 with Scala 3.7.4.

Most of the changes are in the scripted tests, since a lot of them were explicitly looking for files in the `target/scala-2.12` directory. To make these work for both sbt 1 (using Scala 2.12) and sbt 2 (using Scala 3.7), I updated them to use globs since the target directory pattern differs between sbt 1 and 2.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -6,8 +6,9 @@ import Keys.*
 
 val scala212 = "2.12.20"
 val scala213 = "2.13.18"
-val scala3   = "3.3.7"
-val allScala = Seq(scala212, scala213, scala3)
+val scala33  = "3.3.7"
+val scala37  = "3.7.4"
+val allScala = Seq(scala212, scala213, scala33)
 
 val akkaVersion               = "2.6.20"
 val akkaHttpVersion           = "10.2.10"
@@ -62,7 +63,7 @@ inThisBuild(
       )
     ),
     versionScheme            := Some("pvp"),
-    ConsoleHelper.welcomeMessage(scala212, scala213, scala3),
+    ConsoleHelper.welcomeMessage(scala212, scala213, scala33),
     resolvers += Resolver.sonatypeCentralSnapshots
   )
 )
@@ -152,7 +153,7 @@ lazy val macros = project
   .disablePlugins(AssemblyPlugin)
   .settings(
     libraryDependencies ++= {
-      if (scalaVersion.value == scala3) {
+      if (scalaVersion.value == scala33) {
         Seq(
           "com.softwaremill.magnolia1_3" %% "magnolia" % magnoliaScala3Version
         )
@@ -291,15 +292,19 @@ lazy val codegenSbt = project
   .enablePlugins(BuildInfoPlugin)
   .disablePlugins(AssemblyPlugin)
   .settings(
-    skip             := (scalaVersion.value != scala212),
-    ideSkipProject   := (scalaVersion.value != scala212),
+    skip             := !crossScalaVersions.value.contains(scalaVersion.value),
+    ideSkipProject   := !crossScalaVersions.value.contains(scalaVersion.value),
     buildInfoKeys    := Seq[BuildInfoKey](version),
     buildInfoPackage := "caliban.codegen.sbt",
     buildInfoObject  := "BuildInfo"
   )
   .settings(
-    sbtPlugin          := true,
-    crossScalaVersions := Seq(scala212),
+    sbtPlugin                     := true,
+    crossScalaVersions            := Seq(scala212, scala37),
+    pluginCrossBuild / sbtVersion := (scalaBinaryVersion.value match {
+      case "2.12" => sbtVersion.value
+      case _      => "2.0.0-RC7"
+    }),
     libraryDependencies ++= Seq(
       "dev.zio" %% "zio-config"          % zioConfigVersion,
       "dev.zio" %% "zio-config-magnolia" % zioConfigVersion,
@@ -341,7 +346,7 @@ lazy val catsInterop = project
   .disablePlugins(AssemblyPlugin)
   .settings(
     libraryDependencies ++= {
-      if (scalaVersion.value == scala3) Seq()
+      if (scalaVersion.value == scala33) Seq()
       else Seq(compilerPlugin(("org.typelevel" %% "kind-projector" % "0.13.4").cross(CrossVersion.full)))
     } ++ Seq(
       "org.typelevel" %% "cats-effect"      % catsEffect3Version,
@@ -376,7 +381,7 @@ lazy val tapirInterop = project
   .disablePlugins(AssemblyPlugin)
   .settings(
     libraryDependencies ++= {
-      if (scalaVersion.value == scala3) Seq()
+      if (scalaVersion.value == scala33) Seq()
       else Seq(compilerPlugin(("org.typelevel" %% "kind-projector" % "0.13.4").cross(CrossVersion.full)))
     } ++
       Seq(
@@ -400,7 +405,7 @@ lazy val http4s = project
   .disablePlugins(AssemblyPlugin)
   .settings(
     libraryDependencies ++= {
-      if (scalaVersion.value == scala3) Seq()
+      if (scalaVersion.value == scala33) Seq()
       else Seq(compilerPlugin(("org.typelevel" %% "kind-projector" % "0.13.4").cross(CrossVersion.full)))
     } ++
       Seq(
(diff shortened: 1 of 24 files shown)


codegen-sbt/src/sbt-test/codegen/gen-client-task/project/plugins.sbt
```diff
@@ -3,4 +3,7 @@ sys.props.get("plugin.version") match {
   case _       => sys.error("""|The system property 'plugin.version' is not defined.
                          |Specify this property using the scriptedLaunchOpts -D.""".stripMargin)
 }
-addSbtPlugin("org.scalameta" % "sbt-scalafmt" % "2.5.5")
+
+conflictWarning := ConflictWarning("warn", Level.Warn, false)
+resolvers += Resolver.sonatypeCentralSnapshots
+addSbtPlugin("org.scalameta" % "sbt-scalafmt" % "2.5.5+25-26c3c942-SNAPSHOT")
```
mrdziuban: There's no release of `sbt-scalafmt` that supports sbt 2 RC versions yet -- https://github.com/scalameta/sbt-scalafmt/pull/395#issuecomment-3265581259

The `conflictWarning` override is necessary to work around conflicts introduced by `sbt-scalafmt`:

```
Modules were resolved with conflicting cross-version suffixes in warn:
   com.github.plokhotnyuk.jsoniter-scala:jsoniter-scala-core _3, _2.13
   org.scala-lang.modules:scala-collection-compat _3, _2.13
```

They happen because `sbt-scalafmt` [pulls in `scalafmt-dynamic`](https://github.com/scalameta/sbt-scalafmt/blob/26c3c9423deb7ad6ce78609c81aded97c9e3429d/build.sbt#L52-L53) with the `_2.13` suffix, [which pulls in `coursier`](https://github.com/scalameta/scalafmt/blob/v3.9.10/build.sbt#L77), which pulls in both [`jsoniter-scala-core`](https://github.com/coursier/coursier/blob/v2.1.24/project/modules/coursier0.sc#L12) and [`scala-collection-compat` via its dependency on `coursier-util` through `coursier-cache`](https://github.com/coursier/coursier/blob/v2.1.24/project/modules/util0.sc#L10).

mrdziuban:
I'm pretty sure the test failure is an intermittent sbt error. I saw it locally when running `scripted` and it always went away when I ran it again. I've opened an issue against sbt: https://github.com/sbt/sbt/issues/8331

I can comment out `scripted3_sbt2_jdk` for now if you'd prefer not to have a flaky test suite, just let me know

ghostdogpr:
Thanks for working on this, very nice work!
Maybe it's be better to wait for missing dependencies (fmt) and issues being resolved before merging. Or do you personally need Caliban with sbt 2?

build.sbt
```diff
@@ -6,8 +6,9 @@ import Keys.*
 
 val scala212 = "2.12.20"
 val scala213 = "2.13.16"
-val scala3   = "3.3.6"
-val allScala = Seq(scala212, scala213, scala3)
+val scala33  = "3.3.6"
```
ghostdogpr: Shall we name them `scalaLts` and `scalaNext`? (so that we don't need to rename them later)

mrdziuban: My hesitation is that Scala 3.8 will be released and become LTS soon-ish, while [based on this comment](https://github.com/sbt/sbt/issues/8073#issuecomment-3309244277) it looks like sbt will remain on Scala 3.7 to continue supporting Java 8, so the `scalaNext` name would stop making sense.

Maybe we should do `scalaLts` (which will become Scala 3.8.x) and `scala37` (which will remain whatever sbt uses)?

mrdziuban:
> Maybe it's be better to wait for missing dependencies (fmt) and issues being resolved before merging. Or do you personally need Caliban with sbt 2?

Either way is fine by me, I'm not actively using sbt 2 in my codebase yet but wanted to get this working so I could test it out. If it's all the same to you my inclination would be to merge (with the flaky test commented out) sooner rather than later to avoid maintaining the open PR for too long. [The scalafix folks confirmed that a new release is coming soon](https://discord.com/channels/632642981228314653/632683673417678860/1424869879147663502) and I'm planning to take a closer look at the sbt bug this week.

build.sbt
ghostdogpr: how about `scala3Lts` and `scala3ForSbt` then?

mrdziuban: 👍 good idea, done

ghostdogpr:
Since both fixes are coming soon I think it's okay to wait (no plan for big changes in Caliban in the near future so it should be fine)

mrdziuban:
Sounds good, I'll keep an eye out for new versions of scalafmt and sbt and upgrade accordingly. I just pushed one more commit to mark the `caliban` key as `transient` so its result isn't cached. I found this was necessary when testing the sbt 2 plugin in my codebase -- without the `transient` annotation, the generated sources weren't compiled

ghostdogpr:
The issue you had in your codebase didn't happen with our scripted tests?

mrdziuban:
No and I'm not 100% sure why -- I'm guessing it has to do with sequencing and running compile/caliban multiple times. I haven't seen any other issues in my codebase since adding `transient`.

codegen-sbt/src/sbt-test/compiletime-codegen/test-compile/build.sbt
```diff
@@ -81,6 +113,130 @@ lazy val root =
         val content    = IO.read(file(backupFile))
         val newContent = content.replace(previousValue, newValue)
         IO.write(file(initialFile), newContent)
+      },
+      TaskKey[Unit]("check-generated-files-pre-compile") := {
```
guizmaii: Amazing work!

I'm just not a huge fan of replacing the tests with this sbt code (but I guess you had to do this)

I asked to @eed3si9n if it was possible to keep the tests (almost) as they were and just reuse them both with sbt 1 & 2.
See https://x.com/guizmaii/status/1976641699569959202

the issue being the Scala version being hard coded in the tests

eed3si9n: I mostly don't use Twitter anymore. There are dozens of sbt plugins that are cross compiled to both sbt 1.x and 2.x currently. Could you open a discussion on sbt/sbt if there's specific support missing from sbt plz?

eed3si9n: See also https://www.scala-sbt.org/2.x/docs/en/changes/migrating-from-sbt-1.x.html#changes-to-target. There's glob support on `exists` in both sbt 1.x and 2.x to support cross testing.

mrdziuban: Thanks for the info @eed3si9n and @guizmaii for the push to come up with a better approach! I just pushed a commit to use globs instead of defining the checks in `build.sbt`.

There were a couple types of commands/checks in the `compiletime-codegen/test-compile` test where I still had to add logic to `build.sbt` to handle globs:

1. `copyFileWithOptions`
    1. Updated to support the first argument being a glob
    2. It enforces that the glob resolves to exactly one file
2. `newer`
    1. Defined a custom `checkFileNewer` check in `build.sbt`
    2. It supports both arguments being globs
    3. It enforces that the globs resolve to at most one file, but zero files is okay
        1. This aligns with how `newer` works in sbt, it gracefully handles either file not existing

So @eed3si9n it may be nice to add glob support to `newer`, and maybe `copy-file` and `must-mirror` as well, to further support cross-building.

mrdziuban:
I think this should be good to go now @ghostdogpr. I updated to sbt 2.0.0-RC7, which includes my fix for the intermittent test failure -- https://github.com/sbt/sbt/pull/8332 -- and sbt-scalafmt 2.5.6 (already in use on the `series/2.x` branch) is published for sbt 2, so there's no need for the snapshot version anymore.

eed3si9n:
Due to https://github.com/sbt/sbt/issues/8391, this is PR contains a plugin, I'd recommend using RC6 or next RC8.

mrdziuban:
Good to know, thanks @eed3si9n. RC6 has the issue with intermittent failures on `clean`, so I'll wait for RC8.

ghostdogpr:
@mrdziuban that's great, thanks!

I was actually about to release Caliban 3.0.0 but I think I'll wait a little bit since it would be nice to have this too. Would it be possible to rebase this branch to target `series/3.x`? There might be a few conflicts because of https://github.com/ghostdogpr/caliban/pull/2796 but the changes on codegen-sbt are small, it's mostly a change of package/imports.

mrdziuban:
> Would it be possible to rebase this branch to target `series/3.x`?

Sure thing, just rebased it. Do you want to wait for the release of sbt 2.0.0-RC8 before merging?

ghostdogpr:
I can merge this now and just wait for sbt 2.0.0-RC8 before releasing the version. Thanks!

PR #2804, opened 2025-11-26
Update zio-http to 3.6.0

SvenW:
Updating zio-http to handle [heartbeating](https://github.com/zio/zio-http/pull/3703) in SSE based subscriptions with SSE comments instead. PR coming that's using this as well


ghostdogpr:
Thanks!

Issue #2685, opened 2025-06-28
TextEventStream test is flaky


ghostdogpr:
Marked it as ignored because it fails too often.

PR #2806, opened 2025-11-27
Heartbeating for SSE based subscriptions

SvenW:
This adds heartbeating for SSE subscriptions. With the new zio-http version these heartbeats are now sent as SSE comments (ignored by the client but keeps the connection up) making sure we don't close the connection if no messages are passed within the default timeouts

### DIFF
--- a/adapters/quick/src/main/scala/caliban/quick/SseConfig.scala
+++ b/adapters/quick/src/main/scala/caliban/quick/SseConfig.scala
@@ -1,0 +1,14 @@
+package caliban.quick
+
+import zio._
+
+case class SseConfig(
+  heartbeatInterval: Option[Duration]
+) {
+  def withHeartbeatInterval(time: Duration): SseConfig =
+    copy(heartbeatInterval = Some(time))
+}
+
+object SseConfig {
+  def default: SseConfig = SseConfig(Some(10.seconds))
+}
--- a/adapters/quick/src/main/scala/caliban/quick/SseConfig.scala
+++ b/adapters/quick/src/main/scala/caliban/quick/SseConfig.scala
@@ -2,6 +2,17 @@ package caliban.quick
 
 import zio._
 
+/**
+ * Configuration settings for Server-Sent Events (SSE) connections.
+ *
+ * This configuration controls parameters like the frequency of heartbeat messages
+ * sent to keep the connection alive.
+ *
+ * @param heartbeatInterval The interval at which a 'heartbeat' (a comment line)
+ * should be sent to the client to prevent connection
+ * timeouts. [[scala.None]] means no automatic heartbeats
+ * will be sent by the server.
+ */
 case class SseConfig(
   heartbeatInterval: Option[Duration]
 ) {
--- a/adapters/quick/src/main/scala/caliban/quick/SseConfig.scala
+++ b/adapters/quick/src/main/scala/caliban/quick/SseConfig.scala
@@ -10,5 +10,5 @@ case class SseConfig(
 }
 
 object SseConfig {
-  def default: SseConfig = SseConfig(Some(10.seconds))
+  def default: SseConfig = SseConfig(None)
 }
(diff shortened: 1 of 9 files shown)


ghostdogpr:
The next release is likely to be 3.0.0 so could you target the `series/3.x` branch instead?

SvenW:
> The next release is likely to be 3.0.0 so could you target the `series/3.x` branch instead?

Of course. Do you already have a planned roadmap for the `3.0.0` release as well?

ghostdogpr:
I'm just waiting for a new RC of sbt 2 because of https://github.com/ghostdogpr/caliban/pull/2771#issuecomment-3597676680 and it should be good to release

adapters/quick/src/main/scala/caliban/quick/SseConfig.scala
```diff
@@ -0,0 +1,14 @@
...
+
+case class SseConfig(
+  heartbeatInterval: Option[Duration]
+) {
+  def withHeartbeatInterval(time: Duration): SseConfig =
+    copy(heartbeatInterval = Some(time))
+}
+
+object SseConfig {
+  def default: SseConfig = SseConfig(Some(10.seconds))
```
kyri-petrou: I'm not sure whether enabling them by default is the correct approach. I think it's better for users to explicitly enable them?

core/src/main/scala/caliban/HttpUtils.scala
```diff
@@ -61,7 +62,7 @@ private[caliban] object HttpUtils {
             case Left(err) => GraphQLResponse(ObjectValue(List(fieldName -> NullValue)), List(err))
           }
         case _                                                    => ZStream.succeed(resp)
-      }).map(v => toSse(v.toResponseValue)) ++ ZStream.succeed(done)
+      }).map(v => toSse(v.toResponseValue)).mergeHaltLeft(heartbeater.getOrElse(ZStream.empty)) ++ ZStream.succeed(done)
```
kyri-petrou: Can we avoid the merging if the heartbeater is None?

kyri-petrou (commented):
Thanks! Some minor comments below, although:
1. Is it possible to add a test for this or is it too difficult?
2. Can we document this new functionality? Both in the scaladoc of the config and the page

SvenW:
> Thanks! Some minor comments below, although:
> 
> 1. Is it possible to add a test for this or is it too difficult?

Noticed that the [TextEventStream test was flaky](https://github.com/ghostdogpr/caliban/issues/2685) so didn't look into that yet, but might be worth it?
> 2. Can we document this new functionality? Both in the scaladoc of the config and the page

Added some docs now regarding the configuration at least

kyri-petrou:
> Noticed that the https://github.com/ghostdogpr/caliban/issues/2685 so didn't look into that yet, but might be worth it?

@SvenW I see. I think it's still worth adding a test. I imagine it won't be straightforward to write a test for this, so it'd be good to have one in place. If it turns out to be flaky as well we can ignore it until the underlying bug is resolved

SvenW:
> > Noticed that the #2685 so didn't look into that yet, but might be worth it?
> 
> @SvenW I see. I think it's still worth adding a test. I imagine it won't be straightforward to write a test for this, so it'd be good to have one in place. If it turns out to be flaky as well we can ignore it until the underlying bug is resolved

Agreed. I've added a test now but I realized that sttp doesn't support SSE comments fully either so `ServerSentEvent` will just be an empty event using that client in the tests. Not sure how we should handle that in this PR though. I'll try to dig around to understand why that's the case in sttp. Looks like it's been brought up at least https://github.com/softwaremill/sttp-model/issues/379

SvenW:
What do you say @kyri-petrou? Do you want me to change anything else? Sorry for the ping btw

kyri-petrou (approved):
Thank you, LGTM 🎉

kyri-petrou:
@ghostdogpr happy to merge if you're OK with it as well

ghostdogpr:
Let's go, thanks!

PR #2818, opened 2025-12-07
Allow null variables for nullable @oneOf types

### LINKED ISSUE #2812: Nullable `@oneOf` parameterization using variable is rejected
There seems to be a misinterpretation of the specification concerning passing of `@oneOf` input objects via variables.

Relevant schema excerpt:
```
enum DatabaseOwnerType {
  Organization
  Contractor
}

input DatabaseOwnershipFilterInput @oneOf {
  ownerType: DatabaseOwnerType
  contractorIds: [ID!]
}

type Database {
  id: ID!
  name: String!
}

type DatabaseCollection {
  id: ID!
  name: String!
  databases(ownership: DatabaseOwnershipFilterInput): [Database!]
}

type Queries {
  databaseCollections: [DatabaseCollection!]
}
```

The ownership argument in `databases` field is optional, so it should be possible to use a variable of (nullable) type `DatabaseOwnershipFilterInput` to parameterize a query. So
```
query($ownership: DatabaseOwnershipFilterInput) {
  databaseCollections {
    id
    name
    databases(ownership: $ownership) {
      id
      name
    }
  }
}
```
should be admissible and should accept variable sets `{"ownership": null}` and, say, `{"ownership": {"ownerType": "Contractor"}}`.

Yet it is rejected with an error: `Variable 'ownership' cannot be nullable.`
The specification does indeed prohibit nullable variables to be used to parameterize _fields_ of `@oneOf` input objects. Here, however, it is the `@oneOf` input object itself that is parameterized, not its field.

guymers:
A nullable variable should be able to be passed to a @oneOf type that is also nullable.

fixes #2812

### DIFF
--- a/core/src/main/scala/caliban/validation/Validator.scala
+++ b/core/src/main/scala/caliban/validation/Validator.scala
@@ -368,17 +368,19 @@ object Validator {
             case _: Type.ListType  => false
           }
           val v2         =
-            if (isObjField && t.exists(_._isOneOfInput))
-              Some(
-                failWhen(v.variableType.nullable)(
-                  s"Variable '${v.name}' cannot be nullable.",
-                  "Variables used for OneOf Input Object fields must be non-nullable."
-                ) *> validateOneOfInputValue(
-                  context.variables.getOrElse(v.name, NullValue),
-                  s"Variable '${v.name}'"
-                )
-              )
-            else None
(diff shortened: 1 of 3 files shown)


ghostdogpr (approved):
Thanks!

PR #2820, opened 2025-12-09
Fix a code sample in the oneof input documentation

nigredo-tori:
`ById` and `ByName` are intended to be cases of the `AuthorInput` ADT, see the Scala 3 example below.


ghostdogpr (approved):
Thanks!

PR #2822, opened 2025-12-11
Account for directive variables even for seen fragments

ya-at:
#### Problem

Caliban incorrectly reports that the variable `y` is unused in the following GraphQL query:

```graphql
query($x: Boolean!, $y: Boolean!) {
  characters {
    ...q @include(if: $x)
    ...q @include(if: $y)
  }
}
fragment q on Character {
  __typename
  name
}
```

However, this is a valid query, as both variables (`x` and `y`) are used to control the inclusion of the same fragment (`q`). The Apollo Router can form such query to the subgraph.

This leads to unnecessary errors in Caliban, potentially blocking valid operations.

#### Solution

This PR updates Caliban's variable usage detection logic to properly recognize cases where the same fragment is included multiple times with different variables. This ensures that variables like `y` are not falsely flagged as unused.

### DIFF
--- a/core/src/main/scala/caliban/validation/Validator.scala
+++ b/core/src/main/scala/caliban/validation/Validator.scala
@@ -176,8 +176,8 @@ object Validator {
           if (set ne Nil) collectValues(set)
         case f: FragmentSpread =>
           val name = f.name
+          collectDirectives(f.directives)
           if (seen.add(name)) {
-            collectDirectives(f.directives)
             val f0 = context.fragments.getOrElseNull(name)
             if (f0 ne null) {
               collectDirectives(f0.directives)
--- a/core/src/test/scala/caliban/validation/ValidationSpec.scala
+++ b/core/src/test/scala/caliban/validation/ValidationSpec.scala
@@ -16,12 +16,18 @@ object ValidationSpec extends ZIOSpecDefault {
   private val gql         = graphQL(resolverWithSubscription)
   private val interpreter = gql.interpreter
 
+  def execute(
+    query: String,
+    variables: Map[String, InputValue] = Map.empty
+  ): IO[ValidationError, Option[CalibanError]] =
+    interpreter.flatMap(_.execute(query, variables = variables)).map(_.errors.headOption)
+
   def check(
     query: String,
     expectedMessage: String,
     variables: Map[String, InputValue] = Map.empty
   ): IO[ValidationError, TestResult] = {
-    val io = interpreter.flatMap(_.execute(query, variables = variables)).map(_.errors.headOption)
+    val io = execute(query, variables)
     io.map(assert(_)(isSome(hasField[CalibanError, String]("msg", _.msg, equalTo(expectedMessage)))))
   }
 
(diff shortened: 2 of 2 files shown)


ghostdogpr (approved):
Thanks!

ya-at:
@ghostdogpr, hi! Is there any chance the release will happen soon?

ghostdogpr:
I'm waiting for a new milestone of sbt 2 to release caliban 3.0 but you can use the latest snapshot in the meantime.

ya-at:
> I'm waiting for a new milestone of sbt 2 to release caliban 3.0 but you can use the latest snapshot in the meantime.

Thanks for the context on the 3.0 release.
For teams like ours that rely on Maven Central and cannot use snapshots, a new release containing the latest main branch would be incredibly valuable.

Since 3.0 is a major version jump, it will likely require more significant migration efforts (not sure here though).
A final patch or minor release for the 2.x series would provide a stable, up-to-date base for teams until they are ready to make that transition.

Would you consider cutting a patch or minor release from the current main?
This would bundle the recent fixes and provide a stable version for teams that cannot use snapshots until 3.0 is ready.

Of course, if the 3.0 release is imminent (e.g., in the next week), then it might not be necessary.
Could you share an approximate timeline for 3.0?

ghostdogpr:
> Thanks for the context on the 3.0 release. For teams like ours that rely on Maven Central and cannot use snapshots, a new release containing the latest main branch would be incredibly valuable.

Out of curiosity, what's the relation between relying on Maven Central and not being able to use snapshots? Caliban snapshots are uploaded to Maven Central. The latest snapshot is `2.11.1+103-41350788-SNAPSHOT`, you can depend on it and if I were to make a new release right now, it would be the exact same binary.

> Since 3.0 is a major version jump, it will likely require more significant migration efforts (not sure here though). A final patch or minor release for the 2.x series would provide a stable, up-to-date base for teams until they are ready to make that transition.

3.0 is mostly a major version change because of the sttp upgrade from 3.x to 4.x which only impacts caliban client (and the impact is not that large), plus we removed deprecated functions so I expect the upgrade to be very simple.

> Would you consider cutting a patch or minor release from the current main? This would bundle the recent fixes and provide a stable version for teams that cannot use snapshots until 3.0 is ready.
> 
> Of course, if the 3.0 release is imminent (e.g., in the next week), then it might not be necessary. Could you share an approximate timeline for 3.0?

I have no control over when the new milestone of sbt 2.x will be released. But I am not against making a new 2.x release if it takes too long. Let's wait an extra week and I will do one next weekend if it's still not there.

ya-at:
> Out of curiosity, what's the relation between relying on Maven Central and not being able to use snapshots? 

We use `https://repo.maven.apache.org/maven2`, but there are no snapshots available in the `com/github/ghostdogpr/caliban_2.13/` [directory](https://repo.maven.apache.org/maven2/com/github/ghostdogpr/caliban_2.13/) on that repository.

As I understand it, we should use `https://central.sonatype.com/repository/maven-snapshots` instead, so that snapshots are available. For example, the snapshot version `2.11.1+95-14092844-SNAPSHOT` can be found here:  
`https://central.sonatype.com/repository/maven-snapshots/com/github/ghostdogpr/caliban_2.13/2.11.1+95-14092844-SNAPSHOT/caliban_2.13-2.11.1+95-14092844-SNAPSHOT.pom`

ghostdogpr:
Yeah that's right. I guess the term "Maven Central" is ambiguous because https://central.sonatype.com/ is calling itself "Maven Central". If you use sbt you can add `resolvers += Resolver.sonatypeCentralSnapshots` so that it will find snapshots.

PR #2836, opened 2025-12-23
Optimize ArgBuilder derivation performance for case classes with many fields

ya-at:
## Problem

The current `ArgBuilder.gen` macro (based on Magnolia) generates deeply nested code for case classes with many fields,
leading to poor runtime performance (and maybe poor compile-time).

The flamegraph below shows the excessive call stack depth in the generated code.

<details>
<summary>Full flamegraph for ArgBuilder</summary>
Image

</details>

<details>
<summary>Detailed flamegraph</summary>
Image
</details>

### Performance Impact

Benchmarks using real-world case class and data show significant runtime performance degradation:

```
Benchmark                                         Mode  Cnt       Score       Error  Units
ArgBuilderBenchmark.benchmarkDecodeArgBuilder     avgt    5  265174.702 ± 19267.890  ns/op
ArgBuilderBenchmark.benchmarkDecodeFastArgBuilder avgt    5   25945.506 ±   698.670  ns/op
```

The optimized implementation is ~10x faster for the big case class.
(By the way, using pattern matching instead of `if-then-else` gives ~125'000 ns/op, so it's 4-5x slower)

### Compile-time Issues

The attached flamegraph demonstrates excessive nesting in the generated code, which impacts:
- Compilation time for schemas with complex input types
- IDE responsiveness when working with derived ArgBuilder instances
- Overall developer experience

## Solution

Introduce `ArgBuilder.genFast` — a custom macro implementation that generates optimized code
specifically tailored for Caliban's ArgBuilder typeclass.

In Scala 3, `genFast` is an alias to `gen` for API compatibility, as the Scala 3 derivation is already well-optimized.


ghostdogpr:
Hmm, so I have mixed feelings about this. While it's great if it works for you, there are several issues if we want to include this in Caliban:
- I don't think there should be 2 ways to do the same thing. If it works better than Magnolia, why keep the Magnolia one? Like why would a user use `gen` rather than `genFast`? Which means we need the same behavior and stability as the existing one.

- Curious why `ArgBuilder` and not `Schema`? Doesn't `Schema` show similar performance?

- The main reason I used Magnolia in the first place was to avoid writing macros (I don't know macros at all especially Scala 2 ones) so I worry about the maintenance burden of merging that code. For that reason this library existed in the past: https://github.com/zio-archive/caliban-deriving but went unmaintained. Also Scala 2 usage is and will be declining so it's not really something I want to invest time in.

I'd like to hear from other maintainers: @paulpdaniels @kyri-petrou thoughts on that?

ya-at:
> If it works better than Magnolia, why keep the Magnolia one? Like why would a user use `gen` rather than `genFast`?

The `gen` implementation serves as the smaller, more readable reference implementation. Its primary value lies in being easier to understand, debug, and verify for correctness. The `genFast` implementation is an optimized version that should, in theory, produce identical results but through a more complex, high-performance mechanism.

Therefore, a user would choose `gen` when priority is placed on transparency and simplicity, such as during development, debugging, or for use cases where absolute peak performance is not critical. The `genFast` implementation should be chosen when its performance characteristics are required (e.g. for huge case classes), under the assumption it behaves identically to the reference implementation (though currently, there is a known divergence in behaviour for `enum` types).

So, the default is `gen`, but for places where performance matters choose `genFast`.

> Also Scala 2 usage is and will be declining so it's not really something I want to invest time in.

I had the same thought while creating the PR. However, according to the [VirtusLab survey](https://lp.virtuslab.com/wp-content/uploads/2025/02/Scala-Projects-Maintenance-Report.pdf) (page 11), 70% of projects still use Scala 2.

> Curious why ArgBuilder and not Schema? Doesn't Schema show similar performance?

I looked at `ArgBuilder` because we saw a tall tower in the flamegraph produced by it. With `Schema`, we didn't have such issues.

As I understand it, `Schema.gen` uses a mutable `Array` under the hood. That's interesting; I'll explore this area. Maybe it's possible to rewrite the current `ArgBuilder.gen` using only Magnolia and get some performance benefits (although I believe a custom macro is the fastest option anyway).

kyri-petrou:
@ya-at would it be possible to provide the code you used for the benchmark? I'd like to try out a few things.

The part that surprises me a lot is that `Schema` derivation doesn't seem to have the same issue (or at least, nobody reported it so far) and just want to make sure that there isn't some issue with how we use magnolia for the `ArgBuilder`

kyri-petrou:
@ya-at could you take a look at #2839? If you could benchmark that code to see if that fixes the issue that'd be great

ya-at:
> @ya-at could you take a look at #2839? If you could benchmark that code to see if that fixes the issue that'd be great

Here are my benchmark results using synthetic data (benchmark source code is in this PR).
I also [replaced](https://github.com/ghostdogpr/caliban/pull/2836/files#diff-a8bfb49a61601b08740c2a8b2c4288cf3602bfa76c99c5695d2fbd304daba3b1R55-R75) `constructMonadic` with `rawConstruct` (similar to what's in #2839) to measure the performance difference, and the improvement is impressive: #2839 looks very good.

**Results with `constructMonadic`:**
```
Benchmark                                        Mode  Cnt       Score       Error  Units
FastArgBuilderBenchmark.benchmarkArgBuilder      avgt    5  164863.195 ± 11697.795  ns/op
FastArgBuilderBenchmark.benchmarkFastArgBuilder  avgt    5   16815.370 ±   265.832  ns/op
```

**Results with `rawConstruct`:**
```
Benchmark                                        Mode  Cnt      Score     Error  Units
FastArgBuilderBenchmark.benchmarkArgBuilder      avgt    5  25859.346 ± 617.367  ns/op
FastArgBuilderBenchmark.benchmarkFastArgBuilder  avgt    5  16968.020 ± 258.527  ns/op
```

Although my macro is somewhat complicated, it offers much better performance. Therefore, I think we should close this PR and use `rawConstruct`.
My next step is to validate these results with real data and investigate the remaining performance difference between the builders.

kyri-petrou:
@ya-at I rerun the benchmark in your PR using the code that was merged in (I saw that in your benchmarks you used a different code that relies on `constructRaw`). AFAICT the current implementation is as fast as it can be:

```
[info] Benchmark                                        Mode  Cnt      Score      Error  Units
[info] FastArgBuilderBenchmark.benchmarkArgBuilder      avgt    5  11557.946 ±   53.337  ns/op
[info] FastArgBuilderBenchmark.benchmarkFastArgBuilder  avgt    5  15180.090 ± 1139.583  ns/op
```

I think it's safe at this point to assume that the issue was with `constructMonadic`, and that the magnolia derivation doesn't add much / any overhead compared to a macro-based implementation.

ya-at:
> @ya-at I rerun the benchmark in your PR using the code that was merged in (I saw that in your benchmarks you used a different code that relies on `constructRaw`). AFAICT the current implementation is as fast as it can be:
> 
> ```
> [info] Benchmark                                        Mode  Cnt      Score      Error  Units
> [info] FastArgBuilderBenchmark.benchmarkArgBuilder      avgt    5  11557.946 ±   53.337  ns/op
> [info] FastArgBuilderBenchmark.benchmarkFastArgBuilder  avgt    5  15180.090 ± 1139.583  ns/op
> ```
> 
> I think it's safe at this point to assume that the issue was with `constructMonadic`, and that the magnolia derivation doesn't add much / any overhead compared to a macro-based implementation.

Yes, the macro-based implementation is hard to maintain (and for some reason, it's slower in this benchmark—perhaps because of the `lazy val`).

That said, the macro also checks the body of `@GQLDefault`. Is it possible to validate it at compile time using Magnolia too?

kyri-petrou:
> Is it possible to validate it at compile time using Magnolia too?

@ya-at as far as I know, no. But what we've been doing with Scala 3 is adding small macros (that are easy to maintain) that do compile-time checks alongside the generic derivation.

Unfortunately we can't do the same for Scala 2 but I think that's a price we're willing to pay in order to keep the code as maintainable as possible given that none of the core maintainers is experienced with Scala 2 macros and given that we some times need to extend the derivation code to support new features (e.g., `oneOfInput`s)

PR #2839, opened 2025-12-25
Align Scala 2 & 3 `ArgBuilder` derivation code

kyri-petrou:
This PR is an attempt at fixing the issue described in #2836 by aligning the `ArgBuilder` derivation code with Scala 3. Instead of using `constructMonadic` (which seems to be very prominent in the flamegraphs provided in the PR, we manually construct the argument list and pass it to `rawConstruct`.

Hopefully this fixes the issue, but it's hard to confirm without the benchmarking code / repro.

@ghostdogpr I targeted series/2.x for this in case you want to release it early

### DIFF
--- a/core/src/main/scala-2/caliban/schema/ArgBuilderDerivation.scala
+++ b/core/src/main/scala-2/caliban/schema/ArgBuilderDerivation.scala
@@ -54,9 +54,10 @@ trait CommonArgBuilderDerivation {
       }
 
     private def fromFields(fields: Map[String, InputValue]): Either[ExecutionError, T] = {
-      var i    = 0
-      val l    = params.length
-      val arr  = Array.ofDim[Any](l)
+      val params = this.params
+      var i      = 0
+      val l      = params.length
+      val arr    = Array.ofDim[Any](l)
       while (i < l) {
         val (label, default, builder) = params(i)
         val field                     = fields.getOrElseNull(label)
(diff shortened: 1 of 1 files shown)


core/src/main/scala-2/caliban/schema/ArgBuilderDerivation.scala
```diff
@@ -16,6 +16,7 @@ trait CommonArgBuilderDerivation {
 
   type EitherExecutionError[A] = Either[ExecutionError, A]
 
+  @deprecated("Kept for binary compatibility, to be removed")
```
kyri-petrou, line 19: @ghostdogpr we can remove this implicit in `series/3.x`

ghostdogpr (approved):
Looks good! Let's see if @ya-at confirms the improvement.

core/src/main/scala-2/caliban/schema/ArgBuilderDerivation.scala
ghostdogpr: yep, I will

ya-at:
> Looks good! Let's see if @ya-at confirms the improvement.

Yes, very good.
The benchmark results of rawConstruct vs constructMonadic ([source](https://github.com/ghostdogpr/caliban/pull/2836#issuecomment-3691391241)):
```
Benchmark                                        Mode  Cnt       Score       Error  Units
FastArgBuilderBenchmark.constructMonadic         avgt    5  164863.195 ± 11697.795  ns/op
FastArgBuilderBenchmark.rawConstruct             avgt    5   25859.346 ±   617.367  ns/op
```

PR #2846, opened 2026-01-02
VariablesCoercer was being too clingy with other operations' variables

li-nkSN:
caliban/core/src/main/scala/caliban/parsing/VariablesCoercer.scala (line 48)

  val variableDefinitions = doc.operationDefinitions.flatMap(_.variableDefinitions)

  This collects variable definitions from ALL operations in the document, not just the one being executed.

  The Execution Flow

  1. GraphQL.scala:138 - coerceVariables(doc, request.variables) is called
  2. Note: request.operationName is available but NOT passed to coerceVariables
  3. GraphQL.scala:152 - VariablesCoercer.coerceVariables(variables, doc, ...) - no operationName parameter
  4. VariablesCoercer.scala:48 - collects ALL variable definitions across ALL operations
  5. Lines 52-94 validate that ALL variables have values, causing the bug

  The Fix Required

  1. Pass operationName from GraphQLRequest to coerceVariables
  2. Modify VariablesCoercer.coerceVariables to filter variable definitions to only the executed operation

### DIFF
--- a/core/src/main/scala/caliban/GraphQL.scala
+++ b/core/src/main/scala/caliban/GraphQL.scala
@@ -124,14 +124,20 @@ trait GraphQL[-R] { self =>
               )(overallWrappers, request)
           }
 
-        private def coerceVariables(doc: Document, variables: Map[String, InputValue], operationName: Option[String])(implicit
-          trace: Trace
+        private def coerceVariables(doc: Document, variables: Map[String, InputValue], operationName: Option[String])(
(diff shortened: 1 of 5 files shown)


ghostdogpr:
This change is not binary compatible as shown by the Mima failure so it will have to be in 3.x so I reviewed https://github.com/ghostdogpr/caliban/pull/2847. I also am not planning to releasing 2.x unless a severe bug is detected.

PR #2866, opened 2026-01-31
TypeVisitor loops through arguments

hwielenberg:
Fixes a bug where the TypeVisitor/Transformer did not loop through arguments of fields.

### DIFF
--- a/core/src/main/scala/caliban/GraphQL.scala
+++ b/core/src/main/scala/caliban/GraphQL.scala
@@ -69,7 +69,7 @@ trait GraphQL[-R] { self =>
         schemaBuilder.mutation.flatMap(_.opType.name),
         schemaBuilder.subscription.flatMap(_.opType.name),
         schemaBuilder.schemaDescription
-      ) :: schemaBuilder.types.map(transformer.typeVisitor.visit).flatMap(_.toTypeDefinition) ++ additionalDirectives
+      ) :: schemaBuilder.types.flatMap(_.toTypeDefinition) ++ additionalDirectives
         .map(_.toDirectiveDefinition),
       SourceMapper.empty
     )
--- a/core/src/main/scala/caliban/transformers/Transformer.scala
+++ b/core/src/main/scala/caliban/transformers/Transformer.scala
@@ -39,11 +39,9 @@ abstract class Transformer[-R] { self =>
       case other                                                          => other
     }
 
-  @deprecated("Use transformStep instead")
-  def apply[R1 <: R](step: ObjectStep[R1], field: Field): ObjectStep[R1] = transformStep(step, field) match {
-    case os: ObjectStep[R1] => os
-    case _                  => throw new IllegalStateException("Transformer.apply expected an ObjectStep after transformation")
-  }
+  @deprecated("Use transform instead")
+  def apply[R1 <: R](step: ObjectStep[R1], field: Field): ObjectStep[R1] =
+    if (materializedTypeNames(step.name)) transformObjectStep(step, field) else step
 
   @deprecated("Use transformObjectStep instead")
   protected def transformStep[R1 <: R](step: ObjectStep[R1], field: Field): ObjectStep[R1] =
--- a/core/src/main/scala/caliban/transformers/Transformer.scala
+++ b/core/src/main/scala/caliban/transformers/Transformer.scala
@@ -1,11 +1,11 @@
 package caliban.transformers
 
-import caliban.InputValue
+import caliban.{ InputValue, ResponseValue, Value }
 import caliban.execution.Field
 import caliban.introspection.adt._
 import caliban.parsing.adt.Directive
 import caliban.schema.Annotations.GQLDirective
-import caliban.schema.Step
+import caliban.schema.{ PureStep, Step }
 import caliban.schema.Step.{ FunctionStep, MetadataFunctionStep, NullStep, ObjectStep }
 
 import scala.collection.compat._
@@ -28,12 +28,14 @@ abstract class Transformer[-R] { self =>
   protected def transformObjectStep[R1 <: R](step: ObjectStep[R1], field: Field): ObjectStep[R1] = step
 
   protected def transformFunctionStep[R1 <: R](step: FunctionStep[R1], field: Field): FunctionStep[R1] = step
+  protected def transformPureStep(step: PureStep, field: Field): PureStep                              = step
 
   def transform[R1 <: R](step: Step[R1], field: Field): Step[R1] =
     step match {
       case objStep: ObjectStep[R1] if materializedTypeNames(objStep.name) => transformObjectStep(objStep, field)
       case objectStep: ObjectStep[R1]                                     => objectStep
       case funcStep: FunctionStep[R1]                                     => transformFunctionStep(funcStep, field)
+      case pureStep: PureStep                                             => transformPureStep(pureStep, field)
       case other                                                          => other
     }
 
@@ -84,7 +86,7 @@ object Transformer {
       if (f.isEmpty) Empty else new RenameType(f.toMap)
   }
 
-  final private class RenameType(map: Map[String, String]) extends Transformer[Any] {
+  final private class RenameType(map: Map[String, String]) extends Transformer[Any] with TransformFunctionStepAdapter {
 
     val typeVisitor: TypeVisitor = {
       val renameType = { (t: __Type) =>
@@ -97,6 +99,7 @@ object Transformer {
       TypeVisitor.modify(renameType) |+| TypeVisitor.enumValues.modify(renameEnum)
     }
 
+    private val inverseMap                       = map.map(_.swap)
     protected val typeNames: Set[String] = map.keySet
 
     override protected def transformObjectStep[R1 <: Any](step: ObjectStep[R1], field: Field): ObjectStep[R1] =
@@ -104,6 +107,39 @@ object Transformer {
         case null    => step
         case newName => step.copy(name = newName)
       }
+
+    override protected def transformPureStep(step: PureStep, field: Field): PureStep =
+      step.value match {
+        case value: Value.EnumValue =>
+          map.get(value.value) match {
+            case Some(newEnumName) if field.fieldType.innerType.allEnumValues.exists(_.name == newEnumName) =>
+              PureStep(Value.StringValue(newEnumName))
+            case _                                                                                          => step
+          }
+        case _                      => step
+      }
+
+    protected def transformArg(value: InputValue, definition: __InputValue): InputValue = {
+      val innerType = definition._type.innerType
+      value match {
+        case InputValue.ListValue(values)   => InputValue.ListValue(values.map(transformArg(_, definition)))
+        case InputValue.ObjectValue(fields) =>
+          InputValue.ObjectValue(fields.view.map { case (k, v) =>
+            val memberDef = innerType.allInputFields.find(_.name == k)
+            val newValue  = memberDef.map(transformArg(v, _)).getOrElse(v)
+            k -> newValue
+          }.toMap)
+
+        case Value.EnumValue(enumValue) =>
+          inverseMap.get(enumValue) match {
+            case Some(oldEnumValue) if definition._type.innerType.allEnumValues.exists(_.name == enumValue) =>
+              Value.EnumValue(oldEnumValue)
+            case _                                                                                          => value
+          }
+        case _                          => value
+      }
+    }
+
   }
 
   object RenameField {
@@ -125,7 +161,9 @@ object Transformer {
       if (f.isEmpty) Empty else new RenameField(tuplesToMap2(f: _*))
   }
 
-  final private class RenameField(visitorMap: Map[String, Map[String, String]]) extends Transformer[Any] {
+  final private class RenameField(visitorMap: Map[String, Map[String, String]])
+      extends Transformer[Any]
+      with TransformFunctionStepAdapter {
     private val transformMap = swapMap2(visitorMap)
 
     val typeVisitor: TypeVisitor = {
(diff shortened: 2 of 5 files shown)


core/src/main/scala/caliban/GraphQL.scala
```diff
@@ -69,7 +69,7 @@ trait GraphQL[-R] { self =>
         schemaBuilder.mutation.flatMap(_.opType.name),
         schemaBuilder.subscription.flatMap(_.opType.name),
         schemaBuilder.schemaDescription
-      ) :: schemaBuilder.types.map(transformer.typeVisitor.visit).flatMap(_.toTypeDefinition) ++ additionalDirectives
+      ) :: schemaBuilder.types.flatMap(_.toTypeDefinition) ++ additionalDirectives
```
hwielenberg, line 72: I'm not sure why this was added - but it doesn't seem right to me:
* All the types should have already been transformed. So this does it just again. (Unless there is another mismatch between the types returned by `schemaBuilder.types` and which types are used in `caliban.introspection.adt.TypeVisitor#visit`
* This is really bad performance wise. We recursively iterate through the tree of each type for each type. So this is O(n^2)

But including it would also cause the test to pass.

core/src/test/scala/caliban/transformers/TransformerSpec.scala
```diff
@@ -40,6 +40,49 @@ object TransformerSpec extends ZIOSpecDefault {
...
+        val transformed: GraphQL[Any] = api.transform(
+          Transformer.RenameType(
+            "Inactive" -> "Archived"
+          )
+        )
+
+        val rendered = transformed.render
+        for {
+          interpreter <- transformed.interpreter
+          result      <- interpreter.execute("""{ a(byStateFilter: {states:[Active]}) }""").map(_.data.toString) // Providing either Inactive or Archived leads to different errors...
```
hwielenberg: >// Providing either Inactive or Archived leads to different errors...

I think this is a different bug. If the `typeVisitor` part of the transformer allows for renames to enums the `transformStep` probably should make appropriate changes as well to the enum values.

core/src/main/scala/caliban/GraphQL.scala
ghostdogpr: Yeah I think you're right, there is no need to do it twice.

core/src/test/scala/caliban/transformers/TransformerSpec.scala
ghostdogpr: Yeah looks like currently `transformStep` does nothing about enums, only objects.

core/src/main/scala/caliban/GraphQL.scala
kyri-petrou: Weird, this was added recently via a PR that doesn't seem very relevant (see https://github.com/ghostdogpr/caliban/pull/2510/changes#diff-afbb4c407919be09385f8e83a79d1305ad3138f92f3c40331ba39b76e4b5a41aR72)

@paulpdaniels do you happen to recall why you added this transformation here?

kyri-petrou: By the way it's worth noting that this is a `lazy val`, so the performance cost should me amortised (assuming that the `GraphQL` object is not being created dynamically)

paulpdaniels: I don't remember the exact issue, it related to how the schema got rendered with the node interface types. I _think_ it was because the visit code wasn't actually applied previously when doing schema rendering. An easy tell though should be to remove it and see if the relay specs still pass

hwielenberg: All tests pass in this PR. 

I checked out ca3fdc4570dad97536e9cfd52fe611900fa7e350 reverted this change and ran all tests again. All tests still pass.

hwielenberg:
I added a bunch more tests and found bugs with:
  * renaming input fields - which I fixed 
  * renaming enums - both as input and output. working on a fix

core/src/main/scala/caliban/transformers/Transformer.scala
```diff
@@ -23,10 +23,29 @@ abstract class Transformer[-R] { self =>
...
-  protected def transformStep[R1 <: R](step: ObjectStep[R1], field: Field): ObjectStep[R1]
+  private lazy val materializedTypeNames = typeNames
 
-  def apply[R1 <: R](step: ObjectStep[R1], field: Field): ObjectStep[R1] =
-    transformStep(step, field)
+  protected def transformObjectStep[R1 <: R](step: ObjectStep[R1], field: Field): ObjectStep[R1] = step
+
+  protected def transformFunctionStep[R1 <: R](step: FunctionStep[R1], field: Field): FunctionStep[R1] = step
+
+  def transform[R1 <: R](step: Step[R1], field: Field): Step[R1] =
```
hwielenberg, line 33: Adding `transformFunctionStep` is not strictly necessary as I can get to the FunctionStep through the ObjectStep. But then the short circuit (`materializedTypeNames`) becomes an issue.

Unless I somehow collect all possible InputTypes that are nested within an ObjectType there is no way to know which typeNames are relevant for renaming input fields.

kyri-petrou:
LMK when this PR is ready for review and tests passing; I'd like to take a look at it

hwielenberg:
@kyri-petrou It's ready.

hwielenberg:
@kyri-petrou Friendly reminder 😄

core/src/main/scala/caliban/transformers/Transformer.scala
```diff
@@ -23,10 +23,29 @@ abstract class Transformer[-R] { self =>
...
+      case pureStep: PureStep                                             => transformPureStep(pureStep, field)
+      case other                                                          => other
+    }
+
+  @deprecated("Use transform instead")
   def apply[R1 <: R](step: ObjectStep[R1], field: Field): ObjectStep[R1] =
-    transformStep(step, field)
+    if (materializedTypeNames(step.name)) transformObjectStep(step, field) else step
+
+  @deprecated("Use transformObjectStep instead")
```
kyri-petrou, line 46: Add `@deprecatedOverriding` and add version

```scala
  @deprecated("Use transformObjectStep instead", "3.0.1")
  @deprecatedOverriding("Override transformObjectStep instead", "3.0.1")
```

core/src/main/scala/caliban/transformers/Transformer.scala
```diff
@@ -23,10 +23,29 @@ abstract class Transformer[-R] { self =>
...
+  def transform[R1 <: R](step: Step[R1], field: Field): Step[R1] =
+    step match {
+      case objStep: ObjectStep[R1] if materializedTypeNames(objStep.name) => transformObjectStep(objStep, field)
+      case objectStep: ObjectStep[R1]                                     => objectStep
+      case funcStep: FunctionStep[R1]                                     => transformFunctionStep(funcStep, field)
+      case pureStep: PureStep                                             => transformPureStep(pureStep, field)
+      case other                                                          => other
+    }
+
+  @deprecated("Use transform instead")
```
kyri-petrou, line 42: Same

core/src/main/scala/caliban/transformers/Transformer.scala
```diff
@@ -47,8 +66,6 @@ object Transformer {
     val typeVisitor: TypeVisitor = TypeVisitor.empty
 
     protected val typeNames: Set[String] = Set.empty
-
-    protected def transformStep[R1](step: ObjectStep[R1], field: Field): ObjectStep[R1] = step
```
kyri-petrou, line 51: Override `def transform` so that we no-op for the `Empty` transformer

core/src/main/scala/caliban/transformers/Transformer.scala
```diff
@@ -80,13 +97,47 @@ object Transformer {
...
+    override protected def transformObjectStep[R1 <: Any](step: ObjectStep[R1], field: Field): ObjectStep[R1] =
       map.getOrElse(step.name, null) match {
         case null    => step
         case newName => step.copy(name = newName)
       }
+
+    override protected def transformPureStep(step: PureStep, field: Field): PureStep =
+      step.value match {
+        case value: Value.EnumValue =>
+          map.get(value.value) match {
```
kyri-petrou, line 112: You can `import caliban.syntax._`, and then use the `getOrElseNull` extension method to avoid allocations of `Some` (in other places as well)

core/src/main/scala/caliban/transformers/Transformer.scala
```diff
@@ -80,13 +97,47 @@ object Transformer {
...
+          }
+        case _                      => step
+      }
+
+    protected def transformArg(value: InputValue, definition: __InputValue): InputValue = {
+      val innerType = definition._type.innerType
+      value match {
+        case InputValue.ListValue(values)   => InputValue.ListValue(values.map(transformArg(_, definition)))
+        case InputValue.ObjectValue(fields) =>
+          InputValue.ObjectValue(fields.view.map { case (k, v) =>
```
kyri-petrou, line 125: `fields.transform` is more efficient than `view.map{...}.toMap` (Check in the rest of the code as well, there are a few places this can be changed)

core/src/main/scala/caliban/transformers/Transformer.scala
```diff
@@ -128,11 +181,33 @@ object Transformer {
...
+    protected def transformArg(value: InputValue, definition: __InputValue): InputValue = {
+      val innerType = definition._type.innerType
+      innerType.name
+        .flatMap(transformMap.get)
+        .map { mapping =>
+          value match {
+            case InputValue.ListValue(values)   =>
+              InputValue.ListValue(values.map(transformArg(_, definition)))
+            case InputValue.ObjectValue(fields) =>
+              InputValue.ObjectValue(fields.view.map { case (k, v) =>
```
kyri-petrou, line 199: No need for `view.map { ... }.toMap`. Just `.map` is enough

core/src/main/scala/caliban/transformers/Transformer.scala
```diff
@@ -80,13 +97,47 @@ object Transformer {
...
+      }
+
+    protected def transformArg(value: InputValue, definition: __InputValue): InputValue = {
+      val innerType = definition._type.innerType
+      value match {
+        case InputValue.ListValue(values)   => InputValue.ListValue(values.map(transformArg(_, definition)))
+        case InputValue.ObjectValue(fields) =>
+          InputValue.ObjectValue(fields.view.map { case (k, v) =>
+            val memberDef = innerType.allInputFields.find(_.name == k)
+            val newValue  = memberDef.map(transformArg(v, _)).getOrElse(v)
```
kyri-petrou, line 127: There are a few usages of `.map { ... }. getOrElse(...)` that can be substituted with `fold`

core/src/main/scala/caliban/transformers/Transformer.scala
```diff
@@ -402,13 +477,37 @@ object Transformer {
...
+          introspectionField <- parentType.allFields.find(_.name == field.name)
+        } yield transformArgs(args, introspectionField.allArgs)).getOrElse(args)
+
+        original.step(updatedArgs)
+      }
+
+    private def transformArgs(args: Map[String, InputValue], allArgs: List[__InputValue]): Map[String, InputValue] =
+      args.view.map { case (argName, input) =>
+        argName -> allArgs.find(_.name == argName).fold(input)(transformArg(input, _))
+      }.toMap
```
kyri-petrou, line 507: Given how much in the hot-path this is, I would make the following change to `__Field`:

```scala
  private[caliban] lazy val allArgsMap = allArgs.view.map(a => a.name -> a).toMap

  private[caliban] lazy val allArgNames: Set[String] = allArgsMap.keySet
```

And then refactor these 2 methods as:

```scala
    override protected def transformFunctionStep[R1 <: Any](
      original: FunctionStep[R1],
      field: Field
    ): FunctionStep[R1] =
      FunctionStep[R1] { args =>
        val newArgs = field.parentType match {
          case Some(parentType) =>
            parentType.getFieldOrNull(field.name) match {
              case null       => args
              case introField => transformArgs(args, introField.allArgsMap)
            }
          case _                => args
        }
        original.step(newArgs)
      }

    private def transformArgs(
      args: Map[String, InputValue],
      allArgs: Map[String, __InputValue]
    ): Map[String, InputValue] =
      args.transform { (argName, input) =>
        allArgs.getOrElseNull(argName) match {
          case null  => input
          case value => transformArg(input, value)
        }
      }
```

core/src/main/scala/caliban/transformers/Transformer.scala
```diff
@@ -402,13 +477,37 @@ object Transformer {
...
+
+    override protected def transformObjectStep[R1 <: R](step: ObjectStep[R1], field: Field): ObjectStep[R1] =
+      right.transformObjectStep(left.transformObjectStep(step, field), field)
+
+    override protected def transformPureStep(step: PureStep, field: Field): PureStep =
+      right.transformPureStep(left.transformPureStep(step, field), field)
+  }
+
+  private trait TransformFunctionStepAdapter extends Transformer[Any] {
+    override protected def transformFunctionStep[R1 <: Any](
```
kyri-petrou, line 491: Also make this `final`

core/src/main/scala/caliban/transformers/Transformer.scala
```diff
@@ -23,10 +23,29 @@ abstract class Transformer[-R] { self =>
    */
   protected def typeNames: collection.Set[String]
 
-  protected def transformStep[R1 <: R](step: ObjectStep[R1], field: Field): ObjectStep[R1]
+  private lazy val materializedTypeNames = typeNames
```
kyri-petrou, line 26: I'd remove this and define `typeNames` as `lazy val` where needed (I believe in the `CombinedTransformer` only).

kyri-petrou (commented):
I left a few comments, mostly for performance reasons as this PR adds quite a bit of overhead to these 2 Transformers and we want to minimize its impact as much as possible

PR #2869, opened 2026-02-02
Better error handling for stitching.RemoteResolver

hwielenberg:
If the proxy server returns gql errors they are no longer silently swallowed but instead thrown as an ExecutionError.

Improves error messages for similar errors.


ghostdogpr (approved):
Thanks!

PR #2874, opened 2026-02-06
add a built-in `ArgBuilder` for enums

dontgitit:
- adds two new `enumString` methods on `ArgBuilder` to that accept `map`/`flatMap`-like functions and apply them to a graphql enum value
- adds a new `enumJava` method on `ArgBuilder` meant to automatically construct an `ArgBuilder[T]` for some `java.lang.Enum[T]`

### DIFF
--- a/core/src/main/scala/caliban/schema/ArgBuilder.scala
+++ b/core/src/main/scala/caliban/schema/ArgBuilder.scala
@@ -153,6 +153,10 @@ trait ArgBuilderInstances extends ArgBuilderDerivation {
     case BooleanValue(value) => Right(value)
     case other               => Left(InvalidInputArgument("Boolean", other))
   }
+  lazy val enum: ArgBuilder[String]                    = {
+    case EnumValue(value) => Right(value)
+    case other            => Left(InvalidInputArgument("Enum", other))
+  }
 
   private abstract class TemporalDecoder[A](name: String) extends ArgBuilder[A] {
     protected[this] def parseUnsafe(input: String): A
--- a/core/src/main/scala/caliban/schema/ArgBuilder.scala
+++ b/core/src/main/scala/caliban/schema/ArgBuilder.scala
@@ -153,7 +153,7 @@ trait ArgBuilderInstances extends ArgBuilderDerivation {
     case BooleanValue(value) => Right(value)
     case other               => Left(InvalidInputArgument("Boolean", other))
   }
-  lazy val enum: ArgBuilder[String]                    = {
+  lazy val `enum`: ArgBuilder[String]                  = {
     case EnumValue(value) => Right(value)
     case other            => Left(InvalidInputArgument("Enum", other))
   }
--- a/core/src/main/scala/caliban/schema/ArgBuilder.scala
+++ b/core/src/main/scala/caliban/schema/ArgBuilder.scala
@@ -15,6 +15,7 @@ import scala.annotation.implicitNotFound
 import scala.collection.immutable.ListMap
 import scala.collection.mutable
 import scala.collection.mutable.ListBuffer
+import scala.reflect.ClassTag
 import scala.util.Try
 import scala.util.control.{ NoStackTrace, NonFatal }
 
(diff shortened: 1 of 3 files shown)


core/src/main/scala/caliban/schema/ArgBuilder.scala
```diff
@@ -153,6 +153,10 @@ trait ArgBuilderInstances extends ArgBuilderDerivation {
     case BooleanValue(value) => Right(value)
     case other               => Left(InvalidInputArgument("Boolean", other))
   }
+  lazy val enum: ArgBuilder[String]                    = {
```
dontgitit: can also potentially provide something like this for Java enums:
```scala
  implicit def enumArgBuilder[T <: java.lang.Enum[T]: ClassTag]: ArgBuilder[T] =
    enum.map(v => java.lang.Enum.valueOf[T](summon[ClassTag[T]].runtimeClass.asInstanceOf[Class[T]], v))
```

not entirely sure how to do that for Scala enums since `scala.reflect.Enum` doesn't provide `valueOf` (but I think actual subclasses do?)

dontgitit: or  maybe something like this:
```scala
implicit def enumArgBuilder[T <: java.lang.Enum[T]: ClassTag]: ArgBuilder[T] =
  `enum`.flatMap { value =>
    val cls = summon[ClassTag[T]].runtimeClass.asInstanceOf[Class[T]]
    Try {
      java.lang.Enum.valueOf[T](cls, value)
    }.toEither.left.map { t =>
      ExecutionError(s"'$value' is not a valid value of ${cls.getSimpleName}", innerThrowable = Some(t))
    }
  }
```

core/src/main/scala/caliban/schema/ArgBuilder.scala
```diff
@@ -153,6 +153,10 @@ trait ArgBuilderInstances extends ArgBuilderDerivation {
     case BooleanValue(value) => Right(value)
     case other               => Left(InvalidInputArgument("Boolean", other))
   }
+  lazy val `enum`: ArgBuilder[String]                  = {
```
ghostdogpr: Is the goal to use this + `map` for enums?

dontgitit: that's correct, that was the intention!

i guess maybe to make that clearer, it should be something like the following...?

```scala
private lazy val enumBuilder: ArgBuilder[String]                  = {
   case EnumValue(value) => Right(value)
   case other            => Left(InvalidInputArgument("Enum", other))
}

def `enum`[A](f: T => A): ArgBuilder[A] = enumBuilder.map(f)

def `enum`[A](f: T => Either[ExecutionError, A]): ArgBuilder[A] = enumBuilder.flatMap(f)
```

ghostdogpr: That would be `f: String => ...`, right? Maybe we can call it `enumString` and drop the backticks.

dontgitit: sorry, yes, the `T` was meant to be `String`

core/src/main/scala/caliban/schema/ArgBuilder.scala
```diff
@@ -153,6 +154,25 @@ trait ArgBuilderInstances extends ArgBuilderDerivation {
     case BooleanValue(value) => Right(value)
     case other               => Left(InvalidInputArgument("Boolean", other))
   }
+  private lazy val enumBuilder: ArgBuilder[String]     = {
+    case EnumValue(value) => Right(value)
+    case other            => Left(InvalidInputArgument("Enum", other))
+  }
+
+  def enumString[A](f: String => A)(implicit d: DummyImplicit): ArgBuilder[A] = enumBuilder.map(f)
```
dontgitit, line 162: added `DummyImplicit` to resolve:
```
double definition:
[error] def enumString[A](f: String => Either[caliban.CalibanError.ExecutionError,A]): caliban.schema.ArgBuilder[A] in trait ArgBuilderInstances and
[error] def enumString[A](f: String => A): caliban.schema.ArgBuilder[A] in trait ArgBuilderInstances
[error] have same type after erasure: (f: Function1): caliban.schema.ArgBuilder
[error] trait AutoArgBuilderDerivation extends ArgBuilderInstances with LowPriorityDerivedArgBuilder
[error]       ^
[error] one error found
```

core/src/main/scala/caliban/schema/ArgBuilder.scala
```diff
@@ -153,6 +154,25 @@ trait ArgBuilderInstances extends ArgBuilderDerivation {
...
+    Try {
+      Enum.valueOf[T](cls, value)
+    }.toEither.left.map { t =>
+      val validValues = cls.getEnumConstants.map(_.name()).mkString(", ")
+      ExecutionError(
+        s"'$value' is not a valid value of ${cls.getSimpleName}. Valid values are: [$validValues]",
+        innerThrowable = Some(t)
+      )
+    }
+  }
```
dontgitit, line 175: added an implementation for java enums but didn't make it `implicit`
also not sure if exposing the valid values in the error message is a good idea

dontgitit: LLMs say this is a better way for Scala 3:
```scala
import scala.deriving.Mirror
import scala.compiletime.{constValueTuple, summonInline}

// This works for any Scala 3 Enum (Mirror.SumOf)
inline def enumFromStringScala3[T](value: String)(using m: Mirror.SumOf[T]): Either[String, T] =
  // 1. Get all label names (e.g., "Red", "Green", "Blue") at compile time
  val labels = constValueTuple[m.MirroredElemLabels].toList.asInstanceOf[List[String]]
  
  // 2. Find the index of the matching string
  labels.indexOf(value) match
    case -1 => 
      Left(s"'$value' is not a valid value. Valid options: ${labels.mkString(", ")}")
    case index =>
      // 3. Construct the enum instance using the ordinal index
      // 'fromOrdinal' is generated by the compiler for Mirrors
      Right(m.fromOrdinal(index).asInstanceOf[T])
```

should I add this to a new `ArgBuilderVersionSpecific`?

ghostdogpr: Is that for Scala 3 enums? ArgBuilder derivation already handles them.

dontgitit: Yeah it's for Scala 3 enums; I guess it makes sense that ArgBuilder derivation already handles it since I guess the derivation probably works for any arbitrary sum type...?

ghostdogpr:
Thanks!

PR #2908, opened 2026-03-29
Make some Caliban code public to prevent duplication in kyo-caliban

ghostdogpr:
See https://github.com/getkyo/kyo/pull/1490


core/src/main/scala/caliban/Configurator.scala
```diff
@@ -35,7 +35,7 @@ object Configurator {
     def toLayer: ULayer[Unit] = ZLayer.scoped(ref.locallyScoped(self))
   }
 
-  private[caliban] val ref: FiberRef[ExecutionConfiguration] =
+  val ref: FiberRef[ExecutionConfiguration] =
```
kyri-petrou: With this one, instead of making it public it might be worth it to add a method that proxies the `ref.locally` method. AFAICT that's the only usage in kyo

ghostdogpr: Done

PR #2932, opened 2026-05-01
Honor covariant narrowing of interface fields per spec

### LINKED ISSUE #2931: Interface fields are dropped when subtypes covariantly narrow them, and validation rejects narrowed subtypes
## Summary

Caliban's interface derivation and schema validation both reject covariant narrowing of interface fields, even though [the GraphQL spec explicitly permits it](https://spec.graphql.org/October2021/#IsValidImplementationFieldType()):

> An object field type is a valid sub-type if it is the same type as the interface field type, or it is a possible type of the interface field type, or it is a non-null variant of a valid sub-type, or … (see spec for the full rule, including interface chains).

Two bugs interact:

### Bug 1 — interface derivation drops covariantly narrowed fields

[`DerivationUtils.mkInterface`](https://github.com/ghostdogpr/caliban/blob/v3.0.0/core/src/main/scala/caliban/schema/DerivationUtils.scala#L85) computes interface fields by intersecting subtypes' fields and calling [`Types.unify`](https://github.com/ghostdogpr/caliban/blob/v3.0.0/core/src/main/scala/caliban/schema/Types.scala#L185), which only handles same-type or `NON_NULL` wrapping — not interface↔implementer relationships. When subtypes narrow a field, the field is dropped from the interface entirely.

### Bug 2 — validator rejects narrowed-subtype fields

[`SchemaValidator.isValidSubtype`](https://github.com/ghostdogpr/caliban/blob/v3.0.0/core/src/main/scala/caliban/validation/SchemaValidator.scala#L241) checks:

```scala
Types.same(supertypeFieldType, objectFieldType) ||
supertypePossibleTypes.exists(Types.same(_, objectFieldType)) ||
isNonNullableSubtype(supertypeFieldType, objectFieldType)
```

When the supertype field is `NON_NULL[Interface]` and the object field is `NON_NULL[Implementer]`, all three branches fail: the `possibleTypes` lookup happens on the `NON_NULL` wrapper (whose `possibleTypes` is `None`) instead of on the inner interface. The validator never consults the implementer's `interfaces` list either.

## Repro

```scala
import caliban.schema.Annotations.GQLInterface
import caliban.schema.Schema.auto.*
import caliban.{ GraphQL, RootResolver, graphQL }

@GQLInterface
sealed trait Animal { def species: Species }

@GQLInterface
sealed trait Species { def name: String }

case class Mammal(species: Mammalia) extends Animal
case class Bird(species: Aves) extends Animal

case class Mammalia(name: String) extends Species
case class Aves(name: String) extends Species

case class Query(animals: List[Animal])

@main def run(): Unit =
  val gql = graphQL(RootResolver(Query(List(Mammal(Mammalia(\"cat\")), Bird(Aves(\"sparrow\"))))))
  println(gql.render)
```

### Expected

The rendered schema preserves the interface field:

```graphql
interface Animal { species: Species! }
type Mammal implements Animal { species: Mammalia! }
type Bird implements Animal { species: Aves! }
```

This is valid per the spec (covariant narrowing of an interface field).

### Actual

1. **Bug 1**: `Animal` is rendered with **no fields** because `Types.unify(Mammalia!, Aves!) == None`. (If you query `{ animals { species { name } } }` you get `Field 'species' does not exist on type 'Animal'`.)

2. **Bug 2**: If you work around bug 1 by adding the field manually via `withAdditionalTypes` / a `Transformer`, the schema *validation* then fails with `Field 'species' in Object 'Mammal' is an invalid subtype` — `isValidSubtype` doesn't recognize that `Mammalia` implements `Species`.

## Suggested fix

For bug 1: `Types.unify` (or `mkInterface`) should fall back to the **declared abstract type** on the trait when subtype field types disagree but a common supertype exists in the impls' `interfaces` lists.

For bug 2: `isValidSubtype` should additionally check whether the object field type's `interfaces` (transitively) include the supertype field type, after unwrapping `NON_NULL` consistently on both sides.

## Versions

- caliban 3.0.0
- Scala 3.6.4

Happy to follow up with a PR if you'd like.

kroeberk:
## Summary

Fixes the two interacting bugs described in #2931 around covariant narrowing of interface fields, per [spec §IsValidImplementationFieldType()](https://spec.graphql.org/October2021/#IsValidImplementationFieldType()).

### `Types.unify` — preserve narrowed-but-related interface fields

When two sub-type field types differ but transitively share a common interface, `unify` now returns that interface instead of `None`. NON_NULL wrapping is preserved when both sides agree. This keeps the field on the parent interface (using the shared interface as the abstract type) rather than dropping it from the rendered schema entirely.

### `SchemaValidator.isValidSubtype` — spec-compliant sub-type check

Previously, an interface field of `Foo!` could not be implemented by `Bar!` even when `Bar implements Foo`, because the `possibleTypes` lookup happened on the NON_NULL wrapper (whose `possibleTypes` is `None`) and the implementer's `interfaces` list was never consulted.

The check is reworked to follow the spec algorithm:

- Strip NON_NULL on both sides (allowing the implementing field to be non-null when the interface field is nullable, but not the reverse).
- Recurse into LIST item types.
- For an interface supertype field, accept any object/interface that is in `possibleTypes` *or* transitively implements the supertype via `interfaces()`.

## Test plan

- [x] `core/testOnly caliban.validation.ValidationSchemaSpec caliban.schema.TypesSpec` — new tests cover the unify fallback (with and without NON_NULL) and the validator accepting a non-null narrowed sub-type field.
- [x] Full `core/test` on Scala 2.13 — all 548 tests pass, no regressions.
- [x] Full `core/test` on Scala 3.3.7 — all 597 tests pass, no regressions.
- [x] `sbt fmt` clean.

Closes #2931

### DIFF
--- a/core/src/main/scala/caliban/schema/Types.scala
+++ b/core/src/main/scala/caliban/schema/Types.scala
@@ -220,23 +220,25 @@ object Types {
       }
 
   /**
-   * Finds the closest common interface (or interface chain ancestor) shared by `t1` and `t2`,
-   * by walking each type's `interfaces()` lists transitively.
+   * Finds the closest type reachable from both `t1` and `t2` via their `interfaces()` chain
+   * (including each type itself, so that a direct implementer/interface pair returns the
+   * interface). BFS order is used so the closest match wins; a `visited` set guards against
+   * cyclic interface graphs.
    *
    * Used by [[unify]] as a fallback when two types differ but covariantly narrow a shared interface
    * field (per https://spec.graphql.org/October2021/#IsValidImplementationFieldType()).
    */
   private def commonInterface(t1: __Type, t2: __Type): Option[__Type] = {
-    def ancestors(t: __Type): List[__Type] = {
-      val direct = t.interfaces().toList.flatten
-      direct ::: direct.flatMap(ancestors)
+    @tailrec
+    def selfAndAncestors(queue: List[__Type], acc: List[__Type]): List[__Type] = queue match {
+      case Nil       => acc.reverse
+      case h :: rest =>
+        if (acc.exists(same(_, h))) selfAndAncestors(rest, acc)
+        else selfAndAncestors(rest ::: h.interfaces().toList.flatten, h :: acc)
     }
-    val a1                                 = ancestors(t1)
-    if (a1.isEmpty) None
-    else {
-      val a2 = ancestors(t2)
-      a1.find(i1 => a2.exists(i2 => same(i1, i2)))
-    }
+    val a1                                                                     = selfAndAncestors(List(t1), Nil)
+    val a2                                                                     = selfAndAncestors(List(t2), Nil)
+    a1.find(i1 => a2.exists(i2 => same(i1, i2)))
   }
 
   def extractCommonDescription(l: List[__Field]): Option[String] =
--- a/core/src/main/scala/caliban/schema/Types.scala
+++ b/core/src/main/scala/caliban/schema/Types.scala
@@ -208,10 +208,36 @@ object Types {
     if (same(t1, t2)) Option(t1)
     else
       (t1.kind, t2.kind) match {
-        case (__TypeKind.NON_NULL, _) => t1.ofType.flatMap(unify(_, t2))
-        case (_, __TypeKind.NON_NULL) => t2.ofType.flatMap(unify(_, t1))
-        case _                        => None
+        case (__TypeKind.NON_NULL, __TypeKind.NON_NULL) =>
+          for {
+            a <- t1.ofType
+            b <- t2.ofType
+            u <- unify(a, b)
+          } yield u.nonNull
+        case (__TypeKind.NON_NULL, _)                   => t1.ofType.flatMap(unify(_, t2))
+        case (_, __TypeKind.NON_NULL)                   => t2.ofType.flatMap(unify(_, t1))
+        case _                                          => commonInterface(t1, t2)
       }
+
+  /**
+   * Finds the closest common interface (or interface chain ancestor) shared by `t1` and `t2`,
+   * by walking each type's `interfaces()` lists transitively.
+   *
+   * Used by [[unify]] as a fallback when two types differ but covariantly narrow a shared interface
+   * field (per https://spec.graphql.org/October2021/#IsValidImplementationFieldType()).
+   */
+  private def commonInterface(t1: __Type, t2: __Type): Option[__Type] = {
+    def ancestors(t: __Type): List[__Type] = {
+      val direct = t.interfaces().toList.flatten
+      direct ::: direct.flatMap(ancestors)
+    }
+    val a1                                 = ancestors(t1)
+    if (a1.isEmpty) None
+    else {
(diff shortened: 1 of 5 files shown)


core/src/main/scala/caliban/schema/Types.scala
```diff
@@ -208,11 +208,37 @@ object Types {
...
       }
 
+  /**
+   * Finds the closest common interface (or interface chain ancestor) shared by `t1` and `t2`,
+   * by walking each type's `interfaces()` lists transitively.
+   *
+   * Used by [[unify]] as a fallback when two types differ but covariantly narrow a shared interface
+   * field (per https://spec.graphql.org/October2021/#IsValidImplementationFieldType()).
+   */
+  private def commonInterface(t1: __Type, t2: __Type): Option[__Type] = {
```
ghostdogpr, line 231: If `t1 implements t2` (or vice-versa), this will return `None`, shouldn't we return `t2`?

core/src/main/scala/caliban/schema/Types.scala
```diff
@@ -208,11 +208,37 @@ object Types {
...
+    def ancestors(t: __Type): List[__Type] = {
```
ghostdogpr: Maybe add a `visited` check to prevent infinite loops

core/src/main/scala/caliban/schema/Types.scala
kroeberk: Good catch — fixed in 382c910. `commonInterface` now starts each BFS at the type itself, so a direct implementer/interface pair unifies to the interface (added a test for it).

core/src/main/scala/caliban/schema/Types.scala
kroeberk: Done — same commit switched the traversal to BFS with a `visited` accumulator, plus a regression test exercising a self-referential interface pair.

core/src/main/scala/caliban/validation/SchemaValidator.scala
```diff
@@ -230,20 +230,40 @@ private[caliban] object SchemaValidator {
...
 
-        def isValidSubtype(supertypeFieldType: __Type, objectFieldType: __Type) = {
-          val supertypePossibleTypes = supertypeFieldType.possibleTypes.toList.flatten
-
-          Types.same(supertypeFieldType, objectFieldType) ||
-          supertypePossibleTypes.exists(Types.same(_, objectFieldType)) ||
-          isNonNullableSubtype(supertypeFieldType, objectFieldType)
+        def implementsTransitively(t: __Type, ancestor: __Type): Boolean = {
+          val direct = t.interfaces().toList.flatten
+          direct.exists(d => Types.same(d, ancestor) || implementsTransitively(d, ancestor))
```
ghostdogpr: I think this should prevent cycles too

kroeberk: Done in 3498ed1 — same BFS+`visited` pattern as `commonInterface`, with a regression test.

ghostdogpr (approved):
Thanks!

Issue #1743, opened 2023-05-31
Supporting Scala.js / Scala Native on the server


kubukoz:
Hi! I'm working on the Smithy4s interop for Caliban on the server, in which you could derive a Schema for a Smithy service.

This works fine on the JVM, but Smithy4s supports the two other platforms as well (Native only on Scala 3, though), and I would love to get at least some of that working with Caliban. This would clearly require Caliban to cross-compile as well.

I see that there's Scala.js in the build already, but it seems to be limited to the laminext and client modules.

As far as blockers/difficulties go, I'm only aware of:

- Timestamp types which don't have a portable implementation
- ???

If there's anything else you know of, such as any required dependencies that aren't being cross-published, let me know please! I'm happy to get the ball rolling on adding the build axes and getting the shared part of the server to cross-publish. Specifically I'm interested in `caliban-cats`.

ghostdogpr:
We had Scala.js cross-compiling for the server a long time ago but [dropped it](https://github.com/ghostdogpr/caliban/pull/324) because it was annoying dependency-wise and there was nobody using it.

What would be the use case? None of the http libraries behind the adapters support Scala.js nor Scala Native AFAIK.

kubukoz:
Http4s works on both of these platforms :)

ghostdogpr:
Oh I see, I was looking at the weird versions here https://mvnrepository.com/artifact/org.http4s/http4s-dsl

ghostdogpr:
The issues I see for the core module only:
- magnolia: we're still using the old version, the new one by softwaremill supports Scala.js and Native, hopefully it doesn't break anything
- play-json only supports Native and Scala.js on Scala 3 in a RC version
- zio-query and zio-json don't support Scala Native

ghostdogpr:
Since my previous message, we switched to the latest magnolia and there's a version of play-json for Scala.js. That means we should be able to support Scala.js easily now

ghostdogpr:
For Scala native, i see zio-Json supports it now so the only blocker might be zio-query which is probably easy to port.

virtualirfan:
`zio-query`, as of [0.7.5](https://github.com/zio/zio-query/releases/tag/v0.7.5), now supports `scala-native`.

PR #2980, opened 2026-05-29
Cross-compile core to Scala Native

ghostdogpr:
Part of https://github.com/ghostdogpr/caliban/issues/1743

I am not sure if I will merge this yet. Let me know if there is any interest for it.


kyri-petrou:
I agree, let's see if there's any interest in this before merging it in. I had issues with maintaining SN builds as they tend to break within patch releases of SN. Might not be the case here but still probably worth to see if there are people interested in it first.

Having said that, if we can remove reflection I think that'd be great (assuming it doesn't require very complex macros, but maybe that's not too bad now we have AI to help write those)

ghostdogpr:
> Having said that, if we can remove reflection I think that'd be great (assuming it doesn't require very complex macros, but maybe that's not too bad now we have AI to help write those)

Yeah I didn't even remember we had this, I was a bit shocked 😆

PR #2983, opened 2026-05-30
Fix FragmentValidator cache and optimize FieldMap

### LINKED ISSUE #2948: FragmentValidator caches keyed only by hashCode — collisions can flip validation result
Was tracing fragment validation paths and noticed the three memoization caches in `FragmentValidator.findConflictsWithinSelectionSet` are keyed by `Int` (the hash) without any equality check on the underlying selection set. On a hash collision the cache happily returns the previous validation result for a different input.

The caches:

https://github.com/ghostdogpr/caliban/blob/12a663612ca830a5ad96104a801b583790997ca2/core/src/main/scala/caliban/validation/FragmentValidator.scala#L19-L22

```scala
// NOTE: We use the `hashCode()` as the key since it's much more performant
val shapeCache   = mutable.Map.empty[Int, Chunk[String]]
val parentsCache = mutable.Map.empty[Int, Chunk[String]]
val groupsCache  = mutable.Map.empty[Int, Chunk[Set[SelectedField]]]
```

And the lookups (`sameResponseShapeByName`, `sameForCommonParentsByName`, `requireSameNameAndArguments`):

https://github.com/ghostdogpr/caliban/blob/12a663612ca830a5ad96104a801b583790997ca2/core/src/main/scala/caliban/validation/FragmentValidator.scala#L23-L41

```scala
def sameResponseShapeByName(set: Iterable[Selection], parentType: __Type): Chunk[String] = {
  val keyHash = MurmurHash3.unorderedHash(set)
  shapeCache.get(keyHash) match {
    case Some(value) => value    // returned without checking 'set' itself matches
    case None        => ...
  }
}
```

Two failure modes on collision:
- **accepts-invalid**: a query with conflicting field types passes validation because the cache returns an empty error list cached for some prior valid selection set.
- **rejects-valid**: a valid query is rejected because the cache returns errors cached from some prior invalid set.

Probability is low for typical workloads, but it's the birthday paradox at ~65K unique 32-bit hashes, and a hostile client can grind selection sets to collide deliberately.

A `mutable.Map[Iterable[Selection], ...]` (or with the actual `List(set)` as the key) would be the obvious fix. The performance note in the comment is fair, but most of the cost is paid up-front when computing the hash anyway — using the value itself as key keeps the lookup O(1) for non-colliding inputs and only adds an `equals` on the rare collision. Happy to PR if you'd like.

kyri-petrou:
/fixes #2948

Also added a benchmark for this because our current benchmarks didn't stress the fragment validation cache at all. After adding the benchmark I realised that the `FieldMap` was extremely slow, primarily due to unnecessarily concatting sets. 

Since `FieldMap` was only being used internally (there's no need for users to call those methods) I deprecated all the public methods on it and made the object package-private (which is bin-compat). Whenever we're happy with breaking binary compatibility, we can remove all the deprecated methods.

With the new methods on `FieldMap` and some other minor optimization, fragment validation is ~4-5x faster based on the benchmark that I added

### DIFF
--- a/benchmarks/src/main/scala/caliban/validation/FragmentValidatorBenchmark.scala
+++ b/benchmarks/src/main/scala/caliban/validation/FragmentValidatorBenchmark.scala
@@ -1,0 +1,164 @@
+package caliban.validation
+
+import caliban._
+import caliban.parsing.Parser
+import caliban.schema.RootType
+import org.openjdk.jmh.annotations._
+
+import java.util.concurrent.TimeUnit
+
+@State(Scope.Thread)
+@BenchmarkMode(Array(Mode.Throughput))
+@OutputTimeUnit(TimeUnit.SECONDS)
+@Warmup(iterations = 5, time = 1, timeUnit = TimeUnit.SECONDS)
+@Measurement(iterations = 5, time = 1, timeUnit = TimeUnit.SECONDS)
+@Fork(2)
+class FragmentValidatorBenchmark {
+  import FragmentValidatorBenchmark._
+
+  @Benchmark
+  def fragmentConflicts(): Any =
+    run(Validator.validateAll(parsedQuery, rootType))
+}
+
+object FragmentValidatorBenchmark {
+  import caliban.schema.Schema.auto._
+  import caliban.schema.ArgBuilder.auto._
+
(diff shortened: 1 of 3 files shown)


ghostdogpr (approved):
It's actually simpler, nice.

PR #2984, opened 2026-05-30
Cache commonly used `IntNumber`s

kyri-petrou:
This is a trick used both for `java.lang.Integer` and `scala.math.BigInt`. For commonly used ints, we cache them so we reduce allocations during runtime.

While users might not use many `int`s in their schemas, the reason we want to do this is because `IntNumber` is created very frequently due to `PathValue` which is created for each array in the executor. Some other implementations (such as `java.lang.Integer` cache both negative and positive ints in that range, but I don't think it's necessary for us to do it since negative values are not as common (as they don't exist in the path)

### DIFF
--- a/core/src/main/scala/caliban/Value.scala
+++ b/core/src/main/scala/caliban/Value.scala
@@ -7,8 +7,9 @@ import caliban.rendering.ValueRenderer
 import com.github.plokhotnyuk.jsoniter_scala.core.JsonValueCodec
 import zio.stream.Stream
 
-import scala.collection.mutable
 import scala.collection.compat._
+import scala.collection.mutable
+import scala.runtime.AbstractFunction1
 import scala.util.control.NonFatal
 import scala.util.hashing.MurmurHash3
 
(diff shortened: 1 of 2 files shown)


core/src/main/scala/caliban/Value.scala
```diff
@@ -169,17 +169,35 @@ object Value {
...
+        if (value >= 0 && value < MaxCachedInt) cache(value)
+        else new IntNumber(value)
+    }
+
     final case class IntNumber(value: Int)       extends IntValue with PathValue {
       override def toInt: Int       = value
       override def toLong: Long     = value.toLong
       override def toBigInt: BigInt = BigInt(value)
-      override def toString: String = value.toString
+      override def toString: String = String.valueOf(value)
```
kyri-petrou, line 195: Not really impactful since we're not really using the `toString` method of these case classes, but I thought to do it anyways.

The `toString` method boxes the primitive, so it's better to use `String.valueOf` that doesn't cause any boxing

ghostdogpr (approved):
LGTM. Looks like something is wrong with downloading dependencies from Maven, we can retry later.

ghostdogpr:
Mima is unhappy

PR #2993, opened 2026-06-14
Update `safeName` to not use `$` as suffix

### LINKED ISSUE #2988: `$` suffix added by `safeName` produces warnings on Scala 3.9
The [`safeName` method appends `$` to a name](https://github.com/ghostdogpr/caliban/blob/series/3.x/codegen/src/main/scala/caliban/codegen/package.scala#L72) if it matches a value in `caseClassReservedFields`. Unfortunately this causes warnings on Scala 3.9.0-RC1:

```scala
-- [E230] Syntax Warning: /path/to/my/project/target/scala-3.9.0-RC1/src_managed/main/caliban-codegen-sbt/com/example/DateOperator.scala:11:14
11 |  case object equals$ extends DateOperator { val value: String = "equals$" }
   |              ^^^^^^^
   |              The identifier `equals$` should not contain `$`, which is reserved for internal compiler use.
```

It's easy enough to suppress these with the compiler option below, but it would be nice to not have to. Could the `caseClassReservedFields` branch also wrap the name in backticks like the `reservedKeywords` branch does?

```scala
scalacOptions += s"-Wconf:src=${sourceManaged.value}&msg=identifier.*should not contain `\\$$`:s"
```

mrdziuban:
Fixes #2988 

Updates `def safeName` to not use `$` as a suffix when it encounters a reserved keyword.

### DIFF
--- a/codegen/src/main/scala/caliban/codegen/package.scala
+++ b/codegen/src/main/scala/caliban/codegen/package.scala
@@ -67,8 +67,8 @@ package object codegen {
   val doubleQuotes = "\""
 
   def safeName(name: String): String =
-    if (name == "_") "_$" // scala 3 does not allow a name of `_`
+    if (name == "_") "`__`" // scala 3 does not allow a name of `_`
     else if (reservedKeywords.contains(name) || name.endsWith("_")) s"`$name`"
-    else if (caseClassReservedFields.contains(name)) s"$name$$"
+    else if (caseClassReservedFields.contains(name)) s"${name}_"
     else name
 }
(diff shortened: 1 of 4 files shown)


codegen/src/main/scala/caliban/codegen/package.scala
```diff
@@ -67,8 +67,8 @@ package object codegen {
   val doubleQuotes = "\""
 
   def safeName(name: String): String =
-    if (name == "_") "_$" // scala 3 does not allow a name of `_`
+    if (name == "_") "`__`" // scala 3 does not allow a name of `_`
```
mrdziuban, line 70: Just `__` (without backticks) is apparently not a valid method name when the method has an explicit return type specified

```scala
// This is fine
scala> def __ = ()
def __: Unit

// But this doesn't work
scala> def __: Unit = ()
-- [E040] Syntax Error: --------------------------------------------------------
1 |def __: Unit = ()
  |        ^^^^
  |'=' expected, but identifier found; identifier ends in colon, did you mean `__`: in backticks?
```

ghostdogpr (approved):

ghostdogpr:
Thanks!

PR #3001, opened 2026-06-20
Make Scala 3 lazy vals JDK26+ compatible

kyri-petrou:
Scala 3.3.x lazy vals emit warnings in JDK26, and will no longer work in upcoming JDK versions due to their usage of `sun.misc.Unsafe`.

In Scala 3.3.8 we can generate lazy vals that use `VarHandle`s instead when using the `-Yfuture-lazy-vals` flag with a `-release:{X >= 9}` (we use `-release:17` so we're OK)

Note that Mima is complaining after this change because lazy vals created a `<clinit>`, but it might not need a minor release for it (see [here](https://github.com/scala/scala3-lts/pull/637))

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -770,6 +770,7 @@ lazy val commonSettings = Def.settings(
         "-Werror",
         "-explain-types",
         s"-${if (minor >= 5) "X" else "Y"}kind-projector",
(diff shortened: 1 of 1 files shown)


ghostdogpr (approved):

He-Pin:
nice, pekko just did the same

PR #3016, opened 2026-07-07
Fix nested lists computing item-nullability at the wrong level

ghostdogpr:
## Problem

In `Executor`, `reduceListStep` computed `areItemsNullable` from `currentField.fieldType`, and nested `ListStep`s recursed through `reduceStep` with the **same** `currentField`. Because a `Field` carries the entire declared type (all list wrappers at once), every nesting level read the *outer* list's element-nullability.

For a field typed `[[Int!]]`, the inner list was therefore treated as having nullable items. When a non-null inner item (`Int!`) failed, the error was caught and substituted as `null` in the item slot — placing `null` in a non-null position — instead of bubbling up to null the (nullable) inner list. The symmetric `[[Int]!]` case over-propagated.

## Fix

Thread the **actual type at each position** through `reduceStep` as a new `fieldType` parameter, descending exactly one list level per nested `ListStep`:

- `reduceListStep` computes `elementType = Types.listOf(fieldType)` once, derives `areItemsNullable` from it, and reduces each item with that descended `elementType`.
- Every other call site passes the type for its position: top-level `request.field.fieldType`, object fields `f.fieldType`, `@stream` item reductions the descended item type, and query/function/metadata/extension/subscription wrappers pass it through unchanged.

Invariant: `fieldType` is always the GraphQL type of the value the step resolves to at that position.

## Tests

Added *"die inside a non-nullable item of a nested nullable list"* in `ExecutionSpec`: `test: List[Task[List[UIO[Int]]]]` (→ `[[Int!]]!`, nullable inner lists, non-null items). A die in an inner item now yields `{"test":[[1,2],null]}` (inner list nulled) rather than the buggy `{"test":[[1,2],[3,null]]}`.

`core/testOnly caliban.execution.ExecutionSpec caliban.execution.IncrementalDeliverySpec` passes (75 tests).

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -786,7 +786,9 @@ lazy val enableMimaSettingsJVM =
     mimaPreviousArtifacts  := previousStableVersion.value.map(organization.value %% moduleName.value % _).toSet,
     mimaBinaryIssueFilters := Seq(
       ProblemFilters.exclude[DirectMissingMethodProblem]("caliban.*.<clinit>"),
-      ProblemFilters.exclude[DirectMissingMethodProblem]("mdg.engine.proto.reports.*.<clinit>")
+      ProblemFilters.exclude[DirectMissingMethodProblem]("mdg.engine.proto.reports.*.<clinit>"),
+      // private internal method; extra param added to fix nested-list null propagation
+      ProblemFilters.exclude[DirectMissingMethodProblem]("caliban.execution.Executor#StepReducer.reduceStep")
     )
   )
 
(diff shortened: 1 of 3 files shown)


build.sbt
```diff
@@ -786,7 +786,9 @@ lazy val enableMimaSettingsJVM =
     mimaPreviousArtifacts  := previousStableVersion.value.map(organization.value %% moduleName.value % _).toSet,
     mimaBinaryIssueFilters := Seq(
       ProblemFilters.exclude[DirectMissingMethodProblem]("caliban.*.<clinit>"),
-      ProblemFilters.exclude[DirectMissingMethodProblem]("mdg.engine.proto.reports.*.<clinit>")
+      ProblemFilters.exclude[DirectMissingMethodProblem]("mdg.engine.proto.reports.*.<clinit>"),
+      // private internal method; extra param added to fix nested-list null propagation
+      ProblemFilters.exclude[DirectMissingMethodProblem]("caliban.execution.Executor#StepReducer.reduceStep")
```
ghostdogpr, line 791: @kyri-petrou this is fine, right?

PR #3019, opened 2026-07-08
Accept integer literals for ID arguments

ghostdogpr:
## Problem

`ValueValidator.validateScalar` accepted only `StringValue`/`NullValue` for the built-in `ID` scalar and rejected everything else:

```scala
case "ID" =>
  argValue match {
    case _: StringValue | NullValue => unit
    case t => failValidation(s"$errorContext has invalid type $t", "Expected 'ID'")
  }
```

Per the GraphQL spec, the `ID` scalar's input coercion accepts **both String and Int** literals (e.g. graphql-js `GraphQLID.parseLiteral` accepts `IntValueNode` and `StringValueNode`). So a valid query like `{ node(id: 4) }` was rejected at validation with `Expected 'ID'`.

This also created an inconsistency: the variables path (`VariablesCoercer`) has no special `ID` case and passes the value through untouched, so the same integer supplied via a variable was accepted while the inline literal was rejected.

## Fix

Accept `IntValue` for `ID` as well:

```scala
case "ID" =>
  argValue match {
    case _: StringValue | _: Value.IntValue | NullValue => unit
    case t => failValidation(s"$errorContext has invalid type $t", "Expected 'ID'")
  }
```

## Tests

Added *"integer literal is accepted for an ID argument"* in `ValidationSpec`, using a custom `ID`-named scalar whose `ArgBuilder` accepts both String and Int, then executing `{ node(id: 4) }` and asserting no errors and `{"node":"4"}`. Discriminating — before the fix, validation rejected it with `Expected 'ID'`.

`core/testOnly caliban.validation.ValidationSpec` passes (44 tests).

### DIFF
--- a/core/src/main/scala/caliban/validation/ValueValidator.scala
+++ b/core/src/main/scala/caliban/validation/ValueValidator.scala
@@ -149,8 +149,8 @@ private object ValueValidator {
         }
       case "ID"      =>
         argValue match {
-          case _: StringValue | NullValue => unit
-          case t                          => failValidation(s"$errorContext has invalid type $t", "Expected 'ID'")
+          case _: StringValue | _: Value.IntValue | NullValue => unit
+          case t                                              => failValidation(s"$errorContext has invalid type $t", "Expected 'ID'")
         }
       case "Int"     =>
         argValue match {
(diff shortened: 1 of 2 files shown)


ghostdogpr:
Closing: accepting Int literals for the ID scalar regresses UUID default-value validation (a UUID `@GQLDefault("1")` would slip past interpreter construction and only fail at request time), and the upside is marginal — only user-defined string-backed ID scalars benefit, since the built-in UUID ArgBuilder rejects ints regardless. Not worth the safety tradeoff.

PR #3052, opened 2026-07-08
Update jsoniter number parser implementation to match jsoniter-circe

kyri-petrou:
Seems the code we were using is a bit outdated, I updated to match what's also found in jsoniter-circe now

### DIFF
--- a/core/src/main/scala/caliban/interop/jsoniter/jsoniter.scala
+++ b/core/src/main/scala/caliban/interop/jsoniter/jsoniter.scala
@@ -123,7 +123,7 @@ private[caliban] object ValueJsoniter {
         StringValue(in.readString(null))
       case x if x == '-' || (x >= '0' && x <= '9') =>
         in.rollbackToken()
-        numberParser(in)
+        parseNumber(in)
       case 'n'                                     =>
         in.readNullOrError(NullValue, "unexpected JSON value")
       case 'f' | 't'                               =>
@@ -169,7 +169,7 @@ private[caliban] object ValueJsoniter {
         StringValue(in.readString(null))
       case x if x == '-' || (x >= '0' && x <= '9') =>
         in.rollbackToken()
-        numberParser(in)
+        parseNumber(in)
       case 'n'                                     =>
         in.readNullOrError(NullValue, "unexpected JSON value")
       case 'f' | 't'                               =>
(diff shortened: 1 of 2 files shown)


plokhotnyuk:
I'm going to add `JsonRead.readNumber: java.lang.Number` that will return the most suitable implementation from `java.lang.Integer`, `java.lang.Long`, `java.math.BigInteger`, `java.lang.Float`, `java.lang.Double`, and `java.math.BigDecimal` in the next minor version of jsoniter-scala (v2.39.0)

plokhotnyuk:
I've released [v2.39.0](https://github.com/plokhotnyuk/jsoniter-scala/releases/tag/v2.39.0) with `JsonRead.readNumber(): java.lang.Number` implemented while only for  `java.lang.Long`, `java.math.BigInteger`, and `java.math.BigDecimal` as output.

That already speeds up and simplifies circe booster: https://github.com/plokhotnyuk/jsoniter-scala/commit/870a1f11bf9a4ea96e63655fc9a0a1007329ddba

The next step will be improving `JsonRead.readNumber(): java.lang.Number`, `JsonRead.readBigInt(): BigInt`, and `JsonRead.readBigDecimal(): BigDecimal`  implementations to avoid double scanning for small numbers with less than 19 digits.

@kyri-petrou @ghostdogpr Thanks for your support and donations!

core/src/main/scala/caliban/interop/jsoniter/jsoniter.scala
```diff
@@ -203,32 +203,19 @@ private[caliban] object ValueJsoniter {
...
+      case l: java.lang.Long        =>
+        val asLong = l.longValue()
+        val asInt  = asLong.toInt
+        // Check if the number can fit in an Int (most common case)
+        if (asInt.toLong == asLong) Value.IntValue.IntNumber(asInt)
+        else Value.IntValue.LongNumber(asLong)
+      case bd: java.math.BigDecimal =>
+        Value.FloatValue.BigDecimalNumber(bd)
+      case bi: java.math.BigInteger =>
+        Value.IntValue.BigIntNumber(bi)
```
kyri-petrou, line 217: @plokhotnyuk I'm guessing that you're not planning on returning any other instances of `java.lang.Number` in the future, is that right?

If that's the case, it might be worth to document this in the scaladoc of `readNumber`. Or if you're planning on returning any different types, let me know so I can add a default case

PR #3063, opened 2026-07-14
Cap JVM memory in CI test jobs to avoid OOM kills

ghostdogpr:
CI test jobs started OOM-killing (`Received "killed" signal`) on every PR and on `series/3.x`. It's not the code: `cimg/openjdk:17.0-node` is a rolling tag and a recent republish changed the JVM's default memory ergonomics, so the sbt launcher + forked test JVM now exceed the 8GB `large` runner.

Cap the heap of both JVMs to 2G, CI-only (local `.sbtopts` stays 3G), on the five heavy `rootJVM*/test` jobs. Metaspace is left uncapped on purpose — a `MaxMetaspaceSize` limit makes sbt throw `OutOfMemoryError: Metaspace`.


paulpdaniels:
Should we see about porting tests to GitHub CI?

ghostdogpr:
> Should we see about porting tests to GitHub CI?

Hmm it seems okay with this change. Let's see if there are more issues.

PR #3074, opened 2026-07-24
Fix scope closing issue with http4s tests

kyri-petrou:
There's this annoying issue with the http4s which I can't really figure out why it's happening, but the tests are taking 60-70 seconds to run because closing the server resource blocks indefinitely until the suite gives up waiting for it.

The weirdest thing is that the server does shutdown, so I'm not sure exactly why is this happening (maybe something to do with `TestClock` or so. Anyways, this fixes it. It's not great, but at least the tests run quickly now

### DIFF
--- a/adapters/http4s/src/test/scala/caliban/Http4sAdapterSpec.scala
+++ b/adapters/http4s/src/test/scala/caliban/Http4sAdapterSpec.scala
@@ -66,6 +66,12 @@ object Http4sAdapterSpec extends ZIOSpecDefault {
       .provideLayerShared(apiLayer)
   }
 
+  /**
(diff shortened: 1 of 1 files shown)


ghostdogpr:
Ah I think I had that exact issue with another library I was working on that was using Ember. I vaguely remember that the Ember server stays up as long as there is something connected to it until some timeout is reached, so I had to either reduce that shutdown timeout or make sure I killed any client that still had a dangling connection established.

kyri-petrou:
> I had to either reduce that shutdown timeout

@ghostdogpr yeah the weird thing is that I tried this but also didn't work for some reason.

ghostdogpr (approved):
Maybe add a comment saying why we did this so it doesn't get "simplified" in the future

PR #3112, opened 2026-09-08
New phase handlers

paulpdaniels:
# Summary
Rework of the `GatewayWrapper` into `PhaseHooks` and `PhaseHandler`s. The main goal was to provide a flexible means of constructing hooks that are executed during specific phases of the execution pipeline. The previous approach was simpler but the `wrap` API wasn't able to scale with the other extensibility points (see observeEvent and overrideLabels). The new API specifically addresses this by providing a typesafe, composable, performant interface to support hooks.

Hooks are provided to the gateway either with the `@@` operator common to the ZIO ecosystem or with the more explicit `withPhaseHooks`. `PhaseHooks` provides all of the available extensibility hooks as constructor methods that can be composed sequentially using `++`, and each constructor takes a `PhaseHandler` which implements the specific logic of that phase wrapper. Examples of its use can be seen in the reworked `GatewayTracing` and the `GatewayMetrics`.

### Copilot Summary:

This pull request refactors the gateway observability and tracing integration to use a new "aspect" and "phase hook" model instead of the previous "wrapper" approach. It updates documentation, example apps, and the OpenTelemetry tracing implementation to use this new model, and enhances the tracing hooks for more precise and composable observability. The changes also improve the flexibility of how tracing and metrics are attached to gateway phases and clarify their composition.

**Observability and Tracing Model Refactor:**

* The documentation (`gateway-examples/README.md`) and example gateway apps now describe and use observability "aspects" and "phase hooks" instead of "wrappers". This clarifies how to attach metrics and tracing, and demonstrates the new composition model using `@@` for aspects. [[1]](diffhunk://#diff-67943560924bbb41ab0aa632a7581db3d5bc90a27c8ec5cbc508ae883032d215L32-R77) [[2]](diffhunk://#diff-5d2dfcdf1e38f3e1a7af6f829433270525049b68ed0f5367685ebe294c799647L25-R25)
* The tracing integration (`GatewayTracing`) is rewritten to provide a `hooks` bundle and an `aspect` for easy attachment to gateways. The previous `wrapper` is removed, and tracing is now implemented as a set of phase hooks, improving clarity and composability. [[1]](diffhunk://#diff-1496b819e06527c6d6b782a3a3c720000d1745f9d206cd62e8cf14d3d7ffb1c1L4-R134) [[2]](diffhunk://#diff-1496b819e06527c6d6b782a3a3c720000d1745f9d206cd62e8cf14d3d7ffb1c1L83-R146)
* The tracing phase hooks now more precisely scope spans to gateway phases (e.g., request, routing, subgraph call, retry, etc.), and correctly handle span context propagation and parent/child relationships. [[1]](diffhunk://#diff-1496b819e06527c6d6b782a3a3c720000d1745f9d206cd62e8cf14d3d7ffb1c1L4-R134) [[2]](diffhunk://#diff-1496b819e06527c6d6b782a3a3c720000d1745f9d206cd62e8cf14d3d7ffb1c1L83-R146)
* The test suite is updated to use the new aspect model and to verify the correct structure and parentage of tracing spans, especially for subscriptions. [[1]](diffhunk://#diff-4f4a9636bed5a72cb05c174b59b4e1a7da8a4d8fe14586ba8768c1441e101426L46-R46) [[2]](diffhunk://#diff-4f4a9636bed5a72cb05c174b59b4e1a7da8a4d8fe14586ba8768c1441e101426R56) [[3]](diffhunk://#diff-4f4a9636bed5a72cb05c174b59b4e1a7da8a4d8fe14586ba8768c1441e101426L68-R71) [[4]](diffhunk://#diff-4f4a9636bed5a72cb05c174b59b4e1a7da8a4d8fe14586ba8768c1441e101426L83-R86)

**Example Gateway Improvements:**

* The managed federation gateway example is updated to use the new `RemoteGraphQLConfig` and to demonstrate custom subgraph endpoint routing and header forwarding, providing a more realistic example of advanced configuration. [[1]](diffhunk://#diff-bd51a2f279f6c76d169f096010c1794813036b695df2d4fe1619496e08de0a97L4-R7) [[2]](diffhunk://#diff-bd51a2f279f6c76d169f096010c1794813036b695df2d4fe1619496e08de0a97L15-R38)

**Technical Improvements:**

* Tracing span creation now uses ZIO scopes for better resource management, and context propagation is handled more robustly, preventing trace context leaks from clients. [[1]](diffhunk://#diff-1496b819e06527c6d6b782a3a3c720000d1745f9d206cd62e8cf14d3d7ffb1c1L110-R177) [[2]](diffhunk://#diff-1496b819e06527c6d6b782a3a3c720000d1745f9d206cd62e8cf14d3d7ffb1c1L130-L170)

These changes modernize and clarify the observability integration for Caliban gateway users, making it easier to compose metrics, tracing, and custom hooks, and to reason about the lifecycle of gateway requests and subgraph calls.

**References:**  
[[1]](diffhunk://#diff-67943560924bbb41ab0aa632a7581db3d5bc90a27c8ec5cbc508ae883032d215L32-R77) [[2]](diffhunk://#diff-5d2dfcdf1e38f3e1a7af6f829433270525049b68ed0f5367685ebe294c799647L25-R25) [[3]](diffhunk://#diff-bd51a2f279f6c76d169f096010c1794813036b695df2d4fe1619496e08de0a97L4-R7) [[4]](diffhunk://#diff-bd51a2f279f6c76d169f096010c1794813036b695df2d4fe1619496e08de0a97L15-R38) [[5]](diffhunk://#diff-1496b819e06527c6d6b782a3a3c720000d1745f9d206cd62e8cf14d3d7ffb1c1L4-R134) [[6]](diffhunk://#diff-1496b819e06527c6d6b782a3a3c720000d1745f9d206cd62e8cf14d3d7ffb1c1L83-R146) [[7]](diffhunk://#diff-1496b819e06527c6d6b782a3a3c720000d1745f9d206cd62e8cf14d3d7ffb1c1L110-R177) [[8]](diffhunk://#diff-1496b819e06527c6d6b782a3a3c720000d1745f9d206cd62e8cf14d3d7ffb1c1L130-L170) [[9]](diffhunk://#diff-4f4a9636bed5a72cb05c174b59b4e1a7da8a4d8fe14586ba8768c1441e101426L46-R46) [[10]](diffhunk://#diff-4f4a9636bed5a72cb05c174b59b4e1a7da8a4d8fe14586ba8768c1441e101426R56) [[11]](diffhunk://#diff-4f4a9636bed5a72cb05c174b59b4e1a7da8a4d8fe14586ba8768c1441e101426L68-R71) [[12]](diffhunk://#diff-4f4a9636bed5a72cb05c174b59b4e1a7da8a4d8fe14586ba8768c1441e101426L83-R86)


ghostdogpr:
I was traveling for a few days but I'll check it tomorrow or Friday!

gateway-examples/README.md
```diff
@@ -29,34 +29,52 @@ sbt "gatewayExamples/runMain example.gateway.federation.FederationGatewayApp"
...
+The `gateway-tracing` module provides `GatewayTracing.aspect`. Aspects compose with `@@`, so tracing and metrics can be
 installed together:
 
 ```scala
 import caliban.gateway.GatewayMetrics
 import caliban.gateway.tracing.GatewayTracing
 
-val observed = Gateway.compose(first, rest: _*) @@
-  (GatewayMetrics.wrapper |+| GatewayTracing.wrapper)
+val observed = Gateway.compose(first, rest: _*) @@ (GatewayMetrics.aspect @@ GatewayTracing.aspect)
```
ghostdogpr, line 53: Are the parens needed?

gateway/src/test/scala/caliban/gateway/GatewayAspectSpec.scala
```diff
@@ -0,0 +1,374 @@
...
+        observed <- direct.get
+        scoped   <- inScope.get
+        sequence <- order.get
+      } yield assertTrue(
+        response.errors.isEmpty,
+        observed.map(event => event.operationType -> event.outcome) ==
+          Vector(Some(OperationType.Query) -> GatewayWrapper.Outcome.Success),
+        observed.forall(event => event.document.isDefined && event.executionRequest.isDefined),
+        scoped == observed,
+        // Load-bearing: "scoped-out" precedes "scope-closed", i.e. the scope outlives the handler's own outgoing side.
```
ghostdogpr, line 254: The dreaded "load-bearing" term 😆 
To get rid of Claude's annoying way of writing, I run this skill https://github.com/cursor/plugins/blob/main/pstack/skills/unslop/SKILL.md on code comments and docs.

gateway-examples/README.md
```diff
@@ -29,34 +29,52 @@ sbt "gatewayExamples/runMain example.gateway.federation.FederationGatewayApp"
...
+    )
+  ) @@ GatewayMetrics.aspect
 ```
 
 `GatewayWrapper.Event` is one lifecycle algebra; `Event.Attempt` represents each physical HTTP attempt, including attempt zero.
-A wrapper receives the event, the effect, and a typed completion function in one `wrap` call. This keeps result metadata inside
-the span or metric scope that owns it and lets custom logging, profiling, policy, or telemetry integrations use the same seam.
+A phase handler sees its own event type, so `PhaseHooks.Attempt` receives an `Event.Attempt` without matching on the whole
+algebra. `PhaseHandler.incoming` runs before the phase, `PhaseHandler.outgoing` sees its typed result, `PhaseHandler.apply`
+does both and can carry state between them, and `PhaseHandler.around` encloses the phase effect for integrations such as
```
ghostdogpr, line 74: There's no `around`

vuepress/docs/docs/gateway.md
```diff
@@ -210,25 +210,27 @@ type Product {
...
+import caliban.gateway.{ Gateway, PhaseHandler, PhaseHooks }
 import zio.Task
 
 def activeLabels(request: GraphQLRequest): Task[Set[String]] = ???
 
-val progressiveOverrides = GatewayWrapper.overrideLabels[Any] { (request, labels) =>
-  activeLabels(request).map(_ intersect labels)
-}
+val progressiveOverrides = PhaseHooks.OverrideLabels(
+  PhaseHandler.overrideLabels[Any] { (request, labels) =>
```
ghostdogpr, line 223: Code seems wrong (no `overrideLabels` on `PhaseHandler`)

gateway/src/main/scala/caliban/gateway/OperationEvent.scala
```diff
@@ -0,0 +1,20 @@
...
+import caliban.parsing.adt.{ Document, OperationType }
+
+/**
+ * Everything observable about one operation, delivered to the `observeOperation` phase once per request.
+ *
+ * `document` and `executionRequest` are absent whenever the request never reached execution — a preparation failure, a
+ * timeout, a shutdown, or an interruption — so that traffic is reported rather than dropped. A handler that needs
+ * timing brackets it itself, keeping the start in its own `Ctx0`.
+ */
+case class OperationEvent(
```
ghostdogpr, line 14: nit: `final`

gateway/src/main/scala/caliban/gateway/PhaseHooks.scala
```diff
@@ -0,0 +1,253 @@
...
+  /** Hooks with every phase empty. The identity of `++`, and what a gateway starts with. */
+  val empty: PhaseHooks[Any] = new PhaseHooks[Any]()
+
+  /**
+   * Brackets opening the source of one subscription: the subscribe call to the owning subgraph and the upstream
+   * connection it needs. Runs once per subscription, inside its admission permit, and handler work counts against
+   * [[GatewaySubscriptionConfig]]`.setupTimeout`. The outgoing [[GatewayWrapper.Result]] says whether the source
+   * opened.
+   */
+  def SubscriptionSetup[R](handler: PhaseHandler[R, Event.SubscriptionSetup.type, Nothing, Result]): PhaseHooks[R] =
```
ghostdogpr, line 109: These constructors are weirdly cased, usually we start with lowercase

gateway/src/main/scala/caliban/gateway/internal/execution/RemoteSubgraphExecutor.scala
```diff
@@ -209,11 +215,14 @@ private[gateway] final class RemoteSubgraphExecutor[-R](
     replaySafe: Boolean,
     attempt: Int
   )(implicit trace: Trace): ZIO[R, SubgraphExecutor.Failure, GraphQLResponse[CalibanError]] = {
-    val transport   = wrapper.attemptHeaders(name, attempt, headers).flatMap(send(body, _))
+    val transport   =
+      phases.attemptHeaders.runWith(Event.AttemptHeaders(name, attempt, headers))(attempt =>
+        send(body, attempt.headers)
```
ghostdogpr, line 220: Maybe name this `attempt` differently because it shadows the `Int`

gateway/src/main/scala/caliban/gateway/GatewayAspect.scala
```diff
@@ -0,0 +1,19 @@
+package caliban.gateway
+
+abstract class GatewayAspect[-R] { self =>
+  def @@[R1 <: R](that: GatewayAspect[R1]): GatewayAspect[R1] =
+    new GatewayAspect[R1] {
+      private[gateway] def apply[R2 <: R1](gateway: Gateway[R2]): Gateway[R2] =
+        that(self(gateway))
+    }
+
+  private[gateway] def apply[R1 <: R](gateway: Gateway[R1]): Gateway[R1]
```
ghostdogpr, line 10: Should this be public?

gateway/src/main/scala/caliban/gateway/PhaseHandler.scala
```diff
@@ -0,0 +1,191 @@
...
+  private case class IncomingOutgoing[-R, Ev, +Err, Out, Ctx0](
+    incoming: Ev => ZIO[R, Err, (Ev, Ctx0)],
+    outgoing: (Ev, Ctx0, Out) => ZIO[R, Nothing, Unit]
+  ) extends PhaseHandler[R, Ev, Err, Out] {
+    override val enabled: Boolean = true
+
+    def runWith[R1 <: R, E >: Err, A](event: Ev)(fn: Ev => ZIO[R1, E, A])(
+      result: Exit[E, A] => Out
+    )(implicit trace: Trace): ZIO[R1, E, A] = incoming(event).flatMap { case (ev, ctx) =>
+      fn(ev).onExit(exit => outgoing(ev, ctx, result(exit)))
```
ghostdogpr, line 136: The outgoing callbacks accumulate until every incoming hook finishes. If an incoming hook is interrupted, these callbacks won't run. So for example, a gauge would not be decremented after a timeout. I think we need to ensure the corresponding outgoing runs after each incoming succeeds.

gateway/src/main/scala/caliban/gateway/GatewayMetrics.scala
```diff
@@ -40,75 +40,112 @@ object GatewayMetrics {
...
-    labels: Set[MetricLabel],
+    labels: Event => Set[MetricLabel],
     detailLabels: Result => Set[MetricLabel],
     totalLabels: Result => Set[MetricLabel]
-  )(effect: ZIO[R, E, A])(result: Exit[E, A] => Result)(implicit
-    trace: Trace
-  ): ZIO[R, E, A] =
-    ZIO.uninterruptibleMask { restore =>
+  ): PhaseHandler[Any, Event, Nothing, Result] =
+    PhaseHandler((ev: Event) => enterTrack(active, labels(ev)).map(ev -> _))((ev, ctx: Long, out: Result) =>
```
ghostdogpr, line 101: `labels(ev)` is called twice, could be once

gateway/src/main/scala/caliban/gateway/internal/execution/RemoteSubgraphExecutor.scala
```diff
@@ -143,22 +143,28 @@ private[gateway] final class RemoteSubgraphExecutor[-R](
                    })
                    else ZIO.succeed(Nil)
       effectful <- config.effectfulHeaders.mapError(SubgraphExecutor.HeaderFailure(_))
-      values    <- wrapper.outboundHeaders(name, outboundHeaders(incoming, effectful))
-    } yield values)(ZIO.succeed(_))
+      values    <-
+        phases.outboundHeaders.runWith(Event.OutboundHeaders(name, outboundHeaders(incoming, effectful)))(Exit.succeed)(
```
ghostdogpr, line 147: Would be nice to avoid allocations when there is no hook

gateway/src/main/scala/caliban/gateway/internal/execution/RemoteSubgraphExecutor.scala
```diff
@@ -143,22 +143,28 @@ private[gateway] final class RemoteSubgraphExecutor[-R](
...
+    headers.map(values => copied(admission, phases, Some(values)))
 
   override def subscribe(
     request: GraphQLRequest
   )(implicit trace: Trace): ZIO[R with Scope, Throwable, ZStream[Any, Throwable, GraphQLResponse[CalibanError]]] = {
     val open = for {
       values <- headers.mapError(_ => SubscriptionTermination.Source)
-      traced <- wrapper.attemptHeaders(name, 0, values)
+      traced <-
+        phases.attemptHeaders.runWith(Event.AttemptHeaders(name, 0, values))(Exit.succeed)((_: Exit[Nothing, Any]) =>
```
ghostdogpr, line 161: Would be nice to avoid allocations when there is no hook

gateway/src/main/scala/caliban/gateway/internal/OperationHooks.scala
```diff
@@ -65,7 +66,12 @@ private[gateway] final class OperationHooks[-R](
     else {
       val unresolved = labels.map(_.value)
       OperationHooks
-        .run(wrapper.activeOverrideLabels(request, unresolved), OperationHooks.OverrideLabelResolutionFailure)
+        .run(
+          phases.overrideLabels
```
ghostdogpr, line 70: Would be nice to avoid allocations when there is no hook

gateway/src/main/scala/caliban/gateway/GatewayWrapper.scala
```diff
@@ -1,98 +1,18 @@
...
-          self.attemptHeaders(subgraph, attempt, headers).flatMap(that.attemptHeaders(subgraph, attempt, _))
-
-        override def activeOverrideLabels(request: GraphQLRequest, labels: Set[String])(implicit
-          trace: Trace
-        ): ZIO[R1, Throwable, Set[String]] =
-          self.activeOverrideLabels(request, labels).zipWith(that.activeOverrideLabels(request, labels))(_ ++ _)
-      }
-}
-
 object GatewayWrapper {
```
ghostdogpr, line 16: It's a bit weird to keep this file and object if we don't have `GatewayWrapper` anymore

gateway-examples/src/main/scala/example/gateway/federation/ManagedGatewayApp.scala
```diff
@@ -15,10 +16,18 @@ object ManagedGatewayApp extends ZIOAppDefault {
     (zio.Config.string("GRAPH_REF") zipWith zio.Config.secret("KEY"))(Config.apply)
       .nested("APOLLO")
 
+  private val remoteConfig =
+    RemoteGraphQLConfig.default.withExecution(_.forwardIncomingHeaders("Authorization", "Authorization"))
```
ghostdogpr, line 20: `Authorization` is there twice?

gateway-examples/src/main/scala/example/gateway/federation/ManagedGatewayApp.scala
```diff
@@ -1,9 +1,10 @@
 package example.gateway.federation
 
 import caliban.QuickAdapter
-import caliban.gateway.{ Gateway, Supergraph }
+import caliban.gateway.{ Gateway, RemoteGraphQLConfig, Supergraph }
 import zio.Config.Secret
 import zio.{ Console, ZIO, ZIOAppDefault }
+import sttp.client4._
```
ghostdogpr, line 7: Unused

gateway/src/main/scala/caliban/gateway/PhaseHandler.scala
```diff
@@ -0,0 +1,191 @@
...
+        remaining match {
+          case Nil                                          =>
+            fn(ev0).onExit { exit =>
+              val value = result(exit)
+              ZIO.foreachDiscard(stack.reverse)(_.apply(ev0, value))
+            }
+          case Incoming(incoming) :: next                   =>
+            incoming(ev0).flatMap(loop(next, _, stack))
+          case IncomingOutgoing(incoming, outgoing) :: next =>
+            incoming(ev0).flatMap { case (ev1, ctx) =>
```
ghostdogpr, line 174: Similar issue with interruption

gateway/src/main/scala/caliban/gateway/PhaseHandler.scala
```diff
@@ -0,0 +1,191 @@
...
+      def loop(
+        remaining: List[PhaseHandler[R, Ev, Err, Out]],
+        ev0: Ev,
+        stack: List[(Ev, Out) => ZIO[R1, Nothing, Unit]]
+      ): ZIO[R1, E, A] =
+        remaining match {
+          case Nil                                          =>
+            fn(ev0).onExit { exit =>
+              val value = result(exit)
+              ZIO.foreachDiscard(stack.reverse)(_.apply(ev0, value))
```
ghostdogpr, line 169: I think we don't need the reverse, it's already in the order we want here

gateway/src/main/scala/caliban/gateway/PhaseHandler.scala
```diff
@@ -0,0 +1,191 @@
...
+            incoming(ev0).flatMap { case (ev1, ctx) =>
+              val partial = (ev2: Ev, out: Out) => outgoing(ev2, ctx, out)
+              loop(next, ev1, partial :: stack)
+            }
+          case Combined(handlers) :: next                   =>
+            loop(handlers ++ next, ev0, stack)
+          // The scope has to outlive the handler's own outgoing callback, which runs inside `runWith`, so it cannot be
+          // shared with the rest of the chain and closed at the end of it.
+          case Scoped(handler) :: next                      =>
+            ZIO.scoped[R1](handler.runWith[Scope with R1, E, A](ev0)((ev1: Ev) => loop(next, ev1, stack))(result))
```
ghostdogpr, line 183: How about
```scala
 def loop[R2 <: R1](
    remaining: List[PhaseHandler[R2, Ev, Err, Out]],
    ev0: Ev,
    stack: List[(Ev, Out) => ZIO[R2, Nothing, Unit]]
  ): ZIO[R2, E, A] =
```
and then here
```scala
case Scoped(handler) :: next                      =>
  ZIO.scoped[R2](loop[R2 with Scope](handler :: next, ev0, stack))
```

gateway/src/main/scala/caliban/gateway/internal/execution/RemoteSubgraphExecutor.scala
```diff
@@ -397,7 +406,7 @@ private[gateway] object RemoteSubgraphExecutor {
     endpoint: Uri,
     backend: SttpClient,
     config: RemoteGraphQLConfig[R],
-    wrapper: GatewayWrapper[R],
+    phases: PhaseHooks[R],
```
ghostdogpr, line 409: How about naming this `hooks`?

ghostdogpr: In every other files too (I didn't repeat the comment)

gateway/src/main/scala/caliban/gateway/GatewayAspect.scala
```diff
@@ -0,0 +1,19 @@
+package caliban.gateway
+
+abstract class GatewayAspect[-R] { self =>
```
ghostdogpr, line 3: I wonder if we even need that. We could just make `@@` take hooks?

gateway-examples/README.md
paulpdaniels: Ah yeah, Claude went rogue when I had it finish the hook port and started creating additional operators. I superseded this with `scoped`

vuepress/docs/docs/gateway.md
paulpdaniels: True, I removed this because I decided to keep the `PhaseHandler` companion clean and only have the fixed hook constructors in the `PhaseHooks` companion

gateway/src/main/scala/caliban/gateway/PhaseHooks.scala
paulpdaniels: they felt too much like members when I initially did that. Would you be ok if it was `with<Constructor>`?

gateway/src/main/scala/caliban/gateway/PhaseHandler.scala
paulpdaniels: How would we handle the result? Would you just pass the `Exit.Failure(cause)` from the interrupted incoming hook into the `result` method? Or change the signature to take an `Option`?

gateway/src/main/scala/caliban/gateway/internal/execution/RemoteSubgraphExecutor.scala
paulpdaniels: I'll see if the signature will work with `by-name` arguments

gateway/src/main/scala/caliban/gateway/GatewayWrapper.scala
paulpdaniels: Yeah I'll move these under `Aspect`

gateway-examples/src/main/scala/example/gateway/federation/ManagedGatewayApp.scala
paulpdaniels: Whoops, that was a panic add because I accidentally uploaded a live key + configuration

gateway/src/main/scala/caliban/gateway/GatewayAspect.scala
paulpdaniels: I kept the distinction, so that `GatewayAspect` could be used more broadly as a configuration utility. Since you could define aspects that supply other `with` configurations. Or do you mean like an overload that directly takes `PhaseHooks`? Like how on `GraphQL` we have `Aspect` and `Wrapper` overloads?

gateway/src/main/scala/caliban/gateway/PhaseHooks.scala
ghostdogpr: Hmm `PhaseHooks.subscriptionSetup(...)` seems ok to me? It's on an object so it looks like a normal constructor.

gateway-examples/README.md
paulpdaniels: I believe they are otherwise the composition wasn't working correctly, let me try again, not sure if I need the `[Upper, Lower]` trick from ZIO.

gateway/src/main/scala/caliban/gateway/PhaseHandler.scala
ghostdogpr: Yeah I think you can just pass the failure through result as an Exit.Failure(cause)

ghostdogpr: We probably need a test for this case

gateway/src/main/scala/caliban/gateway/GatewayAspect.scala
ghostdogpr: Yeah I thought an overload could take `PhaseHooks`, but also that we could delete that since it has no other use now. We could add it later if needed?

gateway/src/main/scala/caliban/gateway/PhaseHooks.scala
paulpdaniels: Ok fixing

gateway/src/main/scala/caliban/gateway/GatewayAspect.scala
paulpdaniels: Removing altogether, we'll just have `PhaseAspects` for now, and we can add a `GatewayAspect` back if we find people want / need it.

gateway/src/main/scala/caliban/gateway/GatewayMetrics.scala
paulpdaniels: Hmm would there be an issue if event was changed and so the label computation also changed?

gateway/src/main/scala/caliban/gateway/internal/execution/RemoteSubgraphExecutor.scala
paulpdaniels: I'm not sure how we'd avoid the allocation, since it's in a `flatMap`, if it's empty it should collapse to `Exit.succeed(Event.AttemptHeaders(name, 0, values))` or do you mean the `Event` itself?

gateway/src/main/scala/caliban/gateway/internal/OperationHooks.scala
paulpdaniels: Ok I'll explode it into a conditional so we don't create the `OverrideLabels` object

gateway/src/main/scala/caliban/gateway/GatewayWrapper.scala
paulpdaniels: Follow-up: I moved it under `PhaseHooks` and removed Aspect altogether.

gateway/src/main/scala/caliban/gateway/PhaseHandler.scala
paulpdaniels: If the incoming hook didn't run (was interrupted) I don't think we would want to run the outgoing, since then we would have the inverse problem (also don't think we can because we wouldn't have a context value). So the only cases not currently covered is if the effect is cancelled _between_ when `incoming` and `fn` run, since `onExit` should be guaranteed to run if `fn` has started. What if I wrapped the whole block in an uninterruptibleMask instead?

```scala
ZIO.uninterruptibleMask { restore => 
  incoming(event).flatMap { case (ev, ctx) => restore(fn(ev)).onExit(exit => outgoing(ev, ctx, result(exit))) }
}
```

gateway/src/main/scala/caliban/gateway/internal/execution/RemoteSubgraphExecutor.scala
paulpdaniels: Ok I just split the callsite into `if / else` to avoid the allocation.

gateway/src/main/scala/caliban/gateway/PhaseHandler.scala
paulpdaniels: I did away with the whole stack optimization, instead I'll just rely on the ZIO interruption masking at each handler. It is probably something we can revisit if it causes a lot of overhead, but for now it seems to better ensure correctness.

gateway/src/main/scala/caliban/gateway/PhaseHandler.scala
paulpdaniels: Removed in favor of a simpler mechansim.

gateway/src/main/scala/caliban/gateway/internal/execution/RemoteSubgraphExecutor.scala
ghostdogpr: Yeah I meant the `Event`

gateway/src/main/scala/caliban/gateway/GatewayMetrics.scala
ghostdogpr: Ah.. maybe

gateway-examples/README.md
```diff
@@ -29,34 +29,61 @@ sbt "gatewayExamples/runMain example.gateway.federation.FederationGatewayApp"
...
+```scala
+import caliban.gateway.{ Gateway, GatewayMetrics, PhaseHandler, PhaseHooks }
+import zio.ZIO
+
+val logged = Gateway.compose(first, rest: _*)
+  .withPhaseHooks(
+    PhaseHooks.SubgraphCall(
+      PhaseHandler.outgoing((ev, result) => ZIO.logInfo(s"${ev.subgraph} -> ${result.outcome.label}"))
+    )
+  ) @@ GatewayMetrics.aspect
```
ghostdogpr: The examples are still old code (also in the docs)

gateway-examples/README.md
```diff
@@ -29,34 +29,61 @@ sbt "gatewayExamples/runMain example.gateway.federation.FederationGatewayApp"
...
+- `PhaseHandler.incoming` runs before the phase and can rewrite the event. `incomingDiscard` is the same without the rewrite.
+- `PhaseHandler.outgoing` sees the phase's typed result.
+- `PhaseHandler.apply` does both and can carry state from one side to the other, which is what brackets the phase effect.
+- `PhaseHandler.scoped` binds a `Scope` to that bracket, so a resource acquired on the way in outlives the outgoing side.
+
+`GatewayTracing` is built from the last two. It opens a span on the way in and records the phase's outcome on it on the
+way out. That keeps result metadata inside the span or metric scope that owns it, and gives custom logging, profiling, policy, and
+telemetry integrations the same seam to hang off.
+
+Events carry bounded metadata only, never raw GraphQL documents, variables, headers, or response bodies. Closed event values
```
ghostdogpr: This sentence is no longer true.

gateway/src/main/scala/caliban/gateway/PhaseHooks.scala
```diff
@@ -0,0 +1,371 @@
...
+      )
+}
+
+/**
+ * Constructors for single-phase [[PhaseHooks]], one per phase of the gateway lifecycle.
+ *
+ * Each builds hooks that attach a [[PhaseHandler]] to one phase and leave the rest empty. Combine them with `++` and
+ * attach the result with `Gateway#withPhaseHooks` or `Gateway#@@`. Combining accumulates
+ * rather than replaces, so several integrations can observe the same phase. Within a phase both the incoming and the
+ * outgoing sides run in the order they were combined. A [[PhaseHandler.scoped]] handler is the exception: it nests the
```
ghostdogpr: Outgoing run in reverse order, no?

gateway-tracing/src/main/scala/caliban/gateway/tracing/GatewayTracing.scala
```diff
@@ -1,170 +1,236 @@
...
-          span(
-            "caliban.gateway.subgraph",
-            SpanKind.INTERNAL,
+  private val propagation = TraceContextPropagator.default
+
+  /**
+   * The spans this integration records, as phase hooks.
+   *
+   * The request-level SERVER span hangs off `observeOperation`, the outermost phase, so routing and the operation cache
+   * fall inside it rather than beside it. Some hookscarry no span of their own and are left untouched: the request
```
ghostdogpr: typo: hookscarry

gateway-tracing/src/main/scala/caliban/gateway/tracing/GatewayTracing.scala
```diff
@@ -1,170 +1,236 @@
...
 
-          if (result.outcome != Outcome.Success) {
-            attributes.put("error.type", result.outcome.label)
-            span.setStatus(StatusCode.ERROR)
-          }
+  /**
+   * Opens a span around one phase and records that phase's own outcome on it before the span closes.
+   *
+   * The span has to enclose the phase effect, which is what a [[PhaseHandler]] with both an incoming and an outgoing
+   * side is for; the cheaper incoming-only handlers cannot express it. `result` adapts hooksthat report something
```
ghostdogpr: typo: hooksthat

gateway/src/main/scala/caliban/gateway/internal/execution/SubgraphExecutor.scala
```diff
@@ -35,20 +35,20 @@ private[gateway] trait SubgraphExecutor[-R] {
...
   override def forSubscription(implicit trace: Trace)                    =
-    underlying.forSubscription.map(new ObservedSubgraphExecutor(name, _, wrapper))
+    underlying.forSubscription.map(new ObservedSubgraphExecutor(name, _, hooks))
   override def subscribe(request: GraphQLRequest)(implicit trace: Trace) = underlying.subscribe(request)
 
   def execute(request: GraphQLRequest, operationType: OperationType)(implicit
     trace: Trace
   ): ZIO[R, SubgraphExecutor.Failure, GraphQLResponse[CalibanError]] =
-    if (!wrapper.enabled) underlying.execute(request, operationType)
+    if (!hooks.enabled) underlying.execute(request, operationType)
```
ghostdogpr: Shall we check if the specific phase is enabled instead?

gateway/src/main/scala/caliban/gateway/internal/OperationCache.scala
```diff
@@ -80,9 +80,9 @@ private[gateway] final class OperationCache[K, E, V, -R] private (
   private def observe[R0 <: R, E0, A](
     value: CacheResult
   )(effect: ZIO[R0, E0, A])(implicit trace: Trace): ZIO[R0, E0, A] =
-    if (!wrapper.enabled) effect
+    if (!hooks.enabled) effect
```
ghostdogpr: Check specific phease

ghostdogpr (commented):
Maybe the unslop of docs went too far 😄  How about applying only on docs changed in this PR? Otherwise it changes a lot.

gateway/src/main/scala/caliban/gateway/PhaseHooks.scala
```diff
@@ -0,0 +1,369 @@
...
+      Result.fromExit(_)(Result.fromResponse, _ => Result(Outcome.InternalError))
+    )
+}
+
+/**
+ * Constructors for single-phase [[PhaseHooks]], one per phase of the gateway lifecycle.
+ *
+ * Each builds hooks that attach a [[PhaseHandler]] to one phase and leave the rest empty. Combine them with `++` and
+ * attach the result with `Gateway#withPhaseHooks` or `Gateway#@@`. Combining accumulates
+ * rather than replaces, so several integrations can observe the same phase. Within a phase both the incoming and the
```
ghostdogpr, line 88: Isn't only the outgoing reverse?

vuepress/docs/docs/gateway.md
```diff
@@ -208,27 +205,30 @@ type Product {
...
+import caliban.gateway.{ Gateway, PhaseHandler, PhaseHooks }
+import caliban.gateway.PhaseHooks.Event
 import zio.Task
 
 def activeLabels(request: GraphQLRequest): Task[Set[String]] = ???
 
-val progressiveOverrides = GatewayWrapper.overrideLabels[Any] { (request, labels) =>
-  activeLabels(request).map(_ intersect labels)
-}
+val progressiveOverrides = PhaseHooks.OverrideLabels(
```
ghostdogpr, line 220: `overrideLabels`

gateway-examples/README.md
```diff
@@ -29,34 +29,60 @@ sbt "gatewayExamples/runMain example.gateway.federation.FederationGatewayApp"
...
+val logged = Gateway.compose(first, rest: _*)
+  .withPhaseHooks(
+    PhaseHooks.subgraphCall(
+      PhaseHandler.outgoing((ev, result) => ZIO.logInfo(s"${ev.subgraph} -> ${result.outcome.label}"))
+    )
+  ) @@ GatewayMetrics.hooks
+```
+
+`PhaseHooks.Event` is one lifecycle algebra, with `Event.Attempt` standing for each physical HTTP attempt, attempt zero
+included. A phase handler sees only its own event type, so `PhaseHooks.Attempt` receives an `Event.Attempt` and never has to
```
ghostdogpr, line 72: `attempt`

gateway-examples/README.md
```diff
@@ -29,34 +29,60 @@ sbt "gatewayExamples/runMain example.gateway.federation.FederationGatewayApp"
 
 The Federation GraphiQL page is available at <http://localhost:8090/graphiql>.
 
-## Observability wrappers
+## Observability aspects
```
ghostdogpr, line 32: A few occurrences of `aspects` that could be `hooks`

ghostdogpr (commented):
I left a few minor comments. Feel free to merge, I have some minor polishing suggestions but i can do it after it's merged.

